PR #143, opened 2022-09-13
remove sample config from docs

3u13r:
### Proposed change(s)

- update sample config

I don't know what our current approach is to updating the v2.0 docs. Since I wanted to include this also in v2.0 I changed both files.


thomasten:
You also need to remove it from sidebars.js and versioned_sidebars.
Also see the link checker results. There are some links to the deleted page.

> I don't know what our current approach is to updating the v2.0 docs. Since I wanted to include this also in v2.0 I changed both files.

IMO this is the correct approach.

thomasten (approved):

Nirusu (approved):

PR #145, opened 2022-09-13
AB#2415: Don't sign & measure E2E artifacts

Nirusu:
### Proposed change(s)
- Build E2E CLIs as OSS
- Don't sign E2E artifacts

### DIFF
--- a/.github/actions/build_cli/action.yml
+++ b/.github/actions/build_cli/action.yml
@@ -12,6 +12,10 @@ inputs:
     description: "Build CLI for this architecture. [amd64, arm64]"
     required: true
     default: "amd64"
+  enterpriseCLI:
+    description: "Build CLI with enterprise flag."
+    required: false
+    default: "false"
   cosignPublicKey:
     description: "Cosign public key"
     required: false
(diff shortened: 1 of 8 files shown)


Nirusu:
(Draft for now since we want the measurement functionality for releases, but apparently that has been extracted from E2E tests of release-ready states so far)

Nirusu:
Should be ready for review. On demand I separated the measurement action from the E2E test.

Defaults are mostly defined in the action, not in the workflow contrary to many other workflows. I don't think most workflows trying to measure stuff would necessarily need to define all variables... but if we don't like inconsistency, I can move them to the workflow itself and pass it to the action.

malt3 (changes requested):
Please also add an extra step to the release process (e2e tests are now run only for stability, measurement pipelines for GCP and Azure are run to upload measurements):

https://github.com/edgelesssys/constellation/blob/0da373e03688c85f3abb92c2fd6eac33f8a93452/.github/docs/release.md?plain=1#L44

Nirusu:
Added.

malt3 (approved):
LGTM

PR #148, opened 2022-09-13
tidy link checking

thomasten:


.github/docs/README.md
```diff
@@ -92,7 +92,7 @@ Next, add API permissions to Managed Identity:
 * `$GraphAppId` in this article is for Microsoft Graph. Azure AD Graph is `00000002-0000-0000-c000-000000000000`
 * Note that changing permissions can take between few seconds to several hours
 
-Store output of `az ad sp ...` in [GitHub Action Secret](https://github.com/edgelesssys/constellation/settings/secrets/actions) or create a local secret file for act to consume.
+Store output of `az ad sp ...` in GitHub Action Secret or create a local secret file for act to consume.
```
Nirusu, line 95:
```suggestion
Store output of `az ad sp ...` in a GitHub Action Secret called `AZURE_E2E_CREDENTIALS`. If you are using act, you can specify the secret either via `-s AZURE_E2E_CREDENTIALS=value` or by using a local secrets file which act can consume by specifying it with `--secret-file my.secrets`.
```
Or something similar, my suggestion is quite a wordy. 

There's also the `AZURE_E2E_CLIENT_SECRET` or, for GCP, the `GCP_CLUSTER_SERVICE_ACCOUNT` part missing which goes into the Constellation config. We would need to update this, too. Otherwise the tests won't run.

Nirusu (commented):
Link checking changes look LGTM, but we documentation does not appear quite up to date in what secrets we require?

.github/docs/README.md
thomasten: Your suggestion looks ok to me. I don't know about how the other secrets work. Maybe I merge this PR and you can fix all of this in another one?

Nirusu: Sure, we can also do it like this.

Nirusu (approved):

PR #149, opened 2022-09-13
docs: fix command in trusted launch workflow and add fetch-measurements

thomasten:
TrustedLaunchAndConfidentialVmSupported fails on my machine (latest az cli):
```
(InvalidParameter) The value 'TrustedLaunchAndConfidentialVmSupported' for feature 'SecurityType' is invalid. Allowed characters are English alphanumerical characters with spaces and commas. The length of the value cannot exceed 100 characters.
Code: InvalidParameter
Message: The value 'TrustedLaunchAndConfidentialVmSupported' for feature 'SecurityType' is invalid. Allowed characters are English alphanumerical characters with spaces and commas. The length of the value cannot exceed 100 characters.
Target: galleryImage.properties.features
```
Instead I made it configurable via env var.

### DIFF
--- a/docs/docs/workflows/trusted-launch.md
+++ b/docs/docs/workflows/trusted-launch.md
@@ -1,6 +1,6 @@
 # Use Azure trusted launch VMs
 
-Constellation also supports [trusted launch VMs](https://docs.microsoft.com/en-us/azure/virtual-machines/trusted-launch) on Microsoft Azure. Trusted launch VMs don't offer the same level of security as CVMs, but are available in more regions and in larger quantities. The main difference between trusted launch VMs and normal VMs is that the former offer vTPM-based remote attestation. When used with trusted launch VMs, Constellation relies on vTPM-based remote attestation to verify nodes.
+Constellation also supports [trusted launch VMs](https://docs.microsoft.com/en-us/azure/virtual-machines/trusted-launch) on Microsoft Azure. Trusted launch VMs don't offer the same level of security as Confidential VMs, but are available in more regions and in larger quantities. The main difference between trusted launch VMs and normal VMs is that the former offer vTPM-based remote attestation. When used with trusted launch VMs, Constellation relies on vTPM-based remote attestation to verify nodes.
 
 :::caution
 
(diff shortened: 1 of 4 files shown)


hack/importAzure.sh
```diff
@@ -44,6 +44,7 @@ AZURE_IMAGE_NAME="${AZURE_IMAGE_NAME:-upload-target}"
 AZURE_IMAGE_OFFER="${AZURE_IMAGE_OFFER:-constellation}"
 AZURE_IMAGE_DEFINITION="${AZURE_IMAGE_DEFINITION:-constellation}"
 AZURE_SKU="${AZURE_SKU:-constellation-coreos}"
+AZURE_SECURITY_TYPE="${AZURE_SECURITY_TYPE:-ConfidentialVmSupported}"
```
malt3, line 47:
```suggestion
AZURE_SECURITY_TYPE="${AZURE_SECURITY_TYPE:-TrustedLaunch}"
```

I would choose a working security type for TrustedLaunch as default. `ConfidentialVmSupported` seems to only work for CVMs (just like security type `ConfidentialVM`).

thomasten: I wasn't sure who else uses this script and thus what should be the default. Is usage of this script with CVMs outdated and is it safe to change the default behavior?

malt3: The script is only used for trusted launch. I also searched around our repo and this is the only reference. Re-importing images in other scenarios is not required.

docs/docs/workflows/trusted-launch.md
```diff
@@ -32,6 +32,14 @@ The script creates the following resources:
 
 Once the import is completed, use the `ID` of the image version in your `constellation-conf.yaml` for the `image` field. Set `confidentialVM` to `false`.
 
+Fetch the image measurements:
+
+```bash
+IMAGE_VERSION=2.0.0
+URL=https://public-edgeless-constellation.s3.us-east-2.amazonaws.com//CommunityGalleries/ConstellationCVM-b3782fa0-0df7-4f2f-963e-fc7fc42663df/Images/constellation/Versions/$IMAGE_VERSION/measurements.yaml
```
3u13r, line 39: Shoudn't we do the measurement fetching just before initialization, since then the correct measurements should be featched, since we read out the specified image: https://github.com/edgelesssys/constellation/blob/main/cli/internal/cmd/configfetchmeasurements.go#L118

malt3 (approved):
LGTM

PR #151, opened 2022-09-14
AB#2261 Add loadbalancer for control-plane recovery

daniel-weisse:
### Proposed change(s)

- Only start the recovery server on control-plane nodes
  - Worker nodes need a stable control-plane. If one is available no manual recovery is required
  - The server now runs on port `9999` instead of `9000`
- Create a loadbalancer for the recovery server
  - Loadbalancer has a public IP, making recovery of nodes without a public IP address (Azure) simpler

### Checklist

- [ ] Link to Milestone


malt3:
No super relevant, but why do we switch the port?

daniel-weisse:
>No super relevant, but why do we switch the port?

recovery server used the same port as the bootstrapper.
We can't configure two loadbalancing rules for the same port

malt3 (approved):
LGTM. As a next step, we should make the endpoint optional an read it from the ID file.

3u13r:
> > No super relevant, but why do we switch the port?
> 
> recovery server used the same port as the bootstrapper. We can't configure two loadbalancing rules for the same port

Just thinking out loud: Can't we re-use the LB of the first port then? The healthcheck should be the same, the backend also. What is the reason we decided against that?

3u13r (approved):
LGTM

PR #153, opened 2022-09-14
Fix dirty build tag for CLI releases != linux/amd64

Nirusu:
### Proposed change(s)

In our v2.0.0 CLI release, all binaries except the Linux amd64 are tagged with `GitTreeState: dirty` when executing `constellation version`. Presumably, this seems to be the case due to Rekor being downloaded to the constellation root directory, which is not excluded from our git and thus taints the state. Therefore, let's specify it explicitly that we want to download to the build directory and also clean up the binary after we copied it.

It's a bit hard to test before we push another release, but testing it by running the action in act and copying the binaries out of the container seems to work and gives a clean state to the other binaries with this PR.


malt3 (approved):
LGTM

PR #155, opened 2022-09-14
docs: minor fixes in first steps and wording improvements

thomasten:
Most changes should be self-explanatory.
Take a look at the echo commands in the two scripts. I changed it because when copy-pasting it in my terminal, output and commands were mixed and thus the output wasn't easy to read. Is it okay to use echo like this?


malt3 (approved):
LGTM

PR #158, opened 2022-09-15
AB#2262 Automatic recovery

daniel-weisse:
### Proposed change(s)

- Update the `constellation recover` command to be fully automatic
  - Endpoint is read from file if not explicitly set using the `--endpoint`/`-e` flag
  - Recover call is repeated until the recovery loadbalancer no longer has any "healthy" backends

### Additional info
- The actual changeset is rather small: We simply wrap the recover call in a for loop that terminates after a non-retriable error is encountered
- Azure's LB healthprobe frequency was increased to 5 seconds (the maximum)
  - This is needed because otherwise the LB is slow to realise a node was successfully recovered
    - This is no problem on GCP since the default probe frequency is 1 second
- A "service unavailable" error is retried once, this is to avoid aborting the recovery in case the LB routes our request to a node we just recovered
  - For GCP we set a small retry interval of 5 seconds, since that is enough for the LB to remove the node from its backends
  - On Azure we require a larger interval to avoid this issue, this slows down recovery slightly and therefore it is only set if the CSP is Azure

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/azure/loadbalancer.go
+++ b/cli/internal/azure/loadbalancer.go
@@ -98,8 +98,9 @@ func (l LoadBalancer) Azure() armnetwork.LoadBalancer {
 				{
 					Name: to.Ptr(recoveryHealthProbeName),
 					Properties: &armnetwork.ProbePropertiesFormat{
-						Protocol: to.Ptr(armnetwork.ProbeProtocolTCP),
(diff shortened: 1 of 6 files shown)


malt3 (approved):
LGTM

m1ghtym0 (approved):
lgtm

PR #159, opened 2022-09-15
AB#2198: CI/E2E for macOS 

Nirusu:
This is quite a lot of changes at a lot of different places, so I suggest reviewing commit-by-commit for a bit more clarity. Also not sure who wants to review this, so I will just tag some reviewers here ;)

### Proposed change(s)
- Remove many setup steps from actions (dumped them under `setup_linux` for now, it's unused though since Github-hosted runners have most stuff preinstalled anyways)
- Add macOS + multi-arch builds to "Build Constellation binaries" action (cross-compiled under Linux, same as we do for releases)
- Run unit tests for CLI & cdbg/debugd under macOS 
- Add manual e2e test using macOS
- Update sonobuoy to 0.56.10
- Update rekor-cli to 0.12.10
- Some VS Code auto-formatting

All macOS actions use macOS Monterey (12). This is _not_ `macos-latest` which is on Big Sur (11), however since macOS update adoption usually is than for Ubuntu and Monterey is almost a year old already, I think it makes sense to use the more latest version rather than what GitHub declares as latest here.

Note that since that the changes also apply to some release actions we usually only run before we actually release something, I could not test everything in-depth. There could be breakage on private runners, but usually this should be easily resolvable since it's just missing install steps (and they should be under `setup_linux`).

Otherwise, most stuff we use such as Docker Buildx, AWS/Azure/GCloud CLI. CMake, jq, yq and more are already preinstalled in the public runner images. The stuff is often fairly recent since the images are often updated, but not always 100% up-to-date (e.g. GCloud CLI is a few versions older). Not sure why the [Ubuntu 20.04 / `ubuntu-latest`](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#cli-tools) and [Ubuntu 22.04](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#cli-tools) have differences for the GCloud SDK even though it's downloaded externally... but that can/should serve as an exercise on unexpected version dependencies for us.

All stuff that actually ends up in the images and thus in confidential space should be under our control, however, so Go is installed manually every time and for CGo binaries (bootstrapper, disk-mapper etc.) we use Docker anyways. 

For macOS, Docker is missing due to licensing issues and GCloud CLI is also not installed (not sure why). While we can easily install GCloud using an action, I don't personally know about Docker. I think it's technically possible, but not sure about that license issue. We don't need to use the macOS runner for everything, so it should not be an huge issue. For e2e tests, Docker is only required to build the bootstrapper. Here, in case an e2e test is running with an debug-image, we prepend a Linux VM that builds the bootstrapper, uploads it to the Github Actions cache and later download it in the e2e test action instead of building it manually. This step is restricted to macOS runners, so nothing should change for Linux runners.

Apart from that, only minor changes were required. For example: `nproc` does not exist under macOS. While this can be aliased with an equivalent (e.g.: `sysctl -n hw.logicalcpu`), it's often defined redundantly anyways since we only build one job. Apart from that, most changes should be apparent hopefully or restricted to the newly created e2e action.

### DIFF
--- a/.github/actions/azure_login/action.yml
+++ b/.github/actions/azure_login/action.yml
@@ -2,30 +2,14 @@ name: Azure login
 description: "Login to Azure & configure az CLI."
 inputs:
   azure_credentials:
-    description: 'Credentials authorized to create Constellation on Azure.'
+    description: "Credentials authorized to create Constellation on Azure."
     required: true
 runs:
-  using: 'composite'
+  using: "composite"
   steps:
-  - name: Install az CLI
-    run: |
-      echo "::group::Install build dependencies"
-      sudo apt-get update
-      sudo apt-get install ca-certificates curl apt-transport-https lsb-release gnupg -y
-      curl -sL https://packages.microsoft.com/keys/microsoft.asc |
-        gpg --dearmor |
-        sudo tee /etc/apt/trusted.gpg.d/microsoft.gpg > /dev/null
-      AZ_REPO=$(lsb_release -cs)
-      echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ $AZ_REPO main" |
-          sudo tee /etc/apt/sources.list.d/azure-cli.list
-      sudo apt-get update
-      sudo apt-get install azure-cli -y
-      az help
-      echo "::endgroup::"
-    shell: bash
-  # As described at:
-  # https://github.com/Azure/login#configure-deployment-credentials
-  - name: Login to Azure
-    uses: azure/login@24848bc889cfc0a8313c2b3e378ac0d625b9bc16
-    with:
-      creds: ${{ inputs.azure_credentials }}
+    # As described at:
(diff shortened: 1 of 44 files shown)


malt3 (approved):
Broadly lgtm. I think we should try every workflow affected by this before merging (maybe except for the release workflows)

Nirusu:
Found some breakage for CoreOS builder & other-version e2e tests, should all be fixed now. CoreOS was partly my mistake, but there was also some general breakage of the CoreOS builder introduced by eee2df9723a217e6f8b22c0398bd703f0b5c422f, which is now fixed too with an extra commit which I'll sneak in here with this PR.

Tested most actions not involving any releases, looks mostly good now. Will let the CoreOS builder finish running to see if it finally succeeds and if all goes well, merge this on Monday.

PR #160, opened 2022-09-15
Add go-tidy-check workflow

katexochen:
### Proposed change(s)

- Add go-tidy-check action
- Checkout last commit instead of merge commit on pull_request event

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -21,13 +21,15 @@ jobs:
   build-bootstrapper:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+
       - name: Setup Go environment
         uses: actions/setup-go@268d8c0ca0432bb2cf416faae41297df9d262d7f
         with:
           go-version: "1.19.1"
-
-      - name: Check out repository
-        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
 
       - name: Build the bootstrapper
         uses: ./.github/actions/build_bootstrapper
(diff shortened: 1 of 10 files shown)


cli/main.go
```diff
@@ -18,3 +18,5 @@ func main() {
 	}
 	os.Exit(0)
 }
+
+// Trigger.
```
Nirusu, line 22:
```suggestion
```

.github/workflows/build-binaries.yml
```diff
@@ -21,42 +21,58 @@ jobs:
   build-bootstrapper:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout on PR
+        if: github.event_name == 'pull_request'
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
Nirusu, line 28: I _think_ you don't actually need to create two conditional steps here. I looked a bit around on GitHub and found repos that also run on both pull_request and push and specify `ref: ${{ github.event.pull_request.head.sha }}`  for both.

I think it makes sense and can work, since in case of a push, the ref would likely be an empty string, causing the action to use the default behavior (ref = SHA that triggered the action) rather than failing.

Can you please try this? You could keep temporarily remove the pull_request trigger and change it to trigger on a push to your branch while keeping the same
```yaml
       with:
          ref: ${{ github.event.pull_request.head.sha }}
```
variable passthrough.

Nirusu (changes requested):

.github/workflows/test-tidy.yml
```diff
@@ -0,0 +1,52 @@
...
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Checkout on Push
+        if: github.event_name == 'push'
+        uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
+
+      - name: Install Dependencies
+        run: sudo apt-get update && sudo apt-get -y install libcryptsetup-dev libvirt-dev
+
```
Nirusu, line 25:
```suggestion
```

I think you actually don't need this since you are not building anything?

katexochen: If I remember correctly I tried it without and it didn't work. This is also needed for govulncheck and lint action.

Nirusu: Running go lint & go mod tidy in Docker without libvirt-dev installed, I don't see any issues.

Not sure about the linter, but govulncheck does static code checking on binaries and thus requires these packages to build them (it's also OS/arch dependent).

Nirusu (approved):

PR #169, opened 2022-09-19
Remove exporting of PCRs from QEMU metadata API

daniel-weisse:
### Proposed change(s)

Retrieving image measurements is done by a separate tool. We no longer need to start a QEMU cluster to do so.
This PR removes the need to mount a log directory for the metadata container when using terraform to create a QEMU based cluster, since the metadata api no longer writes anything to disk.


malt3:
Can we remove all traces of the log dir? 

https://github.com/edgelesssys/constellation/blob/ref/hack/qemu-metadata-log-dir/terraform/libvirt/variables.tf#L55

Otherwise LGTM

malt3 (approved):
LGTM

PR #172, opened 2022-09-19
Use terraform in CLI to create QEMU cluster

daniel-weisse:
### Proposed change(s)

- Enable `constellation create` to bootstrap a QEMU based cluster
  - Image Path, vCPUs, memory, and state disk size can be configured using `constellation-conf.yaml`
  - The infrastructure is created using the terraform-exec package
    - This package is wrapper for making exec calls on a terraform binary
    - We use either a terraform binary from the users path, if the installed version satisfies our version constraints, or download the newest release to a temporary file
  - When running `constellation create`, the necessary terraform files are saved to the users current working directory, and removed when running `constellation terminate`
- Enable `constellation terminate` to terminate QEMU based clusters
- Move `terraform/libvirt/` to `cli/internal/terraform` as these files are now embedded in the CLI
- Preparations for enabling Terraform to manage creation of Constellation clusters

### Additional info
Creating a qemu cluster is now works like the following:
1. Create a QEMU config using `constellation config generate qemu`
3. Update the config to your preferred configuration and set the path to your qemu image file
4. Run `constellation create -c3 -w2` to create a cluster with 3 control-plane and 2 worker nodes
5. Run `constellation init`
6. Use the cluster
7. Run `constellation terminate` to tear down your cluster and clean up your workspace

### Checklist

- [x] Link to Milestone


.github/docs/qemu.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+</details>
+
+## Update libvirt settings
+
+Open `/etc/libvirt/qemu.conf` and change the following settings:
+
+```shell-session
+security_driver = "none"
```
Nirusu, line 81: Not sure about Fedora/SELinux, but I think on Ubuntu you can get away with keeping AppArmor disabled but defining:
```
"/var/lib/libvirt/images" r,
"/var/lib/libvirt/images/**" rwk,
```
in `/etc/apparmor.d/local/abstractions/libvirt-qemu`.  At least that works on my machine™

Not sure if we want to suggest this to the user, though if it can cause issues on other distros.

Nirusu (commented):
Since we have multi-OS and multi-arch binaries for the CLI, I think we should return an error when trying to create an QEMU instance at least for arm64 since I doubt that's gonna work well with an amd64 image on either KVM or HVF on macOS. Software emulation could work but that's probably a terrible experience.

For macOS on amd64, I think it should be possible to run but the current approach is not quite gonna work. The main issues I see are the hard-coded paths, since libvirt & friends are usually prefixed with `/usr/local/` on macOS when installed with brew (e.g. `/usr/local/var/run/libvirt-sock` instead of just `/var/run/libvirt-sock` on Linux). Docker is the same as on Linux if I recall correctly, even though it's running with HyperKit / in a VM.

However, I can't say this for sure since we would need an bare-metal Intel Mac to test this (or an AWS instance, but they are pricey since you need to pay for 24 hours at least).

If we care, I could dig into this a bit deeper. Or maybe someone else has an Intel Mac lying around. 

If we don't care, I would guard this function too for macOS and restrict running `constellation create qemu` to Linux on amd64 only.

.github/docs/qemu.md
malt3: I like the idea of defining rules for only what is needed for AppArmor. Disabling the security_driver on Ubuntu seems to be what everyone else is doing:

- https://github.com/openshift/installer/blob/master/docs/dev/libvirt/README.md#configure-qemuconf
- https://github.com/dmacvicar/terraform-provider-libvirt/issues?q=security_driver

I think we can test your approach and see if it breaks :)

malt3:
Testing works really well on my machine :)

cli/internal/terraform/input.go
```diff
@@ -0,0 +1,49 @@
...
+	qemuConfigTemplate = `
+constellation_coreos_image = "%s"
+image_format = "%s"
+control_plane_count = %d
+worker_count = %d
+vcpus = %d
+memory = %d
+state_disk_size = %d
+ip_range_start = %d
+machine = "%s"
```
malt3, line 19: nit: I think we can just assume q35 going forward.

cli/internal/terraform/terraform.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/hashicorp/hc-install/product"
+	"github.com/hashicorp/hc-install/releases"
+	"github.com/hashicorp/hc-install/src"
+	"github.com/hashicorp/terraform-exec/tfexec"
+	tfjson "github.com/hashicorp/terraform-json"
+	"github.com/spf13/afero"
+)
+
+const (
+	tfVersion         = ">= 1.2.0"
```
katexochen, line 29: Guess we should pin a concrete version.

katexochen (commented):
Looks good :+1:

cli/internal/terraform/terraform.go
daniel-weisse: The reason for specifying a version constraint instead of pinning a concrete version is to avoid downloading Terraform if a user already has a newer version installed (which should be backwards compatible).

Newest release is `v1.2.9` not sure if we want to set a higher minimum version, but `v1.2.0` works for creating qemu clusters.

.github/docs/qemu.md
Nirusu: I mean, it works (or worked :P) on my machine and many results about this issue mainly revolve around storage pools. The exemption should cover it for our usage (though technically breaking guest-to-guest if a guest can manipulate qemu?), but might not work for other libvirt applications. 

So that's why I can see others suggesting to completely disable it. Or people just choose the lazy way to fix this issue, that wouldn't surprise me either.

There might even be a better, stricter way to fix this to keep the isolation in-tact for our use-case. But my current testing setup is broken for some other reason 🙃

hack/check-licenses.sh
```diff
@@ -27,6 +27,8 @@ while read line; do
           ;;
         github.com/letsencrypt/boulder)
           ;;
+        github.com/hashicorp/*)
+          ;;
```
katexochen, line 31: @thomasten should we just add needed packages here?

katexochen (changes requested):

cli/internal/terraform/loader.go
```diff
@@ -0,0 +1,113 @@
...
+	"fmt"
+	"path"
+	"strings"
+
+	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
+	"github.com/edgelesssys/constellation/v2/internal/file"
+	"github.com/spf13/afero"
+)
+
+//go:embed all:terraform/*
```
Nirusu, line 21:
```suggestion
//go:embed terraform/*
```

All `all:` does is including hidden files (files starting with `.` or `_`). Shouldn't matter for release builds built in a clean Linux environment but can either include trash (e.g. `.DS_Store` on macOS) or even worse: a `.terraform` directory from running Terraform directly in the directory to be embedded that overshadows the real filesystem. 

That can cause some funny non-sense and break stuff in weird ways. Ask me how I know...

Nirusu (changes requested):

hack/check-licenses.sh
thomasten: I wanted to do it like this at first, but then suggested to Daniel to just add hashicorp/* for simplicity. But I don't have a strong opinion on this.

cli/internal/cloudcmd/create.go
```diff
@@ -194,3 +210,28 @@ func (c *Creator) createAzure(ctx context.Context, cl azureclient, config *confi
 
 	return cl.GetState(), nil
 }
+
+func (c *Creator) createQEMU(ctx context.Context, cl qemuclient, name string, config *config.Config, controlPlaneCount, workerCount int,
+) (stat state.ConstellationState, retErr error) {
+	defer rollbackOnError(context.Background(), c.out, &retErr, &rollbackerQEMU{client: cl})
```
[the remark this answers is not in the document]
daniel-weisse: Is this correct?
We dont use the functions context in `createAzure` or `createGCP`. Without testing anything I would suspect reusing the create functions context for the rollbacker to break the rollback action if a user cancels the create by pressing ctrl+c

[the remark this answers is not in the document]
katexochen: You're right. :)

Nirusu:
Would it be possible to add a rollback for the domains?

```
Error: error defining libvirt domain: operation failed: domain 'constell-control-plane-0' already exists with uuid 5f8f75b0-f776-48b6-a3fc-07857b007bb1

  with module.control_plane.libvirt_domain.instance_group[0],
  on modules/instance_group/main.tf line 14, in resource "libvirt_domain" "instance_group":
  14: resource "libvirt_domain" "instance_group" {


Error: error defining libvirt domain: operation failed: domain 'constell-worker-0' already exists with uuid 62c92ea7-43d1-4518-ad53-0c1fdd922ba7

  with module.worker.libvirt_domain.instance_group[0],
  on modules/instance_group/main.tf line 14, in resource "libvirt_domain" "instance_group":
  14: resource "libvirt_domain" "instance_group" {
```

Otherwise LGTM!

daniel-weisse:
> Would it be possible to add a rollback for the domains?

Unless something goes seriously wrong, Terraform should be clean any resources created before an error occurs.
As Terraform handles all interaction with libvirt implementing custom rollback functionality would be a bit annoying since this would add cgo dependencies to the CLI.

Advanced cleanup operations should also become obsolete with https://github.com/edgelesssys/constellation/pull/191 as we can just delete the libvirt container to remove any leftover artifacts.

Nirusu:
Terraform (or more specifically: the Terraform libvirt provisoner) does definitely some cleanup, though the domains remain unfortunately. 

But if it would require custom cleanup code from our side, especially with cgo, then yeah let's just keep it as this. I see "Destroy domain" is already in the docs, so I think that should be fine then.

Nirusu (approved):

katexochen (approved):

PR #174, opened 2022-09-20
Run macos builds as separate jobs

daniel-weisse:
### Proposed change(s)

- Speed up workflow run times by building macOS binaries in separate jobs
  - Reduces build time from 10-12 minutes to ~7 minutes
- Fix cdbg being built for the wrong OS Arch

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Build debugd
         uses: ./.github/actions/build_debugd
 
-  build-cdbg:
+  build-cdbg-linux:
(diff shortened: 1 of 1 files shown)


Nirusu (approved):
I actually had it this way but reverted to keep them in one job since I didn't want to spawn too many parallel runners. I am not sure if it actually stalls another job currently. You can have 20 Linux runners in parallel and we already have many checks on a PR. Your current QEMU PR has 14 parallel jobs, for example. If we have more in the future or other actions running such as e2e tests or from another PR, that could stall some actions due to a lack of available runners.

If other actions don't stall, I agree that should be faster. Good catch with the wrong targetOS.

PR #180, opened 2022-09-20
initial AWS terraform

3u13r:
### Proposed change(s)

- Terraform to create instance IAM roles (aka. serivce accounts attached to the instances)
- Terraform to create a Constellation look-alike infrastructure

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/terraform/aws/iam/main.tf
+++ b/terraform/aws/iam/main.tf
@@ -1,0 +1,189 @@
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 4.0"
+    }
+    random = {
+      source  = "hashicorp/random"
+      version = "3.4.1"
+    }
+  }
+}
+
+# Configure the AWS Provider
+provider "aws" {
+  region = "us-east-2"
+}
+
+locals {
+  uid  = random_id.uid.hex
+  name = "${var.name}-${local.uid}"
+  tag  = "constellation-${local.uid}"
+}
+
+resource "random_id" "uid" {
+  byte_length = 8
+}
+
+resource "aws_iam_instance_profile" "control_plane_instance_profile" {
+  name = "${var.name}_control_plane_instance_profile"
+  role = aws_iam_role.control_plane_role.name
+}
+
+resource "aws_iam_role" "control_plane_role" {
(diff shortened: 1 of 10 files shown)


terraform/aws/iam/main.tf
```diff
@@ -0,0 +1,189 @@
...
+    random = {
+      source  = "hashicorp/random"
+      version = "3.4.1"
+    }
+  }
+}
+
+# Configure the AWS Provider
+provider "aws" {
+  region = "us-east-2"
```
Nirusu, line 16: Should already be a variable?

terraform/aws/main.tf
```diff
@@ -0,0 +1,187 @@
...
+  ports_konnectivity = "8132"
+  ports_verify       = "30081"
+  ports_debugd       = "4000"
+
+  cidr_vpc_subnet_nodes = "192.168.178.0/24"
+  count_control_plane   = 1
+  count_worker          = 1
+  image_id              = "ami-02f3416038bdb17fb" // Ubuntu 22.04 LTS
+  instance_type         = "t2.micro"
+  disk_size             = 30
```
Nirusu, line 37: Should also already be variable for CLI intergration?

Nirusu (commented):
[tfsec](https://github.com/aquasecurity/tfsec) has some suggestions. Not every makes sense in our case I believe (e.g. hard to define CIDR ranges for firewalls when creating a generic cluster with the CLI) but a few might be worth looking into.

Otherwise I added some few suggestions I have while trying this out.

I also would already define some more variables for CLI integration. Even with development left having the instance type, disk size and # of nodes configurable variable seems useful to me.

Nirusu (approved):
In general LGTM.

Maybe two things for consideration:

1. Since AWS has that annoying 32 character limit, maybe we should not include the UID in the name (or shorten it down to four characters)? 
2. Maybe enable Amazon-managed disk encryption by default? In theory it should not be necessary since we encrypt the state disk on our own. But if I see correctly, we also use Azure/Google-managed encryption for the other CSPs. Since it should only minimally affect the latency and an AWS-managed key should be free according to the docs, I think we should enable it nevertheless to silence security scanners and maybe also for compliance reasons (even though technically it really shouldn't matter).

PR #181, opened 2022-09-21
Docs: Suggest terminal commands to install the CLI

Nirusu:
### Proposed change(s)

- Suggest terminal commands to install the CLI
- Point user where to download the sigstore signatures (in case the terminal steps where used and the user did not visit the GitHub release page)


docs/docs/getting-started/install.md
```diff
@@ -15,12 +15,90 @@ Make sure the following requirements are met:
...
 
-Download the CLI executable from the [release page](https://github.com/edgelesssys/constellation/releases). Move the downloaded file to a directory in your `PATH` (default: `/usr/local/bin`) and make it executable by entering `chmod u+x constellation` in your terminal. 
+Download the CLI executable from [GitHub](https://github.com/edgelesssys/constellation/releases). Use `install` to copy the downloaded file to your `PATH` (default: `/usr/local/bin`) and automatically make it executable.
 
 :::note
-Edgeless Systems uses [sigstore](https://www.sigstore.dev/) to sign each release of the CLI. You may want to [verify the signature](../workflows/verify-cli.md) before you use the CLI.
+Edgeless Systems uses [sigstore](https://www.sigstore.dev/) to sign each release of the CLI. You may want to [verify the signature](../workflows/verify-cli.md) before you install and use the CLI.
 :::
 
+You can use the following commands for this:
```
thomasten, line 24: With the new instructions, this is all quite repetitive. I'd just drop the note and shorten all this to
```markdown
The CLI executable is available at [GitHub](https://github.com/edgelesssys/constellation/releases).
Install it with the following commands:
```

docs/docs/getting-started/install.md
```diff
@@ -15,12 +15,90 @@ Make sure the following requirements are met:
...
+
+<tabs>
+<tabItem value="linux-amd64" label="Linux (amd64)">
+
+1. Download the CLI:
+
+```bash
+curl -LO https://github.com/edgelesssys/constellation/releases/download/v2.0.0/constellation-linux-amd64
```
thomasten, line 32:
```suggestion
curl -LO https://github.com/edgelesssys/constellation/releases/latest/download/constellation-linux-amd64
```
This works for me. Can you double-check on macOS?

thomasten (commented):
Thanks! Definitely simplifies this section.

docs/docs/getting-started/install.md
Nirusu: Works! Interesting the parameters are switched here, that's why I thought `latest` in the URL does not work.

Nirusu:
Link checker fails because guix.gnu.org is down, but that's unrelated to this PR.

PR #182, opened 2022-09-21
Upgrade go module to v2

katexochen:
### Proposed change(s)

- Upgrade main go module to v2, fixing Go import issues due to the v2.0.0 tag
- The operator is also upgraded to v2
- Hack folder is left without v2 tag, as it shouldn't be imported anyway
- Scalinggroup resource was modified through running the test


thomasten (commented):
I lack knowledge to review the two node operator scaling commits. Otherwise LGTM

malt3 (approved):
Operator code changes LGTM.

PR #185, opened 2022-09-22
Add gocyclo to golangci-lint config

katexochen:
### Proposed change(s)

- Run gocyclo with golangci-lint, as the check is also used in the goreportcard

**Problems to solve:**
- [ ] access_manager/access_manager.go:109:1                                                    complexity 24 of func `deployKeys` is high (> 15)
- [ ] bootstrapper/cmd/bootstrapper/main.go:47:1                                                complexity 18 of func `main` is high (> 15)
- [ ] bootstrapper/internal/kubernetes/k8sapi/install.go:81:1                                   complexity 22 of func `(*osInstaller).extractArchive` is high (> 15)
- [ ] bootstrapper/internal/kubernetes/kubernetes.go:78:1                                       complexity 31 of func `(*KubeWrapper).InitCluster` is high (> 15)
- [ ] cli/internal/cmd/create.go:50:1                                                           complexity 17 of func `create` is high (> 15)
- [ ] hack/image-measurement/main.go:82:1                                                       complexity 16 of func `(*libvirtInstance).createLibvirtInstance` is high (> 15)
- [ ] hack/pseudo-version/pseudo-version.go:22:1                                                complexity 16 of func `main` is high (> 15)
- [ ] operators/constellation-node-operator/controllers/autoscalingstrategy_controller.go:45:1  complexity 29 of func `(*AutoscalingStrategyReconciler).Reconcile` is high (> 15)
- [ ] operators/constellation-node-operator/controllers/nodeimage_controller.go:79:1            complexity 22 of func `(*NodeImageReconciler).Reconcile` is high (> 15)
- [ ] operators/constellation-node-operator/controllers/nodeimage_controller.go:511:1           complexity 19 of func `(*NodeImageReconciler).createNewNodes` is high (> 15)
- [ ] operators/constellation-node-operator/controllers/nodeimage_controller.go:652:1           complexity 18 of func `nodeImageStatus` is high (> 15)
- [ ] operators/constellation-node-operator/controllers/pendingnode_controller.go:64:1          complexity 16 of func `(*PendingNodeReconciler).Reconcile` is high (> 15)
- [ ] operators/constellation-node-operator/internal/azure/client/nodeimage.go:50:1             complexity 16 of func `(*Client).CreateNode` is high (> 15)
- [ ] operators/constellation-node-operator/main.go:59:1                                        complexity 18 of func `main` is high (> 15)

**How does gocyclo compute complexity?**

> The cyclomatic complexity of a function is calculated according to the following rules:
> 
>  1 is the base complexity of a function
> +1 for each 'if', 'for', 'case', '&&' or '||'

### DIFF
--- a/cli/internal/gcp/client/instances.go
+++ b/cli/internal/gcp/client/instances.go
@@ -148,6 +148,13 @@ func (c *Client) CreateInstances(ctx context.Context, input CreateInstancesInput
 
 // TerminateInstances terminates the clients instances.
 func (c *Client) TerminateInstances(ctx context.Context) error {
+	if err := c.terminateInstanceGroupManagers(ctx); err != nil {
+		return err
+	}
+	return c.deleteInstanceTemplates(ctx)
+}
+
+func (c *Client) terminateInstanceGroupManagers(ctx context.Context) error {
 	ops := []Operation{}
 	if c.workerInstanceGroup != "" {
 		op, err := c.deleteInstanceGroupManager(ctx, c.workerInstanceGroup)
@@ -172,10 +179,12 @@ func (c *Client) TerminateInstances(ctx context.Context) error {
 		c.controlPlaneInstanceGroup = ""
 		c.controlPlanes = make(cloudtypes.Instances)
 	}
-	if err := c.waitForOperations(ctx, ops); err != nil {
-		return err
-	}
-	ops = []Operation{}
+
+	return c.waitForOperations(ctx, ops)
+}
+
+func (c *Client) deleteInstanceTemplates(ctx context.Context) error {
+	ops := []Operation{}
 
 	if c.workerTemplate != "" {
 		op, err := c.deleteInstanceTemplate(ctx, c.workerTemplate)
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -22,6 +22,7 @@ linters:
     - bodyclose
     - errname
     - exportloopref
+    - gocyclo
     - godot
     - gofmt
     - gofumpt
@@ -42,3 +43,6 @@ linters-settings:
     exclude-functions:
       - (*go.uber.org/zap.Logger).Sync
       - (*google.golang.org/grpc.Server).Serve
+  gocyclo:
+    # Minimum code complexity to report, 15 is used by goreportcard.com.
+    min-complexity: 15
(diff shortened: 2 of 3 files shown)


katexochen:
> Some of the functions look fine to me.

@Nirusu Which functions are you talking about exactly? There is also the possibility to disable the check for a function, as I did in the second commit for [NewFromDefault](https://github.com/edgelesssys/constellation/pull/185/files#diff-284021df5d4a42f236e823e18c539a12849770f88d6b1f5a3c06d111b1b90aadR69). But Imo the checker is right for most of the listed functions.

thomasten:
The following is just my opinion. I think there's no right or wrong in this matter.

bootstrapper/cmd/bootstrapper/main.go:47:1 complexity 18 of func main
A dependency injection assembler function. You could split this, but I don't see a readability advantage.

bootstrapper/internal/kubernetes/k8sapi/install.go:81:1 complexity 22 of func (*osInstaller).extractArchive
This is one self-contained task. It would get harder to understand some details if split over multiple functions, but it wouldn't get any easier to understand the high level.

cli/internal/cmd/create.go:50:1 complexity 17 of func create
This looks super simple and clear. It's just UI. Logic is already separated. 

hack/image-measurement/main.go:82:1 complexity 16 of func (*libvirtInstance).createLibvirtInstance
Looks like an aggregate of similar actions.

I stopped here, but I think you can get the idea.

Bottom line: I tend to split functions if they involve very different actions or actions at different levels of abstraction. But it's ok and sometimes I prefer to have longer functions with similar kinds of actions instead of splitting them.

cli/internal/gcp/client/instances.go
```diff
@@ -172,10 +179,12 @@ func (c *Client) TerminateInstances(ctx context.Context) error {
 		c.controlPlaneInstanceGroup = ""
 		c.controlPlanes = make(cloudtypes.Instances)
 	}
-	if err := c.waitForOperations(ctx, ops); err != nil {
-		return err
-	}
-	ops = []Operation{}
+
+	return c.waitForOperations(ctx, ops)
```
thomasten, line 183: Won't this refactoring slow down the termination because it unnecessarily blocks here?

katexochen: It waited before, still waits and it has to wait, because the templates can't be deleted as long as the instance group manager that uses it still exist.

thomasten: I see. I didn't get the diff correctly and thought that there was only one wait before for all ops that now would be two waits.

PR #187, opened 2022-09-22
AB#2377 Repeat benchmarks on CVMs

chmey:
### Proposed change(s)

- Repeated K-Bench benchmarks on CVM for AKS and GKE
- Dropped netperf due to instability
- Reworded docs where necessary.

### Checklist

- [ ] Link to Milestone


chmey:
@m1ghtym0 do you think I could use a log scale for the `internode` graphs too? GKE is much higher than the other data points.

m1ghtym0:
> @m1ghtym0 do you think I could use a log scale for the `internode` graphs too? GKE is much higher than the other data points.

Yes, makes sense.
Did we also change something on the Constellation side? Use CVMs instead of TrustedLaunch for example? Trying to make sense of the changes in Constellation's performance.

chmey:
> > @m1ghtym0 do you think I could use a log scale for the `internode` graphs too? GKE is much higher than the other data points.
> 
> Yes, makes sense. Did we also change something on the Constellation side? Use CVMs instead of TrustedLaunch for example? Trying to make sense of the changes in Constellation's performance.

These tests use `confidentialVM: true` on both Azure and GCP :)

docs/docs/overview/performance.md
```diff
@@ -67,23 +68,22 @@ Deployments: Constellation has the lowest latency for all cases except for scali
...
+K-Bench provides benchmark tests for both, configured as `dp_network_internode` and `dp_network_intranode`. The tests use [`iperf`](https://iperf.fr/) to measure the bandwidth available.
 
 ##### Inter-node
 
-K-Bench has two benchmarks to evaluate the network performance between different nodes.
+Inter-node communication is the network transmission between different Kubernetes nodes.
 
-The first test (`dp_netperf_internode`) uses [`netperf`](https://hewlettpackard.github.io/netperf/) to measure the throughput. Constellation has a slightly lower network throughput than AKS and GKE.
-This can largely be attributed to its [network encryption](../architecture/networking.md).
+The first test (`dp_network_internode`) measures the throughput between nodes. Constellation on Azure and on GCP has a steady inter-node throughput of around 870 Mbps. While that's faster than the average throughput of AKS at 577 Mbps, GKE provides faster networking.
```
m1ghtym0, line 77: Please mention the throughput of GKE as well, feels weird that it's the only one missing the concrete value..

m1ghtym0: Same in the archived ofc

m1ghtym0 (changes requested):

docs/docs/overview/performance.md
```diff
@@ -22,13 +22,15 @@ We used the following configurations for the benchmarks.
 
 - Nodes: 3
 - Machines: `n2d-standard-2`
```
Nirusu, line 24:
```suggestion
- Machines: `n2d-standard-4`
```
I presume

docs/docs/overview/performance.md
```diff
@@ -22,13 +22,15 @@ We used the following configurations for the benchmarks.
...
+- CVM: `true`
 - Kubernetes version: `1.23.6-gke.2200`
```
Nirusu, line 26: That specific version with the GKE tag only applies to GKE, not Constellation.

docs/docs/overview/performance.md
```diff
@@ -22,13 +22,15 @@ We used the following configurations for the benchmarks.
...
 
 #### Constellation and AKS on Azure
 
 - Nodes: 3
-- Machines: `D2a_v4`
+- Machines: `DC4as_v5`
+- CVM: `true`
 - Kubernetes version: `1.23.5`
 - Region: `North Europe`
 - Zone: `2`
```
Nirusu, line 36: Not sure if you pinned the zones for the benchmark, but AFAIK Constellation on Azure can scale to multiple zones by default (there’s no pinned zone), so this information could be slightly wrong for Constellation.

docs/docs/overview/performance.md
chmey: Thanks catching that. If `n2d-standard-4` is the default for Constellation, I will have to repeat the benchmarks on GKE with that instance type. I used `-2` there.

docs/docs/overview/performance.md
chmey: Yeah, this is mostly a result because throughout the edits, we kind of merged the setup specifications between cloud provided Kubernetes and Constellation.

chmey: I think I might exclude the Kubernetes version here as it has probably little impact on the performance. But keep the rest.

docs/docs/overview/performance.md
chmey: No, I did not pin it. In fact I used `westeurope` (need to update the quoted section above) and no pinned zone. 

But this brings up another point: IIRC Azure does not support `DC4as_v5` or maybe any CVMs in Zone 3 of `westeurope`, don't they? So for this run, probably doesn't matter. But for future benchmarks, I'd definitely pin it. Thanks.

docs/docs/overview/performance.md
Nirusu: Yeah, the earlier pre-release versions of Constellation had `n2d-standard-2` as default so it was definitely valid in your previous benchmark. 

Unfortunately, we had to kick out support for `n2d-standard-2` (1c2t) prior to release as updates to Linux/CoreOS included mitigations against Retbleed and caused Linux to disabled SMT, which then caused to fail the K8s pre-flight check.

So `n2d-standard-2` is not supported anymore in 2.0.0 and `n2d-standard-4` is the bottom/default, so comparing it with the same instance type on GKE would be the best I guess. 

I just hope it's not too much effort to recreate the benchmarks...

docs/docs/overview/performance.md
Nirusu: Okay, if Azure only has VMs of this type in one zone in this region then I suppose you're lucky and that should be fine then for now :)

chmey:
- Repeated the benchmarks one more time on GKE with the correct instance size.
- Added the changes you requested and fixed experiment description: Azure Region and Zone, remove Kubernetes version.

Please take a look.

docs/docs/overview/performance.md
```diff
@@ -74,7 +72,7 @@ K-Bench provides benchmark tests for both, configured as `dp_network_internode`
 
 Inter-node communication is the network transmission between different Kubernetes nodes.
 
-The first test (`dp_network_internode`) measures the throughput between nodes. Constellation on Azure and on GCP has a steady inter-node throughput of around 870 Mbps. While that's faster than the average throughput of AKS at 577 Mbps, GKE provides faster networking.
+The first test (`dp_network_internode`) measures the throughput between nodes. Constellation has an inter-node throughput of around 816 Mbps on Azure to 872 Mbps on GCP. While that's faster than the average throughput of AKS at 577 Mbps, GKE provides faster networking at 9.55 Gbps.
```
m1ghtym0, line 75: Wow, the difference between Azure and GCP is insane... that is almost 20x

m1ghtym0 (approved):
LGTM, I'm just a bit worried that the throughput difference between AKS and GCP is correct.

chmey:
> LGTM, I'm just a bit worried that the throughput difference between AKS and GCP is correct.

Understandable - the difference is huge. I have repeated the networking benchmarks once again for AKS with the config (same zone, CVM instances) and it accounted for the same. 

Here is an extract from the inter-node logs:

```
s1:  [  5]   0.00-90.00  sec  6.06 GBytes   578 Mbits/sec   38             sender
s1:  [  5]   0.00-90.00  sec  6.06 GBytes   578 Mbits/sec                  receiver
```

btw, the data points used in the evaluation are archived in an S3 bucket right now, which I/we can always cite from.

It should be more transparent once benchmarking can be done as a CI workflow (#22).

PR #191, opened 2022-09-23
AB#2439 Containerized libvirt

daniel-weisse:
### Proposed change(s)

- Update qemu config to allow setting the libvirt connection URI
  - By default this is left empty, in which case we deploy a containerized libvirt instance

### Additional info
- The containerized libvirt instance exposes its connection over TCP on port `16599`
  - Once the container is up, users can connect to the libvirt instance using `virsh -c "qemu+tcp://localhost:16599/system"`

### Checklist

- [ ] Link to Milestone


cli/internal/libvirt/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM fedora@sha256:486fd5578f93fbc57a519e34ad4b7cac927c3f8a95409baedf0c19e9f287c207 AS deploy
```
3u13r, line 1: I think we should add a short README.md here with use and build instruction like with the qemu metadata docker image.

3u13r (changes requested):

cli/internal/libvirt/libvirt.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+		return err
+	}
+
+	docker, err := docker.NewClientWithOpts(docker.FromEnv, docker.WithAPIVersionNegotiation())
+	if err != nil {
+		return err
+	}
+	defer docker.Close()
+	if err := docker.ContainerRemove(ctx, string(name), types.ContainerRemoveOptions{Force: true}); err != nil {
```
3u13r, line 97: Does `docker.ContainerRemove()` not return e.g. a NotFoundErr that can be checked, or why do we use a file as proxy if the container has already been deleted/does not exist?

cli/internal/libvirt/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+# Prevent cgroup issues on Fedora and configure libvirt
+RUN echo "cgroup_controllers = []" >> /etc/libvirt/qemu.conf && \
+    echo "listen_tls = 0" >> /etc/libvirt/libvirtd.conf && \
+    echo "listen_tcp = 1" >> /etc/libvirt/libvirtd.conf && \
+    echo "tcp_port = \"16599\"" >> /etc/libvirt/libvirtd.conf && \
+    echo "listen_addr = \"localhost\"" >> /etc/libvirt/libvirtd.conf && \
+    echo "auth_tcp = \"none\"" >> /etc/libvirt/libvirtd.conf
+
+COPY ./cli/internal/libvirt/start.sh /start.sh
```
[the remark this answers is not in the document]
daniel-weisse: Use the constellation folder as build context

cli/internal/libvirt/libvirt.go
daniel-weisse: The file is not used as proxy to check if the container is still running, but to ensure we remove the correct container during terminate.
Since the terminate command does not know about the name given to cluster resources during create, we need some additional way to keep this information.
Alternatively we could maybe parse the `terraform.tfvars` file to get the name, not sure if this is the best option though

3u13r: I assume that there's no good way of using terraform itself to handle this container? And simply hard coding the name also has drawbacks. Then my best solution would be to to use something like terraform state show to natively get the name variable.

Nirusu: I would also agree with doing all this in Terraform if possible. 

This Docker creation in Go seems kinda weird when the metadata-grpc server is created in Terraform, but libvirt is created in Go code. I would prefer keeping this consistent and not scattered across two places.

daniel-weisse: Does not work with Terraform since we require the libvirt container to start before everything else does.
However, setting up the libvirt provider requires the libvirt connection URI. This needs to be set in the main terraform file, meaning we need to know the connection URI when we start applying the terraform templates.
This is not possible if the libvirt container itself is deployed using terraform

derpsteb:
Tested on ubuntu 22.04. Works fine. User errors I encountered: Not tagging the docker image correctly and not adding the connection string when inspecting resources with `virsh`. 

One odd thing: I had to change my iptables `filter` table. The `FORWARD` chain was set to `DROP`. Setting it to accept made things work. I don't remember changing iptable rules manually on this machine. So maybe it would make sense to check the settings there? Just thinking out loud.

cli/internal/libvirt/libvirt.go
Nirusu: Mh, there is no dependency management with Terraform and the Docker provider?
Would [defining a different target](https://groups.google.com/g/terraform-tool/c/wKGyI-C03aA/m/zflLYcQLBQAJ) and calling Terraform first on the libvirt target, then the rest of the deployment work?

daniel-weisse: While theoretically possible, this would require a major rewrite to our terraform module.
And I'm unsure if the benefit of not having a separate go module to start the libvirt container is worth the added complexity of the terraform module and qemu terraform template.

Starting the libvirt container as its own instance also gives us the benefit of using the same terraform deployment for containerized libvirt as well as local libvirt installations

3u13r: The last suggestion I thought about was to simply have 2 seperate Terraform files. Since this would necessitate 2 different Terraform workspaces, this is also a sub-optimal solution. Therefore, I'm fine with keeping this part as is. At least until we come up with a solution with no downsides.

3u13r (approved):
LGTM

.github/docs/qemu.md
```diff
@@ -87,7 +94,33 @@ Then restart libvirt
...
+Make sure your rules are not dropping forwarded packages.
+
+List and check your rules:
+
+```shell
+$ sudo iptables -S
+-P INPUT ACCEPT
+-P FORWARD DROP
+-P OUTPUT ACCEPT
+-N DOCKER
```
Nirusu, line 118: A bit confusing that you only need to execute the first line and the rest is the output, but I blame that more on `iptables` itself.

cli/internal/libvirt/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+A container of the image is automatically started by the CLI.
+You can also run the image manually using the following command:
+
+```shell
+docker run -it --rm \
+    --network host \
+    --device /dev/kvm \
+    --privileged true \
+    ghcr.io/edgelesssys/constellation/libvirt:latest
```
Nirusu, line 30:
```suggestion
docker run -it --rm \
    --network host \
    --privileged \
    ghcr.io/edgelesssys/constellation/libvirt:latest
```

Untested, but IIRC --privileged already gives full access to devices from the host.

cli/internal/libvirt/libvirt.go
```diff
@@ -0,0 +1,104 @@
...
+			NetworkMode: container.NetworkMode("host"),
+			Privileged:  true,
+			AutoRemove:  true,
+			Mounts: []mount.Mount{
+				{
+					Type:   mount.TypeBind,
+					Source: "/dev/kvm",
+					Target: "/dev/kvm",
+				},
+			},
```
Nirusu, line 58: Same here, privileged launch should make the manual mount obsolete IIRC.

Nirusu (commented):
Don't forget to update the TODO image definitions later.

PR #195, opened 2022-09-26
Create mkosi image build pipeline

malt3:
### Proposed change(s)

- Create mkosi image build pipeline

### Additional info

- Feel free to review the code already
- Feel free to test the images and report any issues
- You can see a pipeline + images for AWS, GCP, Azure and QEMU here: https://github.com/edgelesssys/constellation/actions/runs/3283621979

### Checklist

- [x] [Secure Boot on Azure can currently not be disabled when creating the cluster with terraform](https://github.com/hashicorp/terraform-provider-azurerm/issues/18747)
  - [x] get fix merged upstream
  - [x] [await new terraform provider release](https://github.com/hashicorp/terraform-provider-azurerm/milestone/206)
- [x] Link to Milestone
- [x] Add license header to all scripts inside of the mkosi folder
- [x] Remove CoreOS build folder
- [x] Deprecate CoreOS config
- [x] Remove all traces of the word CoreOS from the codebase
- [x] e2e + stability tests
- [x] Decide how to work around [replication not being available](https://learn.microsoft.com/en-us/azure/confidential-computing/create-confidential-vm-from-compute-gallery#confidential-vm-images) for Azure ConfidentialVM images
    > This image version can be replicated within the source region but cannot be replicated to a different region or across subscriptions currently.

### To be done after this PR

- [ ] Prepare release pipeline, create production secure boot PKI and devise secure way to handle the keys


image/mkosi/secure-boot/generate_nvram_vars.sh
```diff
@@ -0,0 +1,107 @@
...
+    if test -f "${PKI}/image_SB_VARS.fd"; then
+        echo "OVMF vars already generated."
+        return
+    fi
+    mkdir -p "${PKI}"
+
+    sed \
+        -e 's/^-----BEGIN CERTIFICATE-----$/4e32566d-8e9e-4f52-81d3-5bb9715f9727:/' \
+        -e '/^-----END CERTIFICATE-----$/d' \
+        ${PKI}/db.crt > ${PKI}/PkKek1.oemstr
```
Nirusu, line 22: ??? 

If you keep this (fix for Ubuntu would be great :P) please comment what this does. Also, I could imagine that this here is why this fails on Ubuntu (different PKI?).

image/mkosi/secure-boot/generate_nvram_vars.sh
```diff
@@ -0,0 +1,107 @@
...
+        --vcpus 1 \
+        --osinfo detect=on,require=off \
+        --disk "${image_path},format=raw" \
+        --boot "machine=q35,menu=on,loader=/usr/share/OVMF/OVMF_CODE_4M.ms.fd,loader.readonly=yes,loader.type=pflash,nvram.template=/usr/share/OVMF/OVMF_VARS_4M.ms.fd,nvram=${BASE_DIR}/image.nvram,loader_secure=yes" \
+        --features smm.state=on \
+        --noautoconsole
+    echo 'connect using'
+    echo 'virsh console constell-nvram-gen'
+    echo ''
+    echo 'either load db cert using mokmanager or enroll full PKI using firmware setup'
```
Nirusu, line 69: Grammar / Spelling is bad here.

image/mkosi/secure-boot/generate_nvram_vars.sh
```diff
@@ -0,0 +1,107 @@
...
+    echo '    Choose "Commit Changes and Exit"'
+    echo '    Go to "DB Options">"Enroll Signature">"Enroll Signature using File"'
+    echo '    Select the following cert: /EFI/loader/keys/auto/db.cer'
+    echo '    Choose "Commit Changes and Exit"'
+    echo '    Repeat the last step for the following certs:'
+    echo '    > /EFI/loader/keys/auto/MicWinProPCA2011_2011-10-19.crt'
+    echo '    > /EFI/loader/keys/auto/MicCorUEFCA2011_2011-06-27.crt'
+    echo '    Reboot and continue this script.'
+    echo ''
+    echo 'Press ENTER to continue after you followed one of the guides from above.'
```
Nirusu, line 97: I don't have the visual output in front of me, but I can imagine this is a wall of text that requires careful reading.

I think this could be a great candidate for some ANSI escape code formatting for highlighting file names and differentiating between "mokmanager" and "Enroll full PKI" steps.

image/mkosi/secure-boot/genkeys.sh
```diff
@@ -0,0 +1,65 @@
...
+        cert-to-efi-sig-list -g "${uuid}" "${key}.crt" "${key}.esl"
+    done
+
+    for key in MicWinProPCA2011_2011-10-19.crt MicCorUEFCA2011_2011-06-27.crt MicCorKEKCA2011_2011-06-24.crt; do
+        curl -sL "https://www.microsoft.com/pkiops/certs/${key}" --output "${key}"
+        sbsiglist --owner 77fa9abd-0359-4d32-bd60-28f4e78f784b --type x509 --output "${key%crt}esl" "${key}"
+    done
+
+    # Optionally add Microsoft Windows Production CA 2011 (needed to boot into Windows).
+    cat MicWinProPCA2011_2011-10-19.esl >> db.esl
```
Nirusu, line 44: I don't think we want to boot into Windows for our VMs.

Is this actually required for something or leftovers from the original templates? If required, please adjust comments.

image/mkosi/secure-boot/templates/dev_KEK.conf
```diff
@@ -0,0 +1,20 @@
...
+ distinguished_name     = req_distinguished_name
+ x509_extensions        = v3_req
+ req_extensions         = v3_req
+ prompt                 = no
+
+ dirstring_type = nobmp
+
+ [ req_distinguished_name ]
+ C                      = DE
+ ST                     = Nordrhein Westfalen
```
Nirusu, line 12:
```suggestion
 ST                     = Nordrhein-Westfalen
```
But this is a very minor nit and if you have to regenerate the whole PKI just for this, I wouldn't do that.

image/mkosi/upload/upload_azure.sh
```diff
@@ -0,0 +1,131 @@
...
+        --resource-group "${AZURE_RESOURCE_GROUP_NAME}" \
+        -l "${AZURE_REGION}" \
+        --gallery-name "${AZURE_GALLERY_NAME}" \
+        --gallery-image-definition "${AZURE_IMAGE_DEFINITION}" \
+        --gallery-image-version "${AZURE_IMAGE_VERSION}" \
+        --target-regions ${AZURE_REPLICATION_REGIONS} \
+        ${AZURE_CVM_ENCRYPTION_ARGS} \
+        --replica-count 1 \
+        --replication-mode Full \
+        --os-snapshot "$(az disk list --query "[?name == '${AZURE_DISK_NAME}' && resourceGroup == '${AZURE_RESOURCE_GROUP_NAME^^}'] | [0].id" --output tsv)"
```
Nirusu, line 123: (Whole file)
Personal opinion but for some more complex scripts I would maybe suggesting switching to Python or similar instead of bash for easier maintenance and readability.

image/mkosi/mkosi.conf.d/mkosi.conf
```diff
@@ -0,0 +1,17 @@
+[Distribution]
+Distribution=fedora
+Release=36
+
+[Output]
+Format=gpt_squashfs
```
Nirusu, line 6: Is squashfs going to stay? Is compression by default enabled with gzip (as I thought it was last time)?

If so, switch to zstd if the bootloader does support that. Better compression in total and better decompression performance, which might slightly boost I/O performance from the r/o only device.

If no compression is there, carry on.

image/mkosi/mkosi.skeleton/etc/profile.d/constellation.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+export PATH=/run/state/bin:${PATH}
+export KUBECONFIG=/etc/kubernetes/admin.conf
+alias k=kubectl
```
Nirusu, line 4:
```suggestion
```
Nit but I wouldn't define this on the read-only disk. Maymay maybe only for debug images.

image/mkosi/README.md
```diff
@@ -0,0 +1,187 @@
...
+- Login to Azure (see [here](https://docs.microsoft.com/en-us/cli/azure/authenticate-azure-cli))
+- Prepare secure boot PKI (see `secure-boot/genkeys.sh`)
+- [Prepare virtual machine guest state (VMGS) with customized NVRAM or use existing VMGS blob](#azure-secure-boot)
+
+```sh
+# set these variables
+export AZURE_GALLERY_NAME=
+export AZURE_IMAGE_DEFINITION=
+export AZURE_IMAGE_VERSION=
+export AZURE_VMGS_PATH=
```
Nirusu, line 171: Maybe add some placeholders for them to know what format is required here (e.g. just the gallery name, or the whole gallery URL (sub-)path?).

image/mkosi/mkosi.files/mkosi.azure.conf
```diff
@@ -0,0 +1,3 @@
+[Output]
+KernelCommandLine=constel.csp=azure
```
Nirusu, line 2: I don't know what mkosi does pass to the kernel by default, but I presume they don't automatically add `mitigations=auto,nosmt`as CoreOS does, right? If yes, should we add that for parity?

image/mkosi/secure-boot/templates/dev_KEK.conf
malt3: I will leave it as it is since a PKI for testing is already included in github actions.
Interestingly, I found conflicting results on how to format the state field.
[This intermediate CA used by German Telekom](https://de.ssl-tools.net/subjects/fae8baa6c5fe36847b98f212856735704f86280c) uses `Nordrhein Westfalen` while leaf certificates used for https://www.deutschepost.de use `Nordrhein-Westfalen`.

image/mkosi/mkosi.conf.d/tools.conf
```diff
@@ -0,0 +1,9 @@
+[Content]
+Packages=
+           nano,
+           nano-default-editor,
```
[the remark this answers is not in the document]
katexochen: :heart:

malt3:
> Are these *.vhd blobs really required in the git repo? They are huge for git. If they are, consider using git-lfs for them.

While we can get rid of the *default* VMGS blobs and recreate them on demand, the VMGS blobs that trust our secure boot certs must be stored somewhere. We could move them to a public S3 bucket instead. I'll add this as a TODO. 

> Big ask, but how much work is getting SELinux working kinda properly with this? Some security guidelines do have this as a standard, and it would be cool if could get this in somehow at some point.

I thought about this and we discussed this back when we picked CoreOS (which supports SELinux out of the box). Since we are still based on Fedora, all of the system services should just work if we ensure that our root filesystem keeps SELinux labels from `dnf --installroot`. In addition, we would need to label the various parts of the state partition on the fly and mount it with SELinux in mind. 
If this basic enablement is done, we could enable SELinux and see what breaks. We would definitely have to define reasonable policies for the bootstrapper.
Sadly, this is only where the work begins as SELinux in the container context probably requires more policies. The following part is written to the best of my understanding. I do not have experience in rolling out SELinux for containers.
I expect at least some of our default deployments to break. As an example, Cilium should break as it depends on breaking out of the container and doing a lot of high privilege stuff.
If our deployments are configured to use meaningful policies, we have to decide what to do with our downstream users. By default, enabling SELinux might break a lot of what users tend to do with containers (host mounts is an example).
We would have to either burden our users to define policies for everything OR exclude containers from SELinux (which kind of defeats the purpose)..
I do not fully understand what happens if we set SELinux to be enforcing. What is the default policy for pods that have no securityContext? It could also be that everything container related just works by default (until you add a securityContext to your pods).

Some other thoughts on this:

- GKE does not support SELinux: 
    >  Note: [SELinux](https://selinuxproject.org/page/Main_Page) is not supported on GKE. Use [AppArmor](https://cloud.google.com/container-optimized-os/docs/how-to/secure-apparmor) instead. - https://cloud.google.com/kubernetes-engine/docs/concepts/security-overview
- AKS does support SELinux

## My proposal

SELinux is currently disabled in our CoreOS images so delaying SELinux further is not a regression. We should instead experiment with SELinux in a separate PR and decide if normal user deployments would be affected by that change.

Nirusu:
Are the VMGS required by each Azure user or just for development? In case it's just for development I would just opt for git-lfs, it's definitely the easiest solution and already included in our plan (up to 5 GB in total across all repos I believe). If it's required for the user: I am personally not sure about using public S3 buckets since it's always at the risk of abuse by malicious people spamming requests on your costs, but I guess the details can be discussed later.

Regarding SELinux: Okay, of course it does not to be part of this PR, but I would definitely evaluate this at some point, since we have already seen this (or other hardening solutions) to be part of compliance checks. Was just curious if the change to mkosi itself makes anything easier or not to enable enforced SELinux. 

From what I can see, AppArmor seems easier to integrate for containers with default profiles existing for Docker / containerd and others but integrating AppArmor in a Fedora environment itself doesn't sound like the most fun task.

But yeah, doesn't have to be done right now, we can of course just do it later.

image/mkosi/secure-boot/generate_nvram_vars.sh
malt3: This is actually required for the ovmf vars generator regardless of the platform. The magic value is used as an efivar and used by one of the efi binaries to detect the wanted secure boot key. Will document. And maybe remove if this does not work.

image/mkosi/secure-boot/genkeys.sh
malt3: Keeping the *default* signing keys used in production environments is wanted. Pinging @thomasten for confirmation.

image/mkosi/mkosi.conf.d/tools.conf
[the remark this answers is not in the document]
daniel-weisse: :disappointed:

image/mkosi/mkosi.skeleton/etc/profile.d/constellation.sh
malt3: Why would you not define aliases on read-only disks? I don't see how this would interfere with anything. It just makes debugging a lot easier in my opinion.

image/mkosi/mkosi.skeleton/usr/lib/systemd/system/configure-constel-csp.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Configures constellation cloud service provider environment variable
+
+[Service]
+Type=oneshot
+ExecStart=/bin/bash -c "CSP=$(< /proc/cmdline tr  ' ' '\n' | grep constel.csp | sed 's/constel.csp=//'); echo CONSTEL_CSP=$CSP >> /run/constellation.env"
```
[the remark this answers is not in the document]
malt3: I have a bad feeling about having our go binaries directly parse the kernel commandline but maybe this is okay after all. 

a) could work. Will test it.
b) will implement if a) doesn't work.

image/mkosi/secure-boot/genkeys.sh
thomasten: This script is used for dev/debug images only?
What are the pros/cons of keeping the default keys?

malt3: This script is used by developers to create test keys. The production keys do not exist yet but the secure boot database of the production PKI should be modeled in the exact same way (and also include the same microsoft certs). The only difference should be safer storage of the keys.

> What are the pros/cons of keeping the default keys?

Only con is that you can boot windows using those keys in addition to our own OS. Only pro is to have the same set of keys loaded on all platforms as we can only add our key on Azure (the default keys will always be there).

image/mkosi/upload/upload_azure.sh
malt3: I'll try to refactor this a bit.

image/mkosi/secure-boot/genkeys.sh
thomasten: > pro is to have the same set of keys loaded on all platforms

sounds good

image/mkosi/mkosi.skeleton/etc/profile.d/constellation.sh
Nirusu: It's not interfering, but it's more a general UX argument for me that:

- If I like the alias, I can set it myself easily when debugging stuff in the shell  (if it's not defined on the read-only disk)
- If I don't like the alias, I have to unset this manually every single time (if it's defined on the read-only disk)

I don't have a strong opinion on this, since it's just an alias after all. But personally I prefer keeping the environment as clean as is from personal preferences (even if I left a ❤️ for nano as default editor above previously :P)

image/mkosi/upload/upload_azure.sh
malt3: I tried to simplify this a bit. Rewriting this is python is not a good solution IMO as this script mostly calls `az` with different arguments.

malt3:
Not sure how to fix the vale issue @m1ghtym0

image/mkosi.skeleton/usr/lib/udev/google_nvme_id
```diff
@@ -0,0 +1,247 @@
+
+
```
daniel-weisse, line 2: Are these newlines at the top part of the script, or were they added in by accident?
They also appear in the other gce scripts

internal/config/config.go
```diff
@@ -246,7 +256,8 @@ func Default() *Config {
 				LibvirtURI:            "",
 				LibvirtContainerImage: versions.LibvirtImage,
 				Measurements:          copyPCRMap(qemuPCRs),
-				EnforcedMeasurements:  []uint32{11, 12},
+				EnforcedMeasurements:  []uint32{4, 8, 9, 11, 12, 13, 15},
+				NVRAM:                 "production",
```
daniel-weisse, line 260: Not sure what our policies are for the main branch, but since we don't currently ship production NVRAM I feel like this should be set to `testing` by default.

daniel-weisse (commented):
Nice :small_airplane:

image/README.md
```diff
@@ -1,113 +1,187 @@
...
+
+</details>
+
+<details>
+<summary>Azure</summary>
+
+- Install `az` and `azcopy` (see [here](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli))
+- Login to Azure (see [here](https://docs.microsoft.com/en-us/cli/azure/authenticate-azure-cli))
+- Prepare secure boot PKI (see `secure-boot/genkeys.sh`)
+- [Prepare virtual machine guest state (VMGS) with customized NVRAM or use existing VMGS blob](#azure-secure-boot)
```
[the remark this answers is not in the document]
malt3: You need a secure boot PKI on any platform. If you want to build and start your image on QEMU, you would need to self sign it using your own PKI OR disable secure boot.
The same goes for GCP.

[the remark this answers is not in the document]
malt3: Multiple components inside the image are signed using the db.key. Image creation fails without a PKI.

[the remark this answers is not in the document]
daniel-weisse: But you shouldn't need it to just pack and upload an image to Azure, correct?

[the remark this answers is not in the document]
malt3: I see. If you have a prebuilt image, you can upload it without a PKI. I'll try and move / reword that sentence.

m1ghtym0:
> Not sure how to fix the vale issue @m1ghtym0

I don't understand why vale is failing. I only see the two warnings.  Is reviewdog somehow handling warnings as errors? If that's the case I can disable these warnings.

thomasten:
> > Not sure how to fix the vale issue @m1ghtym0
> 
> I don't understand why vale is failing. I only see the two warnings. Is reviewdog somehow handling warnings as errors? If that's the case I can disable these warnings.

Seems we updated the Vale action to a new major version. In addition to treating Warnings as Errors (which may be configured somehow with the new version), it also includes a vale version that produces some FPs. So I'd suggest to downgrade the action to the previous major version.

docs/styles/Vocab/constellation/accept.txt
```diff
@@ -35,6 +35,7 @@ systemd
 [Uu]nencrypted
 unspoofable
 updatable
+VMs
```
thomasten, line 38:
```suggestion

```
This is one of the FPs. Plural of acronyms should actually be fine.

image/mkosi.skeleton/usr/lib/udev/google_nvme_id
malt3: Thanks. :)

internal/config/config.go
malt3: Every release CLI will have to set this to production then. But I guess this can be automated.. So I'm fine with that :+1:

image/README.md
```diff
@@ -1,113 +1,193 @@
...
-```
+After that, you can build the image with:
 
-## Build an image
+```sh
+# OPTIONAL: to create a debug image, export the following line
+# export BOOTSTRAPPER_BINARY=$(realpath ${PWD}/../../build/debugd)
+# OPTIONAL: specify custom path to secure boot PKI
+# export PKI=/path/to/pki/folder
+sudo make -j $(nproc)
```
daniel-weisse, line 73: This option seems to not be respected by all of the mkosi workflow:

```shell
$ sudo PKI=./pki_testing/ make -j $(nproc)                                                                                                                                                                                                               
mkdir -p /home/daniel/test/constellation/image/mkosi.extra/usr/bin                                                                                                                                                                                        
# for auto enrollment using systemd-boot (not working yet)                                                                                                                                                                                                    
mkdir -p "/home/daniel/test/constellation/image/mkosi.extra/boot/loader/keys/auto"                                                                                                                                                                        
mkdir -p /home/daniel/test/constellation/image/mkosi.extra/usr/sbin
[...]
Can't load key from file '/home/daniel/test/constellation/image/pki/db.key'
809BB8AC1B7F0000:error:80000002:system library:BIO_new_file:No such file or directory:crypto/bio/bss_file.c:67:calling fopen(/home/daniel/test/constellation/image/pki/db.key, r)
809BB8AC1B7F0000:error:10000080:BIO routines:BIO_new_file:no such file:crypto/bio/bss_file.c:75:
```

image/secure-boot/azure/launch.sh
```diff
@@ -0,0 +1,103 @@
...
+      shift # past argument
+      shift # past value
+      ;;
+    -g|--gallery)
+      CREATE_FROM_GALLERY=YES
+      CREATE_FROM_DISK=NO
+      shift # past argument
+      ;;
+    -d|--disk)
+      CREATE_FROM_DISK=YES
```
daniel-weisse, line 22: Unused variable

image/secure-boot/generate_nvram_vars.sh
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env bash
+# Copyright (c) Edgeless Systems GmbH
+#
+# SPDX-License-Identifier: AGPL-3.0-only
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+BASE_DIR=$(realpath "${SCRIPT_DIR}/..")
+PKI=${BASE_DIR}/pki
```
daniel-weisse, line 8: Unused variable

image/secure-boot/genkeys.sh
```diff
@@ -0,0 +1,68 @@
...
+    PKI=${BASE_DIR}/pki
+fi
+if [ -z "${PKI_SET}" ]; then
+    PKI_SET=dev
+fi
+
+gen_pki () {
+    # Only use for non-production images.
+    # Use real PKI for production images instead.
+    count=`ls -1 ${PKI}/*.{key,crt,cer,esl,auth} 2>/dev/null | wc -l`
```
daniel-weisse, line 25:
```suggestion
    count=$(ls -1 ${PKI}/*.{key,crt,cer,esl,auth} 2>/dev/null | wc -l)
```

daniel-weisse: Shellcheck also suggests using `find` instead of `ls`, not sure if it fits here, though

image/secure-boot/signed-shim.sh
```diff
@@ -0,0 +1,50 @@
...
+    exit 1
+fi
+
+# SOURCE is the URL used to download the signed shim RPM
+SOURCE=https://kojipkgs.fedoraproject.org/packages/shim/15.6/2/x86_64/shim-x64-15.6-2.x86_64.rpm
+# EXPECTED_SHA512 is the SHA512 checksum of the signed shim RPM
+EXPECTED_SHA512=971978bddee95a6a134ef05c4d88cf5df41926e631de863b74ef772307f3e106c82c8f6889c18280d47187986abd774d8671c5be4b85b1b0bb3d1858b65d02cf
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+BASE_DIR=$(realpath "${SCRIPT_DIR}/..")
+MKOSI_EXTRA=${BASE_DIR}/mkosi.extra
```
daniel-weisse, line 21: Unused variable

image/README.md
malt3: Thanks. I think we will need to symlink the used pki to `./pki` for mkosi to work. Will update the docs.

3u13r (approved):
LGTM, regarding locally building images for GCP/QEMU.

PR #198, opened 2022-09-27
Mini Constellation

daniel-weisse:
### Proposed change(s)

- Add the `constellation mini` command
  - Use `constellation mini up` to generate a config file, create a 1 control-plane, 1worker node cluster and initialize it
  - Use `constellation mini down` to terminate the cluster and clean up your workspace
 

### Additional info
- Users may specify the `--config` flag when using `mini up` to create a cluster using the given config instead of automatically generating a minimal setup config
- Currently users need to specify the image to use for their cluster using the `--image` flag. Before merging, this option will be replaced by automatically downloading the appropriate QEMU image of a given release.
  - Users who wish to use a different image may do so by creating a config and setting the image as desired

### Checklist

- [x] Link to Milestone


Nirusu:
Still draft so not reviewing it in-depth, but given that Mini Constellation is supposed to be simple to use with a nice UX, I would maybe suggest:

* Check if the user has Docker installed and the daemon is reachable
* Check if the user has hardware virtualization enabled / is supported / KVM exists
* Refuse to run on systems with less than 4 (±1-2 GB reserve) of memory existent in total.
* Display the user a warning if the user has already high memory usage during creation time to avoid them running into the OOM killer quickly. 2 VMs each allocating up to 2 GB of RAM can be a tight memory budget on a machine with 8 GB RAM that runs a desktop OS and does not have any memory saving mechanisms (compression, same-page merging, etc.).

cli/internal/cmd/miniup.go
```diff
@@ -0,0 +1,268 @@
...
+		return config, nil
+	}
+
+	// download image to current directory if it doesn't exist
+	imagePath := "./constellation.qcow2"
+	if _, err := os.Stat(imagePath); err == nil {
+		cmd.Printf("Using existing image at %s\n\n", imagePath)
+	} else if errors.Is(err, os.ErrNotExist) {
+		cmd.Println("Downloading image to ./constellation.qcow2")
+		if err := installImage(cmd.Context(), cmd.OutOrStdout(), versions.ConstellationQEMUImageURL, "./constellation.qcow2"); err != nil {
```
malt3, line 165:
```suggestion
		if err := installImage(cmd.Context(), cmd.OutOrStdout(), versions.ConstellationQEMUImageURL, imagePath); err != nil {
```

malt3:
Testing works great (using a local image).

malt3 (approved):
LGTM

cli/internal/cmd/miniup.go
```diff
@@ -0,0 +1,268 @@
...
+	if err != nil {
+		return fmt.Errorf("creating cluster: %w", err)
+	}
+	cmd.Println("Cluster successfully created.")
+	connectURI := config.Provider.QEMU.LibvirtURI
+	if connectURI == "" {
+		connectURI = libvirt.LibvirtTCPConnectURI
+	}
+	cmd.Println("Connect to the VMs by executing:")
+	cmd.Printf("\tvirsh -c %s\n\n", connectURI)
```
Nirusu, line 79: This implies `virsh` is present on the host system. But it could be replaced by a `docker exec` to the libvirt container.

cli/internal/cmd/miniup.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	return nil
+}
+
+// checkSystemRequirements checks if the system meets the requirements for running a mini Constellation cluster.
+// We do so by verifying that the host:
+// - arch/os is linux/amd64.
+// - has access to /dev/kvm.
+// - has more than 4 CPU cores.
+// - has more than 4GB of memory.
```
Nirusu, line 93: Already discussed, just as a reminder: Also check for free disk space available.

internal/versions/versions.go
```diff
@@ -57,6 +57,10 @@ const (
 	NodeMaintenanceOperatorCatalogImage = "ghcr.io/edgelesssys/constellation/node-maintenance-operator-catalog"
 	NodeMaintenanceOperatorVersion      = "v0.13.1-alpha1"
 
+	// ConstellationQEMUImageURL is the artifact URL for QEMU qcow2 images.
+	// TODO: Replace with actual artifact URL once we have one (S3, Google Bucket or similar).
+	ConstellationQEMUImageURL = "http://localhost/files/constellation.qcow2"
```
Nirusu, line 62: Do we want to change this before the merge? Since we already want to prepare the release this week - next week.

cli/internal/cmd/miniup.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+		return config, nil
+	}
+
+	// download image to current directory if it doesn't exist
+	imagePath := "./constellation.qcow2"
+	if _, err := os.Stat(imagePath); err == nil {
+		cmd.Printf("Using existing image at %s\n\n", imagePath)
+	} else if errors.Is(err, os.ErrNotExist) {
+		cmd.Println("Downloading image to ./constellation.qcow2")
```
Nirusu, line 164:
```suggestion
		cmd.Println("Downloading image to", imagePath)
```

It's practice it's the same output, but still in case this is edited in the future / becomes user-configurable...

cli/internal/cmd/minidown.go
```diff
@@ -0,0 +1,35 @@
...
+		Args:  cobra.ExactArgs(0),
+		RunE:  runDown,
+	}
+
+	return cmd
+}
+
+func runDown(cmd *cobra.Command, args []string) error {
+	err := runTerminate(cmd, args)
+	if removeErr := os.Remove(constants.MasterSecretFilename); removeErr != nil && !os.IsNotExist(removeErr) {
```
Nirusu, line 31: Be careful: Don't remove the master secret unless we know it's actually from a Mini Constellation. Running `./constellation mini down` in an already existing workspace will nuke the master secret.

cli/internal/cmd/miniup.go
```diff
@@ -0,0 +1,268 @@
...
+			return nil, err
+		}
+		if config.GetProvider() != cloudprovider.QEMU {
+			return nil, errors.New("invalid provider for mini constellation cluster")
+		}
+		return config, nil
+	}
+
+	// download image to current directory if it doesn't exist
+	imagePath := "./constellation.qcow2"
```
Nirusu, line 160: Define this as const?

Nirusu (commented):
I did push a change the change to the libvirt container that should fix the KVM permission issues on Debian-based distros (and maybe others). Haven't rebuild the image yet, would be great if you can do that :)

cli/internal/cmd/miniup.go
daniel-weisse: Leftover code from when it was user-configurable (It should no longer be)

internal/versions/versions.go
daniel-weisse: We don't have any images yet, so we would just change it form one placeholder to another until the next release.
Although if you want, we could decide now under which URL we plan to upload the image already set it here.

cli/internal/cmd/minidown.go
daniel-weisse: First of, running `constellation mini down` in the wrong working directory is mostly user error, your OS also doesn't prevent you from running `rm -rf ./*` because you are in the "wrong" directory.

Currently, we don't have a way an easy way to check what kind of cluster the work space is configured for.
We could add additional "safeguards" by first checking if the statefile matches the QEMU provider and the name is set to `mini`

katexochen:
Don't forget to update the changelog. :wink:

Nirusu:
Failing govulncheck should be resolved after rebasing this branch to current main and then merging.

PR #200, opened 2022-09-28
enable konnectivity

3u13r:
### Proposed change(s)

- Route KubeAPI -> kubelet/pod/svc traffic over konnectivity

Note that with these settings it is possible (and likely) that the traffic will do another hop from the worker node kkonnecitivity routed the traffic to. 

As stability test I ran our e2e sonobuoy test on Azure and GCP with 3 control-plane / 3 worker nodes. 

### Checklist

- [x] Link to Milestone

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Local cluster creation based on QEMU
 - Verification of Azure trusted launch attestation keys
 - Kubernetes version v1.25 is now fully supported.
+- Enabled Konnectivity.
 
 ### Changed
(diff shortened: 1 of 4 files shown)


daniel-weisse:
What image did you use for testing?
This one test is failing for me on GCP:

```shell
[sig-network] HostPort validates that there is no conflict between pods with same hostPort but different hostIP and protocol [LinuxOnly] [Conformance]
```

Also, streaming logs (`kubectl logs <pod> -f`) still gets interrupted after ~30 seconds, but I think this PR wasn't supposed to fix that.

daniel-weisse (approved):

Nirusu:
Is this Conformance tested with `constellation init --conformance`?

PR #203, opened 2022-09-29
AB#2444 Verify Azure trusted launch attestation keys

daniel-weisse:
### Proposed change(s)
- Verify attestation keys of Azure trusted launch VMs
  - AK and Cert are loaded from the TPM
  - AkCert is signed by Azure's AME Infra CA, which in turn is signed by `ameroot` which we use as a root of trust

### Additional info
- go-tpm-tools still doesn't provide a way to directly load a cached key, which caused issues with the attestation key stored in the trusted launch TPM
  - go-tpm-tools only supports no auth policy or the endorsment key policy, if a key has another auth policy set, loading the key from template fails, which causes the library to instead overwrite the stored key with a new one
- I have forked the library and added a simple `LoadCachedKey` function, which takes a tpm and handle, and loads the key stored at that handle.

### Checklist

- [ ] Link to Milestone


internal/attestation/azure/trustedlaunch/issuer.go
```diff
@@ -7,31 +7,117 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+type akSigner struct {
+	AkCert []byte
+	CA     []byte
+}
+
+// getAttestationCert returns the DER encoded certificate of the TPM's attestation key and it's CA.
+func (i *Issuer) getAttestationCert(tpm io.ReadWriteCloser) ([]byte, error) {
+	certDER, err := tpm2.NVReadEx(tpm, tpmAkCertIdx, tpm2.HandleOwner, "", 0)
+	if err != nil {
+		return nil, fmt.Errorf("reading attestation key certificate from TMP: %w", err)
```
thomasten, line 60:
```suggestion
		return nil, fmt.Errorf("reading attestation key certificate from TPM: %w", err)
```

internal/attestation/azure/trustedlaunch/validator.go
```diff
@@ -8,39 +8,98 @@ package trustedlaunch
...
+		return nil, fmt.Errorf("parsing attestation key CA certificate: %w", err)
+	}
+
+	rootCA, err := certutil.PemToX509Cert(v.rootCA)
+	if err != nil {
+		return nil, fmt.Errorf("parsing AME root CA: %w", err)
 	}
-	return pubArea.Key()
+	roots := x509.NewCertPool()
+	roots.AddCert(rootCA)
```
thomasten, line 75: This could be done in New or even only once at program start (like global `var ameRootPool = func() *x509.CertPool { ... }()` with panic on error) to catch this logic error early. But not really important, pls do as you prefer.

thomasten (approved):

internal/attestation/azure/trustedlaunch/validator.go
daniel-weisse: Thought about this too, but wasnt sure if we want to have a potential panic call in the Validator init function.
I guess because the conversion of the root ca from PEM should always work, this it should be ok to panic in case of an error?

thomasten: I guess so. @katexochen once suggested to put such `must` code globally so that it runs (and panics on a programming bug) on program start.

thomasten (approved):

Issue #189, opened 2022-09-23
[CLI] Activity indicator for init command


derpsteb:
### Use case
When interacting with cloud provider APIs there are actions that take multiple minutes. In those cases the user might become uncertain if the program is still working. Some kind of animation, like a spinner, would improve UX.

### Describe your solution
Implement a small animation that shows that the program is still running, while waiting for the `init` response to come back.

Feel free to also implement this for other commands like `create` or `terminate`.
### Additional context
- Lots of ASCII Art Spinner inspiration and example code: https://stackoverflow.com/questions/2685435/cooler-ascii-spinners
- The `init` request is sent here: [cli/internal/cmd/init.go](https://github.com/edgelesssys/constellation/blob/main/cli/internal/cmd/init.go#L128)

BaLiKfromUA:
Hey! I would be happy to solve this issue in the scope of Hactoberfest 2022. Can you assign it to me, and I will send my solution in October?

PR #207, opened 2022-10-02
[CLI] Activity indicator for init command

BaLiKfromUA:
### Proposed change(s)
This is my solution to #189 

- created `Spinner` class with basic functionality to reuse it later if needed
- integrated my spinner with `init` command as was proposed in the issue
- covered my spinner with several dummy tests

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/spinner.go
+++ b/cli/internal/cmd/spinner.go
@@ -1,0 +1,67 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+var spinnerStates = []string{"⣾", "⣽", "⣻", "⢿", "⡿", "⣟", "⣯", "⣷"}
+var dotsStates = []string{".", "..", "..."}
+
+type Spinner struct {
+	out      *cobra.Command
+	text     string
+	showDots bool
+	delay    time.Duration
+	wg       *sync.WaitGroup
+	stop     int32
+}
+
+func NewSpinner(c *cobra.Command, text string, showDots bool) *Spinner {
+	return &Spinner{out: c,
+		text:     text,
+		showDots: showDots,
+		wg:       &sync.WaitGroup{},
+		delay:    100 * time.Millisecond,
+		stop:     0,
+	}
+}
+
+func (s *Spinner) Start() {
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+
+	out:
+		for {
+			for i := 0; i < len(spinnerStates); i++ {
+				if atomic.LoadInt32(&s.stop) == 0 {
+					dotsState := ""
+					if s.showDots {
+						dotsState = dotsStates[i%len(dotsStates)]
+					}
+					state := fmt.Sprintf("\r%s %s%s", spinnerStates[i], s.text, dotsState)
+					s.out.Print(state)
+					time.Sleep(s.delay)
+				} else {
+					break out
+				}
+			}
+		}
+
+		dotsState := ""
+		if s.showDots {
+			dotsState = dotsStates[len(dotsStates)-1]
+		}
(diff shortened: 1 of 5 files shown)


BaLiKfromUA:
I hope that my solution is OK-ish for you. Waiting for your feedback.

After all remarks are solved, I can integrate the spinner with `create` and `terminate` commands

cli/internal/cmd/spinner.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	spinnerStates = []string{"⣾", "⣽", "⣻", "⢿", "⡿", "⣟", "⣯", "⣷"}
+	dotsStates    = []string{".", "..", "..."}
+)
+
+type Spinner struct {
```
katexochen, line 23:
```suggestion
type spinner struct {
```
As this is only used within the `cmd` package, it can be private.

cli/internal/cmd/spinner.go
```diff
@@ -0,0 +1,77 @@
...
+	out      *cobra.Command
+	text     string
+	showDots bool
+	delay    time.Duration
+	wg       *sync.WaitGroup
+	stop     int32
+}
+
+func NewSpinner(c *cobra.Command, text string, showDots bool) *Spinner {
```
katexochen, line 32:
```suggestion
func newSpinner(c *cobra.Command, text string, showDots bool) *Spinner {
```
As this is only used within the `cmd` package, it can be private.

cli/internal/cmd/spinner.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+func (s *Spinner) Start() {
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+
+	out:
+		for {
+			for i := 0; i < len(spinnerStates); i++ {
```
katexochen, line 50:
```suggestion
		for i := 0; ; i = (i + 1) % len(spinnerStates) {
```
I would suggest to simplify this loop a little to reduce complexity.

cli/internal/cmd/spinner.go
```diff
@@ -0,0 +1,77 @@
...
+					dotsState := ""
+					if s.showDots {
+						dotsState = dotsStates[i%len(dotsStates)]
+					}
+					state := fmt.Sprintf("\r%s %s%s", spinnerStates[i], s.text, dotsState)
+					s.out.Print(state)
+					time.Sleep(s.delay)
+				} else {
+					break out
+				}
```
katexochen, line 61: Go prefers early returns instead of using else blocks. You can add the break condition to the beginning of your loop like this:
```go
			if atomic.LoadInt32(&s.stop) != 0 {
				break
			}
```

cli/internal/cmd/spinner.go
```diff
@@ -0,0 +1,77 @@
...
+			}
+		}
+
+		dotsState := ""
+		if s.showDots {
+			dotsState = dotsStates[len(dotsStates)-1]
+		}
+		finalState := fmt.Sprintf("\r%s%s", s.text, dotsState)
```
katexochen, line 69: There is a tiny bug in this line, as the previous text isn't fully overwritten in case that the dots aren't printed. It produces the following output:
```
Initializing cluster r 
Your Constellation cluster was successfully initialized.
```
The `r` from the previous iteration isn't overwritten as the string is now shorter than before. You can fix this by just printing a space character after the text.
```suggestion
		finalState := fmt.Sprintf("\r%s%s ", s.text, dotsState)
```

katexochen (changes requested):
Thanks again for your contribution! Looks really good already, your code has great test coverage. Just some minor improvement suggestions from my side. Feel free to add the spinner to create and terminate, too.

BaLiKfromUA:
@katexochen, I believe I finished all stuff that you mentioned. Review again when you have time )

katexochen (approved):
LGTM, great work!

PR #209, opened 2022-10-04
AB#2436: Initial support for Terraform-based AWS Nitro resource creation/termination

Nirusu:
It's time for a fun PR!

### Proposed change(s)
- Add current AWS configuration variables to Constellation config
- Maintain a list of supported instance families or types based on [this list](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances). There is a hack script that mostly does 1:1 copy this list, which then creates a flat slice that is then used for compatibility checks & printing the list of supported instance types (which excludes Graviton instances). This is, in my opinion, the messiest part of this PR and I don't know if there's a better solution that doesn't involve manually curating a list of supported instance type & size.
- Add support for creating/terminating AWS clusters when `CONSTELLATION_AWS_DEV=1` is set
- Add feature to print supported AWS Nitro instance families / types using `constellation config instance-types` when `CONSTELLATION_AWS_DEV=1` is set
- Block off all aTLS related actions for AWS while development is ongoing (no init/verify/...?)
- Minor fixes to the Terraform cleanup (GCP was hardcoded)
- Minor fixes to the Terraform AWS skeleton (e.g. security groups were not applied, this is fixed now)
- Revise the networking setup in the AWS skeleton by using one network load balancer with multiple targets and using a setup of a private subnet and a public subnet with a NAT gateway + internet gateway as advised by the AWS docs:
   - https://aws.amazon.com/premiumsupport/knowledge-center/public-load-balancer-private-ec2/
   - https://docs.aws.amazon.com/vpc/latest/userguide/nat-gateway-scenarios.html 

This dual subnet setup unfortunately creates the requirement that both subnets need to be in the same zone as the load balancer cannot target instances in another zone then the (public) subnet(s) it's assigned to. It could be somewhat relatively fixed easily by creating multiple subnets based on the number of instances. But that can be implemented later, as we will have to do something similar with GCP likely, too. 

### Checklist

- [x] Link to Milestone


internal/config/config.go
```diff
@@ -548,11 +628,62 @@ func validInstanceTypeForProvider(insType string, acceptNonCVM bool, provider cl
...
+		return false
 	default:
 		return false
 	}
 }
 
+// checkIfAWSInstanceTypeIsValid checks if an AWS instance type passed as user input is in one of the many supported instance families or single instances.
+func checkIfAWSInstanceTypeIsValid(userInput string) bool {
+	/*
+		Check for supported instances with the following characterics:
```
3u13r, line 646:
```suggestion
		Check for supported instances with the following characteristics:
```

internal/config/config.go
```diff
@@ -548,11 +628,62 @@ func validInstanceTypeForProvider(insType string, acceptNonCVM bool, provider cl
...
+
+	if isGraviton || !hasEnoughVCPUs {
+		return false
+	}
+
+	// Now check if the user input is a supported family
+	// Note that we cannot directly use the family split from the Graviton check above, as some instances are directly specified by their full name and not just the family in general
+	for _, supportedFamily := range instancetypes.AWSSupportedInstanceTypesOrFamilies {
+		supportedFamilyLowercase := strings.ToLower(supportedFamily)
+		if strings.Contains(userInputLowercase, supportedFamilyLowercase) {
```
3u13r, line 679: I think we should check for equality here.

internal/config/config.go
```diff
@@ -548,11 +628,62 @@ func validInstanceTypeForProvider(insType string, acceptNonCVM bool, provider cl
...
+		return false
+	}
+	if strings.Contains(userInput, ";") {
+		return false
+	}
+
+	userInputLowercase := strings.ToLower(userInput)
+	family := strings.Split(userInputLowercase, ".")[0]
+	isGraviton := strings.LastIndex(family, "g") > 0
+	hasEnoughVCPUs := strings.Contains(userInputLowercase, "xlarge")
```
3u13r, line 669: `m6id.metal` is part of the `AWSSupportedInstanceTypesOrFamilies` list. Is this type allowed?
What could work here is to explicitly disallow `medium`, `micro` `large` and so on.

3u13r (changes requested):

internal/config/config.go
Nirusu: Guess I forgot checking for metal types here. But just adding `metal` in addition to `xlarge` should work better than disallowing the others I guess.

internal/config/config.go
Nirusu: See the comment what makes checking for equality specifically a bit difficult. Would need differentiating between families as input versus full type names.

cli/internal/terraform/terraform/aws/main.tf
```diff
@@ -0,0 +1,301 @@
...
+    subnet_id     = aws_subnet.public.id
+    allocation_id = aws_eip.lb.id
+  }
+
+  tags = {
+    Name = "loadbalancer"
+  }
+
+  enable_cross_zone_load_balancing = true
+}
```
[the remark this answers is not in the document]
Nirusu: Mh, not sure and not super familiar with Terraform conventions.

I think a module makes sense when reusing certain components. It could make sense moving the elastic IP & subnet logic could into a separate module (as for HA creating multiple of them make sense in the future). 

The load balancer only needs to be created once, but for HA would have multiple subnet mappings. I think it could remain here. though.

Does this sound like a good compromise for you?

internal/config/config.go
```diff
@@ -95,6 +98,34 @@ type ProviderConfig struct {
...
+	//   VM instance type to use for Constellation nodes. Needs to be Nitro-enabled. See: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#ec2-nitro-instances
+	InstanceType string `yaml:"instanceType" validate:"aws_instance_type"`
+	// description: |
+	//   Name of the IAM profile to use for the control plane nodes.
+	IAMProfileControlPlane string `yaml:"iamProfileControlPlane" validate:"required"`
+	// description: |
+	//   Name of the IAM profile to use for the worker nodes.
+	IAMProfileWorkerNodes string `yaml:"iamProfileWorkerNodes" validate:"required"`
+	// description: |
+	//   Expected confidential VM measurements.
```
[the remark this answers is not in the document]
Nirusu: Yes, this was just a leftover from copy-paste. Will rename.

internal/config/instancetypes/aws.go
```diff
@@ -0,0 +1,12 @@
...
+*/
+
+//go:generate go run ../../../hack/generate-aws-instancetypes/main.go > aws.go
+// Code generated by hack/generate-aws-instancetypes tool. DO NOT EDIT.
+
+package instancetypes
+
+var AWSSupportedInstanceTypesOrFamilies = []string{"M5", "M5a", "M5ad", "M5d", "M5dn", "M5n", "M5zn", "M6a", "M6i", "M6id", "T3", "T3a", "C5", "C5a", "C5ad", "C5d", "C5n", "C6a", "C6i", "C6id", "Hpc6a", "R5", "R5a", "R5ad", "R5b", "R5d", "R5dn", "R5n", "R6a", "R6i", "R6id", "X2idn", "X2iedn", "X2iezn", "z1d", "D3", "D3en", "I3en", "I4i", "DL1", "G4", "G4ad", "G5", "Inf1", "P4", "VT1", "u-3tb1.56xlarge", "u-6tb1.56xlarge", "u-6tb1.112xlarge", "u-9tb1.112xlarge", "u-12tb1.112xlarge", "p3dn.24xlarge", "m5.metal", "m5d.metal", "m5dn.metal", "m5n.metal", "m5zn.metal", "m6a.metal", "m6i.metal", "m6id.metal", "c5.metal", "c5d.metal", "c5n.metal", "c6a.metal", "c6i.metal", "c6id.metal", "r5.metal", "r5b.metal", "r5d.metal", "r5dn.metal", "r5n.metal", "r6a.metal", "r6i.metal", "r6id.metal", "u-6tb1.metal", "u-9tb1.metal", "u-12tb1.metal", "u-18tb1.metal", "u-24tb1.metal", "x2idn.metal", "x2iedn.metal", "x2iezn.metal", "z1d.metal", "i3.metal", "i3en.metal", "i4i.metal", "g4dn.metal"}
```
katexochen, line 12: Could you add newlines to the generated output? :grinning:

internal/config/config.go
```diff
@@ -471,19 +550,25 @@ func (c *Config) RemoveProviderExcept(provider cloudprovider.Provider) {
...
-		gcpRegex := regexp.MustCompile(`^projects\/constellation-images\/global\/images\/constellation-v[\d]+-[\d]+-[\d]+$`)
-		return !gcpRegex.MatchString(c.Provider.GCP.Image)
+	case c.Provider.AWS != nil:
+		// TODO: Add proper image name validation for AWS when we are closer to release.
+		return true
 	case c.Provider.Azure != nil:
 		azureRegex := regexp.MustCompile(`^\/CommunityGalleries\/ConstellationCVM-b3782fa0-0df7-4f2f-963e-fc7fc42663df\/Images\/constellation\/Versions\/[\d]+.[\d]+.[\d]+$`)
 		return !azureRegex.MatchString(c.Provider.Azure.Image)
+	case c.Provider.GCP != nil:
+		gcpRegex := regexp.MustCompile(`^projects\/constellation-images\/global\/images\/constellation-v[\d]+-[\d]+-[\d]+$`)
```
katexochen, line 560: regexp with `MustCompile` must be global var, not in a func. (not your code, I know)

internal/config/instancetypes/aws.go
Nirusu: Do you mean for the user-visible printed output in the CLI or this just for this file / slice specifically?

daniel-weisse:
I think the files in `cli/internal/terraform/aws/iam` were used for testing only, and are not used in our deployment?
If so, they can be removed

3u13r:
> I think the files in `cli/internal/terraform/aws/iam` were used for testing only, and are not used in our deployment? If so, they can be removed

Please don't remove it. It serves 2 purposes:
1. It is our internal (aka. my) way of creating new roles for us devs, which are consumed by the AWS instances
2. It can be used by third parties as blueprint for their IAM creation.  

Fine with moving it somewhere else, though.

daniel-weisse:
> 1. It is our internal (aka. my) way of creating new roles for us devs, which are consumed by the AWS instances
> 2. It can be used by third parties as blueprint for their IAM creation.

For Azure and GCP we have instructions for setting up permissions using the console or CLI, imo we should add the same for AWS instead of having some terraform template lying in the repo somewhere

Nirusu:
Were would you suggest putting them?

internal/config/config.go
Nirusu: Moved it 👍🏻

katexochen:
> Were would you suggest putting them?

Maybe `/terraform/aws` ?

Nirusu:
If this is indeed on track for v2.2.0, should I remove the `CONSTELLATION_AWS_DEV` feature flag? Or keep it for now in case we won't make it until the next version?

katexochen:
>  should I remove the CONSTELLATION_AWS_DEV feature flag

Would keep until feature is completed. The tag can be easily moved to next version if it doesn't complete until release.

cli/internal/terraform/terraform/aws/variables.tf
```diff
@@ -1,37 +1,37 @@
...
 
 variable "instance_type" {
   type        = string
   description = "Instance type for worker nodes"
-  default     = "t2.micro"
+  default     = "t3.micro"
 }
 
-variable "disk_size" {
+variable "state_disk_size" {
```
3u13r, line 22: This is currently unused. Is the current configured disk size (the local variable) the OS disk and we also need to attach more storage?

3u13r (changes requested):

cli/internal/terraform/terraform/aws/variables.tf
Nirusu: Good catch. From how I interpret the docs, the AMI can define both the OS disk (root device) but also additional EBS storage devices.

For the first, if we provide an AMI we should not need any additional configuration with a read-only disk, I suppose. For the latter, I guess we can add a storage device configurable by the user using the variable and also storage type.

 I'll add this.

Nirusu:
Squashed some fix commits.
Now added:
✅ CSP-encrypted EBS root device (expected to be auto-created by AMI)
✅ State disk (CSP-encrypted on top of what we add later with disk-mapper on our own image, similar as on the other CSPs)

State disk should always be the second disk on the system (e.g. `/dev/sdb`, `/dev/xvdb`, `/dev/hdb`, `/dev/nvme1n1`). AWS takes `/dev/sdb` and converts it into a definition that's suitable for the selected disk type. For the default `gp2`, that seems to be `/dev/nvme1n1` if I am not wrong).

Note that the web UI does _not_ correctly reflect what Linux sees, so for further disk-related development check what you see in Linux, not on the AWS CLI or Console.

cli/internal/terraform/terraform/aws/variables.tf
```diff
@@ -0,0 +1,68 @@
...
+variable "worker_count" {
+  type        = number
+  description = "Number of worker nodes"
+  default     = 1
+}
+
+variable "ami" {
+  type        = string
+  description = "AMI ID"
+  default     = "ami-02f3416038bdb17fb" // Ubuntu 22.04 LTS
```
3u13r, line 49: Should we remove all defaults which are required by the CLI anyway (node numbers, ami, region, instance-type, ...) Or should we keep those if e.g. we want to use the standalone terraform in the future?

daniel-weisse: imo most variables should not have default values in the terraform templates, especially ones like ami, since those would need to be changed with every release

Nirusu: Removed.

internal/config/config.go
```diff
@@ -212,6 +251,16 @@ func Default() *Config {
 		StateDiskSizeGB: 30,
 		DebugCluster:    func() *bool { b := false; return &b }(),
 		Provider: ProviderConfig{
+			AWS: &AWSConfig{
+				Region:                 "",
+				Image:                  "",
+				InstanceType:           "t3.xlarge",
```
malt3, line 257: The default instance type does not support nitro tpm:

```
Error: waiting for Auto Scaling Group (malte-8bc50119-control-plane) capacity satisfied: 1 error occurred:
	* Scaling activity (fad60f97-3445-a48c-c220-6b9716070b1a): Failed: The t3.xlarge instance type does not support an AMI with TPM version v2.0. Specify an instance type that supports TPM version v2.0, and try again. Launching EC2 instance failed.
```

Nirusu: I searched for Amazon Nitro instances but I guess it turns out that AWS Nitro ≠ NitroTPM support:
 
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enable-nitrotpm-prerequisites.html

What a god awful naming scheme for Amazon’s product lineup. Absolute lunacy by Amazon branding this as Nitro yet supporting it on a limited subset of Nitro machines.

Will have to adjust the whole compatibility list later. Thanks Amazon, I hate it.

Nirusu: Fixed. At least it has the benefit that the supported list is now way smaller and thus the check is way easier :)

Nirusu:
@katexochen How do we ignore specific CI warnings for Terraform?
We agreed not to care about public LB and unrestricted ingress/egress as we cannot make sane defaults without knowing the specific setup of the user creating a cluster (they are all authenticated endpoints anyway).

Same with CloudWatch logs. We could automatically create CMK, but it does not really add any purpose as our logs are designed not to leak any sensitive information anyway. So that would just waste the user's money (and you would have to trust AWS not to do anything with your key anyway ;))

internal/config/config.go
```diff
@@ -556,11 +643,59 @@ func validInstanceTypeForProvider(insType string, acceptNonCVM bool, provider cl
...
+				return true
+			}
+		}
+		return false
 	default:
 		return false
 	}
 }
 
+// checkIfAWSInstanceTypeIsValid checks if an AWS instance type passed as user input is in one of the instance families supporting NitroTPM-
```
malt3, line 658: nit
```suggestion
// checkIfAWSInstanceTypeIsValid checks if an AWS instance type passed as user input is in one of the instance families supporting NitroTPM.
```

internal/config/config.go
```diff
@@ -556,11 +643,59 @@ func validInstanceTypeForProvider(insType string, acceptNonCVM bool, provider cl
...
+func checkIfAWSInstanceTypeIsValid(userInput string) bool {
```
malt3, line 659: Some targeted unit tests for this func would be great. Is the AWS instance type format documented somewhere officially? I actually don't know and do not have experience with AWS instance types. If we do custom parsing (instead of listing all available types which - granted - might be a lot), we should ensure that there are no false positives / negatives.

Nirusu: TestValidInstanceTypeForProvider essentially calls this function directly. I could create a targeted test but it would essentially be a copy-paste (or removal of the AWS test cases from this function). Choose what you prefer.

The general naming scheme is described here:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-types.html#AvailableInstanceTypes

Of course in all cleverness Amazon does not always follow the convention with 100% accuracy. For example, there's this whole family of Graviton instances called the A1 family, and none of them has `g` in their name:
https://aws.amazon.com/de/ec2/instance-types/a1/

A for ARM, a for AMD, g for Graviton (which is ARM but okay), _right?_ 🙄 

But they are not on the supported list anyways, luckily. 

I think a better explanation can be found here:
https://stackoverflow.com/a/56880093

On that note, I maybe should switch to M6a as the default value.

From what I can tell, xlarge is almost always 4vCPUs, and everything beyond that are multiplicators of xlarge or metal (which NitroTPM does not support, according to the docs). I haven't seen an exception to this yet, luckily. So everything needs to be xlarge or higher (but not metal!) to account for the 2:1 vCPU ratio.

And from what I can tell, all supported instances are indeed following the 2:1 vCPU ratio:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/cpu-options-supported-instances-values.html

So at least with the current set of supported instance types I don't see where the parser can cause issues since family parsing and size parsing should be mostly okay I believe, and the modificators are part of the family. I need to test case-sensitive quickly, though.

Small downside of the approach right now is that we don't support checking the specific instance type. For example, if someone writes "x2gd.7xlarge" we don't check for that and it will fail at creation. But personally I am not sure if we want to check every size. It's definitely possible, but it's too many families to keep this list up-to-date I guess.

internal/config/config.go
```diff
@@ -556,11 +643,59 @@ func validInstanceTypeForProvider(insType string, acceptNonCVM bool, provider cl
...
+	// Check if instace type has at least 4 vCPUs (= contains "xlarge" in its name)
+	hasEnoughVCPUs := strings.Contains(userDefinedSize, "xlarge")
+	if !hasEnoughVCPUs {
+		return false
+	}
+
+	// Now check if the user input is a supported family
+	// Note that we cannot directly use the family split from the Graviton check above, as some instances are directly specified by their full name and not just the family in general
+	for _, supportedFamily := range instancetypes.AWSSupportedInstanceFamilies {
+		supportedFamilyLowercase := strings.ToLower(supportedFamily)
```
3u13r, line 690: I think we actually need to check the casing here. `C5.xlarge` is not a valid instance type while `c5.xlarge` is.

Nirusu: Should now be enforced by go-validator.

internal/config/config.go
malt3: I see. I think we can leave it as-is for now :) Thanks for the pointers to your research.

malt3 (approved):
My review is limited to the changes due to my review comments. Those parts LGTM.

3u13r (approved):
LGTM

PR #212, opened 2022-10-05
Feat/revive

datosh:
### Proposed change(s)

- enable revive as linter

### Additional info
- Originally triggered because I forgot to comment a public function and wanted our linter to notify me. Saw that we have 200 findings and its not a super quick fix. Added to dev sync agenda to decide if we want to fix.

### Checklist

- [ ] Link to Milestone


operators/constellation-node-operator/internal/azure/client/nodeimage_test.go
```diff
@@ -351,21 +351,21 @@ func TestCapacityPollingHandler(t *testing.T) {
 	assert.Equal(wantCapacity, gotCapacity)
 }
 
-type stubCapacityPoller struct {
```
katexochen, line 354: Imo the stub shouldn't be public, is there anything that returns it?

katexochen (approved):
LGTM :seal:

operators/constellation-node-operator/internal/azure/client/nodeimage_test.go
daniel-weisse: I think the issues comes from public functions returning private types?
If so `NewStubCapacityPoller` should be made private, if possible, so we don't export testing stubs

katexochen: That is what I wanted to suggest. :)

thomasten (approved):

PR #217, opened 2022-10-06
Hide cursor and fix dots

daniel-weisse:
### Proposed change(s)

- Fixes dots not being displayed correctly and hides the cursor while printing them

### DIFF
--- a/cli/internal/cmd/spinner.go
+++ b/cli/internal/cmd/spinner.go
@@ -104,7 +104,7 @@ func spinTTY(out io.Writer, wg *sync.WaitGroup, stop *int32, delay time.Duration
 
 func spinNoTTY(out io.Writer, wg *sync.WaitGroup, _ *int32, _ time.Duration, text string, _ bool) {
 	defer wg.Done()
-	fmt.Fprintln(out, text)
+	fmt.Fprintln(out, text+"...")
 }
 
 type nopSpinner struct{}
--- a/cli/internal/cmd/spinner.go
+++ b/cli/internal/cmd/spinner.go
@@ -12,13 +12,17 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+)
 
-	"github.com/spf13/cobra"
+const (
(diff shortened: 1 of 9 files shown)


datosh (approved):
LGTM

cli/internal/cmd/spinner.go
```diff
@@ -8,39 +8,46 @@ package cmd
...
 
-	"github.com/spf13/cobra"
+const (
+	// hideCursor and showCursor are ANSI escape sequences to hide and show the cursor.
+	hideCursor = "\033[?25l"
+	showCursor = "\033[?25h"
 )
 
 var (
 	spinnerStates = []string{"⣾", "⣽", "⣻", "⢿", "⡿", "⣟", "⣯", "⣷"}
```
datosh, line 24:
```suggestion
	spinnerStates = []string{"⣷", "⣯", "⣟", "⡿", "⢿", "⣻", "⣽", "⣾"}
```

daniel-weisse:
Made some more changes:
* Spinner can be restarted
* Try to detect if out is a tty, if not, restrict messages to once every 20 seconds
  * @Nirusu could you test if this works for darwin?
* Add the `Writer` method to `spinner` directly instead of holding it in a wrapped type

katexochen:
Did you check this works on interrupt? Code lgtm.

daniel-weisse:
>Did you check this works on interrupt?

Works for interrupts, but I just noticed some issues with the slowed down spinner.
Will refactor slightly

Nirusu:
I won’t have time to check this for a couple of days, but you could use the e2e action and check if it disables / slows down the spinner.

It should likely be fine though, the check should generally work for UNIX and the go-isatty package seems to be multi OS aware.

Nirusu (approved):
TTY tested on fish and non-TTY tested on CI, both on macOS.
I hope that covers weird edge cases :D

Personally I think we could append "..." to the non-TTY input just for clarity that this is a longer-waiting process, but I don't consider non-TTY to be super important so decide by yourself.

PR #221, opened 2022-10-07
Wrap references to mini Constellation in tip boxes

daniel-weisse:
### Proposed change(s)

- Make the mini Constellation references a bit more visible by wrapping them in a tip box


m1ghtym0 (approved):
good idea!

PR #225, opened 2022-10-10
Use env to find bash in shebang

katexochen:
### Proposed change(s)

- As `/bin/bash` doesn't exist on all *nixes. :snowflake:


malt3 (approved):
LGTM

PR #229, opened 2022-10-10
AB#2474 Implement List and Self method for AWS

daniel-weisse:
### Proposed change(s)

- Add first support for AWS by implementing `metadata.Self()` and `metadata.List()`
- The parsed information *should* be compatible with the Terraform template in `terraform/aws/`, i.e. instance-roles / uids are read from instance tags, instead of parsed from the instance's name

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/cloud/aws/metadata.go
+++ b/internal/cloud/aws/metadata.go
@@ -11,9 +11,9 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"sync"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/feature/ec2/imds"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
@@ -42,11 +42,17 @@ type Metadata struct {
 	imds imdsAPI
 }
 
-func New(cfg aws.Config) *Metadata {
+// New initializes a new AWS Metadata client using instance default credentials.
+// Default region is set up using the AWS imds api.
+func New() (*Metadata, error) {
+	cfg, err := config.LoadDefaultConfig(context.Background(), config.WithEC2IMDSRegion())
+	if err != nil {
+		return nil, err
+	}
 	return &Metadata{
 		ec2:  ec2.NewFromConfig(cfg),
 		imds: imds.New(imds.Options{}),
-	}
+	}, nil
 }
 
 // Supported is used to determine if metadata API is implemented for this cloud provider.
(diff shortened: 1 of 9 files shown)


internal/cloud/aws/metadata.go
```diff
@@ -0,0 +1,195 @@
...
+	GetMetadata(context.Context, *imds.GetMetadataInput, ...func(*imds.Options)) (*imds.GetMetadataOutput, error)
+}
+
+// Metadata implements core.ProviderMetadata interface for AWS.
+type Metadata struct {
+	ec2  ec2API
+	imds imdsAPI
+}
+
+func New(cfg aws.Config) *Metadata {
```
katexochen, line 45: All top-level, exported names should have doc comments, as should non-trivial unexported type or function declarations. [:link: Go Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments#doc-comments)

katexochen: If possible, don't expose structs from aws sdk outside of this package so other packages don't need to depend on int.

internal/cloud/aws/metadata.go
```diff
@@ -0,0 +1,195 @@
...
+		Name:       name,
+		ProviderID: fmt.Sprintf("aws:///%s/%s", identity.AvailabilityZone, identity.InstanceID),
+		Role:       role.FromString(instanceRole),
+		VPCIP:      identity.PrivateIP,
+	}, nil
+}
+
+func (m *Metadata) getAllInstancesInGroup(ctx context.Context, uid string) ([]types.Instance, error) {
+	var instances []types.Instance
+	nextToken := (*string)(nil)
```
katexochen, line 97: wha? you mean `var nextToken *string`?

internal/cloud/aws/metadata.go
```diff
@@ -0,0 +1,195 @@
...
+	once := sync.Once{}
+	first := func() bool {
+		var ret bool
+		once.Do(func() { ret = true })
+		return ret
+	}
```
katexochen, line 103: nit: This seems over complicated.

internal/cloud/aws/metadata.go
```diff
@@ -0,0 +1,195 @@
...
+	for nextToken != nil || first() {
+		describeOutput, err := m.ec2.DescribeInstances(ctx, &ec2.DescribeInstancesInput{
+			Filters: []types.Filter{
+				{
+					Name:   aws.String("tag:" + tagUID),
+					Values: []string{uid},
+				},
+			},
+			NextToken: nextToken,
+		})
```
katexochen, line 114:
```suggestion

    in := &ec2.DescribeInstancesInput{
			Filters: []types.Filter{
				{
					Name:   aws.String("tag:" + tagUID),
					Values: []string{uid},
				},
			},
			NextToken: nextToken,
		}
	for out, err := m.ec2.DescribeInstances(ctx, in);; out, err = m.ec2.DescribeInstances(ctx, in) {
		...
		if describeOutput.NextToken == nil {
		    break
		}
```
Maybe something like this would be easier to understand?

internal/cloud/aws/metadata.go
```diff
@@ -0,0 +1,195 @@
...
+	if err != nil {
+		return "", err
+	}
+	defer reader.Content.Close()
+	instanceTag, err := io.ReadAll(reader.Content)
+	return string(instanceTag), err
+}
+
+func findTag(tags []types.Tag, wantKey string) (string, error) {
+	for _, tag := range tags {
```
katexochen, line 189: Should we check if `tag == nil` ?

katexochen (changes requested):

internal/cloud/aws/metadata.go
katexochen: Exactly :)

internal/cloud/aws/metadata.go
```diff
@@ -42,11 +42,17 @@ type Metadata struct {
 	imds imdsAPI
 }
 
-func New(cfg aws.Config) *Metadata {
+// New initializes a new AWS Metadata client using instance default credentials.
+// Default region is set up using the AWS imds api.
+func New() (*Metadata, error) {
+	cfg, err := config.LoadDefaultConfig(context.Background(), config.WithEC2IMDSRegion())
```
katexochen, line 48: Context should be given as parameter to the constructor.

internal/cloud/aws/metadata.go
```diff
@@ -0,0 +1,197 @@
...
+)
+
+type ec2API interface {
+	DescribeInstances(context.Context, *ec2.DescribeInstancesInput, ...func(*ec2.Options)) (*ec2.DescribeInstancesOutput, error)
+}
+
+type imdsAPI interface {
+	GetInstanceIdentityDocument(context.Context, *imds.GetInstanceIdentityDocumentInput, ...func(*imds.Options)) (*imds.GetInstanceIdentityDocumentOutput, error)
+	GetMetadata(context.Context, *imds.GetMetadataInput, ...func(*imds.Options)) (*imds.GetMetadataOutput, error)
+}
```
malt3, line 37: We could implement caching around the imds API. This is implemented on Azure: https://github.com/edgelesssys/constellation/blob/e3e7fddf76858d5c4d1d76ad4b74821c035e2800/internal/cloud/azure/imds.go

I view that as an optional feature so feel free to ignore for now.

malt3 (approved):
LGTM. Very clean :)

PR #230, opened 2022-10-11
AB#2480: Increased CRD timeout

datosh:

### Proposed change(s)

- Double CRD timeout to prevent timeouts during `mini up`.

### Additional info
- There is currently no way for users to overwrite this setting if necessary. A better fix would be to allow users to overwrite this value if necessary. Do we want to implement this?

### Checklist

- [ ] Link to Milestone


malt3:
This specifies the maximum wait time between registering a CRD with kubernetes and it being available via the API.
I don't believe we should make this configurable as having a lot of (seemingly unimportant) settings is confusing to end users. Maybe we can set to to a large values (~5min) if anyone ever sees this issue again.

malt3 (approved):
LGTM

PR #231, opened 2022-10-11
Fix CLI version injection in cmake

katexochen:
### Proposed change(s)

- Fix version injection for CLI in cmake file, which was broken by upgrading to v2 Go module


daniel-weisse (commented):
version injection is likely also broken in all Dockerfiles making use of it.
Can you fix those as well?

daniel-weisse (approved):

PR #232, opened 2022-10-11
AB#2479 Implement AWS cloud logging

daniel-weisse:
### Proposed change(s)

- Implements Cloud Logging for AWS instances

### Additional info
- AWS cloud logging requires a Cloudwatch log group to be set up: https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/Working-with-log-groups-and-streams.html
  - Creation of the group should be handled by Terraform
  - Log group should be tagged with the Constellation UID, naming should not matter
  - One log group per Constellation cluster
- Applications write their messages to a log stream, one log group can hold multiple streams
  - One log stream per ec2 instance
  - Log stream is named after the instance writing messages to it

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/internal/cloud/aws/logger.go
+++ b/internal/cloud/aws/logger.go
@@ -32,10 +32,10 @@ type Logger struct {
 	logs          []types.InputLogEvent
 	sequenceToken *string
 
-	flushMux sync.Mutex
+	mux      sync.Mutex
 	interval time.Duration
 	clock    clock.WithTicker
-	wg       *sync.WaitGroup
+	wg       sync.WaitGroup
 	stopCh   chan struct{}
 }
 
@@ -51,7 +51,7 @@ func NewLogger(ctx context.Context) (*Logger, error) {
 		api:      client,
 		interval: time.Second,
 		clock:    clock.RealClock{},
-		wg:       &sync.WaitGroup{},
+		wg:       sync.WaitGroup{},
 		stopCh:   make(chan struct{}, 1),
 	}
 
@@ -67,9 +67,11 @@ func NewLogger(ctx context.Context) (*Logger, error) {
 // Disclose adds a message to the log queue.
 // The messages are flushed periodically to AWS Cloudwatch Logs.
 func (l *Logger) Disclose(msg string) {
+	l.mux.Lock()
+	defer l.mux.Unlock()
 	l.logs = append(l.logs, types.InputLogEvent{
 		Message:   aws.String(msg),
-		Timestamp: aws.Int64(time.Now().UnixMilli()),
+		Timestamp: aws.Int64(l.clock.Now().UnixMilli()),
 	})
 }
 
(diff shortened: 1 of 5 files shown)


internal/cloud/aws/logger.go
```diff
@@ -6,20 +6,192 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+		case <-ticker.C():
+		case <-l.stopCh:
+			return
+		}
+	}
+}
+
+// createStream creates a new log stream in AWS Cloudwatch Logs.
+func (l *Logger) createStream(imds imdsAPI) error {
+	ctx := context.Background()
```
katexochen, line 138: Context should be passed through constructor into this func.

internal/cloud/aws/logger.go
```diff
@@ -6,20 +6,192 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+	if l.groupName == "" {
+		return fmt.Errorf("failed to find log group for UID %s", uid)
+	}
+
+	// create or use existing log stream
+	_, err = l.api.CreateLogStream(ctx, &logs.CreateLogStreamInput{
+		LogGroupName:  &l.groupName,
+		LogStreamName: &l.streamName,
+	})
+	if err != nil {
```
katexochen, line 181:
```suggestion
    in := logs.CreateLogStreamInput{
		LogGroupName:  &l.groupName,
		LogStreamName: &l.streamName,
	}
	if _, err = l.api.CreateLogStream(ctx, &in); err != nil {
```

internal/cloud/aws/logger.go
```diff
@@ -6,20 +6,192 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+		wg:       &sync.WaitGroup{},
+		stopCh:   make(chan struct{}, 1),
+	}
+
+	if err := l.createStream(imds.New(imds.Options{})); err != nil {
+		return nil, err
+	}
+
+	l.wg.Add(1)
+	go l.flushLoop()
```
katexochen, line 63:
```suggestion
	l.startFlushLoop()
```
Imo it is better to keep the concurrency stuff in one place.

internal/cloud/aws/logger.go
```diff
@@ -6,20 +6,192 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+	l.sequenceToken = res.NextSequenceToken
+
+	l.logs = nil
 	return nil
 }
+
+// flushLoop periodically flushes the logs to AWS Cloudwatch Logs.
+func (l *Logger) flushLoop() {
+	defer l.wg.Done()
+	ticker := l.clock.NewTicker(l.interval)
```
katexochen, line 124:
```suggestion
	ticker := l.clock.NewTicker(l.interval)
	defer ticker.Stop()
```

internal/cloud/aws/logger.go
```diff
@@ -6,20 +6,192 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+		// If the flush operation was called on a pre-existing stream,
+		// or another operation sent logs to the same stream,
+		// the sequence token may not be set correctly.
+		// We can retrieve the correct sequence token from the error message.
+		var sequenceErr *types.InvalidSequenceTokenException
+		if errors.As(err, &sequenceErr) {
+			l.sequenceToken = sequenceErr.ExpectedSequenceToken
+			l.flushMux.Unlock()
+			err = l.flushLogs()
+			l.flushMux.Lock()
```
katexochen, line 110: this doesn't work.

daniel-weisse: What do you mean by "this doesn't work"?
Because unless there is some weird side effect created by this, this does what I expect it to do.
`flushLogs` is called again without causing a panic by misusing the mutex.

If the unlocking/relocking is a problem, we could move the repeated flush call to a deferred function call after the mutex has been unlocked.

internal/cloud/aws/logger.go
```diff
@@ -6,20 +6,194 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+
+	return l, nil
 }
 
-// Disclose is not implemented for AWS.
-func (l *Logger) Disclose(msg string) {}
+// Disclose adds a message to the log queue.
+// The messages are flushed periodically to AWS Cloudwatch Logs.
+func (l *Logger) Disclose(msg string) {
+	l.logs = append(l.logs, types.InputLogEvent{
```
katexochen, line 70: I think the write on `l.logs` should be synchronized, too.

katexochen (approved):
Looks great, thanks! :seal: Maybe you could add some simple concurrency tests for the synchronized methods.

PR #235, opened 2022-10-12
Replace deprecated set-output syntax

katexochen:
### Proposed change(s)

- GitHub deprecated `set-output` syntax for workflows/actions yesterday, see [this announcement](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).
- Add dispatch trigger to every workflow. Sometimes workflows aren't triggered because the targeted files aren't changed.

--

Documenting the find/sed command I used here in case I need it again.

```sh
find .github -type f -name "*.yml" -exec sed -i 's/::set-output name=\([^::]*\)::\([^"]*\)/\1=\2 >> $GITHUB_OUTPUT/' {} +
```

### DIFF
--- a/.github/actions/constellation_create/action.yml
+++ b/.github/actions/constellation_create/action.yml
@@ -148,7 +148,7 @@ runs:
       id: constellation-init
       run: |
         constellation init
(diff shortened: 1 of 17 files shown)


datosh (approved):
🚀

PR #242, opened 2022-10-13
Apply UID tags to all applicable GCP resources

daniel-weisse:
### Proposed change(s)

- Apply the `constellation-uid` tag to all applicable resources created during `constellation create`
- Apply `constellation-use` label to loadbalancer resources, so they can be differentiate by means other than their name
- When choosing the Kubernetes loadbalancer, verify the `constellation-uid` label as well as verify that the `constellatio-use` labe is set to `kubernetes`

### Checklist

- [ ] Link to Milestone


daniel-weisse:
I also added the `constellation-role` to the instance labels.
IMO we should keep using instance metadata to retrieve UID and role information instead of checking the labels.
Open to removing role from labels

katexochen:
>  we should keep using instance metadata to retrieve UID and role information

why

daniel-weisse:
I think GCP mostly uses labels for visibility and organization purposes.
Metadata would be the "correct" way to pass information to a VM.

Of course, nothing is preventing us from using labels how we want

katexochen:
> Metadata would be the "correct" way to pass information to a VM.

You cannot do this on a scaling group etc, but we **must** get things like role etc for multiple resources, not only VMs. Therefore metadata is a bad choice, as I already said in our last talk on this matter.

daniel-weisse:
Currently we set `constellation-role`, `kube-env`, and `constellation-uid` using metadata.
If people are in favor of using labels/tags wherever possible, we can completely move role and uid from metadata to labels.

katexochen:
We should use the new tags in the operator to identify the the scale set role etc. without name parsing etc. Do you want to implement it in this PR or in another?

daniel-weisse:
Would prefer to do further refactoring in a new PR

katexochen (approved):
lgtm :tulip:

PR #249, opened 2022-10-13
Fix config check for debug image

katexochen:


Nirusu (commented):
I don't know, I would keep the old check for consistency and just make it case insensitive via `(?i)` (if everything can indeed be insensitive) or partially insensitive for every part we adjust before passing to Terraform.

Personally I don't like using too much Regex for simple tasks, too. But in this case it seems somewhat warranted as a sanity check for the config value, and now it's inconsistent with the GCP implementation.

katexochen:
It checks precisely on the image gallery. Any addition to that wouldn't add benefit.

Nirusu:
Doesn't check if it's a community gallery, though.

katexochen:
You want to cover the case where someone names any resource with the UUID of this gallery?

Nirusu:
Current check does that just fine though... I don't really see the benefit of weakening the check here just to avoid checking casing.

katexochen:
It's not a weakening. It's a code simplification while keeping the check equally strong (based on UUID).

Nirusu:
It's not equally as strong as:
`/communityGalleries/ConstellationCVM-b3782fa0-0df7-4f2f-963e-fc7fc42663df/imaGEes/consteLLation/verSioNs/2.1.0`
would pass the check but does not work in reality ;)

PR #233, opened 2022-10-11
Use Terraform in CLI for Azure

katexochen:
### Proposed change(s)

- The `constellation create` on Azure now uses Terraform to create and destroy cloud resources.
- Removed `endpoint` flag of `constellation init`. IP is now always taken from the `constellation-id.json` file.
- Removed `constellation-state.json` file won't be created anymore. Resources are now managed through Terraform.

### Checklist

- [x] Update [CHANGELOG.md](https://github.com/edgelesssys/constellation/blob/main/CHANGELOG.md)
- [x] Link to Milestone

PR #256, opened 2022-10-13
Case-insensitive replace for every user-input that can go wrong

Nirusu:
### Proposed change(s)
The azurerm provider has it's own convention of enforcing spelling which Azure itself does not enforce or even provide. This is a bit of bad design decision, but we can work around that. While #233 already had basic re-writing, I thought we can get a bit further to fix all edge-cases (at least the ones I see) plus keeping the #249 regex searchable for maintanence. The result is this lunacy of regex.ReplaceAllString :)  

@katexochen Sorry!

### Related issue
- Can supersede #249 

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/cli/internal/cloudcmd/create.go
+++ b/cli/internal/cloudcmd/create.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"net/url"
 	"os"
+	"regexp"
 	"runtime"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


katexochen (approved):

Nirusu:
Will also need to make the measurement fetch code case-insensitive, but I was planning changes for that anyways as part of distribution changes.

Not sure whether I want to merge this before or not.

PR #278, opened 2022-10-14
Use worker count to create workers on azure (instead of control plane count)

malt3:
### Proposed change(s)

- Use worker count to create workers on azure (instead of control plane count)


katexochen (approved):
lgtm

PR #280, opened 2022-10-14
Downgrade vale action

malt3:
### Proposed change(s)

- Downgrade vale action

### Additional info
- Slipped in by https://github.com/edgelesssys/constellation/pull/237
- @thomasten: 
    > Seems we updated the Vale action to a new major version. In addition to treating Warnings as Errors (which may be configured somehow with the new version), it also includes a vale version that produces some FPs. So I'd suggest to downgrade the action to the previous major version.


katexochen (approved):
lgtm.

PR #281, opened 2022-10-14
Allow concurrent actions on the same branch.

malt3:
Actions are free for public repos and we want to see every CI failure

### Proposed change(s)

- Allow concurrent actions on the same branch.


katexochen (approved):
lgtm :sunflower:

PR #285, opened 2022-10-15
Brush over MiniConstellation in docs and README

flxflx:
### Proposed change(s)
* Shorten MiniConstellation ("First steps (local)") page 
* Add links from README
* Take minor updates from Next to 2.1 in docs


m1ghtym0 (approved):
link error is based on new url introduced by this pr

datosh (approved):
LGTM, only two minor suggestions.

internal/versions/versions.go
```diff
@@ -61,7 +61,7 @@ const (
 	LibvirtImage      = "ghcr.io/edgelesssys/constellation/libvirt:v2.1.0@sha256:9769a2b88e2acc0986fdf2ab4a358b60c50a8cd330c183df4435f56e10758b37"
 
 	// ConstellationQEMUImageURL is the artifact URL for QEMU qcow2 images.
-	ConstellationQEMUImageURL = "https://d1gl9j3ejrmbpr.cloudfront.net/mini-constellation-v2.1.0.qcow2"
+	ConstellationQEMUImageURL = "https://d1gl9j3ejrmbpr.cloudfront.net/first-steps-local-v2.1.0.qcow2"
```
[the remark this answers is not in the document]
thomasten: That's not a nit, but a blocker for this PR ;)

README.md
```diff
@@ -56,14 +56,15 @@ Encrypting your K8s is good for:
 <a href="https://landscape.cncf.io/?selected=constellation"><img src="https://raw.githubusercontent.com/cncf/artwork/1c1a10d9cc7de24235e07c8831923874331ef233/projects/kubernetes/certified-kubernetes/versionless/color/certified-kubernetes-color.svg" align="right" width="100px"></a>
 
 * Constellation is a [CNCF-certified][certified] Kubernetes. It's aligned to Kubernetes' [version support policy][k8s-version-support] and will likely work with your existing workloads and tools.
-* Support for Azure and GCP, more to come.
+* Support for Azure and GCP (more to come).
+* Support for local and edge installations with [MiniConstellation][first-steps-local].
```
thomasten, line 60: @datosh @m1ghtym0 Do you think it's ok to promote minicon for edge installations? (And also to compare it with K3s and MicroK8s?) It seems a bit contradictory to our reply to the bare metal issue. Also, as I understood @daniel-weisse, a future edge variant of Constellation would probably not be based on minicon, but something new on the underlying "qemu create".

internal/versions/versions.go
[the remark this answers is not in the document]
flxflx: :o

README.md
datosh: I have the same question. I have added this to our product sync agenda, since I don't have a clear answer myself, but we should find a common understanding here.

daniel-weisse: Agree with Thomas, MiniConstellation should be used for testing only.
We should maybe just add some details for QEMU based Constellation (e.g. MiniConstellation but you perform init yourself), which we can then maybe advertise for edge deployments.
In that case, it would also be great if we can at least offer *some* support for CC in QEMU by supporting SEV-ES, and keeping track of changes to QEMU that would allow us to enable SEV-SNP

flxflx: Ok, I'll remote the edge thing...

PR #296, opened 2022-10-17
Make renovate tidy all go modules

katexochen:

### Proposed change(s)

- Currently, renovate doesn't tidy submodules when updating Go dependencies. This change tries to add this functionality, so you don't have to manually update submods.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -10,9 +10,17 @@
     "dependencies"
   ],
   "postUpdateOptions": [
-    "gomodTidy",
     "gomodUpdateImportPaths"
   ],
+  "postUpgradeTasks": {
+    "commands": [
+      "for mod in $(go list -f '{{.Dir}}' -m); do (cd $mod; go mod tidy); done"
+    ],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -10,9 +10,18 @@
...
   ],
+  "postUpgradeTasks": {
+    "commands": [
+      "for mod in $(go list -f '{{.Dir}}' -m); do (cd $mod; go mod tidy); done"
+    ],
+    "fileFilters": [
+      "**/go.mod",
+      "**/go.sum"
+    ],
+    "executionMode": "update"
```
datosh, line 23:
```suggestion
```
update is default so we can also omit. omitting does not make the config less clear to me.

datosh (approved):
I read that it is possible to download renovate and execute it locally with a dry-run flag.
I have used https://github.com/datosh/some-action as a test repository to tinker with renovate. 

I think the change looks simple enough so we can test this on main!

PR #318, opened 2022-10-18
Update release.md

katexochen:
### Proposed change(s)

- Some improvements from the last release session. Maybe a good preparation for moving the release to CI.


malt3 (approved):
When automating this, we should try to quote everything for safety (and maybe use `${var}` instead of `$var`). But looks good for now :)

PR #319, opened 2022-10-18
Tidy modules on renovate branches

katexochen:

### Proposed change(s)

- Checkout branch instead of head commit, so git is not in detached state.
- Tidy Go mods and commit on renovate branches.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # tag=v3.1.0
         with:
-          ref: ${{ github.event.pull_request.head.sha }}
+          ref: ${{ github.head_ref }}
 
       - name: Setup Go environment
         uses: actions/setup-go@c4a742cab115ed795e34d4513e2cf7d472deb55f # tag=v3.3.1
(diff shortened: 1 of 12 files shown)


malt3 (approved):
LGTM

katexochen:
@malt3 @datosh This changes seems to have a major drawback, as workflow runs aren't triggered after the tidy action pushes new changes, as GitHub recognizes the token and won't run workflows for git actions that come from a GitHub runner. (Thanks @malt3 for researching.) Imo having the CI green after tidying is worse than having to tidy by hand.

We can either

1. Use a personal access token we store in GitHub secrets to commit, so GitHub won't recognize that the change is pushed from within an action, or
2. Live with tidying by hand, or
3. host the renovate bot ourself.

datosh:
> Imo having the CI green after tidying is worse than having to tidy by hand.

I agree.

I think 2) just won't work, if we have to do it manually for each dependency updated. 

1) & 3) seem to have the same security implications. Both 1) and a [self-hosted renovate](https://github.com/renovatebot/github-action#token) require a PAT. I would lean towards 1). 

[I read that GitHub launched fine grained access tokens yesterday.](https://github.blog/2022-10-18-introducing-fine-grained-personal-access-tokens-for-github/) Maybe we can use such a token with limited scope to keep the risks of abuse low.

katexochen:
Okay, I will try (1) tomorrow.

PR #329, opened 2022-10-19
AB#2510 disk-mapper AWS support

daniel-weisse:
### Proposed change(s)

- Add support for AWS to the disk-mapper

### Additional info
The device name of the state disk is expected to be `/dev/sdb`.
Linux may not respect the selected device name and instead choose a difference prefix. E.g. `/dev/sdb` may be available as `/dev/xvdb` or `/dev/hdb`.
On Nitro instances, disks are attached over NVMe, this means disks are named using the `/dev/nvmeXnY` naming scheme.
Naming is not consistent across instance reboots.

However, AWS provides udev rules to fix this problem: https://github.com/amazonlinux/amazon-ec2-utils
When installed, these rules automatically create symlinks for the disks at the path selected as the device name.
This means regardless of where the actual device is, or on what platform we are, when we attach a disk with the device name `/dev/sdb`, and the udev rules are installed, we can use the disk as `/dev/sdb`.

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/disk-mapper/cmd/main.go
+++ b/disk-mapper/cmd/main.go
@@ -15,7 +15,6 @@ import (
 	"net/http"
 	"net/url"
 	"path/filepath"
-	"strings"
 
 	"github.com/edgelesssys/constellation/v2/disk-mapper/internal/mapper"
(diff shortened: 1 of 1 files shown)


malt3 (approved):
LGTM

PR #331, opened 2022-10-19
AB#2508 Use release cosign key only when releasing

datosh:

### Proposed change(s)

- Configured a second cosign key-pair `COSIGN_DEV_*` as action secret to be used for non-release builds
  - Keypair and password was also stored in internal secrets manager

### Additional info
- I tried to use `KEY: ${{ github.ref_protected && 'PROD KEY' || 'DEV KEY' }}`, but `github.ref_protected` [seems to be `false` on protected tags.](https://github.com/datosh/some-action/actions/runs/3281988742/jobs/5404766209) I suspect this is because [protected tags are still in beta](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-tag-protection-rules). I [checked with GitHub folks](https://github.com/community/community/discussions/10906#discussioncomment-3915816). I don't think we need to wait for an answer. We can change implementation later.

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/.github/workflows/generate-measurements.yml
+++ b/.github/workflows/generate-measurements.yml
@@ -58,9 +58,9 @@ jobs:
           azureResourceGroup: ${{ steps.az_resource_group_gen.outputs.res_group_name }}
           coreosImage: ${{ github.event.inputs.coreosImage }}
           isDebugImage: ${{ github.event.inputs.isDebugImage }}
-          cosignPublicKey: ${{ secrets.COSIGN_PUBLIC_KEY }}
-          cosignPrivateKey: ${{ secrets.COSIGN_PRIVATE_KEY }}
-          cosignPassword: ${{ secrets.COSIGN_PASSWORD }}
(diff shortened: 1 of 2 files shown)


thomasten (commented):
LGTM, but not confident enough of my GH actions knowledge to approve.

datosh:
I did some testing in a custom repository beforehand, @thomasten . A minimal example to test this functionality, using the mentioned `ref_protected` condition instead of `startsWith`:
* [On protected branch](https://github.com/datosh/some-action/actions/runs/3281926637/workflow)
* [On not-protected branch](https://github.com/datosh/some-action/actions/runs/3281954344/workflow)
You can see the same workflow file produces a different message in the `ternary.Decide` step.

thomasten (approved):
Thanks, the explanation was helpful to understand this better.

malt3 (approved):
LGTM

PR #333, opened 2022-10-19
AB#2511 Implement minimal feature support for bootstrapper on AWS

daniel-weisse:
### Proposed change(s)

- Bootstrapper no longer panics when CSP is set to `aws`
- Implement missing metadata methods (some stubs) for AWS to satisfy required interfaces 

### Additional info
- While this should allow booting a Constellation image with bootstrapper on AWS, `init` calls will not work and likely cause the bootstrapper to panic
- Missing features for full operation:
  - TPM based attestation
  - Cloud/Metadata functions required to install and run Cilium

### Checklist

- [ ] Link to Milestone


3u13r (approved):
LGTM 
The output of `GetSubnetworkCIDR` is actually not used on azure and will likely also not be used on aws. Will make the necessary changes in my cilium integration PR.

PR #334, opened 2022-10-19
initial draft for automatic updates

3u13r:
Thanks to @malt3 for the joint work on this!

### DIFF
--- a/rfc/updates.md
+++ b/rfc/updates.md
@@ -38,7 +38,7 @@ All Constellation microservices will be bundled into and therefore updated via o
 
 ## Extending the JoinService
 
-The CLI will use a lookup table to map the Kubernetes version from the config to URLs and hashes. Those are sent over during `constellation init` and used by the first bootstrapper. Then, the URLs and hashes are pushed to the `k8s-components-1.23.12` ConfigMap and the Kubernetes version with a reference to the `k8s-components-1.23.12` ConfigMap is pushed to `k8s-versions`.
+The CLI will use a lookup table to map the Kubernetes version from the config to URLs and hashes. Those are sent over during `constellation init` and used by the first Bootstrapper. Then, the URLs and hashes are pushed to the `k8s-components-1.23.12` ConfigMap and the Kubernetes version with a reference to the `k8s-components-1.23.12` ConfigMap is pushed to `k8s-versions`.
 
 ```yaml
 apiVersion: v1
@@ -48,15 +48,16 @@ metadata:
   namespace: kube-system
 data:
   k8s-version: "1.23.12"
-  components: "k8s-components-1.23.12" # This references the ConfigMap below.
+  components: "k8s-components-1.23.12-sha256-8ae09b7e922a90fea7a4259fb096f73e9efa948ea2f09349618102a328c44b8b" # This references the ConfigMap below.
 ```
 
 ```yaml
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: k8s-components-1.23.12
+  name: k8s-components-1.23.12-sha256-8ae09b7e922a90fea7a4259fb096f73e9efa948ea2f09349618102a328c44b8b
   namespace: kube-system
+immutable: true
 data:
   kubeadm:
     url: "https://storage.googleapis.com/kubernetes-release/release/v1.23.12/bin/linux/amd64/kubeadm"
@@ -81,7 +82,28 @@ data:
     hash: "sha256:2f04ce10b514912da87fc9979aa2e82e3a0e431c14fc0ea5c7c7209de74a1491"
 ```
 
-The JoinService will consume the `k8s-components-1.23.12` ConfigMap in addition to the `k8s-version` ConfigMap. If a new node wants to join the cluster, the JoinService looks up the current Kubernetes version and all the component download URLs and hashes and sends them to the joining node.
+The JoinService will consume the `k8s-components-1.23.12` ConfigMap in addition to the `k8s-version` ConfigMap. Currently, the `k8s-version` ConfigMap is mounted into the JoinService pod. We will change that so that the JoinService requests the ConfigMap values via the Kubernetes API. If a new node wants to join the cluster, the JoinService looks up the current Kubernetes version and all the component download URLs and hashes and sends them to the joining node.
+
+## Extending the Bootstrapper
+
+During the cluster initialization we need to create the first ConfigMap with components and hashes.
+We receive all necessary information from the CLI in the first place, since we need to download them to create a initialize the cluster in the first place.
+
+To be able to even update singular components, we need to know if the set of components of a node is the desired one. To achieve that, the Bootstrapper will calculate a hash of all the components' hashes. The node then labels itself during `kubeadm join` with this hash. The hash will later be read by the node operator. The first Bootstrapper needs to be labeled during `kubeadm init`.
+
+```yaml
+apiVersion: kubeadm.k8s.io/v1beta2
+kind: JoinConfiguration
+discovery:
+  bootstrapToken:
+    token: XXX
+    apiServerEndpoint: "XXX"
+    caCertHashes: ["sha256:123456789012345678901234567890564b934be406f13e28f118b32cc0b6e6db"]
+nodeRegistration:
+  name: worker-node-1
+  kubeletExtraArgs:
+    node-labels: "edgeless.systems/kubernetes-components-hash="sha256:8ae09b7e922a90fea7a4259fb096f73e9efa948ea2f09349618102a328c44b8b""
+```
 
 ## Creating an upgrade agent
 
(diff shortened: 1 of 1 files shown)


rfc/updates.md
```diff
@@ -0,0 +1,199 @@
...
+
+```yaml
+upgrade:
+  image: /communityGalleries/ConstellationCVM-b3782fa0-0df7-4f2f-963e-fc7fc42663df/images/constellation/versions/2.3.0
+  measurements:
+    11: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
+    12: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
+```
+
+Instead of having a separate `upgrade` section, we will opt for a declarative approach by updating the existing values of the config file.
```
daniel-weisse, line 150: The original idea behind having a separate upgrade section was to discourage users from accidentally trying to change config values which can not be upgrade, e.g. increasing state disk size or VM instance types.

If we want to integrate upgradeable values into the main config, it should be made clear to the users which values are safe to change, and which are not.
Ideally users dont have to interact with their config file at all, and instead are able to use just the CLI commands to perform upgrades

rfc/updates.md
```diff
@@ -0,0 +1,199 @@
...
+    hash: "sha256sum:0d96a4ff68ad6d4b6f1f30f713b18d5184912ba8dd389f86aa7710db079abcb0"
+  crictl:
+    url: "https://github.com/kubernetes-sigs/cri-tools/releases/download/v1.24.1/crictl-v1.24.1-linux-amd64.tar.gz"
+    hash: "sha256:ebaea1c7b914cdd548012c6cba44f8d760fd0c7915274ffecd6d764957aac83c"
+  cniplugins:
+    url: "https://github.com/containernetworking/plugins/releases/download/v1.1.1/cni-plugins-linux-amd64-v1.1.1.tgz"
+    hash: "sha256:2f04ce10b514912da87fc9979aa2e82e3a0e431c14fc0ea5c7c7209de74a1491"
+```
+
+## Creating a upgrade agent
```
daniel-weisse, line 87:
```suggestion
## Creating an upgrade agent
```

rfc/updates.md
```diff
@@ -0,0 +1,199 @@
...
+The current manual approach is:
+
+* ssh into a control-plane node
+* curl new kubeadm version (see `versions.go`) to `/var/run/state/bin/kubeadmNew`
+* `mv /var/run/state/bin/kubeadmNew /var/run/state/bin/kubeadm && chmod +x /var/run/state/bin/kubeadm`
+* `export PATH=$PATH:/var/run/state/bin/`
+* `kubeadm upgrade plan`
+* `kubeadm upgrade apply vX.Y.Z`
+* bump K8s version in configmap/k8s-version to `vX.Y`
+* [upgrade your cluster](https://constellation-docs.edgeless.systems/constellation/workflows/upgrade) with a new VM image
```
thomasten, line 27:
```suggestion
* [upgrade your cluster](https://docs.edgeless.systems/constellation/workflows/upgrade) with a new VM image
```
nit: please avoid the deprecated subdomain

rfc/updates.md
3u13r: > If we want to integrate upgradeable values into the main config, it should be made clear to the users which values are safe to change, and which are not.

Do you have a implementation in mind that would make it clear(er)? We could move all "update-able" values in their own group. 

> Ideally users dont have to interact with their config file at all, and instead are able to use just the CLI commands to perform upgrades

If the user is not supposed to touch the config file, how should Constellation be integrated in an GitOps style deployment pipeline?

rfc/updates.md
```diff
@@ -0,0 +1,199 @@
...
+
+Since the actual Kubernetes components and Constellation microservice versions are hidden, we will show the user for the actual changes taking place:
+
+```bash
+$ constellation upgrade execute
+Updating wanted Kubernetes version to 1.24.3 ...
+Updating wanted Image to /communityGalleries/ConstellationCVM-b3782fa0-0df7-4f2f-963e-fc7fc42663df/images/constellation/versions/2.3.0
+
+Updating Kubernetes services version to 1.24.5:
+  Autoscaler: 1.24.3 --> 1.24.3
```
datosh, line 185: Should we add an additional note, e.g., `(not updated)`, to clearly show the user that keeping the same version is correct and not a typo/bug?

rfc/updates.md
```diff
@@ -0,0 +1,199 @@
...
+### Constellation microservices
+
+All Constellation microservices will be bundled into and therefore updated via one helm chart.
+
+
+# Automatic Updates
+
+## Extending the JoinService
+
+The CLI will use a lookup table to map the Kubernetes version from the config to URLs and hashes. Those are sent over during `constellation init` and used by the first bootstrapper. Then, the URLs and hashes are pushed to the `k8s-components-1.23.12` ConfigMap and the Kubernetes version with a reference to the `k8s-components-1.23.12` ConfigMap is pushed to `k8s-versions`.
```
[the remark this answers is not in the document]
malt3: Re 2:

A single config map might create merge conflicts in kubernetes. You only want one component modifying each config map.
As we also don't want the operator to know the inner details of kubernetes component versioning, the idea is to have the CLI prepare version config maps and let it hand out a reference to that config to the operator. The  operator is then the only entity modifying the k8s-version configmap.

Re 3:

Using a unique name for components might be a good idea.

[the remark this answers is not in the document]
3u13r: 1. Will add the explanation. 
2. We thought about that too. It is easier to separate the URLs in their own ConfigMap since we need the operator to update the k8s-version but the components list will be created by the Bootstrapper (first list) and CLI (on updates). We don't want to implement any reconcile loops there. 
3. That's a good point. I overlooked that some of the components are not released in sync with Kubernetes itself. We need to re-think some part now, since the check if we need to replace a node needs to consider more than the Kubelet version itself. Maybe the node is labeled with some form of Hash(ComponentsHashes) which then the node operator can check against? @malt3 What do you think?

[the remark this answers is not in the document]
malt3: > That's a good point. I overlooked that some of the components are not released in sync with Kubernetes itself. We need to re-think some part now, since the check if we need to replace a node needs to consider more than the Kubelet version itself. Maybe the node is labeled with some form of Hash(ComponentsHashes) which then the node operator can check against? @malt3 What do you think?

Yes. Labeling a node with a hash is a great idea as it integrates well with the operator. This also make comparing the reference in the k8s-version config map to the node state easier (as we can literally compare the hashes).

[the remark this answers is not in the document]
3u13r: >  I believe this can be done on kubeadm join but will have to check.

https://blog.scottlowe.org/2020/11/30/assigning-node-labels-during-kubernetes-cluster-bootstrapping/
Looks like it. I'd add to the spec that we label each node with `edgeless.systems/kubernetes-components-hash:"sha256:8ae09b7e922a90fea7a4259fb096f73e9efa948ea2f09349618102a328c44b8b"` and rename the components ConfigMap to `k8s-components-1.23.12-8ae09b7e922a90fea7a4259fb096f73e9efa948ea2f09349618102a328c44b8b` or should we crop the hash or just use a UID? (will do the hash for now, can discuss in new comment chain)

rfc/updates.md
```diff
@@ -0,0 +1,193 @@
...
+    hash: "sha256sum:0d96a4ff68ad6d4b6f1f30f713b18d5184912ba8dd389f86aa7710db079abcb0"
+  crictl:
+    url: "https://github.com/kubernetes-sigs/cri-tools/releases/download/v1.24.1/crictl-v1.24.1-linux-amd64.tar.gz"
+    hash: "sha256:ebaea1c7b914cdd548012c6cba44f8d760fd0c7915274ffecd6d764957aac83c"
+  cniplugins:
+    url: "https://github.com/containernetworking/plugins/releases/download/v1.1.1/cni-plugins-linux-amd64-v1.1.1.tgz"
+    hash: "sha256:2f04ce10b514912da87fc9979aa2e82e3a0e431c14fc0ea5c7c7209de74a1491"
+```
+
+The JoinService will consume the `k8s-components-1.23.12` ConfigMap in addition to the `k8s-version` ConfigMap. If a new node wants to join the cluster, the JoinService looks up the current Kubernetes version and all the component download URLs and hashes and sends them to the joining node.
```
malt3, line 84: This places a requirement on the joinservice to now read config maps via the k8s API (instead of reading a mounted file).
I believe that this is the right way to go. Just want to emphasize that change.

datosh:
meeting: do we need to take care of rotating root CA? or other certificates?

malt3 (approved):
LGTM

datosh (approved):

thomasten (approved):

PR #338, opened 2022-10-20
Let tfsec fail soft in CI

katexochen:

### Proposed change(s)

- Let tfsec comment, but not fail CI.


Nirusu (approved):
Just need to make sure someone actually still looks at the results then ;) But I guess ignoring specific rules is also a bit of a pain or cause some accidental oversights by ignoring too much.

Thanks.

katexochen:
> Just need to make sure someone actually still looks at the results then

I guess a fist step would be to not resolve the comments as PR opener but let the reviewer resolve, so at least 2 people looked at it.

PR #341, opened 2022-10-20
Refactor Helm deployments

derpsteb:
### Proposed change(s)
- Use .tgz as serializing format instead of json because Chart dependencies are a private property. Therefore dependencies are lost with a `json.Marshal()` -> `json.Unmarshal()`.
- Move KMS chart into constellation-services chart that will hold all of our microservices in the future.


cli/internal/helm/charts/edgeless/constellation-services/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+name: constellation-services
+description: A chart to deploy all microservices that are part of a valid constellation cluster
+type: application
+version: 0.1.0
+appVersion: "1.16.0"
+
+dependencies:
+  - name: kms
+    version: "0.1.0"
```
daniel-weisse, line 10: How are you determining these version numbers?
I think they should be tied to our current Constellation release version, e.g. v2.2.0-pre on main, and v2.2 .0 for our next release

cli/internal/helm/charts/edgeless/constellation-services/charts/kms/values.yaml
```diff
@@ -1,13 +1,13 @@
...
 # Port on which the service will listen.
-kmsPort: 9000
+port: 9000
 # Name of the ConfigMap that holds measurements and other info.
 joinConfigCMName: join-config
 # Path to which secrets/CMs are mounted.
 serviceBasePath: /var/config
 # Container image.
-kmsImage: setFullImagePathHere
+image: setFullImagePathHere
```
daniel-weisse, line 10: I guess this is a question for everyone:
Do we want to set our images here? Or would it be preferable to pass the values for images from Go code?

daniel-weisse (commented):
Question about the constellation-services values file:
This is currently empty. Is it possible to use this to pass down values to dependency charts?
E.g. set the image for kms in the global value file which is then set in the kms dependency.

Additionally, can it be used to enable/disable specific deployments?
This would make it easy to, for example, not deploy the node operator on QEMU clusters

cli/internal/helm/charts/edgeless/constellation-services/charts/kms/values.yaml
derpsteb: Just for completeness: they are being passed from Go code atm.

https://github.com/edgelesssys/constellation/pull/341/files#diff-540be4cd28ea6a60b8ecfc4c864805bdcb896893cb33f121bfad873a878e2783R110

3u13r: I am kinda torn on this one: bit I think I lean a bit more to moving them from `versions.go` into the helm charts to be able to use the charts standalone.

derpsteb: I like the versions.go file as it makes finding version information very easy.

daniel-weisse: Thoughts on just setting the image to `latest` in the helm charts, but keeping the actual used values in `versions.go`?
Since the most common use case for standalone helm charts is probably managing updates, users would need to adjust the image version in any case

derpsteb: Thinking about this a bit more I like the fact that this doesn't fail silently. You are forced to think about the image version you want to deploy. Isn't that useful to have?

daniel-weisse: Is there such a thing as a "required" field for helm values :thinking:

derpsteb: Possible, but would remove the value from the values.yaml. Is that fine for you?

daniel-weisse: Reading [the helm docs](https://helm.sh/docs/howto/charts_tips_and_tricks/#know-your-template-functions), it sounds to me as if we only need to have the value set to empty, and if not we throw an error.

This would result in even quicker errors (on deploy instead of image pull errors after deployment has already taken place)

derpsteb: Using schemas is probably the most fleshed out solution here. Added just now.

Regarding values in the global scope: With the upcoming join-service migration it should become more apparent why I think it make sense to keep chart specific values in the respective chart dir. The global values file can then be used for shared values. [Branch](https://github.com/edgelesssys/constellation/commit/9caa23acbcc731789f2634ea085b549fe1ec7e4f#diff-e6021db951db1bd26edea8fcab79b3eda4edc89fa94ab5d4cd246e15566c7311).

Not super happy with the way the helm-lint target has to be implemented, but I guess it's manageable.

daniel-weisse (approved):

cli/internal/helm/charts/edgeless/constellation-services/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+name: constellation-services
+description: A chart to deploy all microservices that are part of a valid constellation cluster
+type: application
+version: 2.2.0-pre
```
3u13r, line 5: We need to manually maintain those during the release process, right? Then we should mention this in the `release.md`.

derpsteb: Added.

PR #345, opened 2022-10-21
Feat/triggerd build

datosh:
### Proposed change(s)

-
-

### Checklist

- [ ] Link to Milestone


datosh:
@katexochen these sets of runs should prove that the `path` expressions are preventing our pipelines from running.

[This run shows that only my new action is triggered](https://github.com/edgelesssys/constellation/pull/345/commits/c5a154ed9828fb5e0e369fe2a1c46b813ab2f06a)

[The latest run has all path expressions removed and all actions are triggered](https://github.com/edgelesssys/constellation/pull/345/commits/4374ee6f0c57606a683d62972da90e9f97226106)

I don't exactly understand what is causing this. The [docs state](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore): If you define both branches/branches-ignore and [paths](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore), the workflow will only run when both filters are satisfied.

Not sure if we get something wrong (maybe the bot commit is different? or triggers different events?) or if this is a bug in GitHub actions...

datosh:
Can we close this one? @malt3 wrote that renovate with go module now works correctly.
cc @katexochen

malt3:
Yes!

katexochen:
Yes, can be closed, thanks again. :)

PR #349, opened 2022-10-21
Ref/operator/tag parsing

daniel-weisse:
### Proposed change(s)

- Removes parsing of resource names from the operator, instead rely on tags or labels
  - This comes with a slight caveat for GCP. Since instance group managers can not be labeled, we instead parse the referenced instance template and check if the template belongs to our Constellation, and what role it serves

### Checklist

- [ ] Link to Milestone


malt3 (approved):
LGTM

PR #350, opened 2022-10-21
update k8s conformance folder to the recent submission

thomasten:
I just copied the files from your submission back here, @katexochen. Alternative would be to delete the folder. Back then, @datosh and me decided to keep a copy of it in our own repo, but I think we both didn't have a too strong opinion on this.

### DIFF
--- a/conformance/PRODUCT.yaml
+++ b/conformance/PRODUCT.yaml
@@ -1,8 +1,9 @@
 vendor: Edgeless Systems
 name: Constellation
-version: v1.0.0
+version: v2.1.0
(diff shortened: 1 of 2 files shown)


katexochen (approved):
Also no strong opinions on this one. :shrug:

katexochen:
I think I would tend to just delete it. It is hard to remember to update, as we cannot put it in the manual we have in the conformance repo.

datosh:
Yeah makes sense to remove it! We can bring it back in case we decide to automate it.

PR #339, opened 2022-10-20
AB#2458 AWS NitroTPM attestation

leongross:
### Proposed change(s)

- Add NitroTPM attestation

### Additional info
- NitroTPM does not provide an `EKCert` for verification yet

### Checklist

- [ ] Update [CHANGELOG.md](https://github.com/edgelesssys/constellation/blob/main/CHANGELOG.md)
- [ ] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)
- [ ] Link to Milestone

PR #352, opened 2022-10-21
cilium AWS support

3u13r:
### Proposed change(s)

- Add AWS value set for cilium
- Implement AWS metadata for loadbalancer
- move AWS Terraform from launch config to launch template

Those are the most notable changes. There are also some minor things here and there. The commit messages should give a good overview over what happens in this PR. 

My biggest question is: how do we move forward regarding the mocked attestation on AWS? Should we merge this in full now and replace it late, or should we do something like merge everything from this PR except the mocked attestation?

### Checklist

- [x] Test
- [x] Merge similar values function with Azure
- [x] Extend AWS loadbalancer metadata test
- [x] Link to Milestone

### DIFF
--- a/internal/cloud/gcp/metadata.go
+++ b/internal/cloud/gcp/metadata.go
@@ -83,7 +83,18 @@ func (m *Metadata) Self(ctx context.Context) (metadata.InstanceMetadata, error) 
 	if err != nil {
 		return metadata.InstanceMetadata{}, err
 	}
-	return m.api.RetrieveInstance(ctx, project, zone, instanceName)
+	subnetCIDR, err := m.api.RetrieveSubnetworkAliasCIDR(ctx, project, zone, instanceName)
+	if err != nil {
+		return metadata.InstanceMetadata{}, err
+	}
+
+	instance, err := m.api.RetrieveInstance(ctx, project, zone, instanceName)
+	if err != nil {
+		return metadata.InstanceMetadata{}, err
+	}
+	instance.SubnetworkCIDR = subnetCIDR
+
+	return instance, nil
 }
 
 // GetInstance retrieves an instance using its providerID.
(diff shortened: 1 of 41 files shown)


internal/cloud/gcp/metadata.go
```diff
@@ -83,7 +83,18 @@ func (m *Metadata) Self(ctx context.Context) (metadata.InstanceMetadata, error)
...
+		return metadata.InstanceMetadata{}, err
+	}
+
+	instance, err := m.api.RetrieveInstance(ctx, project, zone, instanceName)
+	if err != nil {
+		return metadata.InstanceMetadata{}, err
+	}
+	instance.SecondaryIPRange = subnetCIDR
+
+	return instance, nil
```
daniel-weisse, line 97: Is the `SecondaryIPRange` information only relevant for listing `Self`, or would we also want this information when calling `List` or `GetInstance`?

daniel-weisse (commented):
Haven't tested any deployments yet, but code looks good to me.

Since https://github.com/edgelesssys/constellation/pull/339 basically only requires a rebase before it can be merged, I would say we wait and then bring Cilium into main with non mocked attestation.
Unless this PR is seriously blocking development right now, in that case I would be ok with merging now, and patching up attestation later.

internal/cloud/gcp/metadata.go
3u13r: Good point. I've only added it to `Self` since I know that this is the only place where it is consumed from. This is of course not a very good reason of only filling it that case. I think from an API/Interface point of view it makes sense to always fill it where possible. Having said that, I'd push this discussion/change to the general metadata refactoring.

daniel-weisse: I can add it to `List` and `GetInstance` :+1:

daniel-weisse (approved):

malt3 (approved):
I would also say #339 can be merged before this. Code LGTM.

PR #359, opened 2022-10-24
Dracut: use inst_rules to install udev rules

malt3:
### Proposed change(s)

- Dracut: use inst_rules to install udev rules

Initramfs generation works. Rules in the initramfs did not work previously anyways so there is no risk of a regression.


Nirusu:
Does this actually fix the rules?

malt3:
> Does this actually fix the rules?

Yes. Please note that on AWS we still need workarounds unless we want to package python in the initramfs

Nirusu:
Cool! 

Yeah I know about AWS. We could also just try to fix the udev rules there (e.g. call our Bash script or just the `nvme` command with the chain we use in the Bash script, since that should virtually replicate the behaviour of the Python script?) or keep the workaround. Only thing I would avoid is having dead rules installed since it might cause unintended behaviour.

But that's unrelated to this PR.

PR #360, opened 2022-10-24
AB#2532 Dont clean up workspace if rollback fails

daniel-weisse:
### Proposed change(s)

If an error occurred during rollback, terraform files were deleted, resulting in dangling resources that are difficult to clean up.
This adds a check to only clean up the workspace if rollback succeeded without errors

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/cli/internal/cloudcmd/clients.go
+++ b/cli/internal/cloudcmd/clients.go
@@ -10,10 +10,11 @@ import (
 	"context"
 
 	"github.com/edgelesssys/constellation/v2/cli/internal/terraform"
+	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
 )
 
 type terraformClient interface {
-	CreateCluster(ctx context.Context, name string, input terraform.Variables) (string, error)
+	CreateCluster(ctx context.Context, provider cloudprovider.Provider, name string, input terraform.Variables) (string, error)
 	DestroyCluster(ctx context.Context) error
 	CleanUpWorkspace() error
 	RemoveInstaller()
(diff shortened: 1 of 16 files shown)


malt3 (approved):
LGTM

Nirusu (changes requested):
I would maybe tell the user that if the rollback fails that they have to manually clean up the working directory. So far it only says "rollback failed" and UX-wise you are kinda in limbo since you don't know what to do know.

For example in `rollbackOnError`:
```
fmt.Println("Automatic rollback failed.")
fmt.Println("Please manually clean up any resources on your CSP and try running 'constellation terminate' again.")
```

Plus don't forget to allow `terminate` to be run when `constellation-id.json` is missing (since that file is only created when it's completed). 

I was accidentally working on this in parallel since I forgot to assign the ticket. You could maybe take a look at what I was thinking and copy it over here (or redo it):
https://github.com/edgelesssys/constellation/tree/fix/rollback

daniel-weisse:
>I was accidentally working on this in parallel since I forgot to assign the ticket. 

Do you want to merge your changes into this PR? If so, just create a PR from your branch to this one, and we can have a look.

Nirusu:
Can do that, will rebase it quickly and make a PR.

daniel-weisse:
>Note that this file is the last one that is cleaned up

Please be aware that this bug fix concerns the `create` command, which only writes an ID file after it exits successfully.
The current behavior removes any terraform files, leaving the user to manually clean up their cloud subscription.

This PR https://github.com/edgelesssys/constellation/pull/362 changes `terminate` to instead rely on the config file, which would allow us to call `terminate` after `create` failed.

malt3 (approved):
I tested 

- Canceling on create
- Continuing rollback
- rerunning terminate after everything is already cleaned up works

Nirusu (approved):

PR #361, opened 2022-10-24
AB#2533 Pin terraform provider hashes

malt3:
### Proposed change(s)

- Pin terraform provider hashes

### Additional info

Manually created using:

```
terraform providers lock -platform=linux_arm64 -platform=linux_amd64 -platform=darwin_arm64 -platform=darwin_amd64 -platform=windows_amd64
```

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/terraform/loader.go
+++ b/cli/internal/terraform/loader.go
@@ -19,6 +19,7 @@ import (
 )
 
 //go:embed terraform/*
+//go:embed terraform/*/.terraform.lock.hcl
 var terraformFS embed.FS
 
 // prepareWorkspace loads the embedded Terraform files,
--- a/cli/internal/terraform/terraform/aws/.terraform.lock.hcl
+++ b/cli/internal/terraform/terraform/aws/.terraform.lock.hcl
@@ -1,0 +1,50 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/hashicorp/aws" {
+  version     = "4.36.1"
+  constraints = "4.36.1"
+  hashes = [
+    "h1:04NI9x34nwhgghwevSGdsjssqy5zzvMsQg2Qjpmx/n0=",
+    "h1:3kC2d7G2MR6Jbbrv8v8L7TWyh1kWLLEwYzjrzxuUSio=",
+    "h1:bVdhic55ukDoSukFwOOqX2q/gZ5efe4aBTMGivEuY4o=",
+    "h1:joU7IYfdBSy43Ln1pDjpDJcDZPcHriVS6MG/OFtR+I4=",
+    "h1:sN3HFvBwuCn+ipD9Ti5OnBJ+V9CzUXviJ2py6tiCK6Q=",
+    "zh:19b16047b4f15e9b8538a2b925f1e860463984eed7d9bd78e870f3e884e827a7",
+    "zh:3c0db06a9a14b05a77f3fe1fc029a5fb153f4966964790ca8e71ecc3427d83f5",
+    "zh:3c7407a8229005e07bc274cbae6e3a464c441a88810bfc6eceb2414678fd08ae",
+    "zh:3d96fa82c037fafbd3e7f4edc1de32afb029416650f6e392c39182fc74a9e03a",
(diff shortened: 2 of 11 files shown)


cli/internal/terraform/loader.go
```diff
@@ -19,6 +19,7 @@ import (
 )
 
 //go:embed terraform/*
+//go:embed terraform/*/.terraform.lock.hcl
```
daniel-weisse, line 22: I don't think we need to explicitly tell Go to embed this file when we already embed the entire folder

cli/internal/terraform/terraform/aws/.terraform.lock.hcl
```diff
@@ -0,0 +1,50 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/hashicorp/aws" {
+  version     = "4.36.1"
+  constraints = "~> 4.0"
```
daniel-weisse, line 6: We should adjust the AWS version constraint to `"4.36.1"` instead of `"~> 4.0"`

daniel-weisse (commented):
I think we also talked about pinning the terraform binary version, not sure if we can provide a hash, though

cli/internal/terraform/loader.go
malt3: I tested this..

> If a pattern names a directory, all files in the subtree rooted at that directory are embedded (recursively), except that files with names beginning with ‘.’ or ‘_’ are excluded. So the variable in the above example is almost equivalent to:
source: https://pkg.go.dev/embed

we could use `//all:go:embed terraform/*` instead. I don't have preference here.

cli/internal/terraform/terraform/aws/.terraform.lock.hcl
malt3: It does not make a difference when it comes to hash pinning but we can do this to prevent accidental upgrades :+1: .

cli/internal/terraform/loader.go
Nirusu: Wouldn't use all here. Back then I opted against this since it might accidentally include other trash or hidden files.

If we include hidden files, it should be explicitly imo.

malt3:
> I think we also talked about pinning the terraform binary version, not sure if we can provide a hash, though

I looked through the source of [terraform-exec](https://github.com/hashicorp/terraform-exec). Our installation method already verifies the hashes of the terraform binary. The hashes are not pinned but signed by hashicorp instead.

[Installer.Ensure](https://pkg.go.dev/github.com/hashicorp/hc-install?utm_source=gopls#Installer.Ensure) calls [Downloader.DownloadAndUnpack](https://pkg.go.dev/github.com/hashicorp/hc-install/internal/releasesjson?utm_source=gopls#Downloader.DownloadAndUnpack) which internally performs the verification.

Nirusu:
Signed vs. pinned hashes has the issue though that supply chain attacks by Hashicorp are possible. If the signing key is compromised, someone can upload a maliciously signed hash and achieve unintended code execution.

Of course it's a more advanced attack, but this issue exists with signed hashes.

malt3:
There is no option for hash pinning in terraform-exec so I don't believe that this is feasible for now. I can open an issue on their Git and ask if such a feature can be added.

Nirusu:
True... We could build a wrapper but that's likely going to be ugly, especially since the terraform-exec / install package also checks for already existing installations so execution paths can vary.

Maybe we can ask. The Terraform package also already supports defining your own PGP key for verification, which is something we could (ab)use for that but would mean we also would need to redistribute their packages signed by us. 

But if they have this option, it shows that they might care about providing a list of pinned hashes, too.

Another question related, do we want to pin the used Terraform version here?:
https://github.com/edgelesssys/constellation/blob/8d6c9e270c29a5fd083e4859a00235e5ed4c1260/cli/internal/terraform/terraform.go#L27

malt3:
Tracking: https://github.com/hashicorp/hc-install/issues/72

malt3:
I moved hash validation of the terraform binary into a separate ticket AB#2534 to unblock this PR.

daniel-weisse (approved):

Nirusu (approved):

PR #362, opened 2022-10-24
Rollback fix: Allow terminate when constellation-id.json is missing

Nirusu:
### Proposed change(s)

- Prompt user they can run `constellation terminate again` to fix a failed rollback
- Make terminate independent of constellation-id.json and instead rely on config (could potentially be completely removed if we know how to deal with QEMU)
- Adapt unit tests to reflect the changes

The TestTerminate does not properly check for the Terraform terminator since the stub does not implement it correctly. It is tested by terraform_test.go. So I just adjusted the TestTerminate test to create a stub file to pass the "cluster does exist" test.

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/cli/internal/cloudcmd/rollback.go
+++ b/cli/internal/cloudcmd/rollback.go
@@ -29,6 +29,8 @@ func rollbackOnError(ctx context.Context, w io.Writer, onErr *error, roll rollba
 	fmt.Fprintln(w, "Attempting to roll back.")
 	if err := roll.rollback(ctx); err != nil {
 		*onErr = multierr.Append(*onErr, fmt.Errorf("on rollback: %w", err)) // TODO: print the error, or return it?
+		fmt.Println("Automatic rollback failed.")
+		fmt.Println("Please manually clean up any resources on your CSP and try running 'constellation terminate' again.")
 		return
 	}
 	fmt.Fprintln(w, "Rollback succeeded.")
--- a/cli/internal/cloudcmd/rollback.go
+++ b/cli/internal/cloudcmd/rollback.go
@@ -28,7 +28,7 @@ func rollbackOnError(ctx context.Context, w io.Writer, onErr *error, roll rollba
 	fmt.Fprintf(w, "An error occurred: %s\n", *onErr)
 	fmt.Fprintln(w, "Attempting to roll back.")
 	if err := roll.rollback(ctx); err != nil {
-		*onErr = multierr.Append(*onErr, fmt.Errorf("Rollback failed: %w", err)) // TODO: print the error, or return it?
+		*onErr = multierr.Append(*onErr, fmt.Errorf("rollback failed: %w", err)) // TODO: print the error, or return it?
 		fmt.Fprintln(w, "Try running 'constellation terminate', or manually clean up dangling resources on your CSP.")
 		return
 	}
--- a/cli/internal/cloudcmd/rollback.go
+++ b/cli/internal/cloudcmd/rollback.go
@@ -28,9 +28,8 @@ func rollbackOnError(ctx context.Context, w io.Writer, onErr *error, roll rollba
 	fmt.Fprintf(w, "An error occurred: %s\n", *onErr)
 	fmt.Fprintln(w, "Attempting to roll back.")
 	if err := roll.rollback(ctx); err != nil {
-		*onErr = multierr.Append(*onErr, fmt.Errorf("on rollback: %w", err)) // TODO: print the error, or return it?
-		fmt.Println("Automatic rollback failed.")
-		fmt.Println("Please manually clean up any resources on your CSP and try running 'constellation terminate' again.")
+		*onErr = multierr.Append(*onErr, fmt.Errorf("Rollback failed: %w", err)) // TODO: print the error, or return it?
+		fmt.Fprintln(w, "Try running 'constellation terminate', or manually clean up dangling resources on your CSP.")
 		return
 	}
 	fmt.Fprintln(w, "Rollback succeeded.")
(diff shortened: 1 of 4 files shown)


cli/internal/cloudcmd/rollback.go
```diff
@@ -29,6 +29,8 @@ func rollbackOnError(ctx context.Context, w io.Writer, onErr *error, roll rollba
 	fmt.Fprintln(w, "Attempting to roll back.")
 	if err := roll.rollback(ctx); err != nil {
 		*onErr = multierr.Append(*onErr, fmt.Errorf("on rollback: %w", err)) // TODO: print the error, or return it?
+		fmt.Println("Automatic rollback failed.")
+		fmt.Println("Please manually clean up any resources on your CSP and try running 'constellation terminate' again.")
```
daniel-weisse, line 33:
```suggestion
		*onErr = multierr.Append(*onErr, fmt.Errorf("Rollback failed: %w", err)) // TODO: print the error, or return it?
		fmt.Fprintln(w, "Try running 'constellation terminate', or manually clean up dangling resources on your CSP.")
```
Line above "Automatic rollback failed" already tells the user roll back failed.
Since the rollback action is only run during `create` we should not tell users to run `terminate` "again"

daniel-weisse (changes requested):
We can easily remove the dependency of terminate on any id or config file by always running `libvirt.Stop`, since the function first checks if the current directory has a name file for the docker container to stop, and if not simply returns nil.

Nirusu:
Not sure if it's trivial. Would require a bit of refactoring since we need the provider name at certain places (TerraformClient requires it, uses it inside `cleanUpWorkspace`).

daniel-weisse:
>since we need the provider name at certain places

Ah , you're right. Simply read your comment about being able to completely remove it if we knew how to deal with QEMU and assumed it wasn't used elsewhere.

Nirusu:
Yeah, I initially thought that was the case but was wrong. Got confused there for a moment 😄

cli/internal/cmd/terminate.go
```diff
@@ -45,13 +44,27 @@ func runTerminate(cmd *cobra.Command, args []string) error {
...
-	var idFile clusterid.File
-	if err := fileHandler.ReadJSON(constants.ClusterIDsFileName, &idFile); err != nil {
+	foundTerraformData, err := checkForTerraformData(fileHandler)
+	if err != nil {
 		return err
 	}
 
+	if !foundTerraformData {
+		return fmt.Errorf("did not find a cluster to terminate")
+	}
```
daniel-weisse, line 54: I don't think we should check for terraform files for 2 reasons:

1. There might be resources we want to remove that are not dependent on terraform (e.g. libvirt container)
2. Terraform won't perform any actions if there are no terraform files in the workspace, meaning we don't need to check ourselves

daniel-weisse: Maybe other people have an opinion on this?
IMO if there is nothing to do, `terminate` should not error

Nirusu: The current behaviour was to error so I just switch to an error for consistency.
Can switch to `fmt.Println` what I initially had. That's fine, too.

malt3: > IMO if there is nothing to do, terminate should not error. 

I agree. Our CLI should be idempotent for resource termination. This means: If there are dangling resources, try cleaning them up. If everything is cleaned up, remove the relevant state files. However, if cleanup partially succeeded, allow a user to intervene and make it safe to rerun `constellation terminate` any number of times (until it returns with exit code 0 and the state file if removed).
If a user (or automated system) reruns `constellation terminate` when there is nothing to terminate, we can print a message to indicate that there is no cleanup to be done but should indicate success using an exit code of 0.

This makes it very easy to integrate the CLI into other tools as we can ensure that there is no state left over from a previous constellation before creating a new one.

Nirusu: Definitely agree with not returning an error personally. Was really just for consistency with the previous behavior. Will change that.

I also like the idea of printing when there is nothing to do, hence I went for it. I think Daniel's point really was if we want to do that based on checking if certain files exist. However, I am not sure how else we can do that without major refactoring and actually checking Terraform and/or Docker to see if there is something to clean up.

Nirusu:
The ID file does not exist when the creation fails, it is created afterwards. Blocking on its non-existence is the current state before this PR, and it cannot cleanup improperly clusters. That’s completely manual or outside the Constellation CLI.

That’s why I went for the Terraform files, since they are essentially the source of truth that something was created or attempted to be created.

We can alternatively tell the user to run terraform terminate. Not sure if a Terraform installed by hc-install persists in the path, though. This will not clean up the .terraform directory and such though (at least I believe?), so not sure if that fully works without any changes.

Nirusu:
An alternative suggestion would maybe be to create some kind of 0-byte mark file on create and as long as it persists in the directory we allow terminate.

This allows for a single source of truth in exchange for a 0 byte file.

If we don’t want to keep that file, we can maybe delete it after successful creation and switch to the ID file. This breaks the single source of truth concept slightly, but has the advantage that these are still files created by our code and not Terraform.

This is then the closest to the current behaviour with the bonus of being able to clean up unfinished clusters, while also in the broader picture avoid calling terminate on something that isn’t Constellation.

The latter is in my opinion a huge plus which the no-op or my current “check what cleanupWorkspace does” would ignore and potentially kill another non-Constellation cluster, especially since we don’t prompt for termination at all (which we maybe should change…).

I think this might be the best of the two worlds we can get here. Let me know what you think.

Nirusu:
After discussion, we will:
-  remove the check
-  prompt for termination (maybe in a separate PR)
-  make the termination independent of the cloud provider 

This will happen in the main PR likely though, so I'll close this.

PR #363, opened 2022-10-24
Add AWS CCM versions

malt3:
### Proposed change(s)

- Add AWS CCM versions

### Checklist

- [x] Link to Milestone


derpsteb:
Following [the docs](https://github.com/edgelesssys/constellation/blob/main/.github/docs/upgrade-kubernetes.md#upgrading-kubernetes-resources) and adding the `-L` flag to the respective curl cmd I saw the same versions/hashes.

PR #369, opened 2022-10-25
Fix "enforceIdKeyDigest" capitalization

malt3:
### Proposed change(s)

- Fix "enforceIdKeyDigest" capitalization


malt3:
This is still not enough

```
./constellation init
Initializing cluster   
Error: init call: rpc error: code = Internal desc = initializing cluster: installing constellation-services: helm install services: ConfigMap in version "v1" cannot be handled as a ConfigMap: json: cannot unmarshal bool into Go struct field ConfigMap.data of type string
```

PR #371, opened 2022-10-25
Fix return value issue

daniel-weisse:
Use indirection to create ec2 client on demand

### DIFF
--- a/internal/attestation/aws/validator_test.go
+++ b/internal/attestation/aws/validator_test.go
@@ -76,8 +76,8 @@ func TestTpmEnabled(t *testing.T) {
 		wantErr bool
 	}{
 		"ami with tpm": {
-			attDoc:  attDocNoTPM,
-			awsAPI:  &stubDescribeAPI{describeImagesTPMSupport: "v2.0"},
+			attDoc: attDocNoTPM,
+			awsAPI: &stubDescribeAPI{describeImagesTPMSupport: "v2.0"},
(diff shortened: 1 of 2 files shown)


internal/attestation/aws/validator_test.go
```diff
@@ -126,8 +128,8 @@ type AWSMetadataStub struct {
 // DescribeImages is a mock function for testing.
 // Although the original function works for multiple supplied image id's, this function ignores this, since in our code we also only get the information for one image.
 func (a *AWSMetadataStub) DescribeImages(ctx context.Context, params *ec2.DescribeImagesInput,
```
leongross, line 130:
```suggestion
func (a *AWSMetadataStub) DescribeImages(ctx context.Context, params *ec2.DescribeImagesInput, optFns ...func(*ec2.Options),
) (*ec2.DescribeImagesOutput, error)
```
Maybe we should wrap up the function back into two lines to make it more readable?

leongross (approved):
LGTM

internal/attestation/aws/validator_test.go
daniel-weisse: This change was caused by my linter. 
If you function declaration is spread over multiple lines, the return values are always on their own separate line.

leongross: Okay, than let's leave it like that.

PR #372, opened 2022-10-25
Test operator code generation is up to date

katexochen:

### Proposed change(s)

- This new CI workflow checks that the generated code in the Constellation node operator is up to date.

### DIFF
--- a/.github/workflows/test-operator-codegen.yml
+++ b/.github/workflows/test-operator-codegen.yml
@@ -1,0 +1,43 @@
+# This workflow check
+name: Operator code generation check
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - "release/**"
+    paths:
+      - "operators/**"
+  pull_request:
+    paths:
+      - "operators/**"
+
+jobs:
+  govulncheck:
+    name: check-codegen
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # tag=v3.1.0
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-operator-codegen.yml
```diff
@@ -0,0 +1,39 @@
...
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - "release/**"
+    paths:
+      - "operator/constellation-node-operator/**"
+  pull_request:
+    paths:
+      - "operator/constellation-node-operator/**"
```
malt3, line 14:
```suggestion
      - "operators/constellation-node-operator/**"
  pull_request:
    paths:
      - "operators/constellation-node-operator/**"
```

.github/workflows/test-operator-codegen.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Setup Go environment
+        uses: actions/setup-go@c4a742cab115ed795e34d4513e2cf7d472deb55f # tag=v3.3.1
+        with:
+          go-version: "1.19.2"
+          cache: true
+
+      - name: Run code generation
+        shell: bash
+        working-directory: "operators/constellation-node-operator"
+        run: make manifests generate
```
malt3, line 35: We could abstract this for every folder in `operators/` (for future expansion). But tbh this is simpler so feel free to leave as it is.

katexochen: Are we sure that the exact same Makefile will exist for every additional opperator?

malt3: If we continue to use kubebuilder (which we really should) then yes.

malt3 (approved):
LGTM

PR #376, opened 2022-10-26
Pin setup-gcloud action to git tag (for renovate)

malt3:
### Proposed change(s)

- Pin setup-gcloud action to git tag (for renovate)


3u13r (approved):
LGTM

PR #378, opened 2022-10-26
Deploy CCM via Helm

derpsteb:
### Proposed change(s)

- Deploy CCM via Helm
- Add unittests that compare all rendered helm charts to known good values.
- Remove helm-lint cmake target & workflow. The unit tests will fail if you add something to constellation-services but don't specify an expected render. The tests take 0.6-0.9 secs on my machine, so I wouldn't implement an extra pipeline for it.

<!-- (uncomment if applicable)
### Related issue
- [AB#2538](https://dev.azure.com/Edgeless/Edgeless/_boards/board/t/Edgeless%20Team/Stories/?workitem=2538)


daniel-weisse (commented):
Before I do any indepth review, would it make sense to either 
1. Have the testdata files available as Go variables, i.e. embed them as string vars/constants
2. Have the testdata available as embedded files

Both options allow us to avoid interaction with the host filesystem during unit testing.

Another thing: Have you considered comparing the testdata in JSON instead of yaml?
This would allow us to use `assert.JSONEq()` which does not rely on 1:1 byte equality, meaning we won't see unexpected errors when data is ordered differently.

derpsteb:
After quick call decided to move unittests to separate PR as more discussion is needed. 
[Unittest PR](https://github.com/edgelesssys/constellation/pull/380).

cli/internal/helm/charts/edgeless/constellation-services/charts/ccm/templates/aws-daemonset.yaml
```diff
@@ -0,0 +1,63 @@
+{{ if eq .Values.csp "AWS" }}
+apiVersion: apps/v1
+kind: DaemonSet
```
daniel-weisse, line 3: Not sure how this would influence ease of use, but, since there is a lot of equality between the CCM deployment of different CSPs, we could merge all ccm daemonsets into a single file and then just set the required values (image, command, CSP specific mounts) conditionally

cli/internal/helm/charts/edgeless/constellation-services/charts/ccm/values.schema.json
```diff
@@ -0,0 +1,108 @@
...
+    "properties": {
+        "csp": {
+            "description": "CSP to which the chart is deployed.",
+            "enum": ["Azure", "GCP", "AWS", "QEMU"]
+        },
+        "subnetworkCIDR": {
+            "description": "CIDR for the subnetwork of the cluster",
+            "type": "string",
+            "examples": ["192.0.2.0/24"],
+            "pattern": "[0-9]{1,3}.[0-9]{1,3}.[0-9]{1,3}.[0-9]{1,3}/[0-9][0-9]"
```
daniel-weisse, line 12:
```suggestion
            "pattern": "[0-9]{1,3}.[0-9]{1,3}.[0-9]{1,3}.[0-9]{1,3}/[0-9]{1,2}"
```

cli/internal/helm/charts/edgeless/constellation-services/charts/ccm/values.schema.json
```diff
@@ -0,0 +1,108 @@
...
+        },
+        "GCP": {
+            "description": "Config values required for deployment on GCP",
+            "type": "object",
+            "properties": {
+                "image": {
+                    "description": "Container image to use for the spawned pods.",
+                    "type": "string",
+                    "examples": ["ghcr.io/edgelesssys/cloud-provider-gcp:latest"],
+                    "pattern": "ghcr.io/edgelesssys/cloud-provider-gcp:.+"
```
daniel-weisse, line 22: While the idea is nice, this prevents users from using their own cloud provider image, which might be desirable.
Same for AWS and Azure

cli/internal/helm/loader.go
```diff
@@ -134,14 +134,48 @@ func (i *ChartLoader) loadConstellationServices(csp cloudprovider.Provider,
...
+			joinServiceVals, ok := vals["join-service"].(map[string]any)
+			if !ok {
+				return helm.Release{}, errors.New("invalid join-service values")
+			}
+			joinServiceVals["enforceIdKeyDigest"] = enforceIDKeyDigest
+
+			ccmVals := vals["ccm"].(map[string]any)
+			ccmVals["Azure"] = map[string]any{
+				"image": versions.VersionConfigs[k8sVersion].CloudControllerManagerImageAzure,
+			}
```
daniel-weisse, line 154:
```suggestion
			vals["ccm"].(map[string]any)["Azure"] = map[string]any{
				"image": versions.VersionConfigs[k8sVersion].CloudControllerManagerImageAzure,
			}
```
Same thing as already mentioned previously.
Also fine to leave as is

daniel-weisse (changes requested):

bootstrapper/internal/kubernetes/kubernetes.go
```diff
@@ -498,12 +473,72 @@ func getIPAddr() (string, error) {
...
+			if err != nil {
+				return nil, fmt.Errorf("marshaling service account key: %w", err)
+			}
+
+			ccmVals := extraVals["ccm"].(map[string]any)
+			ccmVals["GCP"] = map[string]any{
+				"projectID":  projectID,
+				"uid":        uid,
+				"secretData": string(rawKey),
+			}
```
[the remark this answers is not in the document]
derpsteb: @katexochen explained to me yesterday that we should guard map access with a check if the key exists. I would change the accesses to include these checks. I think in those cases the shorter form is not possible?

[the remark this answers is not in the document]
daniel-weisse: Should not matter, but will make things easier.
For this case, if we want to check first if `ccm` exists in the map, we wouldn't need to instantiate an extra variable.
Same for `GCP`.

Example:
```Go
if ccmVals, ok := extraVals["ccm"](map[string]any); ok {
	gcpVals, ok := ccmVals["GCP"]; ok {
		gcpVals = map[string]any{
			"projectID":  projectID,
			"uid":        uid,
			"secretData": string(rawKey),
		}
	}
}
```

cli/internal/helm/charts/edgeless/constellation-services/charts/ccm/templates/aws-daemonset.yaml
derpsteb: I am happy to change this if the majority prefers it that way, but I would prefer to keep them separate. It seems to me that the rendered file is a bit easier to predict this way, over having a lot of if-else statements.

CMakeLists.txt
```diff
@@ -63,7 +63,9 @@ add_test(NAME integration-node-operator COMMAND make test WORKING_DIRECTORY ${CM
...
-add_test(NAME helm-lint COMMAND bash -c "helm lint * --set kms.image='ghcr.io/edgelesssys/constellation/kms:latest' --set join-service.csp='QEMU' \
+add_test(NAME helm-lint COMMAND bash -c "helm lint * --set tags.GCP=true --set kms.image='ghcr.io/edgelesssys/constellation/kms:latest' --set join-service.csp='GCP' \
   --set join-service.enforcedPCRs='[]' --set join-service.image='ghcr.io/edgelesssys/constellation/join-service:latest' --set join-service.measurements='[]' \
-  --set join-service.measurementSalt='deadbeef' --set kms.salt='aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' --set kms.masterSecret='aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'"
+  --set join-service.measurementSalt='deadbeef' --set kms.salt='aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' --set kms.masterSecret='aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' \
+  --set ccm.csp='GCP' --set ccm.subnetworkCIDR='192.0.2.0/24' --set ccm.GCP.image='ghcr.io/edgelesssys/cloud-provider-gcp:latest' --set ccm.GCP.projectID='demoproject-581925' \
+  --set ccm.GCP.uid='foobar' --set ccm.GCP.secretData='jsonstring'"
```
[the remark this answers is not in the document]
derpsteb: Will be removed here: https://github.com/edgelesssys/constellation/pull/380/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a

malt3 (approved):
LGTM

daniel-weisse (approved):

PR #380, opened 2022-10-26
Add helm unittests

derpsteb:
After talking to @katexochen regarding this [comment](https://github.com/edgelesssys/constellation/pull/378#pullrequestreview-1156439752) it seems like keeping the files as-is will be ok.

I added the suggestion of using `YAMLeq`. This should work just like `JSONeq`?

~This PR needs to be rebased once #378 is merged.~ Done.

From original PR:
- Add unittests that compare all rendered helm charts to known good values.
- Remove helm-lint cmake target & workflow. The unit tests will fail if you add something to constellation-services but don't specify an expected render. The tests take 0.6-0.9 secs on my machine, so I wouldn't implement an extra pipeline for it.


cli/internal/helm/testdata/Azure/constellation-services/charts/ccm/templates/gcp-cm.yaml
```diff
@@ -0,0 +1 @@
+
```
malt3, line 1: I'm a little bit confused why these empty files exist but feel free to keep them if it reduces the overall complexity of the test code.

cli/internal/helm/testdata/Azure/constellation-services/charts/kms/templates/daemonset.yaml
```diff
@@ -0,0 +1,63 @@
...
+    matchLabels:
+      k8s-app: kms
+  template:
+    metadata:
+      labels:
+        k8s-app: kms
+    spec:
+      containers:
+        - name: kms
+          image: ghcr.io/edgelesssys/constellation/kmsserver:v2.1.0@sha256:f56f901bd805550ac8232b4f7fdc8091a03ab16de5deddb1edd22b607413f406
```
malt3, line 21: Regarding this image reference and all the others in the testdata:
I still feel like this is a maintenance burden. For testing purposes it would be very beneficial to inject image version lookup tables. I would not like to block this PR though. Maybe we can track this separately? I can create a ticket.

cli/internal/helm/testdata/Azure/constellation-services/charts/ccm/templates/gcp-cm.yaml
derpsteb: Removed the files.

cli/internal/helm/testdata/Azure/constellation-services/charts/kms/templates/daemonset.yaml
derpsteb: I changed the HelmLoader struct to hold image references. The unittests then write bogus values into those fields.
The `New(..)` function needs at least the k8sVersion, unless we are fine with holding a pointer to the whole versionConfigs map in the HelmLoader. Therefore I needed to move the helmLoader instatiation into `initialize(..)`. This slows down the package tests for the `cmd` pkg by 0.1 secs because it loads the actual charts from the embed.FS. I guess this is fine(?), but I wanted to put it out there. 

Happy to hear any alternative solutions :)

malt3: I think what you did now looks good to me. Thanks! :+1:

malt3 (approved):
LGTM

PR #389, opened 2022-10-28
Normalize naming: "sonobuoy fast" -> "sonobuoy quick"

malt3:
### Proposed change(s)

- Normalize naming: "sonobuoy fast" -> "sonobuoy quick"

### Related issue
- https://github.com/edgelesssys/constellation/actions/runs/3341785552/jobs/5533328067
- https://github.com/edgelesssys/constellation/actions/runs/3342423393

### DIFF
--- a/.github/actions/e2e_test/action.yml
+++ b/.github/actions/e2e_test/action.yml
@@ -40,7 +40,7 @@ inputs:
     description: "The resource group to use"
     required: false
   test:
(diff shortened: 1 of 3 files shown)


katexochen:
Ah damn, thanks for fixing. Wouldn't it make more sense to fix in the other direction, as the sonobuoy flag is "quick"?

malt3:
> Wouldn't it make more sense to fix in the other direction, as the sonobuoy flag is "quick"?

Sure thing! Let me quickly change that.

daniel-weisse (approved):

PR #392, opened 2022-10-28
Docs: image changes for Longhorn support

3u13r:
### Proposed change(s)

- Document the changes needed if we want to support Longhorn.


malt3 (approved):
LGTM

PR #395, opened 2022-10-28
RFC: external KMS

thomasten:
My research on Vault and how we could go on with realizing the external KMS feature.
@datosh How well do you know Vault? Can you check my findings to see if I have missed something?

### DIFF
--- a/rfc/external-kms.md
+++ b/rfc/external-kms.md
@@ -10,7 +10,7 @@ Main goals are:
 * The KEK (master secret) never leaves the KMS. Unencrypted DEKs only exist temporarily.
 * Auto recovery of a cluster without manual intervention.
 
-We need to decide whether we can use Hashicorp Vault as a "proxy" that will allow us to support multiple KMSs while only implementing against one API.
+We need to decide on how to implement the KMIP client and what reference KMIP servers we want to implement against in the first iteration.
 
 ## Details on status quo
 
@@ -89,14 +89,29 @@ However, it also doesn't have support for KMIP.
 
 ### Conclusion
 
-We can't use Vault to add KMIP support to Constellation.
+We can't use Vault as a component of Constellation to add KMIP support.
 We also can't use it as a cloud KMS proxy because it only supports GCP.
 
 We could use Vault as DEK storage and use auto unseal with cloud KMS to replace the master secret.
 However, we would need a separate way for KMIP support.
 
-Overall, Vault doesn't seem a good fit for our use case.
 We should probably go with adding native support for external KMSs.
+
+We may use Vault as a reference KMIP server to implement against, but we need to procure Vault Enterprise and the ADP module for it.
+
+## Other tools and libraries
+
+### PyKMIP
+
+[PyKMIP](https://github.com/OpenKMIP/PyKMIP) is a KMIP client library for Python.
+Part of the project is a [KMIP server](https://pykmip.readthedocs.io/en/latest/server.html) for testing.
+We may use it as a reference KMIP server to implement against.
+
+### kmip-go
+
+[kmip-go](https://github.com/ThalesGroup/kmip-go) provides [KMIP protocol structures and can encode](https://pkg.go.dev/github.com/gemalto/kmip-go#example-package-Client) and decode them.
+We can use this as a building block of the Constellation KMIP client.
+It also has a [server type](https://pkg.go.dev/github.com/gemalto/kmip-go#Server) that may be useful in unit tests.
 
 ## Tasks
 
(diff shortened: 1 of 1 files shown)


rfc/external-kms.md
```diff
@@ -0,0 +1,113 @@
...
+* derive the *measurement salt* and the *clusterID*.
+* derive the DEK and set up the state disk.
+
+Note that the Bootstrapper performs these operations itself.
+Particularly, it replicates the DEK derivation of the CKMS in ClusterKMS mode.
+
+The CKMS serves a gRPC API to get DEKs.
+It's designed to support different (cloud) KMS services as backend.
+Currently, only the ClusterKMS backend is used, which uses the master secret from the Kubernetes secrets to derive the DEKs.
+The only clients of the CKMS are the CSI drivers for now.
```
daniel-weisse, line 31: The join-service also makes use of the CKMS to derive DEKs for joining nodes

rfc/external-kms.md
```diff
@@ -0,0 +1,113 @@
...
+Vault does *not* provide any KMIP client functionality.
+
+### Go-KMS-Wrapping
+
+https://github.com/hashicorp/go-kms-wrapping
+
+This library wraps various KMS services and provides a common interface to them.
+It's similar to the CKMS implementation.
+We may consider to replace our own implementation with this.
+However, it also hasn't support for KMIP.
```
daniel-weisse, line 88:
```suggestion
However, it also doesn't have support for KMIP.
```
Or `has no`

daniel-weisse (approved):

m1ghtym0 (commented):
Thank you Thomas for the detailed research.
I agree with your verdict and would go with the native implementation, but I'm no Vault expert either.

rfc/external-kms.md
```diff
@@ -0,0 +1,113 @@
...
+
+Vault enterprise features [seal wrap](https://developer.hashicorp.com/vault/docs/enterprise/sealwrap).
+This is an "extra layer of protection" that uses a KMS to encrypt single values before storing them in Vault.
+The docs promote this as a means of meeting compliance requirements rather than improving security.
+
+### KMIP server
+
+Vault can act as a [KMIP server](https://developer.hashicorp.com/vault/docs/secrets/kmip).
+This means that clients can connect to Vault via both the native API and a KMIP API.
+Vault does *not* provide any KMIP client functionality.
```
datosh, line 79: Maybe I understood the KMIP requirement wrong. How would we use a Vault implementation of a KMIP client? 

I was imagining that we use Vault as the KMIP server for our Constellation KMIP client (a service that runs inside Constellation). This way we have a simple software KMIP server to implement against (you can bootstrap Vault locally for testing).
Customers can choose to also use Vault Enterprise as the KMIP server (my hypothesis is this attractive since many companies already run Vault), or anything else that implements KMIP server, e.g., [Thales HSM](https://cpl.thalesgroup.com/encryption/key-management-interoperability-protocol-kmip).

datosh (commented):
Great summary!
I was not aware that we already have implementations for each CSP. In that case, I agree that integrating with Vault will not give us a speed advantage.

I think we should include a short analysis of what we want to use as a KMIP server in our implementation efforts. I am not aware of a developer-friendly solution.
I still think that Vault would be a good option for us to use as a KMIP server for development and testing. Vault is widely used, well-tested, and easy to automate. While the license cost is moderate ([$20/month?](https://cloud.hashicorp.com/products/vault/pricing)).
Alternatives are:
* [PyKMIP](https://pykmip.readthedocs.io/en/latest/server.html) I am not sure how stable it is, but maybe a good free option
* Self managed HSMs such as [Luna](https://cpl.thalesgroup.com/encryption/key-management-interoperability-protocol-kmip) are quite expensive and hard to operate
  * Some HSMs provide software simulators of their real hardware appliances. Maybe this could be a low-cost & automatable route..?
* SaaS HSMs such as [AWS Cloud HSM](https://cpl.thalesgroup.com/encryption/key-management-interoperability-protocol-kmip) for $2k/month and are okay-ish to automate.

rfc/external-kms.md
thomasten: Oh, I misunderstood you in this regard. I thought you suggested using Vault as a "proxy" that relays requests to a KMIP server. So we would only need to implement against the Vault API instead of the more complex KMIP API.

I agree that Vault could be a good choice for the first KMIP server to implement against. However, the cloud offering doesn't seem to have that feature and the Enterprise edition requires to "contact sales", which may be annoying.

datosh: The **development option** seems to have all the features, but none of the SLAs and seems like a great low-cost solution for us developing this feature: https://cloud.hashicorp.com/products/vault/pricing

thomasten: FAQ on that same page says
> Which Vault Enterprise features are not yet available in HCP Vault?
The **Advanced Data Module (ADP)**, resource limiting, and 3rd party plug-ins are not yet available in HCP Vault.

And KMIP is part of ADP.

m1ghtym0: Ah, I had the same misunderstanding.
Using Vault as our KMIP server for dev and testing makes sense to me even if we have to pay the 20$/month license fee.

datosh (approved):

PR #396, opened 2022-10-28
Remove PublicIP from QEMU metadata

daniel-weisse:
### Proposed change(s)

QEMU was the only CSP still making use of the `PublicIP` field of `InstanceMetadata`, this PR aims to remove that dependency by shifting QEMU's usage from `PublicIP` to `VPCIP`.
This will allow us to completely remove this field from `InstanceMetadata` in https://github.com/edgelesssys/constellation/pull/387

I have also implemented a `GetLoadbalancerEndpoint` function for the qemu metadata api, which just returns the IP of the first control-plane node.
This allows us to remove the `SupportsLoadbalancer` method, since this is a requirement for all new CSP implementations anyways.

### Additional info
- This PR requires new images for the QEMU metadata API and the join-service
- Removed some dead code related to PublicIPs in `internal/cloud/cloudtypes`

### Checklist

- [ ] Link to Milestone


hack/qemu-metadata-api/server/server.go
```diff
@@ -110,6 +111,43 @@ func (s *Server) listPeers(w http.ResponseWriter, r *http.Request) {
 	log.Infof("Request successful")
 }
 
+// getEndpoint returns the IP address of the first control-plane instance.
+// This allows us to fake a load balancer for QEMU instances.
+func (s *Server) getEndpoint(w http.ResponseWriter, r *http.Request) {
```
malt3, line 116: I like this route. We could even replace this part of the metadata API to return a real loadbalancer IP in the future without changing anything in the bootstrapper.

hack/qemu-metadata-api/server/server.go
```diff
@@ -110,6 +111,43 @@ func (s *Server) listPeers(w http.ResponseWriter, r *http.Request) {
...
+			strings.HasSuffix(lease.Hostname, "0") {
+			w.Header().Set("Content-Type", "application/json")
+			if err := json.NewEncoder(w).Encode(lease.IPaddr); err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+			log.Infof("Request successful")
+			return
+		}
+	}
```
malt3, line 145: Could we return a random control-plane node instead? This is not ideal as we do not perform health checks (as a real loadbalancer would do) but almost gives us high availability for free.

daniel-weisse: I chose to return the first control-plane so we have a consistent Kubernetes endpoint across all nodes. Not sure if/how this will break when we instead return a random control-plane node. I figure it would at least cause problems for the first control-plane node.

malt3: Yeah. You're probably right. Let's stay consistent!

malt3 (approved):
LGTM

PR #400, opened 2022-10-28
Upgrade to Go 1.19

katexochen:
### Proposed change(s)

- Update to Go 1.19, changelog: https://tip.golang.org/doc/go1.19
- Use `atomic.Bool` instead of atomic functions for int32 where applicable


3u13r (approved):
LGTM

PR #404, opened 2022-10-28
e2e: add nop payload to only test infra creation

katexochen:


malt3 (approved):
LGTM

PR #406, opened 2022-10-28
bootstrapper: undefine -v flag to avoid collision

3u13r:
### Proposed change(s)

- Temporarily undefine -v flag as it collided with glogs definition. 

As far as I can see, we never used the flag in our images anyway to setting to the flag's default.

The other solution would be to revert [79f52e67cb0d078c951f7ebcff97e6a91b9253a8 ](https://github.com/edgelesssys/constellation/commit/79f52e67cb0d078c951f7ebcff97e6a91b9253a8). In my opinion this is less preferable since this will block AWS support again. 

As soon as https://github.com/google/go-sev-guest/issues/23 is merged and used downstream, we can re-enable the flag.

### DIFF
--- a/bootstrapper/cmd/bootstrapper/main.go
+++ b/bootstrapper/cmd/bootstrapper/main.go
@@ -49,9 +49,12 @@ const (
 
 func main() {
 	gRPCDebug := flag.Bool("debug", false, "Enable gRPC debug logging")
-	verbosity := flag.Int("v", 0, logger.CmdLineVerbosityDescription)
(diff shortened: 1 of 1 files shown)


katexochen (approved):
You checked we haven't got any scripts that use this flag, right?

3u13r:
Yes, checked for `bootstrapper.*-v` in the repo. Also checked the systemd unit starting the bootstrapper. Of course could be that I missed something.

PR #407, opened 2022-10-30
undefine more -v flags due to glog

3u13r:
### Proposed change(s)

- Undefine more -v flags due to glog collision

Need to undefine -v in disk-mapper and join-service as both import atls which depends on go tpm and so on. 
Not sure about if undefining more -v flags is the right way forward, though. Again, the flags should be unused. A user could change the join-service deployment and add the flag, but this is unlikely as they should keep their hands off our deployments anyway. Furthermore, hopefully this is a very temporary fix (tm). 

Also, the AWS support is almost finished. Only testing needed. Reverting https://github.com/edgelesssys/constellation/commit/79f52e67cb0d078c951f7ebcff97e6a91b9253a8 would block AWS, We'll probably need to decide for the next release if we want AWS or not. I guess we could also build a fork of go-tpm-tools which includes the AWS fix, but is purposefully outdated to not include the glog dependency. Interested in opinions here.

### DIFF
--- a/disk-mapper/cmd/main.go
+++ b/disk-mapper/cmd/main.go
@@ -50,10 +50,12 @@ const (
 
 func main() {
 	csp := flag.String("csp", "", "Cloud Service Provider the image is running on")
-	verbosity := flag.Int("v", 0, logger.CmdLineVerbosityDescription)
+	// FIXME: define flag again once its definition no longer collides with glog.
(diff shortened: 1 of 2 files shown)


katexochen:
>  I guess we could also build a fork of go-tpm-tools which includes the AWS fix, but is purposefully outdated to not include the glog dependency

Imo your current way to handle the situation is alright. If we need the flag for something, we could also just rename it. Wouldn't put more effort into the fork.

katexochen (approved):

PR #416, opened 2022-10-31
AB#2566 RFC for image discoverability (description of image version uid)

malt3:
### Proposed change(s)
- RFC for image discoverability (description of image version uid)

### DIFF
--- a/rfc/image-discoverability.md
+++ b/rfc/image-discoverability.md
@@ -77,7 +77,9 @@ s3://<BUCKET-NAME>/images/<IMAGE-VERSION-UID>.json
     "cvm": "azure-cvm-123",
     "trustedlaunch": "azure-trusted-launch-123"
   },
-  "gcp": "gcp-image-123",
+  "gcp": {
+    "sev-es": "gcp-image-123"
+  }
   "qemu": "https://cdn.edgeless.systems/.../qemu.raw"
 }
 ```
--- a/rfc/image-discoverability.md
+++ b/rfc/image-discoverability.md
@@ -114,7 +114,7 @@ The `image` field is independent of the CSP and is a used to discover the CSP-sp
 The CLI can find a CSP- and region specific image reference by looking up the `image version uid` in the following order:
 
 - if a local file `<IMAGE-VERSION-UID>.json` exists, use the lookup table in that file
-- otherwise, load the image lookup table from a well known URL (e.g. `https://cdn.edgeless.systems/images/<IMAGE-VERSION-UID>.json`) and use the lookup table in that file
+- otherwise, load the image lookup table from a well known URL (e.g. `https://cdn.confidential.cloud/images/<IMAGE-VERSION-UID>.json`) and use the lookup table in that file
 - choose the CSP-specific image reference for the current region and security type:
   - On AWS, use the AMI ID for the current region (e.g. `.aws.us-east-1`)
   - On Azure, use the image ID for the security type (CVM or Trusted Launch) (e.g. `.azure.cvm`)
--- a/rfc/image-discoverability.md
+++ b/rfc/image-discoverability.md
@@ -4,8 +4,8 @@ The Constellation OS image build pipeline generates a set of images using a chos
 
 ```mermaid
 graph LR
-    version["input: version (<code>v2.2.0</code>)"] --> imageid["image version uid (<code>v2.2.0-cc0de5c</code>)"]
-    commit["input: Constellation repo commit hash (<code>cc0de5c</code>)"] --> imageid
+    version["input: version (<code>v2.2.0</code>)"] --> imageid["image version uid (<code>v2.2.0</code>)"]
+    commit["input: Constellation repo commit hash (<code>cc0de5c</code>)"] --> pipeline
     imageid --> pipeline{build images}
     pipeline --> awsimage["raw AWS image"]
     pipeline --> azureimage[raw Azure image]
@@ -25,13 +25,13 @@ graph LR
     awsmeasure --> awsmeasurements[measurements.yml]
     azuremeasure --> azuremeasurements[measurements.yml]
     gcpmeasure --> gcpmeasurements[measurements.yml]
-    awsimage1 --> lut{Upload image lookup table to S3<br>key: <code>v2.2.0-cc0de5c</code><br>value: lookup table for image references specific to each CSP}
+    awsimage1 --> lut{Upload image lookup table to S3<br>key: <code>v2.2.0</code><br>value: lookup table for image references specific to each CSP}
     awsimage2 --> lut
     awsimage3 --> lut
     azureimage1 --> lut
     azureimage2 --> lut
     gcpimage1 --> lut
-    awsmeasurements --> s3measurements{"Upload image measurements to S3<br><code>aws/v2.2.0-cc0de5c/measurements.yaml<br>azure/v2.2.0-cc0de5c/measurements.yaml<br>gcp/v2.2.0-cc0de5c/measurements.yaml<br></code>"}
+    awsmeasurements --> s3measurements{"Upload image measurements to S3<br><code>aws/v2.2.0/measurements.yaml<br>azure/v2.2.0/measurements.yaml<br>gcp/v2.2.0/measurements.yaml<br></code>"}
     azuremeasurements --> s3measurements
     gcpmeasurements --> s3measurements
 ```
@@ -46,7 +46,8 @@ The build pipeline takes as inputs:
   - a pseudo-version number (e.g. `v2.3.0-pre-branch-name`) for branch images
 - a commit hash of the Constellation monorepo that is used to build the images (e.g. `cc0de5c68d41f31dd0b284d574f137e0b0ad106b`)
 
-To identify images belonging to one invocation of the build pipeline, the pipeline uses a unique identifier for the set of images, referred to as `image version uid` that combines the version number and the commit hash (e.g. `v2.2.0-cc0de5c68d41f31dd0b284d574f137e0b0ad106b`).
+To identify images belonging to one invocation of the build pipeline, the pipeline uses a unique identifier for the set of images, referred to as `image version uid`.
+This is either the release version number (e.g. `v2.2.0`) or a pseudo version that combines the version number and the commit hash (e.g. `v2.3.0-pre-debug-cc0de5c68d41f31dd0b284d574f137e0b0ad106b`).
 
 The build pipeline produces as outputs:
 
@@ -107,7 +108,7 @@ s3://<BUCKET-NAME>/measurements/<CSP>/<IMAGE-VERSION-UID>/measurements.yaml.sig
 ## CLI image discovery
 
 The CLI needs to be able to discover the image references for a given `image version uid`.
-By default, the CLI will prefill the `image` field of the `constellation-conf.yaml` when `constellation config generate <CSP>` is run with a hardcoded `image version uid` (e.g. `v2.2.0-cc0de5c68d41f31dd0b284d574f137e0b0ad106b`).
+By default, the CLI will prefill the `image` field of the `constellation-conf.yaml` when `constellation config generate <CSP>` is run with a hardcoded `image version uid` (e.g. `v2.2.0`).
 The `image` field is independent of the CSP and is a used to discover the CSP-specific image reference as needed for the following operations:
 
 - `constellation create`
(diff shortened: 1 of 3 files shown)


rfc/image-discoverability.md
```diff
@@ -0,0 +1,127 @@
...
+By default, the CLI will prefill the `image` field of the `constellation-conf.yaml` when `constellation config generate <CSP>` is run with a hardcoded `image version uid` (e.g. `v2.2.0-cc0de5c68d41f31dd0b284d574f137e0b0ad106b`).
+The `image` field is independent of the CSP and is a used to discover the CSP-specific image reference as needed for the following operations:
+
+- `constellation create`
+- `constellation upgrade apply`
+
+The CLI can find a CSP- and region specific image reference by looking up the `image version uid` in the following order:
+
+- if a local file `<IMAGE-VERSION-UID>.json` exists, use the lookup table in that file
+- otherwise, load the image lookup table from a well known URL (e.g. `https://cdn.edgeless.systems/images/<IMAGE-VERSION-UID>.json`) and use the lookup table in that file
```
Nirusu, line 117: Nit:
```suggestion
- otherwise, load the image lookup table from a well known URL (e.g. `https://cdn.confidential.cloud/images/<IMAGE-VERSION-UID>.json`) and use the lookup table in that file
```

rfc/image-discoverability.md
```diff
@@ -0,0 +1,127 @@
...
+## Image measurements
+
+This RFC is not about the image measurements themselves, but about how they are stored and looked up.
+The format of the image measurements is described in the [secure software distribution RFC](secure-software-distribution.md).
+
+The image measurements are stored in a folder structure in S3 that is organized by CSP and `image version uid`.
+
+```
+s3://<BUCKET-NAME>/measurements/<CSP>/<IMAGE-VERSION-UID>/measurements.yaml
+s3://<BUCKET-NAME>/measurements/<CSP>/<IMAGE-VERSION-UID>/measurements.yaml.sig
```
Nirusu, line 102: What do you think about having only one measurement file per version with all CSPs and images included that the CLI parses and compares against? 

The suggested format of the measurements would allow this with few modifications if we already keep CSP and image version in there.

This could be then simplified to `s3://<BUCKET-NAME>/measurements/<IMAGE-VERSION-UID>.yaml(.sig)`

rfc/image-discoverability.md
```diff
@@ -0,0 +1,127 @@
...
+  "qemu": "https://cdn.edgeless.systems/.../qemu.raw"
+}
+```
+
+- For AWS, the image lookup table contains the AMI IDs for each region
+- For Azure, the image lookup table contains the image IDs for the CVM and Trusted Launch images
+- For GCP, the image lookup table contains the image ID
+- For QEMU, the image lookup table contains a URL to the QEMU image
+
+This document is not signed and can be extended in the future to include more image references (e.g. if an image is replicated to a new AWS region).
```
Nirusu, line 90: Not sure about this. Would prefer signing it unless we plan allowing third-parties to host their own lookup table.

A wrong image lookup table could still lead to unintended code execution on `constellation create`. It will fail the PCR test on `constellation init`, but I still would prefer avoiding getting this far if we can.

Nirusu:
What about `constellationVersion` and the other fields we discussed in https://github.com/edgelesssys/constellation/blob/1f8eba37c8b4b4f91e104cea3c7053dd09bfaa32/rfc/updates.md#extending-the-constellation-upgrade-command ?

Would you maybe suggest getting rid of `constellationVersion` and replacing it with `microServiceVersion` version instead?

rfc/image-discoverability.md
malt3: > Would prefer signing it unless we plan allowing third-parties to host their own lookup table.

This would require a separate key pair to prevent any possible confusion between signed measurements and signed image lookup tables. It's also harder to maintain when replicating images in other regions.

> A wrong image lookup table could still lead to unintended code execution on `constellation create`

The lookup table is still fetched via https from our cdn. An attacker with control over our cdn could inject a malicious image reference that is detected on `constellation init`.
I view this as an attack vector with little impact that is very hard to pull off. 

>  unless we plan allowing third-parties to host their own lookup table.

This is not planned but definitely a possibility.

Anyone else with an opinion on this?

rfc/image-discoverability.md
Nirusu: No major benefits, mainly just having one file instead of different paths. W

e wanted to revise the measurement pipeline anyway, so in case we also do Matrix uploads we could consolidate them. This would work similar to the release manifest, which in theory you could also split up in multiple files.

rfc/image-discoverability.md
Nirusu: I know JSON and YAML are compatible, but I am not sure how you want to create a confusion here as the file structure should be vastly different. If that is possible, our parsing logic has some major issues.

rfc/image-discoverability.md
```diff
@@ -0,0 +1,127 @@
...
+    "us-east-1": "ami-123",
+    "us-west-2": "ami-456",
+    "eu-west-1": "ami-789"
+  },
+  "azure": {
+    "cvm": "azure-cvm-123",
+    "trustedlaunch": "azure-trusted-launch-123"
+  },
+  "gcp": "gcp-image-123",
+  "qemu": "https://cdn.edgeless.systems/.../qemu.raw"
```
thomasten, line 81: don't we want to upload the qemu image as compressed qcow2?

rfc/image-discoverability.md
```diff
@@ -0,0 +1,127 @@
...
+The CLI can find a CSP- and region specific image reference by looking up the `image version uid` in the following order:
+
+- if a local file `<IMAGE-VERSION-UID>.json` exists, use the lookup table in that file
+- otherwise, load the image lookup table from a well known URL (e.g. `https://cdn.confidential.cloud/images/<IMAGE-VERSION-UID>.json`) and use the lookup table in that file
+- choose the CSP-specific image reference for the current region and security type:
+  - On AWS, use the AMI ID for the current region (e.g. `.aws.us-east-1`)
+  - On Azure, use the image ID for the security type (CVM or Trusted Launch) (e.g. `.azure.cvm`)
+  - On GCP, use the only image ID (e.g. `.gcp`)
+
+This allows customers to upload images to their own cloud subscription and use them with the CLI by providing the image lookup table as a local file.
```
thomasten, line 123: Is this supposed to be the only way to use a custom built image?

thomasten, line 123: Is the Trusted Launch workflow supposed to be
* download the image
* upload the image to my cloud subscription
* create a local lookup table file

?

rfc/image-discoverability.md
malt3: The root file system itself is a compressed squashfs since switching to mkosi so I used raw images on QEMU for my own testing as there is not a lot to gain. It may still be a good idea to stay with QCOW for consistency.

rfc/image-discoverability.md
```diff
@@ -0,0 +1,127 @@
...
+```
+s3://<BUCKET-NAME>/measurements/<CSP>/<IMAGE-VERSION-UID>/measurements.yaml
+s3://<BUCKET-NAME>/measurements/<CSP>/<IMAGE-VERSION-UID>/measurements.yaml.sig
+```
+
+## CLI image discovery
+
+The CLI needs to be able to discover the image references for a given `image version uid`.
+By default, the CLI will prefill the `image` field of the `constellation-conf.yaml` when `constellation config generate <CSP>` is run with a hardcoded `image version uid` (e.g. `v2.2.0-cc0de5c68d41f31dd0b284d574f137e0b0ad106b`).
+The `image` field is independent of the CSP and is a used to discover the CSP-specific image reference as needed for the following operations:
```
derpsteb, line 109: "`image` field is independent of the CSP"

Should this really be "independent"? It seems to me like the field does in fact depend on the CSP?

rfc/image-discoverability.md
malt3: I see three possible ways to expose this functionality to our users:

1) Our scripts provided to customers that they can use to reupload images also create the local lookup table file. User places the file in the workspace.
2) A larger org may want to have their own "repository". A HTTP server that hosts their lookup tables with their images and uses the same paths as described in this document. Only the base URL would be their own. They could then set the repository in the `constellation-conf.yaml`. I could extend the RFC to explicitly support this.
3) We add an alternative field or a special scheme to be used in the exiting version field: `imageReference` could override the `image` version field with a literal cloud provider image reference OR `gcp://<image-reference>` if the same field should be reused.

1) is the easiest to implement and allows users as well as developers to make local overrides while still using the same versions as the official images. This allows our tools to still reason about what a version is compatible with -- even if the image reference points to a copy of the image.
2) Is only feasible for larger orgs. It would be the most elegant solution in my opinion.
3) Has a large drawback: We cannot reason about the version of the provided image reference anymore. If we want to implement any form of compatibility checks, not knowing what image version a reference belongs to makes this impossible. Even if we use two fields (version and reference) individually, a user could update one field and we would still make wrong decisions based on that.

malt3: Please see my comment on the other remark

rfc/image-discoverability.md
```diff
@@ -0,0 +1,127 @@
...
+
+```
+s3://<BUCKET-NAME>/measurements/<CSP>/<IMAGE-VERSION-UID>/measurements.yaml
+s3://<BUCKET-NAME>/measurements/<CSP>/<IMAGE-VERSION-UID>/measurements.yaml.sig
+```
+
+## CLI image discovery
+
+The CLI needs to be able to discover the image references for a given `image version uid`.
+By default, the CLI will prefill the `image` field of the `constellation-conf.yaml` when `constellation config generate <CSP>` is run with a hardcoded `image version uid` (e.g. `v2.2.0-cc0de5c68d41f31dd0b284d574f137e0b0ad106b`).
```
derpsteb, line 108: Why do we require the commit hash for release images?

I am thinking about a user wanting to upgrade their OS image. As I am understanding this right now they will have to lookup the version and hash. Not really a lot of work if you have the release page open already, but I also don't see why the commit is necessary for release versions as they should all be tagged.

malt3: My thinking is that we really want to discourage manual management of this field. Down the line we want to discover what versions are available and compatible automatically in the CLI. If users see *just* a version field, they feel more inclined to modify it and break compatibility in the process. This also makes it just a little more streamlined for us to decide what commit an image was really built from - git tags can be set and moved later on and the release version string can be set to anything when running the pipeline as it is just a string input for the workflow.

I notice that you are not the first who feels uneasy about this so if the commit hash is super unpopular we can omit it.

malt3:
> What about `constellationVersion` we discussed about in https://github.com/edgelesssys/constellation/blob/1f8eba37c8b4b4f91e104cea3c7053dd09bfaa32/rfc/updates.md#extending-the-constellation-upgrade-command ?
> 
> Would you maybe suggest calling it with `microserviceVersion` version then? Otherwise it would seem confusing to me to have a field called `constellationVersion` and also `image` which specifically holds a version.

That's a good idea in my eyes. @3u13r

rfc/image-discoverability.md
thomasten: Understood. Then I've got no opinion on this.

rfc/image-discoverability.md
malt3: @Nirusu @derpsteb I think I got convinced after all to remove the hash from release image versions.

rfc/image-discoverability.md
malt3: Let's add this as a possible future extension

rfc/image-discoverability.md
malt3: @thomasten any preferences on the preferred implementation of this? If there are none I would leave the RFC as-is and add the customer repository as a possible future extension.

thomasten: sounds good

3u13r (approved):
Current version LGTM from my side

Nirusu (approved):
Still unsure about unsigned software manifests, but given this is just a RFC for now and #446 waits on this PR I think we can go ahead for now and merge this.

The general concept sounds good to me, we can figure out the details later when we start designing and implementing this feature.

datosh (approved):
LGTM! I like to local file overwrite for the look-up table file. Great for users and local testing 👍

PR #418, opened 2022-10-31
Prompt before terminate

Nirusu:
### Proposed change(s)

- Add a prompt to `constellation terminate` before terminating any resources
- Add a `--yes` flag to bypass this for CI and other automation scripts.

You can choose if you'd like the separate parse function (initial commits) or if we just include the check in the terminate function itself (last commit) since it's just one check. The former one is more consistent with the other CLI functions, but we don't need any other flags anyway currently so it's kinda obsolete except for consistency.

### Related issue
I suggested this initially in #360. 
While one could argue that if someone calls `constellation terminate` they would be doing this by intention (similar to `rm`), in my opinion, there's multiple good reasons to prompt the user before terminating:
- We prompt the user on `create` despite `terminate` being the action with significantly worse unintended consequences.
- Prompting before terminating is similar behaviour within the Azure and GCP CLI and Terraform itself. The AWS CLI does not prompt.
- The change from #360 means that `constellation terminate` has no checks on its current directory. It will not only destroy any Constellation cluster without asking, but also _any_ Terraform workspace in the current working directory. This can create significant headaches unintentionally due to us saving on checks; thus we should not automatically allow this.

If anyone wants to skip this, they can bypass this with a `--yes` flag, similarly as in `constellation create`.

- [x] Link to Milestone

### DIFF
--- a/.github/actions/constellation_destroy/action.yml
+++ b/.github/actions/constellation_destroy/action.yml
@@ -4,5 +4,5 @@ runs:
   using: 'composite'
   steps:
   - name: Constellation terminate
-    run: constellation terminate
+    run: constellation terminate --yes
     shell: bash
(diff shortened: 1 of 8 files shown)


Nirusu:
Link checker fails for some reason. I think it can be ignored and re-tried later.

daniel-weisse (approved):

malt3 (approved):
LGTM, too

PR #421, opened 2022-11-01
Install CSI drivers as a CI action

chmey:
### Proposed change(s)

- Provide CI step to install CSI drivers: So far only the K-Bench action uses CSI drivers and it installs the drivers itself. This does not scale well, since other workflows/actions can't reuse the installation action. Therefore I propose to separate the installation from the benchmarking action.
- Also, added a step to delete the CSI PVCs that were created during the e2e tests in `constellation_destroy`. The motivation is to not leave disks behind from our e2e tests at the CSPs. Alternatively, every e2e action that creates disks should be responsible for cleaning them up (as opposed to the central deletion in the termination step). What do you think? I am concerned about the possibility to fail right before the termination. Maybe we should except failure here?


chmey:
> The test run I started on this branch failed with some Python error in the evaluation step: https://github.com/edgelesssys/constellation/actions/runs/3376697696/jobs/5604732308
> 
> Is there a min number of nodes required?

I will look into running it with fewer nodes.

What do you think about the `constellation_terminate` approach of this PR?

katexochen:
> What do you think about deleting the PVCs during cosntellation_destroy as proposed here?

Looks good to me, tested it with an e2e nop run, no problem. :+1:

chmey:
So, from the raw workflow logs, I see the test run created a cluster with 1 control node and 1 worker node:

```
2022-11-02T10:29:12.2936465Z [36;1mconstellation create -c 1 -w 1 --name e2e-test -y[0m
```

The `iperf` benchmarks actually require two distinct nodes as client and server can't be on the same node due to their `anti-affinity`: https://github.com/edgelesssys/k-bench/blob/feat/constellation/config/dp_network_internode/iperf_client_pod.yaml

So yeah, the benchmarks require at least two worker nodes. I will address this in another PR. Thanks!

katexochen (approved):
lgtm :baby_chick:

PR #422, opened 2022-11-01
Take location from state file for az jump host

derpsteb:
### Proposed change(s)

- I was hitting a resource not found error because I was using different locations. This patch should prevent that error from happening.

### DIFF
--- a/hack/azure-jump-host/jump-host-create
+++ b/hack/azure-jump-host/jump-host-create
@@ -5,6 +5,7 @@ set -o pipefail
 SCRIPTDIR="$( dirname --  $(realpath "${BASH_SOURCE}"); )";
 RG=$(jq -r .azureresourcegroup constellation-state.json)
 SUBNET=$(jq -r .azuresubnet constellation-state.json)
+LOCATION=$(jq -r .azurelocation constellation-state.json)
 VNET=${SUBNET%"/subnets/nodeNetwork"}
 PUBKEY=$(cat ~/.ssh/id_rsa.pub)
 DEPLOYMENT_NAME=jump-host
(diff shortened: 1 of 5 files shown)


malt3 (dismissed):
I'm wondering what the jump host can still be used for. I was considering removing it since SSH is not available anymore. If it is useful for your purposes we should keep it :+1: 

<strike>LGTM</strike>

EDIT: whoops. @Nirusu has a point.

hack/azure-jump-host/jump-host-create
```diff
@@ -5,6 +5,7 @@ set -o pipefail
 SCRIPTDIR="$( dirname --  $(realpath "${BASH_SOURCE}"); )";
 RG=$(jq -r .azureresourcegroup constellation-state.json)
 SUBNET=$(jq -r .azuresubnet constellation-state.json)
+LOCATION=$(jq -r .azurelocation constellation-state.json)
```
Nirusu, line 8: What version have you tested this with? We don't have a state file on the current main branch?

derpsteb: I needed this to recover a v2.0 cluster. Maybe this should be on the v2.0 branch? The docs for v2.0 in the recovery section could also point to this hack folder. They currently only state that you should setup a jump host.

malt3: I think that's a good idea. Can you rebase this PR to target `release/v2.0`?

derpsteb:
Moved the change to only happen on `release/v2.0`. 
Since this is no longer needed on main we can remove the script from main.

malt3 (approved):
LGTM

PR #423, opened 2022-11-01
AB#2582: deploy CNM via Helm

derpsteb:
### Proposed change(s)

- Deploy cloud node manager via helm


cli/internal/helm/charts/edgeless/constellation-services/charts/cnm/values.schema.json
```diff
@@ -0,0 +1,58 @@
...
+            },
+            "required": [
+                "image"
+            ]
+        }
+    },
+    "required": [
+        "csp"
+    ],
+    "allOf": [
```
3u13r, line 26: Warning: I'm not that familiar with JSON schemas. 
I was thinking about if we can simplify to basically
1. check if csp is Azure, if true:
2. check if image is set
i.e. is there a way so we don't have another file we have to touch if we add another csp. While I see that this can act as a safeguard for us, for the CNM it does not make sense in my opinion. 

The way I would think of, is to provide a default empty string in the values.yaml. The best solution would involve re-writing this schema in a way which handles unset values.

derpsteb: The CNM will not be required for AWS? Sure, then I could make `image` a top level key and remove the csp field. 

I am unsure if "handling unset values" is something the schema can do for us. It can't do any transformation of the inputs afaik. We could change the schema to allow unset values, but then we would have to change the chart to handle those values. Handling an unset image would mean setting a default value, which is unwanted for the image imo.

daniel-weisse (approved):

cli/internal/helm/charts/edgeless/constellation-services/charts/cnm/values.schema.json
3u13r: Yes the CNM (currently and I think will be) only required for Azure,

Regarding the validation:
I think that a good behavior is that the CNM is optional and only deployed when the image is set. So we have an empty string as default value and the whole azure-daemonset.yaml is wrapped in a `if image == ""`.

daniel-weisse (commented):
There are some stub functions and interfaces that can be adjusted with the deployment of cloud node manager deployment via helm

cli/internal/helm/charts/edgeless/constellation-services/charts/cnm/values.schema.json
derpsteb: Currently subcharts are enabled via tags. If the tag `Azure` is set the CNM will be deployed. This also how we control for example that the CCM is not deployed on QEMU.

Knowing that CNM is only relevant for Azure I removed all CSP-selection related things and just made cnm.image a required value. So if tag `Azure` is set you have to set cnm.image, otherwise you don't.

Does this sound good to you?

3u13r (approved):
LGTM

PR #426, opened 2022-11-02
AB#2571 Make AMI public on creation

malt3:
### Proposed change(s)

- Make AMI public on creation


daniel-weisse (approved):
Looks good in general
Can we have the publish step be conditional so we only publish release images / selected debug images?
I don't think we need to or should publish images we created for testing

malt3:
> Can we have the publish step be conditional so we only publish release images / selected debug images?
I don't think we need to or should publish images we created for testing

Let's do this separately :+1:

PR #429, opened 2022-11-02
Update to Go 1.19.3

Nirusu:
### Proposed change(s)

- Update to Go 1.19.3

### Related issue
- Fixes https://pkg.go.dev/vuln/GO-2022-1095 but we don't have an official Windows build so shouldn't affect us

Not sure if I should add this to the README.md like last time, but I tend not to since I don't see this affecting us?

### Checklist

- [ ] Link to Milestone


katexochen (approved):
lgtm. I agree that we don't have to add this explicitly to the changelog.

datosh (approved):

PR #430, opened 2022-11-02
Fix link checker

Nirusu:
### Proposed change(s)
- Disable medium.com based links (blog.edgeless.systems, blog.sigstore.dev) from link checking until Lychee gains cookie support.

This is due to medium.com having integrated some redirects which require cookies to be set in order to be redirected back to the original URL. Not sure why exactly (L7 protection against bots?) they added this but until then there's not much we can do until Lychee supports storing cookies or we build/use some alternative solution.

### Related issues
- https://github.com/lycheeverse/lychee/issues/645

### Additional info
The check is still broken since www.linux-kvm.org seems to be misconfigured and not serve the full TLS chain currently. I tried to contact them if they are aware of this, but for now I would wait and keep this PR a draft to see if they fix it on their side.

### DIFF
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -1,3 +1,7 @@
 http://localhost:
 https://github.com/edgelesssys/wiki/blob/master/documentation/constellation/customer-onboarding.md
 https://github.com/edgelesssys/wiki/blob/master/documentation/rebasing_forks.md
+# medium.com based links disabled for now until Lychee gains cookie support (medium.com requires cookies to avoid infinite redirects)
+https://blog.edgeless.systems/
+https://blog.sigstore.dev/kubernetes-signals-massive-adoption-of-sigstore-for-protecting-open-source-ecosystem-73a6757da73
+


.lycheeignore
```diff
@@ -1,3 +1,7 @@
 http://localhost:
 https://github.com/edgelesssys/wiki/blob/master/documentation/constellation/customer-onboarding.md
 https://github.com/edgelesssys/wiki/blob/master/documentation/rebasing_forks.md
+# medium.com based links disabled for now until Lychee gains cookie support (medium.com requires cookies to avoid infinite redirects)
```
daniel-weisse, line 4: Maybe add tracking link to the GitHub issue?
Otherwise lgtm

Nirusu: Can add the tracking link.
The KVM website is a bit funky since the certificate is valid, but they don't seem to serve the intermediate certificate in the chain. Your browser may automatically fetch it.

In the terminal, this fails on Ubuntu 22.04 & Fedora 38 even with the most recent ca-certificates:
```
curl -v https://www.linux-kvm.org/page/Main_Page
```

On macOS, this works just fine in contrast for example.

Nirusu:
Since this currently fails all our PRs and I am not sure when linux-kvm.org will fix the links, I'll go ahead and add linux-kvm.org for to the exclude list for now so that we at least have the link checker working properly for now. Especially important when we want to modify the docs for the upcoming release.

malt3 (approved):
LGTM

PR #432, opened 2022-11-02
fix: don't run CI K-Bench with less than 2 worker nodes

chmey:
K-Bench's network benchmarks require two distinct worker nodes.: https://github.com/edgelesssys/k-bench/blob/feat/constellation/config/dp_network_internode/iperf_client_pod.yaml

Thanks @katexochen for the discovery in #421 

### Proposed change(s)

- Add check prior to running the K-Bench action to terminate early if not enough workers are scheduled.

### DIFF
--- a/.github/actions/e2e_test/action.yml
+++ b/.github/actions/e2e_test/action.yml
@@ -56,6 +56,16 @@ runs:
         echo "Invalid input for test field: ${{ inputs.test }}"
         exit 1
 
+    # K-Bench's network benchmarks require at least two distinct worker nodes.
+    - name: Validate k-bench inputs
+      if: inputs.test == 'k-bench'
+      shell: bash
+      run: |
+        if [[ "${{ inputs.workerNodesCount }}" -lt 2 ]]; then
+          echo "::error::Test K-Bench requires at least 2 worker nodes."
+          exit 1
+        fi
+
     - name: Determine build target
       id: determine-build-target
       shell: bash


.github/actions/e2e_test/action.yml
```diff
@@ -56,6 +56,21 @@ runs:
...
+      if: ${{ contains(fromJson('["k-bench"]'), inputs.test) }}
+      shell: bash
+      run: |
+        if [ "$e2eTest" = "k-bench" ] && [ "$constellationWorkerNodes" -lt 2 ]; then
+          echo "::error::Test K-Bench requires at least 2 worker nodes."
+          exit 1
+        fi
+      env:
+        constellationWorkerNodes: ${{ inputs.workerNodesCount }}
+        e2eTest: ${{ inputs.test}}
```
katexochen, line 72:
```suggestion
    - name: Validate k-bench inputs
      if:  inputs.test == 'k-bench'
      shell: bash
      run: |
        if [[ "${{ inputs.workerNodesCount }}" -lt 2 ]]; then
          echo "::error::Test K-Bench requires at least 2 worker nodes."
          exit 1
        fi
```

Would suggest to not over-complicate for now. It might be easier to just add a step for each test that requires it instead of building huge bash constructs.

.github/actions/e2e_test/action.yml
```diff
@@ -56,6 +56,18 @@ runs:
         echo "Invalid input for test field: ${{ inputs.test }}"
         exit 1
 
+    # K-Bench's network benchmarks require at least two distinct worker nodes.
+    - name: Validate k-bench inputs
+      if: inputs.test == 'k-bench'
+      shell: bash
+      run: |
+        if [ "$constellationWorkerNodes" -lt 2 ]; then
```
katexochen, line 64:
```suggestion
        if [[ "${{ inputs.workerNodesCount }}" -lt 2 ]]; then
```

katexochen (changes requested):

katexochen (approved):

PR #436, opened 2022-11-02
AB#2530 CI benchmark: compare to previous results and generate graphs

chmey:
### Proposed changes

- In K-Bench benchmark runs, compare the current benchmark to the previous performance of Constellation on the same cloud provider.
- Output the performance progression in the GitHub action summary as a markdown table (see example below).
- If running on main, update the stored benchmark in S3 for future comparison.
- Generate performance charts as in the [docs](https://docs.edgeless.systems/constellation/overview/performance) and attach them to the workflow run as artifacts.
- Add documentation for benchmarks. Specifically add walk-through on updating the stored benchmark for cloud-provider managed Kubernetes clusters.
- Separated parsing and generation of graphs from K-bench results.
- Add support to parse non-Constellation benchmark results e.g. AKS / GKE.

## Additional Information:
- check the exemplary run, its summary and artifacts (graphs): https://github.com/edgelesssys/constellation/actions/runs/3419603528
# constellation-gcp

<details>

- Commit of current benchmark: 8eb0a6803bc431bcebc2f6766ab2c6376500e106
- Commit of previous benchmark: 8f733daaf5c5509f024745260220d89ef8e6e440

| Benchmark suite | Current | Previous | Ratio |
|-|-|-|-|
| pod_create (ms) | 135 | 198 | 0.682 ⬇️ |
| pod_list (ms) | 100 | 99 | 1.01 ⬆️ |
| pod_get (ms) | 98 | 98 | 1.0 ⬆️ |
| pod_update (ms) | 187 | 132 | 1.417 ⬆️ |
| pod_delete (ms) | 119 | 108 | 1.102 ⬆️ |
| svc_create (ms) | 156 | 149 | 1.047 ⬆️ |
| svc_list (ms) | 97 | 96 | 1.01 ⬆️ |
| svc_get (ms) | 97 | 96 | 1.01 ⬆️ |
| svc_update (ms) | 100 | 101 | 0.99 ⬇️ |
| svc_delete (ms) | 143 | 139 | 1.029 ⬆️ |
| depl_create (ms) | 201 | 218 | 0.922 ⬇️ |
| depl_list (ms) | 101 | 101 | 1.0 ⬆️ |
| depl_update (ms) | 111 | 110 | 1.009 ⬆️ |
| depl_scale (ms) | 391 | 391 | 1.0 ⬆️ |
| depl_delete (ms) | 401 | 402 | 0.998 ⬇️ |
| net_internode_snd (Mbit/s) | 953.0 | 964.0 | 1.01 ⬆️ |
| net_intranode_snd (Mbit/s) | 18500.0 | 18600.0 | 1.01 ⬆️ |
| fio_root_async_R70W30_R (MiB/s) | 0.45 | 0.45| 1.0 ⬆️ |
| fio_root_async_R70W30_W (MiB/s) | 0.20 | 0.20 | 1.0 ⬆️ |
| fio_root_async_R100W0_R (MiB/s) | 0.59 | 0.59 | 1.0 ⬆️ |
| fio_root_async_R0W100_W (MiB/s) | 1.18 | 1.18 | 1.0 ⬆️ |

</details>

### DIFF
--- a/.github/actions/e2e_kbench/README.md
+++ b/.github/actions/e2e_kbench/README.md
@@ -1,0 +1,162 @@
+# K-Bench
+
+## Continuous Benchmarking
+The K-Bench action runs K-Bench benchmarks on Constellation clusters.
+The benchmark suite records storage, network, and Kubernetes API benchmarks.
+
+After testing, the action compares the results of the benchmarks to previous results of Constellation on the same cloud provider. That way, it is possible to evaluate performance progression throughout the development.
+
+The data of previous benchmarks is stored in the private S3 artifact store.
+
+In order to support encrypted storage, the action deploys the [Azure CSI](https://github.com/edgelesssys/constellation-azuredisk-csi-driver) and [GCP CSI](https://github.com/edgelesssys/constellation-gcp-compute-persistent-disk-csi-driver) drivers. It uses a [fork](https://github.com/edgelesssys/k-bench) of VMware's K-Bench. The fork deploys volumes that use the `encrypted-storage` storage class. Also, it has support to authenticate against GCP which is required to update the stored records for GKE.
+
+### Displaying Performance Progression
+The action creates a summary of the action and attaches it the workflow execution log.
+
+The table compares the current benchmark results of Constellation on the selected cloud provider to the previous records (of Constellation on the cloud provider).
+
+The hashes of the two commits that are the base for the comparison are prepended to the table.
+
+Example table:
+
+<details>
+
+- Commit of current benchmark: 8eb0a6803bc431bcebc2f6766ab2c6376500e106
+- Commit of previous benchmark: 8f733daaf5c5509f024745260220d89ef8e6e440
+
+| Benchmark suite | Current | Previous | Ratio |
+|-|-|-|-|
+| pod_create (ms) | 135 | 198 | 0.682 ⬇️ |
+| pod_list (ms) | 100 | 99 | 1.01 ⬆️ |
+| pod_get (ms) | 98 | 98 | 1.0 ⬆️ |
+| pod_update (ms) | 187 | 132 | 1.417 ⬆️ |
+| pod_delete (ms) | 119 | 108 | 1.102 ⬆️ |
+| svc_create (ms) | 156 | 149 | 1.047 ⬆️ |
+| svc_list (ms) | 97 | 96 | 1.01 ⬆️ |
+| svc_get (ms) | 97 | 96 | 1.01 ⬆️ |
+| svc_update (ms) | 100 | 101 | 0.99 ⬇️ |
+| svc_delete (ms) | 143 | 139 | 1.029 ⬆️ |
+| depl_create (ms) | 201 | 218 | 0.922 ⬇️ |
+| depl_list (ms) | 101 | 101 | 1.0 ⬆️ |
+| depl_update (ms) | 111 | 110 | 1.009 ⬆️ |
+| depl_scale (ms) | 391 | 391 | 1.0 ⬆️ |
+| depl_delete (ms) | 401 | 402 | 0.998 ⬇️ |
+| net_internode_snd (Mbit/s) | 953.0 | 964.0 | 1.01 ⬆️ |
+| net_intranode_snd (Mbit/s) | 18500.0 | 18600.0 | 1.01 ⬆️ |
+| fio_root_async_R70W30_R (MiB/s) | 0.45 | 0.45| 1.0 ⬆️ |
+| fio_root_async_R70W30_W (MiB/s) | 0.20 | 0.20 | 1.0 ⬆️ |
+| fio_root_async_R100W0_R (MiB/s) | 0.59 | 0.59 | 1.0 ⬆️ |
+| fio_root_async_R0W100_W (MiB/s) | 1.18 | 1.18 | 1.0 ⬆️ |
+
+</details>
+
+### Drawing Performance Charts
+The action also draws graphs as used in the [Constellation docs](https://docs.edgeless.systems/constellation/next/overview/performance). The graphs compare the performance of Constellation to the performance of managed Kubernetes clusters.
+
+Graphs are created with every run of the benchmarking action. The action attaches them to the `benchmark` artifact of the workflow run.
+
+## Updating Stored Records
+
+### Managed Kubernetes
+One must manually update the stored benchmark records of managed Kubernetes:
+
+### AKS
+Follow the [Azure documentation](https://learn.microsoft.com/en-us/azure/aks/learn/quick-kubernetes-deploy-portal?tabs=azure-cli) to create an AKS cluster of desired benchmarking settings (region, instance types). If comparing against Constellation clusters with CVM instances, make sure to select the matching CVM instance type on Azure as well.
+
+Once the cluster is ready, set up managing access via `kubectl` and take the benchmark:
+```bash
+# Setup
+git clone https://github.com/edgelesssys/k-bench.git
+cd k-bench && git checkout feat/constellation
+./install.sh
+
+# Remove the Constellation encrypted storage class
+# Remember to revert this change before running K-Bench on Constellation!
+yq 'del(.spec.storageClassName)' config/dp_fio/fio_pvc.yaml
+yq 'del(.spec.storageClassName)' config/dp_netperf_internode/netperf_pvc.yml
+yq 'del(.spec.storageClassName)' config/dp_network_internode/netperf_pvc.yaml
+yq 'del(.spec.storageClassName)' config/dp_network_intranode/netperf_pvc.yml
+
+# Run K-Bench
+mkdir -p ./out
+kubectl create namespace kbench-pod-namespace --dry-run=client -o yaml | kubectl apply -f -
+./run.sh -r "kbench-AKS" -t "default" -o "./out/"
+kubectl delete namespace kbench-pod-namespace --wait=true || true
+kubectl create namespace kbench-pod-namespace --dry-run=client -o yaml |  kubectl apply -f -
+./run.sh -r "kbench-AKS" -t "dp_fio" -o "./out/"
+kubectl delete namespace kbench-pod-namespace --wait=true || true
+kubectl create namespace kbench-pod-namespace --dry-run=client -o yaml |  kubectl apply -f -
+./run.sh -r "kbench-AKS" -t "dp_network_internode" -o "./out/"
+kubectl delete namespace kbench-pod-namespace --wait=true || true
+kubectl create namespace kbench-pod-namespace --dry-run=client -o yaml |  kubectl apply -f -
+./run.sh -r "kbench-AKS" -t "dp_network_intranode" -o "./out/"
+
+# Benchmarks done, do processing.
+mkdir -p "./out/AKS"
+mv ./out/results_kbench-AKS_*m/* "./out/kbench-AKS/"
+
+# Parse
+git clone https://github.com/edgelesssys/constellation.git
+mkdir -p benchmarks
+BDIR=benchmarks
+EXT_NAME=AKS
+KBENCH_RESULTS=k-bench/out/
+
+python constellation/.github/actions/e2e_kbench/evaluate/parse.py
+
+# Upload result to S3
+S3_PATH=s3://edgeless-artifact-store/constellation/benchmarks
+aws s3 cp benchmarks/AKS.json ${S3_PATH}/AKS.json
+```
+
+### GKE
+Create a GKE cluster of desired benchmarking settings (region, instance types). If comparing against Constellation clusters with CVM instances, make sure to select the matching CVM instance type on GCP and enable **confidential** VMs as well.
+
+Once the cluster is ready, set up managing access via `kubectl` and take the benchmark:
+```bash
+# Setup
+git clone https://github.com/edgelesssys/k-bench.git
+cd k-bench && git checkout feat/constellation
+./install.sh
+
+# Remove the Constellation encrypted storage class
+# Remember to revert this change before running K-Bench on Constellation!
+yq 'del(.spec.storageClassName)' config/dp_fio/fio_pvc.yaml
+yq 'del(.spec.storageClassName)' config/dp_netperf_internode/netperf_pvc.yml
+yq 'del(.spec.storageClassName)' config/dp_network_internode/netperf_pvc.yaml
+yq 'del(.spec.storageClassName)' config/dp_network_intranode/netperf_pvc.yml
+
+# Run K-Bench
+mkdir -p ./out
+kubectl create namespace kbench-pod-namespace --dry-run=client -o yaml | kubectl apply -f -
+./run.sh -r "kbench-GKE" -t "default" -o "./out/"
+kubectl delete namespace kbench-pod-namespace --wait=true || true
+kubectl create namespace kbench-pod-namespace --dry-run=client -o yaml |  kubectl apply -f -
+./run.sh -r "kbench-GKE" -t "dp_fio" -o "./out/"
+kubectl delete namespace kbench-pod-namespace --wait=true || true
+kubectl create namespace kbench-pod-namespace --dry-run=client -o yaml |  kubectl apply -f -
+./run.sh -r "kbench-GKE" -t "dp_network_internode" -o "./out/"
+kubectl delete namespace kbench-pod-namespace --wait=true || true
+kubectl create namespace kbench-pod-namespace --dry-run=client -o yaml |  kubectl apply -f -
+./run.sh -r "kbench-GKE" -t "dp_network_intranode" -o "./out/"
+
+# Benchmarks done, do processing.
+mkdir -p "./out/GKE"
+mv ./out/results_kbench-GKE_*m/* "./out/kbench-GKE/"
+
+# Parse
+git clone https://github.com/edgelesssys/constellation.git
+mkdir -p benchmarks
+BDIR=benchmarks
+EXT_NAME=GKE
+KBENCH_RESULTS=k-bench/out/
+
+python constellation/.github/actions/e2e_kbench/evaluate/parse.py
+
+# Upload result to S3
+S3_PATH=s3://edgeless-artifact-store/constellation/benchmarks
+aws s3 cp benchmarks/GKE.json ${S3_PATH}/GKE.json
+```
+
+### Constellation
+The action updates the stored Constellation records for the selected cloud provider when running on the main branch.
--- a/.github/actions/e2e_kbench/action.yml
+++ b/.github/actions/e2e_kbench/action.yml
@@ -11,6 +11,7 @@ inputs:
 
 runs:
   using: "composite"
+
   steps:
     - name: Setup python
       uses: actions/setup-python@13ae5bb136fac2878aff31522b9efb785519f984 # tag=v4.3.0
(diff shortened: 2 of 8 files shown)


katexochen:
LGTM so far, not sure if someone with more Python skill want to take @Nirusu @derpsteb? 

The only thing I would like to improve is to reference the sha of the commit used for comparison in the summary output. Currently, the result is hard to interpret as it is unclear when/on which commit the last benchmark was done.

.github/actions/k-bench/evaluate/graph.py
```diff
@@ -0,0 +1,181 @@
...
+            'ENV variables BDIR is required.')
+    return out_dir
+
+
+def bar_chart(data, headers, title='', suffix='', val_label=True, y_log=False):
+    """Draws a bar chart with multiple bars per data point.
+
+    Args:
+        data (list): List of lists of value points.
+        headers (list): List of headers (x-axis).
```
Nirusu, line 83:
```suggestion
        data (dict[str, list]): List of lists of value points.
        headers (dict[str, list]): List of headers (x-axis).
```

Both are not lists.

.github/actions/k-bench/evaluate/graph.py
```diff
@@ -0,0 +1,181 @@
...
+    Returns: out_dir
+    """
+    out_dir = os.environ.get('BDIR', None)
+    if not out_dir:
+        raise TypeError(
+            'ENV variables BDIR is required.')
+    return out_dir
+
+
+def bar_chart(data, headers, title='', suffix='', val_label=True, y_log=False):
```
Nirusu, line 78:
```suggestion
def bar_chart(data, headers, title='', suffix='', y_log=False):
```

Unused variable.

.github/actions/k-bench/evaluate/graph.py
```diff
@@ -0,0 +1,181 @@
...
+    """Draws a bar chart with multiple bars per data point.
+
+    Args:
+        data (list): List of lists of value points.
+        headers (list): List of headers (x-axis).
+        title (str, optional): The title for the chart. Defaults to "".
+        suffix (str, optional): The suffix for values e.g. "MiB/s". Defaults to "".
+        val_label (bool, optional): Put a label of the value over the bar chart. Defaults to True.
```
Nirusu, line 86:
```suggestion
```

Since it's unused.

.github/actions/k-bench/evaluate/parse.py
```diff
@@ -0,0 +1,84 @@
...
+        tests = {ext_provider_name: benchmark_path}
+        out_file_name = "{nm}.json".format(nm=ext_provider_name)
+    else:
+        # Constellation benchmark.
+        benchmark_path = os.path.join(
+            base_path,
+            "kbench-constellation-" + csp,
+        )
+        tests = {f"constellation-{csp}": benchmark_path}
+        out_file_name = "constellation-{csp}.json".format(csp=csp)
```
Nirusu, line 59: You could probably unify the logic here by:
```python
if not ext_provider_name:
   ext_provider_name = f"constellation-{csp}"
```

.github/actions/k-bench/README.md
```diff
@@ -0,0 +1,157 @@
+# K-Bench
+
+## Continous Benchmarking
```
Nirusu, line 3:
```suggestion
## Continuous Benchmarking
```

.github/actions/k-bench/README.md
```diff
@@ -0,0 +1,157 @@
...
+The K-Bench action runs K-Bench benchmarks on Constellation clusters.
+The benchmark suite records storage, network, and Kubernetes API benchmarks.
+
+After testing, the action compares the results of the benchmarks to previous results of Constellation on the same cloud provider. That way, it is possible to evaluate performance progression throughout the development.
+
+The data of previous benchmarks is stored in the public S3 bucket.
```
Nirusu, line 9: Mh, personally my idea was to deprecate the old public S3 bucket slowly. Do we actually publish this data somewhere publicly?

I wonder if we shouldn't just create a private bucket for various artifacts we use internally.

.github/actions/k-bench/README.md
```diff
@@ -0,0 +1,157 @@
...
+git clone https://github.com/edgelesssys/k-bench.git
+cd k-bench && git checkout feat/constellation
+./install.sh
+
+# Remove the Constellation encrypted storage class 
+# Remember to revert this change before running K-Bench on Constellation!
+yq 'del(.spec.storageClassName)' config/dp_fio/fio_pvc.yaml
+yq 'del(.spec.storageClassName)' config/dp_netperf_internode/netperf_pvc.yml
+yq 'del(.spec.storageClassName)' config/dp_network_internode/netperf_pvc.yaml
+yq 'del(.spec.storageClassName)' config/dp_network_intranode/netperf_pvc.yml
```
Nirusu, line 73: You could generally avoid this by assuming that "encrypted-storage" is also the default storage class on Constellation. I think if you follow the docs it actually should be the default one.

.github/actions/k-bench/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Continous Benchmarking
+The K-Bench action runs K-Bench benchmarks on Constellation clusters.
+The benchmark suite records storage, network, and Kubernetes API benchmarks.
+
+After testing, the action compares the results of the benchmarks to previous results of Constellation on the same cloud provider. That way, it is possible to evaluate performance progression throughout the development.
+
+The data of previous benchmarks is stored in the public S3 bucket.
+
+In order to support encrypted storage, the action deploys the [GCP CSI](https://github.com/edgelesssys/constellation-gcp-compute-persistent-disk-csi-driver) and [Azure CSI](https://github.com/edgelesssys/constellation-azuredisk-csi-driver) drivers. It uses a [fork](https://github.com/edgelesssys/k-bench) of VMware's K-Bench. The fork deploys volumes that use the `encrypted-storage` storage class. Also, it has support to authenticate against GCP which is required to update the stored records for GKE.
```
Nirusu, line 11:
```suggestion
In order to support encrypted storage, the action deploys the [Azure CSI](https://github.com/edgelesssys/constellation-azuredisk-csi-driver) and [GCP CSI](https://github.com/edgelesssys/constellation-gcp-compute-persistent-disk-csi-driver) drivers. It uses a [fork](https://github.com/edgelesssys/k-bench) of VMware's K-Bench. The fork deploys volumes that use the `encrypted-storage` storage class. Also, it has support to authenticate against GCP which is required to update the stored records for GKE.
```

Reads better alphabetically (and follow the order below).

.github/actions/k-bench/evaluate/graph.py
```diff
@@ -0,0 +1,181 @@
...
+from collections import defaultdict
+
+import numpy as np
+from matplotlib import pyplot as plt
+
+SUBJECTS = [
+    "constellation-gcp",
+    "GKE",
+    "constellation-azure",
+    "AKS",
```
Nirusu, line 13:
```suggestion
    "Constellation on GCP",
    "GKE",
    "Constellation on Azure",
    "AKS",
```

Or similar. I might even suggest switching the order of GCP and Azure for consistency in our docs but if it's too much effort it's not necessary I guess.

.github/actions/k-bench/evaluate/parse.py
chmey: True, thanks!

.github/actions/k-bench/evaluate/graph.py
chmey: Ah, thanks. Forgot to add that switch to the render function.

.github/actions/k-bench/evaluate/graph.py
chmey: Headers should be a list right? It comes from ` dict.values()`.

.github/actions/k-bench/evaluate/graph.py
chmey: Sure. Since this "constellation-$CSP" format is also used for the directory structure it is a bit messy to change it there and introduce spaces.

Therefore, I suggest just adding a legend list that the graph generator uses. Changing the order shouldn't be of concern though.

chmey:
Updated the stored result JSON documents in S3 to resemble the new structure:

```json
 "subject": "constellation-gcp",
 "commit": "N/A",
 "kbench": {
   "pod_create": 306.521,
    ...
  }
}

```

chmey:
Two test runs that will also update the stored files and put their commit hash:
- GCP: https://github.com/edgelesssys/constellation/actions/runs/3431879073/jobs/5720526028
- Azure: https://github.com/edgelesssys/constellation/actions/runs/3431880996/jobs/5720530163

.github/actions/k-bench/evaluate/graph.py
Nirusu: `dict.values()` doesn't directly return a list. You get some weird [view object](https://docs.python.org/3.11/library/stdtypes.html#dictionary-view-objects) that can also update dynamically.

For example, if you do this:
```python
mydict = {"first": True}
values = mydict.values()
print(values)
mydict["second"] = False
print(values)
```

You get:
```python
dict_values([True])
dict_values([True, False])
```

So my suggestion here technically also wasn't correct.

But if you want to return a list you can just cast it to a list with `list(mydict.values())`. That should be correct then.

chmey:
Updated runs using the new bucket. Please note, that only the runs from these one will round their values.
- GCP: https://github.com/edgelesssys/constellation/actions/runs/3438410209
- Azure: https://github.com/edgelesssys/constellation/actions/runs/3438411735

I will now include the condition to only update the records when running on main, again.

Please review again @Nirusu @katexochen

.github/actions/k-bench/README.md
chmey: You're right, but I won't address it in this PR anymore.

.github/actions/k-bench/evaluate/compare.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+    path_prev, path_curr = get_paths()
+
+    with open(path_prev) as f_prev:
+        bench_prev = json.load(f_prev)
+    with open(path_curr) as f_curr:
+        bench_curr = json.load(f_curr)
+
+    name = bench_curr['subject']
+    if name != bench_prev['subject']:
```
derpsteb, line 70: Presence of these keys is not checked. You could use `in` or `try`..`catch KeyError`.

.github/actions/k-bench/evaluate/compare.py
```diff
@@ -0,0 +1,114 @@
...
+        '- Commit of current benchmark: {ch}'.format(ch=bench_curr['commit']),
+        '- Commit of previous benchmark: {ch}'.format(ch=bench_prev['commit']),
+        '',
+        '| Benchmark suite | Current | Previous | Ratio |',
+        '|-|-|-|-|',
+    ]
+
+    for subtest, _ in bench_prev['kbench'].items():
+        val_prev = bench_prev['kbench'][subtest]
+        val_curr = bench_curr['kbench'][subtest]
```
derpsteb, line 88: Same here.

.github/actions/k-bench/evaluate/graph.py
```diff
@@ -0,0 +1,189 @@
...
+    # Draw legend
+    ax.legend(bars, LEGEND_NAMES)
+    if y_log:
+        ax.set_yscale('log')
+    ax.set_xticks(np.arange(len(headers)))
+    ax.set_xticklabels(headers)
+
+    plt.setp(ax.get_xticklabels(), fontsize=LABEL_FONTSIZE,
+             rotation=LABEL_ROTATE_BY)
+    plt.title(f'{title} ({suffix})')
```
derpsteb, line 132: You're mixing use of `.format` and f-strings here. I think it would be nice to stick to one. I personally prefer f-strings, but chose as you like :)

.github/actions/k-bench/evaluate/compare.py
```diff
@@ -0,0 +1,114 @@
...
+
+    Create a markdown table showing the benchmark progressions.
+
+    Print the result to stdout.
+    """
+    path_prev, path_curr = get_paths()
+
+    with open(path_prev) as f_prev:
+        bench_prev = json.load(f_prev)
+    with open(path_curr) as f_curr:
```
derpsteb, line 66: The files may not exist. Maybe catch the FileNotFoundError?

.github/actions/k-bench/evaluate/graph.py
```diff
@@ -0,0 +1,189 @@
...
+def main():
+    """Read the files and create diagrams."""
+    out_dir = configure()
+    combined_results = defaultdict(dict)
+
+    for test in SUBJECTS:
+        # Read the previous results
+        read_path = os.path.join(
+            out_dir, "{subject}.json".format(subject=test))
+        with open(read_path, 'r') as res_file:
```
derpsteb, line 146: The files may not exist. Catch the error?

.github/actions/k-bench/evaluate/parse.py
```diff
@@ -0,0 +1,83 @@
...
+    # Expect the results in directory:
+    # kbench-EXT_PROVIDER_NAME/
+    benchmark_path = os.path.join(
+        base_path,
+        "kbench-" + ext_provider_name,
+    )
+    tests = {ext_provider_name: benchmark_path}
+    out_file_name = "{nm}.json".format(nm=ext_provider_name)
+
+    if not os.path.exists(benchmark_path):
```
derpsteb, line 58: [mkdir](https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir) can ensure the path exists if you want that. Seems more robust than failing if the path doesn't exist yet?

.github/actions/k-bench/evaluate/parse.py
```diff
@@ -0,0 +1,83 @@
...
+        raise Exception(
+            f'Path to benchmarks {benchmark_path} does not exist.')
+
+    # Parse subtest
+    default_results = default.eval(tests=tests)
+    network_results = network.eval(tests=tests)
+    fio_results = fio.eval(tests=tests)
```
derpsteb, line 65: Maybe change the function name so it doesn't look like the `eval` keyword?

.github/actions/k-bench/evaluate/parse.py
```diff
@@ -0,0 +1,83 @@
...
+
+    combined_results = defaultdict(dict)
+    combined_results["commit"] = commit_hash
+    combined_results["subject"] = next(iter(tests))
```
derpsteb, line 69: Catch the exception or give a default?

.github/actions/k-bench/evaluate/graph.py
chmey: I prefer f strings as well, but I recently read the https://github.com/wemake-services/wemake-python-styleguide and tried to adapt a bit of it ;)

.github/actions/k-bench/evaluate/parse.py
chmey: Yes, cheers!

.github/actions/k-bench/evaluate/parse.py
```diff
@@ -0,0 +1,83 @@
...
+    benchmark_path = os.path.join(
+        base_path,
+        "kbench-" + ext_provider_name,
+    )
+    tests = {ext_provider_name: benchmark_path}
+    out_file_name = "{nm}.json".format(nm=ext_provider_name)
+
+    if not os.path.exists(benchmark_path):
+        raise ValueError(
+            'Benchmarks do not exist at {benchmark_path}.'.format(path=benchmark_path))
```
chmey, line 60: Use `{path)`

.github/actions/k-bench/evaluate/graph.py
```diff
@@ -0,0 +1,193 @@
...
+
+    # Combine the evaluation of the Kubernetes API benchmarks
+    for i, api in enumerate([pod_key2header, svc_key2header, depl_key2header]):
+        api_data = {}
+        for s in SUBJECTS:
+            points = combined_results[s]['kbench']
+            subject_data = [points[h] for h in api]
+            api_data[s] = subject_data
+        hdrs = list(api.values())
+        bar_chart(data=api_data, headers=hdrs,
```
Nirusu, line 161: Nit: A space would be good so that:
"API Latency(ms)" -> "API Latency (ms)"

(same for the other graphs)

Nirusu (commented):
Minor nits regarding type hinting but generally the code looks good to me.

Having non-float numbers for the latency also seems way cleaner now, nice! There's still some slight overlap though which _maybe_ could be minimized by slightly decreasing the font size by 1-2 pt or slightly rotating the values? Though I guess that wouldn't be sufficient for the storage throughput graphs with floats.

I think the one other point I would have is more generally to the benchmarks rather than specific to this PR. The API latency graph seems to be a bit all over the place in terms of the results. In our docs we have completely different results and I am not sure if these are somewhat reproducible. Also I wonder if running k-bench in the CI has some impacts on the flakiness of this graphs versus running it locally. I could see that the shared runners are likely constantly under high load and thus might create higher / non-deterministic numbers than on a self-hosted runner or a local machine.

Do you have some data on this? Because if the values are super flaky then I am not sure if it makes sense to track this in the future.

But don't worry, I don't want you to remove this now :D Just generally curious that when we do comparisons between builds or releases if it makes sense to actually try to interpret this number.

chmey:
> Having non-float numbers for the latency also seems way cleaner now, nice! There's still some slight overlap though which maybe could be minimized by slightly decreasing the font size by 1-2 pt or slightly rotating the values? Though I guess that wouldn't be sufficient for the storage throughput graphs with floats.


The `bar_label` docs (https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.bar_label.html) seem to only allow to add a padding. I will add a padding of 1pt.

For now, when creating a cleaner chart (i.e. for customer or docs), maybe set `var_label` to `False` to not generate these labels. Alternatively, one could increase the distance between the individual bars. In my opinion this would look worse.

> Do you have some data on this? Because if the values are super flaky then I am not sure if it makes sense to track this in the future.

True. Since the API benchmarks involve the machine starting the benchmarks (here: GitHub actions runner) it might add some arbitrary delay. I agree, it is less reproducible that way. My suggestion is: Run the K-Bench benchmarks from a machine that has somewhat stable networking i.e. in office or always in the same cloud zone. I will create a ticket for that, and you can discuss it further.

.github/actions/k-bench/evaluate/parse.py
```diff
@@ -0,0 +1,83 @@
+"""Parse logs of K-Bench tests and generate performance graphs."""
+import json
+import os
+from collections import defaultdict
+from typing import Tuple
+
+from evaluators import default, fio, network
+
+
+def configure() -> Tuple[str, str, str, str | None]:
```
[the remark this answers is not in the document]
chmey: The function raises an exception if the main ENV variables are undefined. It won't return None for them.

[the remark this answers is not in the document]
Nirusu: Ah yes sorry, was a bit too early for me. 
Though would've need to be str, str, str, str | None, str but I see you already fixed that :)

Nirusu:
Yeah, I think we can investigate this at a later point with shared runners & self-hosted runners.

.github/actions/k-bench/evaluate/parse.py
derpsteb: If it's about checking if the path is correct I would expect a check using normpath/realpath with an expected value.

katexochen (approved):
LGTM, thanks for putting in all the effort! :rocket:

Nirusu (approved):
Also LGTM, thanks for the great work!

PR #439, opened 2022-11-03
aws: join serivce support + verify service

3u13r:
### Proposed change(s)

- Bump Join-Service inital AWS support
- Add needed AWS IAM permissions which are attached to the VMs. Need to describe Image due to `func (v *Validator) tpmEnabled(...)` in the AWS validator
- Add support for AWS in verify service.


daniel-weisse (approved):

verify/cmd/main.go
```diff
@@ -34,6 +35,8 @@ func main() {
 
 	var issuer server.AttestationIssuer
 	switch *provider {
+	case "aws":
```
daniel-weisse, line 38: We can replace this switch block with
```Go
switch cloudprovider.FromString(*provider) {
case cloudprovider.AWS:
	issuer = aws.NewIssuer()
case cloudprovider.GCP:
//...

PR #465, opened 2022-11-07
rfc: reproducible-builds

leongross:
### Proposed change(s)

- Create an RFC for our requirements for reproducible builds

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/rfc/reproducible-builds.md
+++ b/rfc/reproducible-builds.md
@@ -4,7 +4,7 @@ This is the only way to ensure a verifiable path from source code to binary.
 Every step of the build process must be deterministic.
 
 ## Definition of Goals
-1. All our OCIs executed in the cluster must be reproducible bit by bit .
+1. All our OCIs executed in the cluster must be reproducible bit by bit.
 1. For that, every compiled executable has to be deterministically compiled.
 1. If we change parts of our codebase (i.e. add several other programming languages) we should not have to make major changes to the image build system.
 1. Since docker does not offer built-in options to remove timestamps (which breaks reproducibility) we have to move to another OCI build system.
@@ -19,14 +19,17 @@ Since podman internally [uses](https://podman.io/blogs/2018/10/31/podman-buildah
 With buildah, Containerfiles/Dockerfiles can be used as usual. This means we can adjust the build to include necessary libraries for i.e CGO builds. Only the build command itself has to be adjusted to omit timestamps.
 
 ## `ko`
-Ko is limited to building OCI images for go applications. By default images such as [distroless](https://github.com/GoogleContainerTools/distroless) are used. These are minimal and hence very small images, that are stripped of anything but runtime dependencies. Problems arise when the default images do not satisfy our dependency needs (as they currently do with the `disk-mapper` which relies on a dynamically built `libcryptsetup` library). To solve this issue we have two options:
+Ko is limited to building OCI images for go applications. By default images such as [distroless](https://github.com/GoogleContainerTools/distroless) are used as base images. These are minimal and hence very small images, that are stripped of anything but runtime dependencies. Problems arise when the default images do not satisfy our dependency needs (as they currently do with the `disk-mapper` which relies on a dynamically built `libcryptsetup` library). To solve this issue we have three options:
 
 1. Build our own base images independent from or inspired by distroless
 2. Fork distroless, edit underlying [bazel dependencies](https://github.com/GoogleContainerTools/distroless/blob/main/debian_archives.bzl), build the image
+3. Use `apko` to build minimal Alpine images. These images can be configured via `apko` and a declarative `*.yaml` config file.
 
-Option `1.` results in a similar maintenance work as using buildah.
+Option `1.`: results in a similar maintenance work as using `buildah`.
 <br>
-Option `2.` results in an even bigger maintenance overhead since we currently do not use bazel at all.
+Option `2.`: results in an even bigger maintenance overhead since we currently do not use `bazel` at all.
+<br>
+Option `3.`: for our current use cases very easy to configure.
 
 ## `kaniko`
 Over time, issues complaining about breaking/inconsistent reproducibility accumulated.
@@ -34,7 +37,7 @@ This seems to happen more or less regularly. We should try to avoid a build syst
 
 ## Steps to Achieve Goals
 
-### Executables
+## Executables
 * Remove metadata from binaries, namely:
   * Timestamps
   * Build IDs
@@ -45,7 +48,7 @@ This seems to happen more or less regularly. We should try to avoid a build syst
 * Eliminate dependencies on libraries (make executable static)
 
 Striping metadata from the binary can be done in the building process.
-This can be achieved by setting the appropriate compiler and linker flags (see [`go tool link`](https://pkg.go.dev/cmd/link) and `[go help build]`(https://pkg.go.dev/cmd/go)).
+This can be achieved by setting the appropriate compiler and linker flags (see [`go tool link`](https://pkg.go.dev/cmd/link) and [`go help build`](https://pkg.go.dev/cmd/go).
 
 * `buildvcs=false`: Omit version control information
 * `-trimpath`: Remove file system paths from executable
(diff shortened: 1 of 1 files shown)


rfc/reproducible-builds.md
```diff
@@ -0,0 +1,66 @@
...
+  * Version numbers
+  * Paths
+  * Anything that might change for another host OS / time / ...
+* Eliminate dependencies on libraries (make executable static)
+
+Striping metadata from the binary can be done in the building process.
+This can be achieved by setting the appropriate compiler flags.
+
+```bash
+$ CGO_ENABLED=0 go build -o <out_name> -buildvcs=false -trimpath -ldflags "-s -w -buildid=''"
```
datosh, line 30: Maybe we could summarize shortly (or link a resource that does) what these flags achieve. I had to look up the meaning of `-s -w`, but `go tool link` provides the information.

rfc/reproducible-builds.md
```diff
@@ -0,0 +1,66 @@
+# Reproducible Builds
+To ensure the security of constellation's supply chain, we need to make our software builds reproducible.
+This is the only way to ensure a verifiable path from source code to binary.
+Every step of the build process must be deterministic.
+
+## Definition of Goals
+1. All our OCIs executed in the cluster must be reproducible bit by bit .
+1. For that, every compiled executable has to be deterministically compiled.
+1. If we change parts of our codebase (i.e. add several other programming languages) we should not have to make major changes to the image build system.
+1. Moving from `Docker` to a reproducible build system should have a minimum overhead. Ideally, we can reuse our existing `Dockerfile`s.
```
datosh, line 10: We could state why we are moving away from `Docker` or more precisely how `Docker` breaks reproducibility.

datosh (commented):
Great summary! This paints one clear path forward to implement reproducible builds consistently.

I have learned in the past that it helps to mention the "paths not taken", to remember why certain tools were not chosen. I often forget the reason two weeks after. A single sentence is often enough as a reminder.

I know we have considered `kaniko` but we have chosen `buildah` in the end. I know there is also `podman` would those be an alternative?

Once we have a decision regarding `ko` we should also include it in the RFC.

rfc/reproducible-builds.md
```diff
@@ -0,0 +1,66 @@
...
+* The timestamps of the files in the image (creation, modification) must be identical for each build.
+* Every component that is shipped with the image has to be identical.
+
+To achieve this we will use [buildah](https://github.com/containers/buildah). It currently meets all the requirements mentioned above.
+
+To ensure that the final image is deterministic, a pattern such as the following should be followed:
+
+```Dockerfile
+FROM <image>@sha256:<hash> as build
+RUN  <install_deps>
```
Nirusu, line 46: How do we ensure that we get the same version from the remote repository? We can pin versions, but does Fedora actually keep older package versions on their repositories? Ubuntu, for example, does not.

daniel-weisse: For any binary that does not use CGO, we should not have any dependency other than Go (and the tools required to install Go).
Go keeps older version available, the version of tools like wget does not matter for the final build, as they are not used by our binary.

Container images that need CGO will require us pinning the versions of the C libraries we link against. These are typically not updated frequently and I think Fedora keeps them around for longer. However, we obviously need to check this first.

Nirusu: The CGo case is where I could see issues. From googling, it seems that Koji has longer term storage of packages, but older packages can indeed disappear on the mirrors.

But I guess we would need to test that ourselves to have a better answer than educated guessing.

malt3: Maybe we can have images with explicitly two layers for CGO:

- All dependencies (shared libraries) on one layer
- Our go binary on the upper layer

This allows us to reuse the dependency layer from our own container registry in catastrophic events.
We could even split this into explicit Dockerfiles for external deps and our own code to make the layers addressable.

thomasten: > This allows us to reuse the dependency layer from our own container registry in catastrophic events. We could even split this into explicit Dockerfiles for external deps and our own code to make the layers addressable.

We chose this approach for EGo and EdgelessDB. (Just for reproducible build of the binaries. The container images are not reproducible.)

rfc/reproducible-builds.md
```diff
@@ -91,3 +93,35 @@ buildah build \
 
 The result is an image with one deterministic layer (pinned by the hash) and deterministic build artifacts.
 Hence, the entire build is reproducible.
+
+### `apko` / `ko`
+To include c libraries into a distroless minimal image, we have to rebuild the this base image.
```
datosh, line 98:
```suggestion
To include c libraries into a distroless minimal image, we have to rebuild the base image.
```

rfc/reproducible-builds.md
```diff
@@ -91,3 +93,35 @@ buildah build \
...
+
+```yaml
+contents:
+  repositories:
+    - https://dl-cdn.alpinelinux.org/alpine/edge/main
+    - https://dl-cdn.alpinelinux.org/alpine/edge/community  # for cryptsetup
+  packages:
+    - alpine-base
+    - cryptsetup-dev  # for disk-mapper
+    - gcompat         # for glibc/musl comparability
```
datosh, line 110:
```suggestion
    - gcompat         # for glibc/musl compatability
```

rfc/reproducible-builds.md
```diff
@@ -91,3 +93,35 @@ buildah build \
...
+
+environment:
+  PATH: /usr/sbin:/sbin:/usr/bin:/bin
+```
+
+Then in our `.ko.yaml`, we can use the newly created image as a base image, also just for certain build ids:
+```yaml
+baseImageOverrides:
+  github.com/edgelesssys/constellation/v2/disk-mapper/cmd: edgelesssys/alpine-libcryptsetup:base
```
datosh, line 119: What are the `apko` commands you have used to get to this point?

I have tried to run `apko`, but it has a dependency on `apk`, so the easiest way is to run this build inside a container as well?

leongross: Good point. I built the image using chainguard's docker image, I will mention that in the rfc.

rfc/reproducible-builds.md
```diff
@@ -0,0 +1,134 @@
...
+
+## OCI build alternatives
+This is a list comparing different build OCI builders. This list does not claim to be complete, since we only focus on points relevant to our current needs.
+
+## `buildah/podman`
+Since podman internally [uses](https://podman.io/blogs/2018/10/31/podman-buildah-relationship.html) buildah to build the image and this rfc only covers building and not execution the names are uses synonymously.
+With buildah, Containerfiles/Dockerfiles can be used as usual. This means we can adjust the build to include necessary libraries for i.e CGO builds. Only the build command itself has to be adjusted to omit timestamps.
+
+## `ko`
+Ko is limited to building OCI images for go applications. By default images such as [distroless](https://github.com/GoogleContainerTools/distroless) are used as base images. These are minimal and hence very small images, that are stripped of anything but runtime dependencies. Problems arise when the default images do not satisfy our dependency needs (as they currently do with the `disk-mapper` which relies on a dynamically built `libcryptsetup` library). To solve this issue we have three options:
```
malt3, line 22:
```suggestion
Ko is limited to building OCI images for go applications. By default images such as [distroless](https://github.com/GoogleContainerTools/distroless) are used as base images. These are minimal and hence very small images, that are stripped of anything but runtime dependencies. Problems arise when the default images do not satisfy our dependency needs (as they currently do with the `disk-mapper` which relies on a dynamically linked `libcryptsetup` library). To solve this issue we have three options:
```

leongross:
> RFC looks good to me. I suggest running a markdown linter once.
> 
> I think the main challenge will be getting the CSI drivers to build using `ko`. Azure's should probably work, the GCP one might be a challenge ([upstream Dockerfile](https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/v1.7.3/Dockerfile), and [our Dockerfile](https://github.com/edgelesssys/constellation-gcp-compute-persistent-disk-csi-driver/blob/master/Dockerfile))

What exactly do you think will be the issue here? On the first look on the gcp Dockerfile it seems like a lot of libraries and tools are copied manually, but as I see it we can install these dependencies by building our own `apko` as mentioned in the rfc.

I see the issue of building our own `libcryptsetup` to disable the use of udev, but I think we will be able to make this library available, but for this we have to make some further research fist.

malt3 (approved):
LGTM. Thanks for the in-depth research!

daniel-weisse (approved):
Managed to build the CSI drivers with packaged cryptsetup, should be good to go :+1:

datosh (approved):

PR #466, opened 2022-11-07
AWS: Don't expose SSH debugging ports on the load balancer

Nirusu:
### Proposed change(s)
- Remove SSH target on the default AWS load balancer 

This was mainly just useful for early bringup and general verification if the LB setup works using ordinary OS images such as Ubuntu. Since we have Constellation images now with our own debug daemon, it's not necessary anymore (and even confusing as you don't have guarantees which worker node you'll actually hit without further configuration - it's behind a load balancer after all).


datosh (approved):
LGTM

Nirusu:
CLA check hangs so I cannot merge 🥲

PR #470, opened 2022-11-07
Remove access manager

datosh:

### Proposed change(s)

- Access manager was removed from deployment some time ago. This change will also remove all traces of access manager and SSH user+key management from code base.
- A Kubernetes native way on accessing nodes was documented. 

### Checklist

- [ ] Link to Milestone


malt3 (approved):
That is a very thorough cleanup and a nice reduction of our codebase. LGTM

docs/docs/workflows/troubleshooting.md
```diff
@@ -40,4 +40,28 @@ Constellation uses the default bucket to store logs. Its [default retention peri
...
+    ```sh
+    kubectl get nodes
+    # or to see more information, such as IPs:
+    kubectl get nodes -o wide
+    ```
+
+2. Connect to the node:
+
+    ```sh
+    kubectl debug node/constell-worker-xksa0-000000 -it --image=busybox
```
malt3, line 56: I like the use of ephemeral containers here.

3u13r (approved):
LGTM, please sync with @daniel-weisse before merging.

malt3:
Just FYI: removing the ssh field from the config struct is technically a breaking change since you cannot take your old config and reuse it. We may want to either keep it in the config but return a warning/error that is very descriptive OR remove it from the config but tell the user exactly what to do if parsing fails.

datosh:
> Just FYI: removing the ssh field from the config struct is technically a breaking change since you cannot take your old config and reuse it. We may want to either keep it in the config but return a warning/error that is very descriptive OR remove it from the config but tell the user exactly what to do if parsing fails.

Ohh good point! I didn't realize that additional fields will also make the parsing fail. I will get this fixed!

daniel-weisse:
Changes look good to me.
As this changes or removes a lot of code that is also changed or removed by https://github.com/edgelesssys/constellation/pull/387 and https://github.com/edgelesssys/constellation/pull/477 I would like to wait at least until the first of the PRs is merged before bringing this PR to main.

datosh:
Sounds good! I'll wait until #387 is merged.

daniel-weisse:
#387 is merged, you will have to fix merge conflicts in `internal/cloud` then the PR should be good to go

malt3:
> brought config structs back in and marked them as deprecated with a link to new ssh docs. what do you think, @malt3 ?

Looks good! Thanks

daniel-weisse (approved):

PR #473, opened 2022-11-08
AWS: Apply security group to worker nodes

Nirusu:
### Proposed change(s)
- Apply the security group to worker nodes on AWS, too

### Additional info
- For a next release we should further harden down the rules for worker nodes on all CSPs. They are only reachable internally and not exposed to the internet, but could pose an issue for lateral movement.


daniel-weisse:
Did you test if this causes any issues, e.g. with loadbalancing?

Nirusu:
Giving it a quick test right now, give me a few mins.

Nirusu:
Online boutique demo using load balancers works 👍🏻

3u13r (approved):
LGTM

daniel-weisse (approved):

PR #477, opened 2022-11-08
AB#2524 Refactor Azure metadata/cloud API

daniel-weisse:
### Proposed change(s)
Follow up to https://github.com/edgelesssys/constellation/pull/387
* Rename `Metadata` struct to `Cloud`
* Unexport functions that are not required to be exported
* Less code duplication
* Filter resources by UID to avoid including resources not associated with the current Constellation, but existing in the same resource group
* Remove SSH keys from all metadata packages

### Checklist

- [x] More testing


debugd/internal/debugd/metadata/cloudprovider/cloudprovider.go
```diff
@@ -91,18 +91,7 @@ func (f *Fetcher) DiscoverLoadbalancerIP(ctx context.Context) (string, error) {
...
-
-	keys := []ssh.UserKey{}
-	for username, userKeys := range self.SSHKeys {
-		for _, keyValue := range userKeys {
-			keys = append(keys, ssh.UserKey{Username: username, PublicKey: keyValue})
-		}
-	}
-
-	return keys, nil
+	return nil, nil
```
katexochen, line 96: In case this shouldn't be called but is still needed for some interface, please panic instead of returning nil. Otherwise I'd suggest to delete.

internal/cloud/azure/cloud.go
```diff
@@ -0,0 +1,428 @@
...
+					if err != nil {
+						return nil, fmt.Errorf("retrieving VM network interfaces: %w", err)
+					}
+					instance, err := convertToInstanceMetadata(*vm, interfaces)
+					if err != nil {
+						return nil, fmt.Errorf("converting VM to instance metadata: %w", err)
+					}
+					instances = append(instances, instance)
+				}
+			}
```
katexochen, line 234: nit: maybe we could first collect scale sets and list instances afterwards to reduce nesting?

katexochen (dismissed):

debugd/internal/debugd/metadata/cloudprovider/cloudprovider.go
daniel-weisse: Will be removed by https://github.com/edgelesssys/constellation/pull/470

internal/cloud/azure/cloud.go
daniel-weisse: While I agree this would reduce nesting, this would also negatively impact performance since we would iterate the list twice, once to collect and once to process, which is unneeded imo.

malt3 (approved):
LGTM

PR #478, opened 2022-11-08
add brief instructions to AWS IAM Terraform script

thomasten:
Following the link from the docs to the folder, it may not be obvious for everyone what to do next. So I suggest adding these instructions.

### DIFF
--- a/hack/terraform/aws/iam/README.md
+++ b/hack/terraform/aws/iam/README.md
@@ -12,7 +12,16 @@ terraform init
 terraform apply -auto-approve -var name_prefix=my_constellation
 ```
 
-Get the profile names from the Terraform output values `control_plane_instance_profile` and `worker_nodes_instance_profile` and add them to your Constellation configuration file.
+You can either get the profile names from the Terraform output values `control_plane_instance_profile` and `worker_nodes_instance_profile` and manually add them to your Constellation configuration file.
+
+Or you can do this with a `yq` command:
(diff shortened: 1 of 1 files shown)


Nirusu:
If we actually promote this script, maybe we can eventually move it out of "hack/"? 

I mainly put it there because I thought it was only for development, but if that's not the case it can have a suitable directory?

thomasten:
> If we actually promote this script, maybe we can eventually move it out of "hack/"?
> 
> I mainly put it there because I thought it was only for development, but if that's not the case it can have a suitable directory?

I don't care too much, but am open to it. What do you suggest?

3u13r:
Just a note: In the end the files will probably end up in `cli/internal/...' for `constellation iam create`. So we don't need to search for a fully permanent directly now. But I also see that referencing the script with the current path is not fully optimal.

datosh:
> Just a note: In the end the files will probably end up in `cli/internal/...` for `constellation iam create`. So we don't need to search for a fully permanent directory now. But I also see that referencing the script with the current path is not fully optimal.

Docs now refer to release branch. So it should be stable for this version of the docs.

Nirusu:
If we plan on moving the files again soon to the CLI then maybe indeed we can just keep them there for now. 

I think either way people will find them mainly through the link in the docs and not through browsing our repository directly, so it should be fine.

PR #487, opened 2022-11-08
Update release.md

thomasten:


.github/docs/release.md
```diff
@@ -3,7 +3,8 @@
 This checklist will prepare `v1.3.0` from `v1.2.0`. Adjust your version numbers accordingly.
 
 1. Merge ready PRs
-2. Create docs release (new major or minor release)
+2. Search the code for TODOs and FIXMEs that should be resolved before releasing.
```
3u13r, line 6: Since this would make the usage of TODOs and FIXMEs official, we should discuss their meaning. 
IMO TODOs and FIXMEs that block a release in some form (regressions, bugs, other unwanted behavior) should not exist. I personally treat both as a "this is important information if you are interested in this code section". This is also how I would categorize all the current TODOs. 

That being said, we can still add this here to regularly have a look at all TODOs. On the other hand, it encourages using TODOs to document release blockers.

thomasten: I also thought about your points but came to the conclusion that not all TODOs may be so obviously relevant at the time of writing as
https://github.com/edgelesssys/constellation/blob/34f729ccd28a9494e1aa03187e9011d8395f7c78/internal/config/config.go#L271
or
https://github.com/edgelesssys/constellation/blob/34f729ccd28a9494e1aa03187e9011d8395f7c78/internal/config/config.go#L582
So IMO it's worth checking them again.

> IMO TODOs and FIXMEs that block a release in some form (regressions, bugs, other unwanted behavior) should not exist.

I disagree. I want to know by just reading the code that we are aware of incomplete work at a certain code site. It's also the most convenient way to let the reviewer know that I postponed something on purpose. However, I also want important TODOs to be backed by tickets and the reviewer should check if the author has created such. I also like the convention TODO(AB#123) for these. (But wouldn't enforce them for all TODOs.)

malt3: Adding TODOs to relevant sections also make it easier to find the places where changes are needed. I also agree that we have enfore creation of tickets for every release blocker.

datosh: I mostly use TODOs to remind me about open tasks for the same PR and resolve all TODOs before submitting or merging PR, and create a ticket for further work instead.

I agree with @malt3 that they help to find places for changes quickly. Fully on board with using `TODO(AB#123)` for achieving this.

Does this mean that there should not exist a TODO without a referenced ticket?

Without opening a can of worms here: How about using GitHub Issues instead so folks find contribution points for Constellation easier? (Question for Dev Sync?)

daniel-weisse: While the idea of using GitHub issues is nice, most of the TODOs in our code are there not because we didn't have the time to implement them yet, but because we are either missing some 3rd party functionality to implement the feature (e.g. more in-depth AWS TPM attestation), or because we did not yet scope out how exactly we want the feature implemented.

Nirusu: What about the following...?

- We use TODO for longer term stuff we want to fix (which, let's be realistic, likely won't be touched until someone is already refactoring something else)
- We use FIXME for stuff that definitely needs to be fixed before release
- We integrate an action into the Build CLI / Bootstrapper (for release) workflow that fails when FIXMEs are still left in the code

Something like this seems the best of both worlds to me. 

Creating a ticket for each release block in combination with a TODO would also work I guess, but further distinguishing between "this is a workaround for something broken in a library we use" vs "this is a temporary workaround for development that should be resolved until release" in code directly would be nice. Hence my suggestion to distinguish by TODO/FIXME.

Completely omitting TODOs/FIXMEs seems somewhat inevitable to me in our current workflow. Since I wrote the TODOs there originally, back when I wrote this code we did not have any support for Constellation images, so the question is how would we handle these cases when we merge everything into main? If we used feature branches and smaller branches of this branch for development I could maybe see it work, but that would require major adjustments to our development workflow.

katexochen (approved):
After our discussion today, I think we can approve this change now. :dove:

PR #511, opened 2022-11-10
deps: don't group major ver upgrades of actions

katexochen:

### Proposed change(s)

- Major version upgrades shouldn't be grouped so one can decide for each action whether to upgrade or not.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -26,7 +26,7 @@
         "^k8s.io",
         "^sigs.k8s.io"
       ],
-      "groupName": "k8s dependencies"
+      "groupName": "K8s dependencies"
     },
     {
       "matchPackagePatterns": [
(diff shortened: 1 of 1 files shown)


datosh (commented):
It seems this is already a [supported feature in Renovate](https://docs.renovatebot.com/presets-default/#separatemultiplemajorreleases).

Seems like a good idea for all ecosystems.

katexochen:
> Seems like a good idea for all ecosystems.

What exactly do you suggest? I'm not sure what this option is doing exactly, but especially things like the SDKs should still be grouped on major releases.

datosh (approved):
Good point. SDKs should probably receive major upgrades in one go. Then your custom solution works better!

PR #513, opened 2022-11-10
Update GitHub workflow runners to Ubuntu 22.04

Nirusu:
Let's see if anything breaks.

### DIFF
--- a/.github/workflows/check-licenses.yml
+++ b/.github/workflows/check-licenses.yml
@@ -35,4 +35,4 @@ jobs:
         run: hack/check-licenses.sh
 
       - name: Check for files without license header
-        run: "! grep -rL --include='*.go' -e'SPDX-License-Identifier: AGPL-3.0-only' -e'DO NOT EDIT'"
+        run: "! grep -rL --include='*.go' -e'SPDX-License-Identifier: AGPL-3.0-only' -e'DO NOT EDIT' | grep ''"
--- a/.github/workflows/check-licenses.yml
+++ b/.github/workflows/check-licenses.yml
@@ -15,7 +15,7 @@ on:
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout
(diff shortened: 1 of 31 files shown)


Nirusu:
grep needed some fixes. Now returns 0 even when nothing is found:
```bash
  The --files-without-match (-L) option has reverted to its behavior
  in grep 3.1 and earlier.  That is, grep -L again succeeds when a
  line is selected, not when a file is listed.  The behavior in grep
  3.2 through 3.4 was causing compatibility problems.
```
https://lists.gnu.org/archive/html/info-gnu/2020-09/msg00009.html

Love fixing compatibility problems by introducing compatibility problems :)

I rewrote the script to actually check the length of the output instead. I think that should work better.

But except from that everything seems to be working okay.

.github/workflows/check-licenses.yml
```diff
@@ -35,4 +35,12 @@ jobs:
...
-        run: "! grep -rL --include='*.go' -e'SPDX-License-Identifier: AGPL-3.0-only' -e'DO NOT EDIT'"
+        run: |
+          FILES_WITHOUT_LICENSE_HEADER=$(grep -rL --include='*.go' -e'SPDX-License-Identifier: AGPL-3.0-only' -e'DO NOT EDIT')
+          if [[ $(echo "$FILES_WITHOUT_LICENSE_HEADER" | sed '/^\s*$/d' | wc -l) -ne "0" ]]; then
+            echo "$FILES_WITHOUT_LICENSE_HEADER"
+            exit 1
+          else
+            exit 0
+          fi
+        shell: bash
```
thomasten, line 46:
```suggestion
        run: "! grep -rL --include='*.go' -e'SPDX-License-Identifier: AGPL-3.0-only' -e'DO NOT EDIT' | grep ''"
```

Nirusu: That's definitely easier!

katexochen (approved):
> Let's see if anything breaks.

Looks good. :grinning:

PR #525, opened 2022-11-11
Increase timeout for fetch-measurements

Nirusu:
### Proposed change(s)
- Bump timeout for fetch-measurements from 3 seconds to 60 seconds.

This should be plenty for slow connections / long-distance traffic.

### Related issues
- See Discord


3u13r (approved):
LGTM

PR #526, opened 2022-11-11
Automatically fetch measurements for Kubernetes artifacts

3u13r:
This is the first (tiny) step towards automated Constellation updates. The hashes will be used in later PRs. For further reference see: https://github.com/edgelesssys/constellation/blob/main/rfc/updates.md .

### Proposed change(s)

- Add the hashes of the kubernetes artifacts to `versions.go`


katexochen (approved):

internal/versions/generateHashes.go
```diff
@@ -0,0 +1,105 @@
...
+		panic("bad status: " + resp.Status)
+	}
+
+	// Writer the body to buffer
+	if _, err = buf.ReadFrom(resp.Body); err != nil {
+		panic(err)
+	}
+
+	// Generate SHA256 hash of the file
+	h := sha256.New()
```
datosh, line 50: Have we considered using the hashes provided by the publishers of the respective project? Or, why did we decide to calculate the hashes ourselves?

For `crictl` e.g. there is:
* release: https://github.com/kubernetes-sigs/cri-tools/releases/download/v1.25.0/crictl-v1.25.0-linux-amd64.tar.gz
* hash256: https://github.com/kubernetes-sigs/cri-tools/releases/download/v1.25.0/crictl-v1.25.0-linux-amd64.tar.gz.sha256
* hash512: https://github.com/kubernetes-sigs/cri-tools/releases/download/v1.25.0/crictl-v1.25.0-linux-amd64.tar.gz.sha512

datosh: Alternatively, we could also do both and compare the result.

datosh (commented):
Love the approach! Using generators & automation to take care of error-prone manual tasks ❤️

internal/versions/generateHashes.go
3u13r: Opted for your 2nd suggestion. Taking both the upstream hash and hashing the file ourselves gives us a compile/generate time check if the hashes match and therefore if our Constellation setup will succeed.

PR #535, opened 2022-11-11
docs: publish access manager removal

thomasten:
Suggest publishing this to v2.2 as access-manager is already deprecated and not deployed. What do you think?


datosh (approved):
Great idea! This helps users to find the correct way to SSH into nodes.

Changes LGTM!

PR #544, opened 2022-11-14
AB#2512 Config secrets via env var & config refactoring

datosh:
### Proposed change(s)

- refactored measurements code
  - `map[uint32][]byte` to use consistent type, makes subsequent changes easier
  - provided functions such as `Zero`, `One` and `AllBytes` to make tests more readable and faster to write
  - moved equality test from upgrade cmd to measurement type
- refactored multierr hashicorp library to use uber multierr library
- refactored config code
  - Added `New` function `WithDefaultOptions` to make usage of config in CLI much more consistent. For example, many subcommands lacked config validation.

### Checklist

- [x] Link to Milestone


datosh:
> Sorry for drive-by review but since this touched quite a bit of code I wanted to work on, it was kinda necessary to look into here.

Not at all - thanks for taking the time to review! 🙂

internal/config/config.go
```diff
@@ -618,110 +518,81 @@ func (c *Config) GetProvider() cloudprovider.Provider {
...
+	var validationErrors error
+	for _, e := range errs {
+		validationErrors = multierr.Append(
+			validationErrors,
+			errors.New(e.Translate(trans)),
+		)
 	}
-	return false
+	return validationErrors
 }
```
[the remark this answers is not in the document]
datosh: My idea was to move the validate helper functions to a dedicated file, but leave the public API of config in its original one. I agree that it is not the cleanest cut.

One could also argue to move the private methods of config back to config.go.

What do the other reviewers think?

internal/attestation/measurements/measurements.go
```diff
@@ -18,45 +18,76 @@ import (
...
+
+// One returns a PCR value with all bits set to one.
+func One() []byte {
+	return AllBytes(0xFF)
+}
+
+// AllBytes returns a PCR value with all bytes set to b.
+func AllBytes(b byte) []byte {
+	return bytes.Repeat([]byte{b}, 32)
+}
```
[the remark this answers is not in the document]
datosh: I introduced them to reduce repeating bytes and make code more readable in general. I think for the code base as it is right now it is a benefit to have these. 

I can't see how they are related to the measurements 11, 12, 13, ... you are talking about. Maybe we can have a chat once the other reviews are in..? 

> So I don't know for sure. We can also just adjust it later.

Yes, if future changes make these unnecessary we should remove them! 👍

internal/config/config.go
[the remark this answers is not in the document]
thomasten: I like having the validate helper functions in another file.
Regarding Validate(), both locations are fine with me and there are reasonable arguments for both.

internal/config/config.go
```diff
@@ -323,198 +329,79 @@ func Default() *Config {
...
+		if err := opt(c); err != nil {
+			return nil, err
+		}
 	}
-
-	// Show single string if only one other provider is defined, show list with brackets if multiple are defined.
-	t, _ := ut.T("more_than_one_provider", fe.Field(), strings.Join(definedProviders, ", "))
-
-	return t
+	return c, nil
```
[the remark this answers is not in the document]
datosh: I like `config.New` as a cleaner implementation for now, and `config.NewWithOptions` as an option going forward. I'll wait for other feedback and will change the implementation if others agree as well!

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- `CONSTELL_AZURE_CLIENT_SECRET_VALUE` as an alternative to provide `provider.azure.clientSecretValue` via environment variable instead.
```
thomasten, line 24: The sentence reads a bit odd. Maybe
```suggestion
Environment variable `CONSTELL_AZURE_CLIENT_SECRET_VALUE` as an alternative way to provide the configuration value `provider.azure.clientSecretValue`.
```

bootstrapper/internal/kubernetes/k8sapi/install_test.go
```diff
@@ -713,10 +713,10 @@ func (w *tarGzWriter) Bytes() []byte {
 
 func (w *tarGzWriter) Close() (result error) {
 	if err := w.tarWriter.Close(); err != nil {
-		result = multierror.Append(result, err)
+		result = multierr.Append(result, err)
```
thomasten, line 716: good catch!

cli/internal/cmd/configfetchmeasurements.go
```diff
@@ -57,22 +58,22 @@ func configFetchMeasurements(cmd *cobra.Command, verifier rekorVerifier, fileHan
 		return err
 	}
 
-	conf, err := config.FromFile(fileHandler, flags.config)
+	config, err := config.New(config.WithDefaultOptions(fileHandler, flags.configPath)...)
```
thomasten, line 61: nit: isn't the old name that doesn't shadow the package name better?

docs/docs/getting-started/first-steps.md
```diff
@@ -71,6 +71,12 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
 
     Run `constellation config instance-types` to get the list of all supported options.
 
+    :::tip
+
+    Alternative, you can leave `clientSecretValue` empty, and provide the secret via `CONSTELL_AZURE_CLIENT_SECRET_VALUE` environment variable.
```
thomasten, line 76:
```suggestion
    Alternatively, you can leave `clientSecretValue` empty and provide the secret via the `CONSTELL_AZURE_CLIENT_SECRET_VALUE` environment variable.
```

docs/docs/getting-started/first-steps.md
```diff
@@ -71,6 +71,12 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
 
     Run `constellation config instance-types` to get the list of all supported options.
 
+    :::tip
```
thomasten, line 74: should this be moved between "Fill the values ..." and "By default, ..."?

internal/attestation/measurements/measurements_test.go
```diff
@@ -321,3 +321,95 @@ func TestMeasurementsFetchAndVerify(t *testing.T) {
...
+	assert.Equal(t, zero, pcr)
+}
+
+func TestOne(t *testing.T) {
+	one := []byte{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF}
+	pcr := One()
+	assert.Equal(t, one, pcr)
+}
+
+func TestMeasurements_EqualTo(t *testing.T) {
```
thomasten, line 365: nit: not sure what our naming convention is for this kind of tests. @katexochen

internal/config/config.go
```diff
@@ -618,110 +518,81 @@ func (c *Config) GetProvider() cloudprovider.Provider {
...
-// IsDebugCluster checks whether the cluster is configured as a debug cluster.
-func (c *Config) IsDebugCluster() bool {
-	if c.DebugCluster != nil && *c.DebugCluster {
-		return true
+	var validationErrors error
+	for _, e := range errs {
+		validationErrors = multierr.Append(
+			validationErrors,
+			errors.New(e.Translate(trans)),
+		)
```
thomasten, line 595: nit: For "seperation of concerns"-like reasons, I prefer the old variant of this function and would move the conversion to error into WithValidate(). But if you decide to keep it like this, please change the doc comment of this func.

thomasten (changes requested):
Just a few smaller things. Nils already brought up good points regarding some of the high-level decisions, which I also thought about while reviewing. However, I didn't come to to a clear preference and I think that this PR is definitely an improvement to the status quo, so I'd be fine with that.

internal/attestation/measurements/measurements.go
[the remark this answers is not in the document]
malt3: Yes, 11, 12 and 13 will eventually have a non-zero value

internal/attestation/measurements/measurements_test.go
katexochen: In most cases we would just have called this `TestEqualTo` I guess, but for me its okay this way, too, as github.com/cweill/gotests (included in the VSCode Go ext) uses this naming convention. Maybe something we could start to adopt...

internal/attestation/measurements/measurements.go
```diff
@@ -18,45 +18,76 @@ import (
 	"net/url"
 
 	"github.com/edgelesssys/constellation/v2/internal/attestation/vtpm"
+	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
 	"github.com/edgelesssys/constellation/v2/internal/sigstore"
 	"gopkg.in/yaml.v2"
 )
 
 // Measurements are Platform Configuration Register (PCR) values.
 type Measurements map[uint32][]byte
```
katexochen, line 27: nit: A common thing in Go would be to call this just `M`, like in `testing.T`, as it its the only exported type in the package and the package name is already descriptive enough...

internal/attestation/measurements/measurements.go
```diff
@@ -18,45 +18,76 @@ import (
...
+// Zero returns a PCR value with all bits set to zero.
+func Zero() []byte {
+	return AllBytes(0x00)
+}
+
+// One returns a PCR value with all bits set to one.
+func One() []byte {
+	return AllBytes(0xFF)
+}
+
```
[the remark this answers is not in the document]
katexochen: I don't like such exported utilities, especially, the name of the function isn't descriptive enough, I would `measurements.Zero` expect to return a zero value of type Measurements, not a slice of bytes. If you don't want to copy code (which isn't a bad think) or don't want to count up to 32 chars (I totally feel that), I'd suggest to provide a single utility like `AllBytes` but with a more descriptive name and comment.

internal/attestation/measurements/measurements.go
```diff
@@ -18,45 +18,76 @@ import (
...
+	case cloudprovider.GCP:
+		return gcpPCRs
+	case cloudprovider.QEMU:
+		return qemuPCRs
+	default:
+		return nil
+	}
+}
+
 var (
```
katexochen, line 60: You could use functions instead of global vars to make this code safer. Global state is never a good idea...

internal/attestation/measurements/measurements_test.go
```diff
@@ -321,3 +321,95 @@ func TestMeasurementsFetchAndVerify(t *testing.T) {
...
 }
+
+func TestAllBytes(t *testing.T) {
+	testCases := map[string]struct {
+		b       byte
+		wantPCR []byte
+	}{
+		"0x00": {
+			b:       0x00,
+			wantPCR: []byte{0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00},
```
katexochen, line 332: Why not use` bytes.Repeat` here?

internal/config/config.go
```diff
@@ -27,6 +28,10 @@ import (
 	en_translations "github.com/go-playground/validator/v10/translations/en"
 )
 
+// Measurements is a required alias since docgen is not able to work with
+// types in other packaes.
```
katexochen, line 33:
```suggestion
// Measurements is a required alias since docgen is not able to work with
// types in other packages.
```

internal/attestation/measurements/measurements.go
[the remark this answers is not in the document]
datosh: Renamed to `PCRWithAllBytes` and removed `Zero` and `One` functions.

internal/config/config.go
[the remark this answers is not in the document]
datosh: Sorry, my answer was probably too short and therefore ambiguous: I totally agree with your suggestion! 

There is no need for the implemented flexibility when no one is using it. Implemented your suggestion 🙂

docs/docs/getting-started/first-steps.md
datosh: I don't have strong feelings for either - I have moved as suggested!

internal/attestation/measurements/measurements.go
datosh: Thanks for the encouragement! I have thought about it but didn't find any instances of it in our code base, so I decided against it. I too like the short form better, so that clients do not stutter.

internal/attestation/measurements/measurements.go
datosh: Good hint! I moved them into `DefaultsFor`. It was the only reference anyway.

internal/attestation/measurements/measurements_test.go
datosh: Since this code tests `AllBytes` (now `PCRWithAllBytes`) which does the repetition I wanted to show the full expected value once.

datosh:
Thanks for all the great feedback. Please take another look!

internal/attestation/measurements/measurements_test.go
katexochen: Imo you can assume these (especially external/stdlib) dependencies to be correct, as they don't belong to your _system under test_.

internal/config/config.go
```diff
@@ -323,198 +329,40 @@ func Default() *Config {
...
-		providerCount++
-	}
-
-	if providerCount < 1 {
-		sl.ReportError(provider, "Provider", "Provider", "no_provider", "")
-	} else if providerCount > 1 {
-		sl.ReportError(provider, "Provider", "Provider", "more_than_one_provider", "")
+// FromFile returns config file with `name` read from `fileHandler` by parsing
+// it as YAML.
+func FromFile(fileHandler file.Handler, name string) (*Config, error) {
```
thomasten, line 334: optional: We can try to prevent developers from accidentially using this func instead of New. I've got two suggestions:
* This func is currently only used by cdbg. If it's acceptable that cdbg validates the config, we could use New there, too, and unexport this func.
* Or we could extend the doc comment to encourage developers to use New in most cases.

internal/config/config.go
```diff
@@ -323,198 +329,40 @@ func Default() *Config {
...
+	c := &Config{}
 
-	// Register Provider validation error types
-	if err := validate.RegisterTranslation("no_provider", trans, registerNoProviderError, translateNoProviderError); err != nil {
+	// Read config file
+	readConfig, err := FromFile(fileHandler, name)
+	if err != nil {
 		return nil, err
 	}
+	*c = *readConfig
```
thomasten, line 357:
```suggestion
	c, err := FromFile(fileHandler, name)
	if err != nil {
		return nil, err
	}
```

internal/config/config.go
```diff
@@ -323,198 +329,40 @@ func Default() *Config {
...
-
-	if err := validate.RegisterTranslation("gcp_instance_type", trans, registerTranslateGCPInstanceTypeError, translateGCPInstanceTypeError); err != nil {
-		return nil, err
-	}
+// New creates a new config by:
+// 1. Reading config file via provided fileHandler from file with name.
+// 2. Read secrets from environment variables.
+// 3. Validate config.
+func New(fileHandler file.Handler, name string) (*Config, error) {
+	c := &Config{}
```
thomasten, line 351:
```suggestion
```

internal/config/config.go
```diff
@@ -73,7 +79,7 @@ type UpgradeConfig struct {
 	Image string `yaml:"image"`
 	// description: |
 	//   Measurements of the updated image.
-	Measurements Measurements `yaml:"measurements"`
+	Measurements measurements.M `yaml:"measurements"`
```
thomasten, line 82: Does this need to be
```suggestion
	Measurements Measurements `yaml:"measurements"`
```
because of docgen?

internal/config/config_doc.go
```diff
@@ -242,8 +242,8 @@ func init() {
 	AzureConfigDoc.Fields[6].Name = "clientSecretValue"
 	AzureConfigDoc.Fields[6].Type = "string"
 	AzureConfigDoc.Fields[6].Note = ""
-	AzureConfigDoc.Fields[6].Description = "Client secret value of the Active Directory app registration credentials."
-	AzureConfigDoc.Fields[6].Comments[encoder.LineComment] = "Client secret value of the Active Directory app registration credentials."
+	AzureConfigDoc.Fields[6].Description = "Client secret value of the Active Directory app registration credentials. Alternative use CONSTELL_AZURE_CLIENT_SECRET_VALUE environment variable."
+	AzureConfigDoc.Fields[6].Comments[encoder.LineComment] = "Client secret value of the Active Directory app registration credentials. Alternative use CONSTELL_AZURE_CLIENT_SECRET_VALUE environment variable."
```
thomasten, line 246: don't forget to regenerate before merge

thomasten (approved):

internal/config/config_doc.go
datosh: Good point! Looking forward to @3u13r 's check there 😉

internal/config/config.go
datosh: not sure why that one instance was changed.

PR #548, opened 2022-11-14
AB#2275 Add azuredisk CSI driver

daniel-weisse:
### Proposed change(s)

- Deploy Azure CSI driver during `constellation init`

### Checklist

- [ ] ~Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)~
  - See [Azure Boards Ticket](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/2615)
- [x] Link to Milestone


cli/internal/helm/loader_test.go
```diff
@@ -64,6 +64,7 @@ func TestTemplate(t *testing.T) {
 		},
 		"Azure": {
 			config: &config.Config{Provider: config.ProviderConfig{Azure: &config.AzureConfig{
+				DeployCSIDriver:      func() *bool { b := true; return &b }(),
```
derpsteb, line 67: The CLI config would need a default value here as well, wouldn't it?

derpsteb (approved):
LGTM :).
Tested the deployment on azure, volumes were created just fine.

Since using go-generate to pull in the chart from the [azdisk-csi-driver repo](https://github.com/edgelesssys/constellation-azuredisk-csi-driver/blob/main/charts/latest/azuredisk-csi-driver/Chart.yaml) is too much overhead with the current repo structure I think it would be nice to have 2-3 sentences on where this chart comes from and how it is adapted, roughly. 

For a chart that is written by us I would like to discuss file naming and the use values a bit, but since the chart is mostly upstream code I think we should keep our changes minimal.

malt3 (approved):
LGTM

PR #549, opened 2022-11-14
Do not quote azure image upload params

malt3:
### Proposed change(s)

- Do not quote azure image upload parameters


daniel-weisse (approved):
Upload works, but there might be other issues with the default replication regions described in the image readme

malt3:
Fixed the replication region issue

PR #550, opened 2022-11-14
Feat/go gen ci check

3u13r:
### Proposed change(s)

Add a CI test which checks if running `go generate ./...` leads to changes. 

Thanks to @katexochen for the help.

### DIFF
--- a/.github/actions/install_docgen/action.yml
+++ b/.github/actions/install_docgen/action.yml
@@ -1,0 +1,26 @@
+name: Install Docgen
+description: |
+  Install Docgen
+runs:
+  using: "composite"
+  steps:
+    - name: Checkout talos
+      uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # tag=v3.1.0
+      with:
+        fetch-depth: 0
+        repository: "siderolabs/talos"
+        ref: "v1.2.6"
+        path: talos
+
+    # This is required to be able to build docgen
+    - name: Remove go.work
+      shell: bash
+      working-directory: talos
+      run: rm go.work*
+
+    - name: Install Docgen
+      shell: bash
+      working-directory: talos/hack/docgen
+      run: |
+        go build -o docgen .
+        mv docgen /usr/local/bin
--- a/.github/workflows/test-go-generate.yml
+++ b/.github/workflows/test-go-generate.yml
@@ -1,0 +1,52 @@
+name: Go generate check
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - "release/**"
+  pull_request:
+
+jobs:
+  go-generate:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # tag=v3.1.0
+        with:
+          path: constellation
+          ref: ${{ github.head_ref }}
+
+      - name: Setup Go
+        uses: actions/setup-go@c4a742cab115ed795e34d4513e2cf7d472deb55f # tag=v3.3.1
+        with:
+          go-version: "1.19.3"
+          cache: true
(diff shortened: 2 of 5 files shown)


.github/actions/install_docgen/action.yml
```diff
@@ -0,0 +1,25 @@
...
+  steps:
+    - name: Checkout talos
+      uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # tag=v3.1.0
+      with:
+        fetch-depth: 0
+        repository: "siderolabs/talos"
+        ref: "v1.2.6"
+        path: talos
+
+    - name: Remove go.work
```
katexochen, line 15: Maybe a good place to leave a comment that this is required for building docgen.

.github/actions/install_docgen/action.yml
```diff
@@ -0,0 +1,25 @@
...
+      shell: bash
+      working-directory: talos
+      run: rm go.work
```
katexochen, line 18:
```suggestion
      run: rm go.work*
```
Surely not needed, but who knows what future Go tooling will do if there is a lonely go.work.sum...

.github/workflows/test-go-generate.yml
```diff
@@ -0,0 +1,40 @@
+name: Go generate check
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - "release/**"
+  pull_request:
```
katexochen, line 9:
```suggestion
  workflow_dispatch:
  push:
    branches:
      - main
      - "release/**"
    paths:
      - "**.go"
      - "**/go.mod"
      - "**/go.sum"
  pull_request:
    paths:
      - "**.go"
      - "**/go.mod"
      - "**/go.sum"
```

This is only needed on changes in Go files, isn't it?

.github/workflows/test-go-generate.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Install stringer
+        shell: bash
+        run: go install golang.org/x/tools/cmd/stringer@latest
+
+      - name: Go generate check
+        shell: bash
+        working-directory: constellation
+        run: |
+          go generate ./...
+          git diff --exit-code
```
katexochen, line 40: This should be done with multi-module support.
```suggestion
      - name: Get Go submodules
        id: submods
        shell: bash
        run: |
          mods=$(go list -f '{{.Dir}}' -m | xargs)
          echo "Found mods: $mods"
          echo "submods=${mods}" >> "$GITHUB_OUTPUT"
       
      - name: Go generate check
        shell: bash
        working-directory: constellation
        run: |
          mods=${{ steps.submods.outputs.submods }}
          result=0
          for mod in $mods; do
              (cd $mod; go generate ./...; git diff --exit-code || result=1)
          done
          exit $result
```

katexochen: Note for myself to check how this behaves if a tool referenced in go generate is not installed...

katexochen (changes requested):

.github/workflows/test-go-generate.yml
katexochen: Ah, surely this is not the case for the generation of the cilium helm chart...

cli/internal/helm/charts/cilium/README.md
```diff
@@ -37,7 +37,7 @@ $ helm install cilium cilium/cilium --namespace=kube-system
 ```
 
 After Cilium is installed, you can explore the features that Cilium has to
-offer from the [Getting Started Guides page](https://docs.cilium.io/en/stable/gettingstarted/).
+offer from the [Getting Started Guides page](https://docs.cilium.io/en/latest/gettingstarted/).
```
daniel-weisse, line 40: This link in the upstream cilium is broken, maybe we should just add it to the ignore list?

3u13r: Already done in .github/workflows/check-links.yml:29

PR #551, opened 2022-11-14
Hardcode azure subscription "secret" to avoid masking

malt3:
### Proposed change(s)

- Hardcode azure subscription "secret" to avoid masking

### Additional info
- I want to use the subscription id as part of a gtihub actions job output
- Job outputs are searched for secrets
- If job outputs contain secrets as a substring, the whole output is replaced with null
  - `'imageLookupTableAzureTrustedLaunch' => {"azure":{"trustedlaunch":"/subscriptions/***/resourceGroups/constellation-images/providers/Microsoft.Compute/galleries/Constellation_Testing/images/feat-os-images-version-uids/versions/2022.1114.151640"}}`
  - `Warning: Skip output 'imageLookupTableAzureTrustedLaunch' since it may contain secret.`

### DIFF
--- a/.github/workflows/e2e-test-daily.yml
+++ b/.github/workflows/e2e-test-daily.yml
@@ -8,7 +8,7 @@ on:
 env:
   ARM_CLIENT_ID: ${{ secrets.AZURE_E2E_CLIENT_ID }}
   ARM_CLIENT_SECRET: ${{ secrets.AZURE_E2E_CLIENT_SECRET }}
-  ARM_SUBSCRIPTION_ID: ${{ secrets.AZURE_E2E_SUBSCRIPTION_ID }}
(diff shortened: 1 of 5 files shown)


Nirusu:
I am not 100% sure if this fix solely is enough as e.g. there's also the userAssignedIdentity which contains the subscription ID. Not sure how sensitive Github's secret redaction is here.

But you can try, I don't mind.

malt3:
I tested this with json secrets. The masking is applied for every line in the secret.

Nirusu:
I don't know for sure, feel free to try.

Nirusu (approved):

malt3:
I refactored my code to use artifacts instead of outputs.

PR #553, opened 2022-11-15
Refactor enforced/expected PCRs

daniel-weisse:
### Proposed change(s)
Refactor how we use expected and enforced measurement.

We currently keep two separate list:
* One `map[uint32][]byte` to hold all expected measurement values
* One `[]uint32` to keep track of indices that are enforced to be equal to the values given in the expected measurement list
  * Internally this slice is converted to `map[uint32]struct{}` for easier comparison

This approach comes with a couple drawbacks:
1. We don't throw any errors if an attestation Validator is configured incorrectly, e.g. the enforced measurement list contains indices that are not present in the expected measurement list. 
2. Our code always carries around 2 variables until they are passed to a validator
3. The config file looks a bit confusing. From one look it is not clear what the enforced list is, and how it corresponds to the expected values.
4. We have to do manual validation that the expected measurements have the expected length of 32 bytes
5. We have to do manual validation that all indices given in the enforced list actually exist in the expected list

This PR proposes to merge both lists into a single `map[uint32]Measurement`, where `Measurement` is the following:
```Go
type Measurement struct {
  Expected [32]byte
  WarnOnly bool
}
```

This fixes multiple problems:
1. Config file looks a lot cleaner
  * Users should immediately be able to tell which values are enforced, and which are not
2. We can skip manual validation, since our unmarshal operation will fail for invalid length measurements, and automatically fill in whether the measurement shall be enforced or not
3. Attestation Validator code is a bit cleaner and can not be configured incorrectly.

Example measurements entry for a config file:
```yaml
provider:
  gcp:
    measurements:
      0:
        expected: DzXCFGCNk8em5ornNZtKi+Wg6Z7qkQfs5CfE3qTkOc8=
        warnOnly: false
      6:
        expected: AQIDBAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
        warnOnly: true
      11:
        expected: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
        warnOnly: false
      12:
        expected: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
        warnOnly: false
      15:
        expected: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
        warnOnly: false
```

#### Measurement signing

With this PR the signed measurements change slightly:
* Measurements include the enforced/warnOnly setting
  * This allows us to more easily update configs using the `fetch-measurements` command
* Measurements are written and signed in 0 indent JSON format
* Expected measurements are encoded as hex instead of base64
* On GCP all PCRs up to PCR[15] are included in the expected measurements
* On Azure all PCRs up to PCR[15], except PCRs 0, 6, and 11 are included in the expected measurements

### Additional info
- The current code takes into account unmarshalling a simple base64 measurements map of `map[uint32]string`, however this sets measurements to not enforced.
  - Allows for backwards compatibility with older configs / clusters

### Checklist

- [ ] Link to Milestone


malt3:
> [...]  or update the way we save expected measurements (e.g. include enforced/not-enforced in the artifact)

This is the best solution in my eyes. The enforced measurements could theoretically change between image versions so we want to ship a list of enforced measurements with actual expected measurements for an image.
We can still have a minimum set of important measurements hardcoded in the CLI that we at least warn about if missing or not set to be enforced.

internal/attestation/measurements/measurements.go
```diff
@@ -130,41 +153,73 @@ func (m Measurements) EqualTo(other Measurements) bool {
...
-		measurement, err := base64.StdEncoding.DecodeString(value)
+		measurement, err := base64.StdEncoding.DecodeString(value.Expected)
 		if err != nil {
 			return err
 		}
-		(*m)[key] = measurement
+
+		if len(measurement) != 32 {
+			return fmt.Errorf("invalid measurement at key %d: invalid length: %d", key, len(measurement))
+		}
```
Nirusu, line 217: Maybe we could wrap this error and catch it to make it user-friendlier?
But we cannot do it in the go validator unfortunately due to the fixed size byte array.

Nirusu (commented):
Generally LGTM for the draft and I like that we now enforce PCRs directly at the PCR validation and not before.

internal/attestation/measurements/measurements.go
```diff
@@ -12,33 +12,53 @@ import (
...
+	PCRIndexOwnerID = tpmutil.Handle(16)
+)
+
 // Measurements are Platform Configuration Register (PCR) values.
-type Measurements map[uint32][]byte
+type Measurements map[uint32]Measurement
+
+// Measurement wraps expected PCR value and whether it is enforced.
+type Measurement struct {
+	Expected [32]byte `yaml:"expected"`
```
thomasten, line 42: Since we're breaking config compatibility anyway, I'd like to propose marshaling this to a hex string, which seems the more natural format for a measurement. This is what we do in our other products for user-facing measurements.
Just want to hear your opinions. It's not super important to me.

internal/attestation/measurements/measurements.go
```diff
@@ -12,33 +12,53 @@ import (
...
+	Enforced bool     `yaml:"enforced"`
```
thomasten, line 43: We would need to ensure that this is a required field. Alternatively, we could "invert" the field to something like `WarnOnly` so the default `false` is secure.

thomasten, line 43: > We can still have a minimum set of important measurements hardcoded in the CLI that we at least warn about if missing or not set to enforced.

I gues this is the solution for *preventing users from accidentially running without important measurements*?
By hardcoded, do you mean prefilled in the generated config with Enforced=true? And if so, should the `Expected` field be omitted/empty/invalid-placeholder-value or the like?

daniel-weisse: >do you mean prefilled in the generated config with Enforced=true

Yes, the expected value would be left empty, or contain a sentinel value that we can catch during config validation.
This way we can print a useful error message to the user

internal/attestation/measurements/measurements.go
daniel-weisse: I like that idea.
Would also make things a bit easier to read, since tools like `tpm2_pcrread` also print the output as hex stings

datosh, line 42: Would it make sense to have a type definition for `[32]byte`? Something like

```go
type PCR [32]byte
```

datosh (commented):
I like the new approach! 
Since we break config compatibility, should we think about moving `measurements` out of the `provider` section and into the root level? It seems that all providers would require a set of measurements. 
Do we expect deployments without them?

thomasten:
> should we think about moving `measurements` out of the `provider` section and into the root level? It seems that all providers would require a set of measurements.
> Do we expect deployments without them?

There might be providers that don't use a TPM for attestation. Maybe in the case of TDX. But our measurements map approach might still fit. Not sure.

daniel-weisse:
>There might be providers that don't use a TPM for attestation. 

I would opt to keep it in the provider specific configuration for now, at least until we know a bit more about how TDX/SEV-SNP attestation will look in the future on different CSPs.
Moving the measurements key to top level would require some code refactoring and adjustments when it comes to prefilling the expected/enforced measurements when generating a config. This should not be a lot, but might be annoying if we have to switch back in the future. (Also annoying for users who have to adjust to the changes in placements multiple times)

.github/actions/constellation_measure/action.yml
```diff
@@ -51,13 +51,33 @@ runs:
...
+            yq '.0.warnOnly = false |
+              .4.warnOnly = false |
+              .8.warnOnly = false |
+              .9.warnOnly = false |
+              .11.warnOnly = false |
+              .12.warnOnly = false |
+              .13.warnOnly = false |
+              .15.warnOnly = false |
+              .15.expected = "0000000000000000000000000000000000000000000000000000000000000000"' \
+              -I 0 -o json -i measurements.json
```
Nirusu, line 80: I think if we already merge the CI change to JSON as part of this PR, we would also need to adjust fetch-measurements (which currently expects .yaml URLs).

daniel-weisse: I would opt to not update the `fetch-measurements` command default URLs until the new measurement pipeline is implemented.
`fetch-measurements` only fetches the signed measuremnts for release images, of which we have non in json format until the next release

Nirusu: I was also thinking about this, but then:

1. Having JSON and the new structure in there for parsing in general would already be part of the new measurement pipeline and help development; it wouldn't be wasted work (unless I am missing something here).
2. The new measurement pipeline would only add another JSON object holding metadata and change the URLs depending on the image UID, so we likely would mainly build stuff on top of what is currently there (and tweak the download URLs).
3. In a strict sense, this PR would break the concept of having a "releasable main" since you adjust the CI (which will likely look completely different with the new pipeline) but don't make adjustments to the downloading code.

internal/attestation/measurements/measurements.go
```diff
@@ -53,8 +53,14 @@ func (m *M) FetchAndVerify(ctx context.Context, client *http.Client, measurement
...
-	if err := yaml.Unmarshal(measurements, m); err != nil {
-		return "", err
+
+	if err := json.Unmarshal(measurements, m); err != nil {
+		if yamlErr := yaml.Unmarshal(measurements, m); yamlErr != nil {
+			return "", multierr.Append(
+				err,
+				fmt.Errorf("trying yaml format: %w", yamlErr),
+			)
+		}
```
Nirusu, line 63: Can we make the legacy fall back dependent on configVersion (since we want to bump it for the next release)?

configVersion == v1 -> use YAML
configVersion == v2 -> use JSON

.github/actions/constellation_measure/action.yml
daniel-weisse: I think there was some miscommunication here.
When I said I would like to not make changes to `fetch-measurements`, I was explicitly talking about the code of that command, more specifically the URL parsing, since that still uses the old public bucket.

I have since adapted the code to be agnostic to yaml/json formating when parsing measurements during `*M.FetchAndVerify()`

internal/attestation/measurements/measurements.go
daniel-weisse: No, since this code does not know about the config

thomasten: configVersion and measurements format also don't have much to do with each other, do they?

malt3: Additional Q: Isn't the current code functionally identical to just using yaml.Unmarshal directly? json is a subset of yaml so every valid json document can be parsed by yaml.Unmarshal.

Nirusu: In the later run they would _kind of_ have to do something with each other.

For newer versions (≥ v2.3.0) it would be cool if fetch-measurements would only allow the new image & measurement format, so that you cannot transparently downgrade both the image + expected measurements to try a downgrade attack. 

On the other hand, if we have backwards compatibility with v2.2.0, we would still need to accept the old format.

Having a generic fallback would skip the "better" security guarantees with the worse ones.

However, if we support configVersion v1 with v2.3.0 images, then configVersion and measurements wouldn't strictly be related with each other I guess.

At least that is my intention here. We could also add it at a later point, though.

Nirusu: In the future and likely before the next release, fetch-measurements is going to need to the image UID to download the measurements anyway, which only the new configVersion will have. 

That was the intention for having a relationship between configVersion <-> measurements; eventually they are going to be related for fetch-measurements, so the idea here was that it can also work as a function toggle for YAML <-> JSON, too. Especially since fetch-measurements should later only allow measurements with a metadata header for images ≥2.3.0.

Though given that as @malt3 pointed out, yaml.Unmarshal likey can decode JSON just fine anyway so that would need to be implemented different anyway I guess.

So I assume we can just do that in an upcoming PR then.

daniel-weisse: >Isn't the current code functionally equivalent to just using yaml.Unmarshal directly? json is a subset of yaml so every valid json document can be parsed by yaml.Unmarshal.

While in theory this would be true, in Go reality this does not work.
It might be possible if we use the Kubernetes yaml implementation instead.

>In the future and likely before the next release, fetch-measurements is going to need to the image UID to download the measurements anyway, which only the new configVersion will have.

I think you are confusing where this code section is implemented. `*m.FetchAndVerify()` is not a function of the `constellation config fetch-measurements`, but a method of `measurements.M`, which simply takes a URL and loads the measurements from there. That code section does not know, and does not need to know, the config version.

daniel-weisse:
>why did we decide to change file formats in the first place? 

0 indent JSON is a more stable file format than YAML, especially if we want to use it for signing measurements, as it does not depend on specific spacing.

internal/attestation/measurements/measurements.go
Nirusu: I see where it's coming from, but not sure I agree that FetchAndVerify should actually belong to `measurements.M` vs belonging to `Config` in the first place since all actual callers are actively there to modify the config.

I don't mind how it's implemented in detail, but later on it should be possible to have metadata enforcement in some way or another.

cli/internal/cloudcmd/upgrade.go
```diff
@@ -112,6 +112,17 @@ func (u *Upgrader) updateMeasurements(ctx context.Context, newMeasurements measu
 		return nil
 	}
 
+	// don't allow potential security downgrades by setting the warnOnly flag to true
+	for k, newM := range newMeasurements {
```
thomasten, line 116: good idea, but can't there be a case where we deliberately stop enforcing a PCR (and maybe enforce another one instead)?

cli/internal/cloudcmd/upgrade.go
```diff
@@ -112,6 +112,17 @@ func (u *Upgrader) updateMeasurements(ctx context.Context, newMeasurements measu
...
+		currentM, ok := currentMeasurements[k]
+		if !ok {
+			continue
+		}
+		if !currentM.WarnOnly && newM.WarnOnly {
```
thomasten, line 121:
```suggestion
		if currentM, ok := currentMeasurements[k]; ok && !currentM.WarnOnly && newM.WarnOnly {
```
or keep it as is if you find this one too hard to read

cli/internal/cloudcmd/validators.go
```diff
@@ -79,63 +78,58 @@ func (v *Validator) UpdateInitPCRs(ownerID, clusterID string) error {
 func (v *Validator) updatePCR(pcrIndex uint32, encoded string) error {
```
thomasten, line 78: update doc comment of this func

internal/attestation/measurements/measurements.go
```diff
@@ -94,58 +70,207 @@ func (m *M) FetchAndVerify(ctx context.Context, client *http.Client, measurement
...
 }
 
-// MarshalYAML overwrites the default behaviour of writing out []byte not as
-// single bytes, but as a single base64 encoded string.
-func (m M) MarshalYAML() (any, error) {
-	base64Map := make(map[uint32]string)
+// GetEnforced returns a list of all enforced Measurements,
+// i.e. all Measurements that are not marked as WarnOnly.
+func (m *M) GetEnforced() []uint32 {
+	enforced := []uint32{}
```
thomasten, line 100:
```suggestion
	var enforced []uint32
```
nit https://github.com/edgelesssys/wiki/blob/master/company/gostyle.md#nil-is-a-valid-slice

internal/attestation/measurements/measurements.go
```diff
@@ -94,58 +70,207 @@ func (m *M) FetchAndVerify(ctx context.Context, client *http.Client, measurement
...
+		}
+	}
+	return enforced
+}
 
-	for key, value := range m {
-		base64Map[key] = base64.StdEncoding.EncodeToString(value[:])
+// SetEnforced sets the WarnOnly flag to true for all Measurements
+// that are NOT included in the provided list of enforced measurements.
+func (m *M) SetEnforced(enforced []uint32) {
```
thomasten, line 111: what if we have an index in the enforced list that's not in m?

internal/attestation/measurements/measurements.go
```diff
@@ -94,58 +70,207 @@ func (m *M) FetchAndVerify(ctx context.Context, client *http.Client, measurement
...
+// WithAllBytes returns a measurement value where all 32 bytes are set to b.
+func WithAllBytes(b byte, warnOnly bool) Measurement {
+	return Measurement{
+		Expected: *(*[32]byte)(bytes.Repeat([]byte{b}, 32)),
+		WarnOnly: warnOnly,
+	}
+}
+
+// DefaultsFor provides the default measurements for given cloud provider.
+func DefaultsFor(provider cloudprovider.Provider) M {
```
thomasten, line 235: we need default (dummy) values for those that previously have been included in EnforcedMeasurements

internal/config/config.go
```diff
@@ -199,9 +196,6 @@ type AzureConfig struct {
 	// description: |
 	//   Expected confidential VM measurements.
 	Measurements Measurements `yaml:"measurements"`
```
thomasten, line 198:
```suggestion
	Measurements Measurements `yaml:"measurements" validate:"required"`
```

internal/watcher/validator.go
```diff
@@ -100,17 +101,24 @@ func (u *Updatable) Update() error {
...
 		return err
 	}
-	u.log.Debugf("New measurements: %v", measurements)
+	u.log.Debugf("New measurements: %+v", measurements)
 
-	var enforced []uint32
-	if err := u.fileHandler.ReadJSON(filepath.Join(constants.ServiceBasePath, constants.EnforcedPCRsFilename), &enforced); err != nil {
-		return err
+	// handle legacy measurement format, where expected measurements and enforced measurements were stored in separate data structures
+	// TODO: remove with v2.4.0
```
thomasten, line 111: reminder: if not already done, create ticket(s) for such TODOs

internal/watcher/validator.go
```diff
@@ -100,17 +101,24 @@ func (u *Updatable) Update() error {
...
+	if _, err := u.fileHandler.Stat(filepath.Join(constants.ServiceBasePath, constants.EnforcedPCRsFilename)); err == nil {
+		u.log.Debugf("Detected legacy format. Loading enforced PCRs...")
+
+		var enforced []uint32
+		if err := u.fileHandler.ReadJSON(filepath.Join(constants.ServiceBasePath, constants.EnforcedPCRsFilename), &enforced); err != nil {
+			return err
+		}
+		measurements.SetEnforced(enforced)
+		u.log.Debugf("Merged measurements with enforced values: %+v", measurements)
 	}
```
thomasten, line 121:
```suggestion
	var enforced []uint32
	if err := u.fileHandler.ReadJSON(filepath.Join(constants.ServiceBasePath, constants.EnforcedPCRsFilename), &enforced); err == nil {
		u.log.Debugf("Detected legacy format.")
		measurements.SetEnforced(enforced)
		u.log.Debugf("Merged measurements with enforced values: %+v", measurements)
	} else if !errors.Is(err, fs.ErrNotExist) {
		return err
	}
```
In most cases, avoid TOCTOU-like code and check the error of the actual action instead.

cli/internal/cloudcmd/upgrade.go
daniel-weisse: Use cases would be very small.
Only two I can really see are.
1. A CSP starts doing weird things in a PCR we currently enforce, resulting in that PCR being unstable.
  * We currently don't enforce any CSP specific PCRs with the exception of PCR[0] on GCP
2. Systemd starts doing weird things in a PCR we currently enforce, resulting in that PCR being unstable.
  * I would trust systemd to not do this suddenly without notice

If any of those 2 cases should happen, we can update the command to allow for changes in specific cases, which would then allow upgrades with a newer CLI version

internal/attestation/measurements/measurements.go
daniel-weisse: Currently does nothing.
I would opt to return an error in that case

internal/watcher/validator.go
daniel-weisse: https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/2641

malt3 (approved):
LGTM. I think we may consider using the kubernetes yaml parser at some point but this is good for now.

Nirusu (changes requested):
E2E tests need fixing (no enforcedMeasurements field anymore).

internal/attestation/measurements/measurements.go
```diff
@@ -94,58 +70,224 @@ func (m *M) FetchAndVerify(ctx context.Context, client *http.Client, measurement
...
+		}
+	}
 
-	for key, value := range m {
-		base64Map[key] = base64.StdEncoding.EncodeToString(value[:])
+	// set enforced measurements from list
+	for _, idx := range enforced {
+		measurement, ok := (*m)[idx]
+		if !ok {
+			return fmt.Errorf("measurement %d not in list, but set to enforced", idx)
```
thomasten, line 129: Suggest rearranging the code to do the check first so that m is not changed on error. This may not be needed for the current users of M, but avoids surprises and is a good practice in general.

internal/config/validation.go
```diff
@@ -210,3 +211,34 @@ func (c *Config) translateAzureInstanceTypeError(ut ut.Translator, fe validator.
...
+
+	t, _ := ut.T("no_placeholders", msg)
+	return t
+}
+
+func getPlaceholderEntries(measurements Measurements) []uint32 {
+	var placeholders []uint32
+
+	for idx, measurement := range measurements {
+		if bytes.Equal(measurement.Expected[:], bytes.Repeat([]byte{0x12, 0x34}, 16)) {
```
thomasten, line 238: consider exporting measurements.PlaceholderMeasurement and using it here

.github/actions/constellation_create/action.yml
```diff
@@ -67,6 +67,8 @@ runs:
       run: |
         constellation config generate ${{ inputs.cloudProvider }}
 
+        e2eMeasurements = ""
+
```
Nirusu, line 71: What does this do or is supposed to do?

thomasten (approved):

Nirusu (approved):

PR #557, opened 2022-11-15
AB#2616: Fix "stuck" Terraform states as best as we safely can

Nirusu:
### Proposed change(s)
- Move Constellation Terraform workspace to `./constellation-workspace` (thus simplify cleanup code and avoid destroying random Terraform clusters)
- Make "create" on unclean states retryable by checking if file on disk == expected file to write
- Make "destroy" work with missing Terraform plugins by calling "terraform init" before trying to "terraform destroy".

When reviewing, please read the issues I was trying to solve and see if that makes sense as a whole. Practical testing with random abortions and CTRL+C hitting might be useful.

Other ideas to the issues are also greatly appreciated.

### Related issue
To understand the changes here, let me explain the issue(s).
Currently, when you do `constellation create` and either:
- rapidly abort with CTRL+C
- Have wonky or no internet
- Terraform fails to download the provider plugins for some other reason

You will end up in a state where create fails, the rollback fails and terminate also fails. You are essentially stuck until you manually delete all Terraform files and directories in the current directory (`.terraform .terraform.hcl.lock modules main.tf variables.tf output.tf terraform.tfstate terraform.tfvars`).

So the first idea is to move all Terraform related files into a separate Terraform workspace directory. This is a feature Terraform officially supports and should makes things way cleaner. It unfortunately breaks version compatibility, but many changes aiming for v2.3.0 do and this can be easily migrated if we have an update flow. Might also be useful if we use Terraform for other features (IAM).

The second idea is that every file we try to write from the embedded filesystem and the .tfvars file should not error when they already exist. Terraform is able to continue applying when it has a .tfstate and the variables match. 

However, here I wasn't sure if we want to allow overwriting the files from the embedded filesystem to the disk or if we should error on something unknown. I opted for checking for equality and failing to allow the rollback mechanism clean up the previous state just in case we have a mismatch. This has the disadvantage that there's a slight window where e.g. `.terraform.lock.hcl` could be created, not written and blocks the rollback/terminate process.

The last idea is to allow the cleanup even when Terraform plugins are missing. There are many ways to do that and almost all are flawed without requiring manual user interaction. What I compromised on here is trying to initialize Terraform first and then try to destroy an existing cluster. The downside of this is when `terraform init` already fails during create for any case which is not CTRL+C (or we just hit the moment where not all files are written to disk), it will likely also fail on terminate. So we are stuck again until the user manually cleans up.

An alternative solution I was considering was to skip the `terraform destroy` when there's a lack of a .tfstate. The major benefit here would be that `terraform init` is not required for this and thus would allow even offline cleanup and fixing even more "stuck states". However, we already had the discussion using the state file as a proxy of existing once and did not want to go that route. This could break some expectations or statuses Terraform has (e.g. running in the background, .tfstate not written to disk yet). 

So I went for `terraform init` before `terraform destroy` it is, in trade off for slightly more "stuck" states where manual user interaction is required. 

In any case, even if we are in another "stuck" state now, manual cleanup is as easy as deleting `constellation-terraform` instead of finding all files and directories, both visible and hidden. Thus it should be an improvement over what we had before.

### DIFF
--- a/cli/internal/cmd/create.go
+++ b/cli/internal/cmd/create.go
@@ -12,6 +12,7 @@ import (
 	"io/fs"
 
 	"github.com/edgelesssys/constellation/v2/cli/internal/cloudcmd"
+	"github.com/edgelesssys/constellation/v2/cli/internal/terraform"
 	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
 	"github.com/edgelesssys/constellation/v2/internal/config"
 	"github.com/edgelesssys/constellation/v2/internal/constants"
@@ -126,7 +127,7 @@ func create(cmd *cobra.Command, creator cloudCreator, fileHandler file.Handler, 
 	idFile, err := creator.Create(cmd.Context(), provider, conf, flags.name, instanceType, flags.controllerCount, flags.workerCount)
 	spinner.Stop()
 	if err != nil {
-		return err
+		return translateCreateErrors(cmd, err)
 	}
 
 	if err := fileHandler.WriteJSON(constants.ClusterIDsFileName, idFile, file.OptNone); err != nil {
@@ -207,6 +208,27 @@ func checkDirClean(fileHandler file.Handler) error {
 	}
 
 	return nil
+}
+
+func translateCreateErrors(cmd *cobra.Command, err error) error {
+	switch {
+	case errors.Is(err, terraform.ErrTerraformWorkspaceDifferentFiles):
+		cmd.PrintErrln("\nYour current working directory contains an existing Terraform workspace which does not match the expected state.")
+		cmd.PrintErrln("This can be due to a mix up between providers, versions or an otherwise corrupted workspace.")
+		cmd.PrintErrln("Before creating a new cluster, try \"constellation terminate\".")
+		cmd.PrintErrf("If this does not work, either move or delete the directory %q.\n", constants.TerraformWorkingDir)
+		cmd.PrintErrln("Please only delete the directory if you made sure that all created cloud resources have been terminated.")
+		return err
+	case errors.Is(err, terraform.ErrTerraformWorkspaceExistsWithDifferentVariables):
+		cmd.PrintErrln("\nYour current working directory contains an existing Terraform workspace which was initiated with different input variables.")
+		cmd.PrintErrln("This can be the case if you have tried to create a cluster before with different options which did not complete, or the workspace is corrupted.")
+		cmd.PrintErrln("Before creating a new cluster, try \"constellation terminate\".")
+		cmd.PrintErrf("If this does not work, either move or delete the directory %q.\n", constants.TerraformWorkingDir)
+		cmd.PrintErrln("Please only delete the directory if you made sure that all created cloud resources have been terminated.")
+		return err
+	default:
+		return err
+	}
 }
 
 func must(err error) {
--- a/cli/internal/terraform/loader.go
+++ b/cli/internal/terraform/loader.go
@@ -7,8 +7,10 @@ SPDX-License-Identifier: AGPL-3.0-only
 package terraform
 
 import (
+	"bytes"
 	"embed"
 	"errors"
+	"fmt"
 	"io/fs"
 	"path"
 	"path/filepath"
@@ -41,7 +43,24 @@ func prepareWorkspace(fileHandler file.Handler, provider cloudprovider.Provider,
 			return err
 		}
 		fileName := strings.Replace(filepath.Join(workingDir, path), rootDir+"/", "", 1)
-		return fileHandler.Write(fileName, content, file.OptMkdirAll)
+		if err := fileHandler.Write(fileName, content, file.OptMkdirAll); errors.Is(err, afero.ErrFileExists) {
+			// If a file already exists, check if it is identical. If yes, continue and don't write anything to disk.
+			// If no, don't overwrite it and instead throw an error. The affected file could be from a different version,
+			// provider, corrupted or manually modified in general.
+			existingFileContent, err := fileHandler.Read(fileName)
+			if err != nil {
+				return err
+			}
+
+			if !bytes.Equal(content, existingFileContent) {
+				return fmt.Errorf("trying to overwrite existing Terraform file with different version")
+			}
+			return nil
+		} else if err != nil {
+			return err
+		}
+
+		return nil
 	})
 }
 
--- a/cli/internal/terraform/loader.go
+++ b/cli/internal/terraform/loader.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"io/fs"
 	"path"
+	"path/filepath"
 	"strings"
 
 	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
@@ -24,7 +25,7 @@ var terraformFS embed.FS
 
 // prepareWorkspace loads the embedded Terraform files,
 // and writes them into the workspace.
-func prepareWorkspace(fileHandler file.Handler, provider cloudprovider.Provider) error {
+func prepareWorkspace(fileHandler file.Handler, provider cloudprovider.Provider, workingDir string) error {
 	// use path.Join to ensure no forward slashes are used to read the embedded FS
 	rootDir := path.Join("terraform", strings.ToLower(provider.String()))
 	return fs.WalkDir(terraformFS, rootDir, func(path string, d fs.DirEntry, err error) error {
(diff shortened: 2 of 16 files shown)


Nirusu:
Added two-three other commits in which I split up CreateCluster into PrepareWorkspace and CreateCluster, with the rollback only being active after a successful PrepareWorkspace (e.g.: no unexpected files in the Terraform workspace).

Advantage here is that writing from the Terraform cluster creation to avoid destroying something in rollback from an unknown state we are not expect.

Downside is that it's another "stuck state" in case a file is corrupted or the user did "constellation create", aborted, changed the variables in constellation-conf.yaml and then retried "constellation create" again.

So not sure whether I want to keep the commit, but I can drop it if it's no good.

cli/internal/terraform/loader.go
```diff
@@ -39,27 +42,29 @@ func prepareWorkspace(fileHandler file.Handler, provider cloudprovider.Provider)
...
 			if err != nil {
 				return err
 			}
-			fileName := strings.TrimPrefix(path, rootDir+"/")
-			return ignoreFileNotFoundErr(fileHandler.RemoveAll(fileName))
-		}); err != nil {
+
+			if !bytes.Equal(content, existingFileContent) {
+				return fmt.Errorf("trying to overwrite existing Terraform file with different version")
+			}
```
daniel-weisse, line 56: I feel like reading both files and comparing their byte content isnt the ideal way to check for equality, but should be OK for this case, since this code is only used by the CLi

internal/constants/constants.go
```diff
@@ -77,6 +77,8 @@ const (
 	AdminConfFilename = "constellation-admin.conf"
 	// MasterSecretFilename filename of Constellation mastersecret.
 	MasterSecretFilename = "constellation-mastersecret.json"
+	// TerraformWorkingDir is the directory name for the TerraformClient workspace.
+	TerraformWorkingDir = "./constellation-terraform"
```
daniel-weisse, line 81:
```suggestion
	TerraformWorkingDir = "constellation-terraform"
```

daniel-weisse (approved):

cli/internal/terraform/loader.go
Nirusu: Mh, not sure to be honest. Initially I had the idea to compare SHA256 hashes, but that essentially ends up also being a byte-by-byte comparison.

I think the main issue here is that both files are loaded into memory. However, luckily Terraform files aren't this large so it shouldn't be a major issue I guess.

Nirusu:
Okay, in the end I decided to go for allowing to overwrite existing files to get rid of more annoying issues as long as there is no `terraform.tfstate`. This should not be an issue for the Constellation CLI flow given that this is the state file we use and files are likely almost the same for a completed cluster.

In case Terraform ran, we go back to the verify flow. In theory we can remove this and just take whatever is in there, however I am not sure if we want to allow this (even for certain safety reasons).

In addition, to further discourage reluctant users from not messing around in the directory, I opted to create a "DO_NOT_MODIFY_THIS_DIRECTORY_MANUALLY" warning file. I think this should really be the last warning not to play around with the files in there, as they might be overwritten or deleted depending on the actions an user takes. 

As a benefit for all this effort, the CLI should generally be more reliable to handle interruptions and ideally be able to get a clean unpacked state whenever it is interrupted and re-ran again. When a Terraform state actually exists is the point we stop modifying files and only allow verification as we don't want to have dangling cloud resources.

`constellation terminate` itself will only work when Terraform is able to initialize the providers. However, when a Terraform state file exists, that should be the case so create -> terminate should work in case overwriting is blocked from this point on.

Let me know what you think about it.

Nirusu:
Still personally indecisive about whether to allow overwriting or not, but I think for this PR we can use just not do it at first.

This is better than the current behaviour and should only corrupt if the process gets very abruptly at the beginning *or* if you CTRL+C, kill the rollback and instead change the variables and re-try again.

Then we require manual cleanup. The main issue is really the latter case which I could see as slightly annoying, but unless the variables are corrupt, terminate and rollback should work here.

I think if we still run into issues too often here we can investigate allowing overwriting, but I think there's a bunch of minor issues if we go down that road (e.g. inconsistent states, "dirty" flags, loosening terminate to allow deleting incomplete directories) that would need some discussion (or maybe even better: refactoring to avoid them in the first place)

As a compromise I added some user-friendly warnings on how to proceed if they encounter this issue.

Let me know what you think.

cli/internal/terraform/loader.go
Nirusu: @thomasten Any suggestion on how to do file equality comparisons here? Is bytes.Equal over the whole content fine (they are not large files) versus using a hash or similar?

cli/internal/cmd/create.go
```diff
@@ -209,6 +210,27 @@ func checkDirClean(fileHandler file.Handler) error {
...
+		cmd.PrintErrln("This can be due to a version mix up or a corrupted workspace.")
+		cmd.PrintErrln("Before creating a new cluster, try \"constellation terminate\".")
+		cmd.PrintErrf("If this does not work, either move or delete the directory \"%s\".\n", constants.TerraformWorkingDir)
+		cmd.PrintErrln("Please only delete the directory if you made sure that all created cloud resources have been terminated.")
+		return err
+	case terraform.ErrTerraformWorkspaceExistsWithDifferentVariables:
+		cmd.PrintErrln("\nYour current working directory contains an existing Terraform workspace which was initiated with different input variables.")
+		cmd.PrintErrln("This can be the case if you have tried to create a cluster before with different options which did not complete, or the workspace is corrupted.")
+		cmd.PrintErrln("Before creating a new cluster, try \"constellation terminate\".")
+		cmd.PrintErrf("If this does not work, either move or delete the directory \"%s\".\n", constants.TerraformWorkingDir)
```
daniel-weisse, line 226:
```suggestion
		cmd.PrintErrf("If this does not work, either move or delete the directory %q.\n", constants.TerraformWorkingDir)
		cmd.PrintErrln("Please only delete the directory if you made sure that all created cloud resources have been terminated.")
		return err
	case terraform.ErrTerraformWorkspaceExistsWithDifferentVariables:
		cmd.PrintErrln("\nYour current working directory contains an existing Terraform workspace which was initiated with different input variables.")
		cmd.PrintErrln("This can be the case if you have tried to create a cluster before with different options which did not complete, or the workspace is corrupted.")
		cmd.PrintErrln("Before creating a new cluster, try \"constellation terminate\".")
		cmd.PrintErrf("If this does not work, either move or delete the directory %q.\n", constants.TerraformWorkingDir)
```

cli/internal/cmd/create.go
```diff
@@ -209,6 +210,27 @@ func checkDirClean(fileHandler file.Handler) error {
...
+func translateCreateErrors(cmd *cobra.Command, err error) error {
+	switch err {
+	case terraform.ErrTerraformWorkspaceDifferentFiles:
+		cmd.PrintErrln("\nYour current working directory contains an existing Terraform workspace which does not match the expected state.")
+		cmd.PrintErrln("This can be due to a version mix up or a corrupted workspace.")
+		cmd.PrintErrln("Before creating a new cluster, try \"constellation terminate\".")
+		cmd.PrintErrf("If this does not work, either move or delete the directory \"%s\".\n", constants.TerraformWorkingDir)
+		cmd.PrintErrln("Please only delete the directory if you made sure that all created cloud resources have been terminated.")
+		return err
+	case terraform.ErrTerraformWorkspaceExistsWithDifferentVariables:
```
daniel-weisse, line 222:
```suggestion
	switch {
	case errors.Is(err, terraform.ErrTerraformWorkspaceDifferentFiles):
		cmd.PrintErrln("\nYour current working directory contains an existing Terraform workspace which does not match the expected state.")
		cmd.PrintErrln("This can be due to a version mix up or a corrupted workspace.")
		cmd.PrintErrln("Before creating a new cluster, try \"constellation terminate\".")
		cmd.PrintErrf("If this does not work, either move or delete the directory \"%s\".\n", constants.TerraformWorkingDir)
		cmd.PrintErrln("Please only delete the directory if you made sure that all created cloud resources have been terminated.")
		return err
	case errors.Is(err, terraform.ErrTerraformWorkspaceExistsWithDifferentVariables):
```
Allows us to also catch wrapped errors

daniel-weisse (approved):

cli/internal/terraform/loader.go
thomasten: Generally, prefer simple implementations over (prematurely) optimized, but more complex ones. (See our review guide.) So IMO this is the right implementation in this case.
If you would write a generic file comparison func, you would read the files in chunks. I think you would only use hashes in special case, e.g., if you have a large file that you want to compare against multiple others. But even in this case, this isn't always the best solution because you need to read all the other files completely even if they already differ in the first chunk.

PR #558, opened 2022-11-15
warn about function argument count over 5

datosh:

### Proposed change(s)

- As discussed in architecture WG, here is a proposal to lint functions >5 args.

### Checklist

- [ ] Link to Milestone


.github/workflows/test-lint.yml
```diff
@@ -53,4 +53,4 @@ jobs:
         with:
           skip-pkg-cache: true
           skip-build-cache: true
-          args: --timeout=15m ${{ steps.submods.outputs.submods }}
+          args: --new-from-rev=HEAD~ --timeout=15m ${{ steps.submods.outputs.submods }}
```
datosh, line 56: @katexochen is this what you had in mind regarding "only scan new code"? 
Not sure if I did it right, because a [manual trigger failed](https://github.com/edgelesssys/constellation/actions/runs/3471511080/jobs/5801096211). But it may be just because I used the wrong event to trigger the pipeline?

katexochen:
```
with:
         ....

          # Optional: show only new issues if it's a pull request. The default value is `false`.
          # only-new-issues: true
```

datosh:
Manually triggered lint run for this branch: https://github.com/edgelesssys/constellation/actions/runs/3478006214

katexochen:
> https://github.com/edgelesssys/constellation/actions/runs/3478006214

Shouldn't the test succeed with `only-new-issues`?

daniel-weisse (approved):
I support this change :+1:

datosh:
> > https://github.com/edgelesssys/constellation/actions/runs/3478006214
> 
> Shouldn't the test succeed with `only-new-issues`?

As I understand the `golangci-lint` docs it only works when executed in the context of a PR. Since I didn't change any source files in this PR, the workflow wasn't triggered, and I had to run it manually, so it scanned everything.

datosh:
@katexochen see the last two runs in this PR where I changed and reverted a .go file as a means to test this.

malt3 (approved):
LGTM

PR #563, opened 2022-11-16
Merge v2.2.1 changes back to main

datosh:
### Proposed change(s)

- Merge v2.2.1 changes back to main

### Checklist

- [ ] Link to Milestone


daniel-weisse (approved):
In the future we might want to consider not increasing helm chart versions for services that did not change.
But we should have a in-depth discussion about something like that first

datosh:
> In the future we might want to consider not increasing helm chart versions for services that did not change. But we should have a in-depth discussion about something like that first

Good point. I put them on the agenda for next dev sync.

PR #566, opened 2022-11-16
Feat/init cluster download components with hash

3u13r:
### Proposed change(s)

- Pin and verify Kubernetes components with hash
- Send Kubernetes components list from CLI to first bootstrapper. Those are used during Init()

Note, all other bootstrappers besides the first one initialized still directly uses the `VersionsConfig` map. This will be changed in further PRs. We first need to modify the JoinService. For further reference see "rfc/updates.md".

### Checklist

- [x] Link to Milestone


bootstrapper/internal/kubernetes/k8sapi/k8sutil.go
```diff
@@ -75,34 +78,30 @@ func NewKubernetesUtil() *KubernetesUtil {
 	}
 }
 
+// InstallComponentsFromCLI installs the kubernetes components passed from the CLI.
+func (k *KubernetesUtil) InstallComponentsFromCLI(ctx context.Context, kubernetesComponents versions.ComponentVersions) error {
```
daniel-weisse, line 82: nit: Naming doesn't fit that well.
IMO the bootstrapper, or any parts of it, should not know about the existence of the CLI, just that it should install some components it received over gRPC

bootstrapper/internal/kubernetes/k8sapi/k8sutil.go
```diff
@@ -75,34 +78,30 @@ func NewKubernetesUtil() *KubernetesUtil {
...
+	for _, component := range kubernetesComponents {
+		if err := k.inst.Install(
+			ctx, component.URL, component.Hash, component.InstallPath, component.Extract,
+		); err != nil {
```
daniel-weisse, line 86: Since we are trying to keep the number of function arguments at a reasonable level, would it make sense to just pass the `component` struct here, instead of pulling the values from it and passing them one by one?

bootstrapper/internal/kubernetes/k8sapi/k8sutil.go
3u13r: Yes, I agree. My plan was to rename this to `InstallComponents` once we can deprecate the other function. So this is more of a placeholder name for now, but I'm open to any other suggestions here.

daniel-weisse: Name this function `InstallComponents` and the other one `_DeprecatedInstallComponents` :wink:

bootstrapper/internal/kubernetes/k8sapi/k8sutil.go
3u13r: I also thought about that, but decided against it. My thought process was that we differentiate between `InstallComponents` and `Install` i.e. `Install` has a more general Interface. Also my question is if we should always optimize the number of parameters, or should we just stay below 5?

bootstrapper/internal/kubernetes/k8sapi/k8sutil.go
3u13r: But the original `InstallComponents` _will be_ deprecated but currently is still in use. Therefore I opted for the `FunctionNew()` and `Function` approach rather than `Function` and `FunctionOld()`. The idea being to not alter the "old" interface when it's being removed anyway.

bootstrapper/internal/kubernetes/k8sapi/k8sutil.go
daniel-weisse: >i.e. `Install` has a more general Interface

Does it need to be more general? 
And is it really more general if all of the parameters are just variables of `versions.ComponentVersions`?
IMO this isn't as much about keeping low number of arguments, but about knowing exactly what variables are needed for the function.

E.g. a function with the parameters `(context.Context, string, string, string, bool)` isn't as clear as `(context.Context, versions.ComponentVersions)`.
Especially since the later option allows us to set doc comments on the individual parts.

daniel-weisse (approved):

malt3 (approved):
LGTM

PR #574, opened 2022-11-17
ci: use /usr/bin/env instead of /bin/env

katexochen:


malt3 (approved):
LGTM! Thanks

PR #578, opened 2022-11-17
ci: use shfmt fork

katexochen:

### Proposed change(s)

- Shfmt didn't recognize files like `mkosi.finalize` because of the dot in the filename. Upstream issue: https://github.com/mvdan/sh/issues/944
- Also using shfmt for shellcheck input to ensure these files are checked, too.


malt3 (approved):
LGTM

PR #603, opened 2022-11-21
Surpress argument-limit errors and add TODO

derpsteb:
### Proposed change(s)

- Linting on main currently reports errors for the argument-limit rule. However, we decided to only enforce the limit of 5 arguments for new code and refactor old code piece-by-piece. To make sure the linter on main stays green this patch disables the revive linter on the relevant lines only and adds a TODO comment.


katexochen:
Imo this shouldn't be the way...

katexochen (approved):

derpsteb:
While talking about @katexochen's concerns the following points came up:
- We currently have no immediate solution for some of the functions. Refactoring them probably needs disproportionately large changes.
- Introducing new functions that are similar to existing ones (e.g. like Kubewrapper.InitCluster) is not possible without breaking the linting rule.

I am merging the PR nevertheless to stop the linter from failing on main. We may want to revisit our use of the linter rule and these comments in the future if it turns out that we keep adding exceptions to the current ones.

derpsteb:
Waiting for #606 to be merged so go-generate is green again.

PR #612, opened 2022-11-21
re-run go generate

datosh:

### Proposed change(s)

- go generate check is failing - is this a change we want to have? Or has something else changed we should fix? 
-

### Checklist

- [ ] Link to Milestone


daniel-weisse:
If this is important, you can merge.
https://github.com/edgelesssys/constellation/blob/feat/re-generate/cli/internal/helm/testdata/Azure/constellation-services/charts/azuredisk-csi-driver/templates/storageclass_integrity.yaml will also need to be updated.
Wondering why the CI is not failing the unit tests are not failing for this PR.
Will need to check tomorrow

PR #614, opened 2022-11-21
Use config structs to limit nr of function args

derpsteb:
### Proposed change(s)

- I forgot to run golangci-lint on the sub packages when introducing the revive-ignore comments. This PR introduces some config structs for the missing functions.


malt3 (approved):
LGTM

PR #621, opened 2022-11-22
Remove kubernetesServicesVersion from upgrade RFC.

derpsteb:
### Proposed change(s)

While discussing the reasons for the separation between `microserviceVersion`  and `kubernetesServicesVersion` we realized that it is not necessary to distinguish between the two, since images for each service are set in the CLI anyways. Deployments were similar for all versions up until now.

Because it seems to me that it will make things easier for us and the user I propose to remove the separation.

### DIFF
--- a/rfc/updates.md
+++ b/rfc/updates.md
@@ -173,8 +173,7 @@ we should add comments to those fields who will not update the cluster.
 
 ```yaml
 kubernetesVersion: 1.24.3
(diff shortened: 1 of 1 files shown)


malt3:
I approve of this change but would wait until after todays meeting

3u13r (approved):
LGTM

PR #624, opened 2022-11-22
AB#2579 Add constellation iam create command

stdoutput:
### Proposed change(s)

- Add `constellation iam create` command to the Constellation CLI, automatically sets up the required IAM configuration based on the paramters supplied by the user by using Terraform scripts.
- Command output will consist of the values for the `constellation-conf.yml` that need to be filled in by the user.

### GCP
- Command: `constellation iam create gcp`
- You must be authenticated with the [GCP CLI](https://cloud.google.com/sdk/gcloud) in the used shell context.
- Available flags are:
 * `zone`: Required. [GCP zone](https://cloud.google.com/compute/docs/regions-zones#available) you want to create the configuration in.
 * `serviceAccountID`: Required. ID for the service account that will be created. Must match `^[a-z](?:[-a-z0-9]{4,28}[a-z0-9])$`
 * `projectID`: Required. ID of the GCP project the configuration will be created in.
 * `yes`: Skips confirmation steps.
### AWS
- Command: `constellation iam create aws`
- You must be authenticated with the [AWS CLI](https://aws.amazon.com/en/cli/) in the used shell context.
- Available flags are:
 * `zone`: Required. [AWS availability zone](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-availability-zones) you want to create the configuration in.
 * `prefix`: Required.  Name prefix which will be used to create the IAM resources.
 * `yes`: Skips confirmation steps.
 ### Azure
- Command: `constellation iam create azure`
- You must be authenticated with the [Azure CLI](https://learn.microsoft.com/en-us/cli/azure/install-azure-cli) in the used shell context.
- Available flags are:
 * `region`: Required. [Azure region](https://azure.microsoft.com/en-us/explore/global-infrastructure/geographies/#choose-your-region) you want to create the configuration in.
 * `resourceGroup`: Required.  Name of the resource group the IAM resources will be created in.
 * `servicePrincipal`: Required.  Name of the service principal that will be created.
 * `yes`: Skips confirmation steps.
 
### Checklist

- [ ] Link to Milestone


cli/internal/terraform/loader.go
```diff
@@ -27,11 +28,26 @@ var ErrTerraformWorkspaceDifferentFiles = errors.New("creating cluster: trying t
...
+func prepareWorkspace(action int, fileHandler file.Handler, provider cloudprovider.Provider, workingDir string) error {
+	var rootDir string
+	// select the correct path for the Terraform files depending on the action
+	switch action {
+	case IAMAction:
+		// use path.Join to ensure no forward slashes are used to read the embedded FS
+		rootDir = path.Join("terraform", strings.ToLower(provider.String()), "iam")
+	case ClusterAction:
+		// use path.Join to ensure no forward slashes are used to read the embedded FS
+		rootDir = path.Join("terraform", strings.ToLower(provider.String()))
```
Nirusu, line 47:
```suggestion
	case IAMAction:
		// use path.Join to ensure no forward slashes are used to read the embedded FS
		rootDir = path.Join("terraform", "iam", strings.ToLower(provider.String()))
	case ClusterAction:
		// use path.Join to ensure no forward slashes are used to read the embedded FS
		rootDir = path.Join("terraform", strings.ToLower(provider.String()))
```

Would suggest changing terraform/gcp/iam -> terraform/iam/gcp so we don't necessarily extract it for the cluster creation.

cli/internal/terraform/loader.go
```diff
@@ -27,11 +28,26 @@ var ErrTerraformWorkspaceDifferentFiles = errors.New("creating cluster: trying t
 //go:embed terraform/*/.terraform.lock.hcl
 var terraformFS embed.FS
 
+const (
+	IAMAction = iota
+	ClusterAction
+)
```
katexochen, line 34:
```suggestion
type Action int

const (
	IAMAction Action = iota
	ClusterAction
)
```

cli/internal/terraform/terraform/gcp/iam/main.tf
```diff
@@ -0,0 +1,69 @@
+terraform {
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = "~> 4.43.0"
```
katexochen:
```suggestion
      version = "4.43.0"
```
We try to always pin versions.

cli/internal/terraform/terraform/gcp/iam/.terraform.lock.hcl
```diff
@@ -0,0 +1,26 @@
...
+provider "registry.terraform.io/hashicorp/google" {
+  version     = "4.43.0"
+  constraints = "~> 4.43.0"
+  hashes = [
+    "h1:JpShTtgnxpiIVnr0R2Lccrh84mnrf7Z1/v/yw0UZ1gI=",
+    "h1:OTNLlWoTq+SdbbtgLK7uFVAn3aP9QVjIGYU2ndKsz+Q=",
+    "h1:PSIkDVwksHe9oZd+XP369N8U+6/+SPF8Z5wHkcwmWKw=",
+    "h1:gmUUWhuuY/YRIllvVBRGl1kUHqsNBHQ/4BHdwKQbzXQ=",
+    "h1:mheXqRMLbMeTr8/E6UakMhWwIL0HqwIHYBE2u2Sbldg=",
+    "zh:0b424cab24856dc47177733145fa61b731f345a6a42a0c0b7910ccfcf4e8c8a2",
```
katexochen: Not sure, but this looks like not all platforms are pinned as required. Command is:
```
terraform providers lock -platform=linux_arm64 -platform=linux_amd64 -platform=darwin_arm64 -platform=darwin_amd64 -platform=windows_amd64
```
I already updated these in main.

katexochen (commented):
Just some early feedback. Great code quality! :dodo:

cli/internal/cmd/iamcreate.go
```diff
@@ -0,0 +1,21 @@
...
+
+// NewIAMCreateCmd returns a new cobra.Command for the iam create parent command. It needs another verb, and does nothing on its own.
+func newIAMCreateCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Create IAM configuration on a cloud platform for your Constellation cluster",
+		Long:  "Create IAM configuration on a cloud platform for your Constellation cluster.",
+		Args: cobra.MatchAll(
+			cobra.ExactArgs(0),
+		),
```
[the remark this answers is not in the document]
stdoutput: Agreed! Also removed this on the other commands which used the combinator whilst only having 1 argument rule [here](https://github.com/edgelesssys/constellation/pull/624/commits/7d38edfc9fa375db02fc5bece474a08a19f4ef11)

cli/internal/cmd/iamcreate.go
```diff
@@ -0,0 +1,21 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// NewIAMCreateCmd returns a new cobra.Command for the iam create parent command. It needs another verb, and does nothing on its own.
+func newIAMCreateCmd() *cobra.Command {
```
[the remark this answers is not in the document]
stdoutput: Agreed! I discussed with Leo but we wanted to keep the 1 command / file structure. Maybe we should discuss the structure of this folder generally. It might make sense to somehow group parents with their subcommands?

cli/internal/cmd/iamcreategcp.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// Write back values
+	dec, err := base64.StdEncoding.DecodeString(iamFile.GCPOutput.ServiceAccountKey)
+	if err != nil {
+		return err
+	}
+
+	outJson := make(map[string]string)
+	err = json.Unmarshal([]byte(dec), &outJson)
+	if err != nil {
```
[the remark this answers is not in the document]
stdoutput: Probably bad naming here. I just need to unpack the values from the b64-encoded JSON object anywhere. Rephrased this [here](https://github.com/edgelesssys/constellation/pull/624/commits/f639b15552af849fabe1ee2f38cbb1e8d318f786)

cli/internal/cmd/iamcreategcp.go
```diff
@@ -0,0 +1,188 @@
...
+	return nil
+}
+
+// parseGcpFlags parses and validates the flags of the iam create gcp command
+func parseGcpFlags(cmd *cobra.Command) (gcpFlags, error) {
+	zone, err := cmd.Flags().GetString("zone")
+	if err != nil {
+		return gcpFlags{}, fmt.Errorf("parsing zone string: %w", err)
+	}
+	if !regexp.MustCompile(`^\w+-\w+-[abc]$`).MatchString(zone) { // Put into constants package later on
```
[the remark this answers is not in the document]
stdoutput: Agreed! I tried to copy the same style that was used in the other commands. Actually, since we only use them exactly there, putting them into the constants module doesn't make much sense at all

cli/internal/cmd/iamcreategcp.go
```diff
@@ -0,0 +1,188 @@
...
+		return gcpFlags{}, fmt.Errorf("invalid zone string: %s", zone)
+	}
+
+	region, err := cmd.Flags().GetString("region")
+	if err != nil {
+		return gcpFlags{}, fmt.Errorf("parsing region string: %w", err)
+	}
+	if region == "" {
+		// Infer region from zone
+		zoneParts := strings.Split(zone, "-")
```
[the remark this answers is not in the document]
stdoutput: I probably dont get what strings.Cut does exactly here. From the docs I think it would cut a zone string, e.g. `"europe-west1-a"` (containing 2 dashes) at the first occurance of a dash, so I would be left with `["europe", "west1-a"]` (Which doesnt really make sense for the above usage?)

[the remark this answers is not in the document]
katexochen: You're right, missed that second dash. :sweat_smile:

Nirusu:
Can you please fix the linter warnings first before I review? :)

Nirusu:
Regarding the "Check licenses" step, can you add this as the header to each of the file reported in there (= new *.go files created as part of your PR)?

```bash
/*
Copyright (c) Edgeless Systems GmbH
SPDX-License-Identifier: AGPL-3.0-only
*/

```

Then the check should pass.

cli/internal/cloudcmd/iam.go
```diff
@@ -0,0 +1,85 @@
...
+			return iamid.File{}, err
+		}
+		defer cl.RemoveInstaller()
+		return c.createGCP(ctx, cl, iamConfig)
+	default:
+		return iamid.File{}, fmt.Errorf("unsupported cloud provider: %s", provider)
+	}
+}
+
+// createGcp creates the IAM configuration on GCP.
```
Nirusu, line 58:
```suggestion
// createGCP creates the IAM configuration on GCP.
```

cli/internal/cmd/iamcreategcp.go
```diff
@@ -0,0 +1,215 @@
...
+	gcpFlags, err := parseGcpFlags(cmd)
+	if err != nil {
+		return err
+	}
+
+	// verify CP auth.
+	isAuthed, err := checkGcpAuth()
+	if err != nil {
+		return err
+	}
```
Nirusu, line 67: Can you change the capitalization from Gcp to GCP?

cli/internal/cmd/iamcreategcp.go
```diff
@@ -0,0 +1,215 @@
...
+
+	email, err := cmd.Output()
+	if err != nil {
+		return false, fmt.Errorf("could not retrieve gcloud login: %w", err)
+	}
+
+	if !regexp.MustCompile(`^\w+([-+.']\w+)*@\w+([-.]\w+)*\.\w+([-.]\w+)*$`).MatchString(string(email)) {
+		return false, fmt.Errorf("invalid gcloud user")
+	}
+	*/
```
Nirusu, line 212: Mh, currently we just let Terraform handle this and return its error message. 

If the error message looks clean enough, I would say let Terraform handle this. If not, then maybe a catch layer would be sufficient based on Terraform's error message?

cli/internal/terraform/terraform/iam/gcp/.terraform.lock.hcl
```diff
@@ -0,0 +1,42 @@
...
+    "h1:JpShTtgnxpiIVnr0R2Lccrh84mnrf7Z1/v/yw0UZ1gI=",
+    "h1:JzDVTkBVHTw67N7S9bD/+u7U3aipjutF4MZx0eplONw=",
+    "h1:OTNLlWoTq+SdbbtgLK7uFVAn3aP9QVjIGYU2ndKsz+Q=",
+    "h1:PSIkDVwksHe9oZd+XP369N8U+6/+SPF8Z5wHkcwmWKw=",
+    "h1:d6Ds7aGeW5M8+AgWfkfbh3HTf3tP8ePjYGhoJnOxklU=",
+    "h1:enGPHgPHCbmLU8BIHLHcOwRlbQnPPQzdmXNoFVs1lIg=",
+    "h1:gmUUWhuuY/YRIllvVBRGl1kUHqsNBHQ/4BHdwKQbzXQ=",
+    "h1:hMhGerpfzsbfQ3DFsAGwJYpSIK6an+wMGx2I/lwaNuY=",
+    "h1:iotFu763eGWVC4zfXkNuo3RUeQLDvwVX2E7R/2sNxCg=",
+    "h1:mheXqRMLbMeTr8/E6UakMhWwIL0HqwIHYBE2u2Sbldg=",
```
Nirusu, line 17: I am not the expert on Terraform lock files here but I am surprised there's one less hash than in ../../gcp/.terraform.lock.hcl.

Are all platforms included (macOS arm64 + amd64, Linux arm64 + amd64, Windows amd64)? If yes, just click resolved.

cli/internal/terraform/terraform_test.go
```diff
@@ -83,12 +88,82 @@ func TestPrepareCluster(t *testing.T) {
...
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+			require := require.New(t)
+
+			c := &Client{
+				tf:         &stubTerraform{},
+				file:       file.NewHandler(tc.fs),
+				workingDir: constants.TerraformWorkingDir,
```
Nirusu, line 158:
```suggestion
				workingDir: constants.TerraformIAMWorkingDir,
```

?

cli/internal/terraform/terraform_test.go
```diff
@@ -214,6 +296,122 @@ func TestCreateCluster(t *testing.T) {
...
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+			require := require.New(t)
+
+			c := &Client{
+				tf:         tc.tf,
+				file:       file.NewHandler(tc.fs),
+				workingDir: constants.TerraformWorkingDir,
```
Nirusu, line 399:
```suggestion
				workingDir: constants.TerraformIAMWorkingDir,
```
?

cli/internal/terraform/loader_test.go
```diff
@@ -71,14 +85,14 @@ func TestLoader(t *testing.T) {
 
 			file := file.NewHandler(afero.NewMemMapFs())
 
-			err := prepareWorkspace(file, tc.provider, constants.TerraformWorkingDir)
+			err := prepareWorkspace(tc.action, file, tc.provider, constants.TerraformWorkingDir)
```
Nirusu, line 88: I would split this up based on action so that 
cluster -> TerraformWorkingDir
iam -> TerraformIAMWorkingDir

cli/internal/cmd/iamcreategcp.go
stdoutput: I think relying on the TF error messages is enough here. It was my initial idea to add custom checks, since I didn't know Terraform had such concise error messages. I think they are more than sufficient for the UX we try to achieve.  Therefore I removed the custom check now

cli/internal/terraform/terraform/iam/gcp/.terraform.lock.hcl
stdoutput: The version locks wildly confuse me. I used the exact same command snippet as for the others. Can someone maybe share the command they have been using for the other Terraform configs?

cli/internal/terraform/loader_test.go
stdoutput: Is there a typo in the suggested change then? It would make sense to leave the 4th paramter (`constants.TerraformWorkingDir`) out then, right?

Nirusu: Already told you this but for the other reviewers: 

`prepareWorkspace` is not bound to TerraformClient specifically, so either both have it hardcoded depending on the action or both should get it supplied as a parameter to keep them roughly in sync.

I think I have a slight preference to pass this as an argument, but as long as it is consistent I don't mind too much.

3u13r (changes requested):
Overall good work! Regarding change requests nothing too major.

cli/internal/cmd/iamcreateaws.go
```diff
@@ -0,0 +1,128 @@
...
+		Use:   "aws",
+		Short: "Create IAM configuration on AWS for your Constellation cluster",
+		Long:  "Create IAM configuration on AWS for your Constellation cluster.",
+		Args:  cobra.ExactArgs(0),
+		RunE:  runIAMCreateAWS,
+	}
+
+	cmd.Flags().String("prefix", "", "Name prefix for all resources.")
+	must(cobra.MarkFlagRequired(cmd.Flags(), "prefix"))
+	cmd.Flags().String("zone", "", "AWS availability zone the resources will be created in (e.g. us-east-2a). Find available zones here: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-availability-zones. Note that we do not support every zone / region. Find a list of currently supported regions here: https://docs.edgeless.systems/constellation/next/getting-started/first-steps.")
```
3u13r, line 29:
```suggestion
	cmd.Flags().String("zone", "", "AWS availability zone the resources will be created in (e.g. us-east-2a). Find available zones here: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-availability-zones. Note that we do not support every zone / region. You can find a list of all supported regions in our docs.")
```
I'm on the fence about this. On the one hand we shouldn't directly link to docs, especially to the "next" version of them, since the docs can and will change rapidly. On the other hand, it is nice to dive the user a direct hint. 

Since the user should follow the docs anyway when creating a Constellation, I think that it is ok to generally refer them to the docs.

cli/internal/cmd/iamcreateaws.go
```diff
@@ -0,0 +1,128 @@
...
+		return awsFlags{}, fmt.Errorf("parsing zone string: %w", err)
+	}
+	if strings.HasPrefix(zone, "eu-central-1") {
+		region = "eu-central-1"
+	} else if strings.HasPrefix(zone, "us-east-2") {
+		region = "us-east-2"
+	} else if strings.HasPrefix(zone, "ap-south-1") {
+		region = "ap-south-1"
+	} else {
+		return awsFlags{}, fmt.Errorf("invalid AWS region, to find a correct region please refer to our docs (https://docs.edgeless.systems/constellation/next/getting-started/first-steps) and https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-availability-zones")
```
3u13r, line 107: (similar to above with regards to linking to our docs)

3u13r (changes requested):
Last change request from my side. Otherwise looking good.

katexochen (changes requested):

cli/internal/cmd/iamcreateaws.go
```diff
@@ -0,0 +1,128 @@
...
+	spinner.Stop()
+	if err != nil {
+		return err
+	}
+
+	cmd.Printf("region:\t%s\n", awsFlags.region)
+	cmd.Printf("zone:\t%s\n", awsFlags.zone)
+	cmd.Printf("iamProfileControlPlane:\t%s\n", iamFile.AWSOutput.ControlPlaneInstanceProfile)
+	cmd.Printf("iamProfileWorkerNodes:\t%s\n", iamFile.AWSOutput.WorkerNodeInstanceProfile)
+	cmd.Println("Your IAM configuration was created successfully. Please fill the above values into your configuration file.")
```
datosh, line 83: As a potential follow up task: Have we thought about filling in the values directly into the configuration file? I imagine most users want to do this anyway, and `iam create` could get a flag `--no-fill-in` to prevent this in case this is not desired.

stdoutput: I talked to Leo about this and the current decision was that anyone should be able to setup the IAM on an independent machine. However, I was also a supporter of directly filling the values into the config or at least offer a flag to enable automatic filling. For now, I think we should add the IAM command as is and create another ticket on the board.

cli/internal/terraform/terraform/iam/gcp/main.tf
```diff
@@ -2,7 +2,7 @@ terraform {
   required_providers {
     google = {
       source  = "hashicorp/google"
-      version = "4.44.1"
+      version = "4.44.0"
```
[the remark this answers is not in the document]
stdoutput: iirc, @katexochen has made a point for pinning provider versions. Should the renovate-bot update these versions automatically? If yes, does it also create the new hashes?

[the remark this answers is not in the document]
katexochen: Renovate updated these versions on main in https://github.com/edgelesssys/constellation/pull/666. It also updates the hashes. If you didn't add a provider, you should should just use the version and hashes from main when rebasing.

cli/internal/cmd/iamcreateaws.go
3u13r: Tracked in AB#2706 now

3u13r (approved):
LGTM

cli/internal/cmd/iamcreateaws.go
datosh: That makes sense!

PR #628, opened 2022-11-23
Remove obsolote revive comments

derpsteb:
### Proposed change(s)

- Since we removed the rule we don't need these comments anymore.


katexochen (approved):
Thx :)

PR #635, opened 2022-11-23
Write version API RFC

malt3:
### Proposed change(s)

- Write version API RFC

### Additional info

- This is a really early draft idea. I am not proposing it at this point in time.
- The document exists as a PR so it is easier to comment on.
- Not every aspect has to be implemented right away

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/rfc/update-api.md
+++ b/rfc/update-api.md
@@ -12,23 +12,26 @@ Design goals:
     - CLI versions
     - Kubernetes versions
 
-The folowing HTTP endpoints are available:
+The following HTTP endpoints are available:
 
-- `GET /constellation/v1/updates/latest/` contains files showing the latest version available
+- `GET /constellation/v1/updates/<stream>/latest/` contains files showing the latest version available
     - `image.json` contains the latest image version
     - `microservice.json` contains the latest microservice version
     - `cli.json` contains the latest cli version
     - `kubernetes.json` contains the latest supported version of Kubernetes
-- `GET /constellation/v1/updates/major/<major-version>/` contains files with version information for this major version
+- `GET /constellation/v1/updates/<stream>/major/<major-version>/` contains files with version information for this major version
     - `image.json` contains a list of all minor image versions that belong to a major version
     - `microservice.json` contains a list of all minor microservice versions that belong to a major version
     - `cli.json` contains a list of all minor cli versions that belong to a major version
     - `kubernetes.json` contains a list of all supported minor version of Kubernetes that belong to a major version
-- `GET /constellation/v1/updates/minor/<minor-version>/` contains files with version information for this minor version
+- `GET /constellation/v1/updates/<stream>/minor/<minor-version>/` contains files with version information for this minor version
     - `image.json` contains a list of all patch image versions that belong to a minor version
     - `microservice.json` contains a list of all patch microservice versions that belong to a minor version
     - `cli.json` contains a list of all patch cli versions that belong to a minor version
     - `kubernetes.json` contains a list of all supported patch version of Kubernetes that belong to a minor version
+
+`stream` is used to distinguish between different release streams. For example, `stable` and `beta` could be two different streams.
+Currently, only `stable` is supported. The parameter exists to allow for future expansion.
 
 ## Examples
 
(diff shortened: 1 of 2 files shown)


datosh (commented):
Great idea. I think this is both flexible and simple enough for our use case.

One thing that came to mind is that our microservices already use Helm, which is the only system that already provides some sort of update discovery.

```sh
$ helm search repo nginx-ingress --version 1.10 -l
NAME                    CHART VERSION   APP VERSION     DESCRIPTION                                       
stable/nginx-ingress    1.10.3          0.25.0          An nginx Ingress controller that uses ConfigMap...
stable/nginx-ingress    1.10.2          0.25.0          An nginx Ingress controller that uses ConfigMap...
stable/nginx-ingress    1.10.1          0.25.0          An nginx Ingress controller that uses ConfigMap...
stable/nginx-ingress    1.10.0          0.25.0          An nginx Ingress controller that uses ConfigMap...

helm search repo nginx-ingress --version 1 -l
NAME                    CHART VERSION   APP VERSION     DESCRIPTION                                       
stable/nginx-ingress    1.41.3          v0.34.1         DEPRECATED! An nginx Ingress controller that us...
stable/nginx-ingress    1.41.2          v0.34.1         An nginx Ingress controller that uses ConfigMap...
stable/nginx-ingress    1.41.1          v0.34.1         An nginx Ingress controller that uses ConfigMap...
[... lots more ...]
```

Would it make sense to use this functionality? Either by dropping microservices from this RFC or by using the information helm provides as a backend to generate this information?

3u13r:
> Would it make sense to use this functionality? Either by dropping microservices from this RFC or by using the information helm provides as a backend to generate this information?

Since the helm charts are bundled as part of the CLI, we must upgrade it in order to upgrade the helm deployments. Also, we currently don't host them in a helm repository which is searchable.

rfc/update-api.md
```diff
@@ -0,0 +1,105 @@
...
+    - Kubernetes versions
+
+The folowing HTTP endpoints are available:
+
+- `GET /constellation/v1/updates/latest/` contains files showing the latest version available
+    - `image.json` contains the latest image version
+    - `microservice.json` contains the latest microservice version
+    - `cli.json` contains the latest cli version
+    - `kubernetes.json` contains the latest supported version of Kubernetes
+- `GET /constellation/v1/updates/major/<major-version>/` contains files with version information for this major version
```
3u13r, line 22: I'm thinking about the tradeoff between defining different endpoints and just having one endpoint and do the version parsing client side. I think that one endpoint with all version is more generic, which benefits third parties. 

Also defining one endpoint and doing client side parsing allows for more flexibility.

rfc/update-api.md
```diff
@@ -0,0 +1,105 @@
...
+    - `image.json` contains a list of all minor image versions that belong to a major version
+    - `microservice.json` contains a list of all minor microservice versions that belong to a major version
+    - `cli.json` contains a list of all minor cli versions that belong to a major version
+    - `kubernetes.json` contains a list of all supported minor version of Kubernetes that belong to a major version
+- `GET /constellation/v1/updates/minor/<minor-version>/` contains files with version information for this minor version
+    - `image.json` contains a list of all patch image versions that belong to a minor version
+    - `microservice.json` contains a list of all patch microservice versions that belong to a minor version
+    - `cli.json` contains a list of all patch cli versions that belong to a minor version
+    - `kubernetes.json` contains a list of all supported patch version of Kubernetes that belong to a minor version
```
3u13r, line 31: Should we combine some of those? I.e. I can only obtain a new microservice version through updating the CLI. So it is more of a meta information about the specific CLI version. I think we should at least somehow reflect which microservice version and Kubernetes version a specific CLI supports.

malt3: I think we should not impose version constraints early on in this API. Keeping them separate allows us to change compatibility checks between resources more easily in the CLI.

rfc/update-api.md
malt3: > Also defining one endpoint and doing client side parsing allows for more flexibility.

Multiple endpoints give us more flexibility for each resource type.
We could add metadata like changelogs, download URLs or other info that is unique to one kind of resource.

malt3:
@katexochen would like to name this API `version`(s) API. I think that's a good idea as it describes clearly what the API does / doesn't do.

PR #638, opened 2022-11-24
Remove client pkg from kubectl pkg

derpsteb:
## Proposed changes 

The nested client pkg was necessary to implement a generator pattern. The generator was necessary as the Kubewrapper type expects a k8sapi.Client object during instantiation. However, the required kubeconfig is not ready during Kubewrapper creation. This patch relies on an Initialize function to set the Kubeconfig and hands over an empty struct during Kubewrapper creation. This allows us to remove the extra Client abstraction.

The deleted testcases were only covering the Apply/GetObject code. This code was based on our custom marshaler for k8s structs and is not necessary anymore as we moved our deployments to Helm. The uncovered code exclusively calls the kubernetes.Interface functions.

e2e test manual for [aws](https://github.com/edgelesssys/constellation/actions/runs/3539346542), [gcp](https://github.com/edgelesssys/constellation/actions/runs/3539345333/jobs/5941136780) and [azure](https://github.com/edgelesssys/constellation/actions/runs/3539344112/jobs/5941134330).


bootstrapper/internal/kubernetes/k8sapi/kubectl/kubectl.go
```diff
@@ -11,110 +11,96 @@ import (
...
-		return err
-	}
+	deployments := k.AppsV1().Deployments(namespace)
 
-	if err = client.AddTolerationsToDeployment(ctx, tolerations, name, namespace); err != nil {
+	// retry resource update if an error occurs
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		result, err := deployments.Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get Deployment to add toleration: %v", err)
```
daniel-weisse, line 92:
```suggestion
			return fmt.Errorf("failed to get Deployment to add toleration: %w", err)
```

bootstrapper/internal/kubernetes/k8sapi/kubectl/kubectl.go
```diff
@@ -123,30 +109,71 @@ func (k *Kubectl) AddTolerationsToDeployment(ctx context.Context, tolerations []
...
-	if err != nil {
-		return err
-	}
+	deployments := k.AppsV1().Deployments(namespace)
+
+	// retry resource update if an error occurs
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		result, err := deployments.Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get Deployment to add node selector: %v", err)
```
daniel-weisse, line 118:
```suggestion
			return fmt.Errorf("failed to get Deployment to add node selector: %w", err)
```

bootstrapper/internal/kubernetes/k8sapi/kubectl/kubectl.go
```diff
@@ -123,30 +109,71 @@ func (k *Kubectl) AddTolerationsToDeployment(ctx context.Context, tolerations []
...
+		err := k.WaitForCRD(ctx, crd)
 		if err != nil {
 			return err
 		}
 	}
 	return nil
 }
+
+// WaitForCRD waits for the given CRD to be established.
+func (k *Kubectl) WaitForCRD(ctx context.Context, crd string) error {
```
daniel-weisse, line 148:
```suggestion
// waitForCRD waits for the given CRD to be established.
func (k *Kubectl) waitForCRD(ctx context.Context, crd string) error {
```
Does not need to be exposed

bootstrapper/internal/kubernetes/k8sapi/kubectl/kubectl.go
```diff
@@ -123,30 +109,71 @@ func (k *Kubectl) AddTolerationsToDeployment(ctx context.Context, tolerations []
...
+		switch event.Type {
+		case watch.Added, watch.Modified:
+			crd := event.Object.(*apiextensionsv1.CustomResourceDefinition)
+			if crdHasCondition(crd.Status.Conditions, apiextensionsv1.Established) {
+				return nil
+			}
+		case watch.Deleted:
+			return fmt.Errorf("crd %q deleted", crd)
+		case watch.Error:
+			return fmt.Errorf("crd %q error: %v", crd, event.Object)
```
daniel-weisse, line 166:
```suggestion
		case watch.Error:
			return fmt.Errorf("crd %q error: %+v", crd, event.Object)
```
Not sure about returning the full Object in the error message.
Open to leaving it tho, if it helps with debugging.

daniel-weisse (commented):
Nice :)
Some small things

daniel-weisse:
Running [`punused`](https://github.com/bep/punused) the following kubectl methods and functions are reported as not in use, and can / should probably be removed:

```
bootstrapper/internal/kubernetes/k8sapi/k8sutil.go:55:2 method WaitForCRDs is unused
bootstrapper/internal/kubernetes/k8sapi/k8sutil.go:56:2 method ListAllNamespaces is unused
bootstrapper/internal/kubernetes/k8sapi/kubectl/kubectl.go:26:5 variable ErrKubeconfigNotSet is unused
bootstrapper/internal/kubernetes/k8sapi/kubectl/kubectl.go:137:19 method (*Kubectl).WaitForCRDs is unused
``` 

Also left over from other changes the `GCPGuestAgentDaemonset` is now unused.

bootstrapper/internal/kubernetes/k8sapi/kubectl/kubectl.go
```diff
@@ -11,110 +11,96 @@ import (
...
-	}
+	return &Kubectl{}
 }
 
-// Apply will apply the given resources using server-side-apply.
-func (k *Kubectl) Apply(resources kubernetes.Marshaler, forceConflicts bool) error {
-	if k.kubeconfig == nil {
-		return ErrKubeconfigNotSet
+// Initialize sets sets all required fields so the Kubectl client can be used.
+func (k *Kubectl) Initialize(kubeconfig []byte) error {
```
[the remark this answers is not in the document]
derpsteb: I tried to do that, but we need dependency injection for the Kubewrapper tests. If we were to initialize the kubectl object [here](https://github.com/edgelesssys/constellation/blob/main/bootstrapper/internal/kubernetes/kubernetes.go#L155), the tests would expect a file on the filesystem. This seemed like an antipattern to me. I nonetheless feel like it would improve the overall code quality.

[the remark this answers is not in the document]
daniel-weisse: Maybe replacing the `client` variable with a `newClient` function might do the trick.
Could also be somewhat unwieldy.
Maybe keeping it as is works, while keeping in mind that we want to refactor the entire package in the future, especially since we moved a large part of it to helm

bootstrapper/internal/kubernetes/k8sapi/kubectl/kubectl.go
derpsteb: Code was removed.

derpsteb:
Left ListAllNamespaces. punused does not seem to recognize it's use [here](https://github.com/edgelesssys/constellation/blob/main/bootstrapper/internal/kubernetes/kubeWaiter/waiter.go#L45).

daniel-weisse (approved):
Looks good to me

PR #640, opened 2022-11-24
Fix potential data race when accessing a validators OID

daniel-weisse:
### Proposed change(s)

- Add a lock to `*Updatable.OID()` to prevent data race when calling the function in parallel to `*Updatable.Update()`, which can happen if a client connects to the server at the same time the sever is updating its validator

### Checklist

- [ ] Link to Milestone


katexochen (approved):
lgtm

PR #641, opened 2022-11-24
Add new generate measurements matrix CI/CD action (now with AWS support)

Nirusu:
### Proposed change(s)
- Remove old `constellation_measure` and `generate_measurement` actions 
- Introduce a new matrix measurement generation pipeline that does all CSPs at the same time (based on E2E tests this time with a `nop` test for code reuse)
- Add AWS support 
- Use new measurement format in JSON
- Add a "measurements with metadata" header which is manually feeded into `pcr-reader` in the CI
- Verify the expected measurements from the "Build OS image" pipeline with the same subset retrieved from a running cluster (separate new hack tool called `pcr-compare`)
- Sign & Upload the images to the new CDN S3 bucket

The TODO post this PR is to implement `fetch-measurements` to be compatible to this (use CDN URLs, image UID and also parse the metadata header).

An example run can be found here: https://github.com/edgelesssys/constellation/actions/runs/3542049575

### Additional info
- For AWS, PCR 0 _seems_ to be stable. It is some 16 bytes values that's measured in there. But we don't know what it is exactly, so don't enforce it for now. (Hey AWS, can we please have some more documentation on NitroTPM? Would be great 🥹)
- For AWS, PCR 1 is special compared to other CSPs. It contains EBS storage device configuration and is not stable (likely contains some resource IDs which differs for each created cluster?). So it's stripped from the output.
- Example output looks like this:
```json
{
    "csp": "azure",
    "image": "debug-v2.3.0-pre.0.20221122094954-29ff6cb78630",
    "measurements": {
        "1": {
            "expected": "3d458cfe55cc03ea1f443f1562beec8df51c75e14a9fcf9a7234a13f198e7969",
            "warnOnly": true
        },
        "11": {
            "expected": "0000000000000000000000000000000000000000000000000000000000000000",
            "warnOnly": false
        },
        "12": {
            "expected": "0425c52a9bd01b5e581b250093c0cfdd6bb2347829c3b9b3049f627b73e6980d",
            "warnOnly": false
        },
        "13": {
            "expected": "0000000000000000000000000000000000000000000000000000000000000000",
            "warnOnly": false
        },
        "14": {
            "expected": "d7c4cc7ff7933022f013e03bdee875b91720b5b86cf1753cad830f95e791926f",
            "warnOnly": true
        },
        "15": {
            "expected": "0000000000000000000000000000000000000000000000000000000000000000",
            "warnOnly": false
        },
        "2": {
            "expected": "3d458cfe55cc03ea1f443f1562beec8df51c75e14a9fcf9a7234a13f198e7969",
            "warnOnly": true
        },
        "3": {
            "expected": "3d458cfe55cc03ea1f443f1562beec8df51c75e14a9fcf9a7234a13f198e7969",
            "warnOnly": true
        },
        "4": {
            "expected": "dfd42fd6f8ed69c29ee325ab894d6fbcf74ac813c1c11db698ff12d1e216e89b",
            "warnOnly": false
        },
        "5": {
            "expected": "ee817a78e022c74c8db133aac21b1b3e90e990ef5e6e188b0bc6cb6e19ee0381",
            "warnOnly": true
        },
        "7": {
            "expected": "346547a8ce5957af27e552427d6b9e6d9cb502f0156e9155380451eea1b3f0ed",
            "warnOnly": true
        },
        "8": {
            "expected": "0000000000000000000000000000000000000000000000000000000000000000",
            "warnOnly": false
        },
        "9": {
            "expected": "97754437fbea2d903e50eb9201c156fd2620842a4fadc01f16618f73610ccbdc",
            "warnOnly": false
        }
    }
}
```


.github/workflows/generate-measurements.yml
```diff
@@ -72,23 +89,83 @@ jobs:
...
-          cosignPublicKey: ${{ startsWith(github.ref, 'refs/heads/release/v') && secrets.COSIGN_PUBLIC_KEY || secrets.COSIGN_DEV_PUBLIC_KEY }}
-          cosignPrivateKey: ${{ startsWith(github.ref, 'refs/heads/release/v') && secrets.COSIGN_PRIVATE_KEY || secrets.COSIGN_DEV_PRIVATE_KEY }}
-          cosignPassword: ${{ startsWith(github.ref, 'refs/heads/release/v') && secrets.COSIGN_PASSWORD || secrets.COSIGN_DEV_PASSWORD }}
-          awsAccessKeyID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          awsSecretAccessKey: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          awsDefaultRegion: ${{ secrets.AWS_DEFAULT_REGION }}
-          awsBucketName: ${{ secrets.PUBLIC_BUCKET_NAME }}
+          test: "nop"
+
+      # Check /docs/secure_software_distribution.md#sign-measurements
```
daniel-weisse, line 94: This file does not exist

hack/pcr-compare/main.go
```diff
@@ -0,0 +1,160 @@
...
+
+	parsedActualMeasurements, err := parseActualMeasurements(os.Args[2])
+	if err != nil {
+		panic(err)
+	}
+
+	mapExpectedMeasurements := createMapFromStrings(parsedExpectedMeasurements)
+	mapActualMeasurements := convertBytesMeasurementsToHexMeasurements(parsedActualMeasurements)
+
+	foundMismatch := compareMeasurements(mapExpectedMeasurements, mapActualMeasurements)
```
daniel-weisse, line 60: This code seems unnecessarily complex.

Why not just unmarshal the measurement files directly to `measurements.M`, delete the keys we don't want (or copy just the keys we want), and then call `expectedMeasurements.EqualTo(actualMeasurements)` ?

We could even go with removing this tool completetly, and just use [`yq compare`](https://mikefarah.gitbook.io/yq/v/v3.x/commands/compare) to handle the task for us

internal/attestation/measurements/measurements.go
```diff
@@ -38,6 +38,13 @@ const (
 // M are Platform Configuration Register (PCR) values that make up the Measurements.
 type M map[uint32]Measurement
 
+// WithMetadata is a struct supposed to provide CSP & image metadata next to measurements.
+type WithMetadata struct {
```
daniel-weisse, line 42: This data type currently has no use in our main code.
Do we expect to use it in the future?
I could see adapting the `measurements.M` unmarshal methods to either take a file with metadata or without, in which case we probably don't have to export this type.

daniel-weisse (changes requested):
>For AWS, PCR 0 seems to be stable. [...] so don't enforce it for now.

We don't want to enforce any values we can't reproduce locally. (The only exception being GCP PCR[0]).
If its stable across different VMs, we can add it to the list of `warnOnly` measurements

>The TODO post this PR is to `implement fetch-measurements` to be compatible to this

Since this TODO is a hard release blocker for v2.3, can you please add a ticket for this?

internal/attestation/measurements/measurements.go
Nirusu: `fetch-measurements` would likely require it in some way.

.github/workflows/generate-measurements.yml
```diff
@@ -1,66 +1,83 @@
...
       - name: Create Azure resource group
+        if: matrix.provider == 'azure'
         id: az_resource_group_gen
-        if: ${{ github.event.inputs.cloudProvider == 'azure' }}
         shell: bash
         run: |
           uuid=$(cat /proc/sys/kernel/random/uuid)
           name=e2e-test-${uuid%%-*}
-          az group create --location westus --name "$name" --tags e2e
+          az group create --location northeurope --name "$name" --tags e2e
```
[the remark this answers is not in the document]
Nirusu: No that's from rebasing I guess.

.github/docs/release.md
```diff
@@ -87,11 +87,11 @@ This checklist will prepare `v1.3.0` from `v1.2.0`. Adjust your version numbers
...
         ```
 
-    12. [Generate measurements](/.github/workflows/generate-measurements.yml) for the images on each CSP.
+    12. [Generate measurements](/.github/workflows/generate-measurements.yml) for the images.
 
         ```sh
-           gh workflow run generate-measurements.yml --ref release/v$minor -F cloudProvider=azure -F osImage=v$ver -F isDebugImage=false
-           gh workflow run generate-measurements.yml --ref release/v$minor -F cloudProvider=gcp -F osImage=v$ver -F isDebugImage=false
+           gh workflow run generate-measurements.yml --ref release/v$minor -F osImage=v$ver -F isDebugImage=false
+           gh workflow run generate-measurements.yml --ref release/v$minor -F osImage=v$ver -F isDebugImage=false
```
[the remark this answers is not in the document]
Nirusu: Yes (though it actually misses a parameter here, will add that).

hack/pcr-compare/main.go
Nirusu: To be honest I would keep it in Go since it's cleaner than `yq compare`, has nicer output and would avoid more Bash hackery in the actions. Also the final consumer is going to be Go code using the same structs so it would make sense to have some common ground here.

But I'll try to strip down the code, I originally wrote it before we had `measurements.M`and `EqualTo`.

.github/workflows/generate-measurements.yml
Nirusu: Good thing it was there all the time, that's copy-paste :) 
Will remove it.

.github/workflows/generate-measurements.yml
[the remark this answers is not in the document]
Nirusu: Actually no, it's copy-paste from other actions (E2E tests) also northeurope.
So what's our preference here?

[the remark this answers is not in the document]
daniel-weisse: If our other actions use northeurope, this should also do so

hack/pcr-compare/main.go
Nirusu: Okay, I simplified this quite a bit. Now both inputs use the same functions for everything, equality is checked with `EqualTo` but I do the same output since I hope that's nice to look at for debugging.

Check if you think it's fine that way.

Nirusu: Okay, with cooler output it becomes slightly more complicated, though the actual check is still done by `EqualTo` so it should fail and abort the pipeline even if the more complex output logic has some mistakes.

On the other hand though, `yq compare` is deprecated and does not exist anymore:
https://mikefarah.gitbook.io/yq/upgrading-from-v3#comparing-yaml-files

Of course we could diff but I don't know, I still like the Go code more.

daniel-weisse: Looks a lot better now, thanks.

IMO the extra comparison is not that important, since we already print the files multiple times during the workflow, but if you think it adds value we can keep it.

malt3 (approved):
I am a bit late to this PR but LGTM

daniel-weisse (approved):

hack/pcr-compare/main.go
Nirusu: Alright.

 If this thing makes too much trouble in the future we can just remove it and replace it I guess. But I like it for fast debugging (though it looks cooler locally to be honest since Github Actions strips the colors).

PR #643, opened 2022-11-24
deps: disable major upgrades for k8s.io/client-go

katexochen:
There are older versions tagged like v12.0.0, but latest version is v.0.25.4.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -132,6 +132,15 @@
       "prPriority": 15
     },
     {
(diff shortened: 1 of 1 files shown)


malt3 (approved):
Kubernetes view on Go versioning is interesting. LGTM.

katexochen:
looks like this change isn't working as expected... renovate still tries to upgrade to v11. :roll_eyes:

PR #644, opened 2022-11-24
AB#2645 Add warning about non retriable error during init

daniel-weisse:
### Proposed change(s)

- Add a simple warning about `constellation init` being non retriable if the command fails at a certain step

### Additional info
- I currently marked only one error as non retriable, if there are any others we should catch please say so
- Additional information for failures during other steps of `init` might be helpful
  -  If you notice any let me know, or create a PR in the future

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/cli/internal/cmd/init.go
+++ b/cli/internal/cmd/init.go
@@ -9,6 +9,7 @@ package cmd
 import (
 	"context"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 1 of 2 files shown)


Nirusu:
Generally I would say LGTM for now...?

I think when we do more fine-grained progress reports we can also think which errors we could mark as unrecoverable and which not...?

thomasten (approved):
I agree with Nils

Nirusu (approved):

PR #646, opened 2022-11-24
rfc: include upgrade check command

3u13r:
### Proposed change(s)

- Add the discussed changes to the update rfc.

### DIFF
--- a/rfc/updates.md
+++ b/rfc/updates.md
@@ -201,14 +201,25 @@ Note that:
 * `constellation upgrade check`
 * `constellation upgrade apply`
 
-When `constellation upgrade check` is called it checks if the current CLI embeds helm charts and kubernetes components which are newer than the ones defines in the `constellation-config.json`. If this is the case, it prints a detailed list of all the components that will be updated. Moreover, it checks via the update API (see: rfc/update-api.json) for new image patch versions as they are always compatible with inside a minor version.
+When `constellation upgrade check` is called it checks if the current CLI includes helm charts and kubernetes components that are newer than the ones configured in `constellation-config.json`.
+If this is the case, the CLI prints a list of all components that will be updated.
+Moreover, it checks for new image patch versions via the update API (see: rfc/update-api.json).
+Image patch versions are forward compatible within one minor version.
 
-Lastly, it uses the update API (see: rfc/update-api.json) if there is a newer CLI version available. If this is the case, it will print the latest version and omit ony of the points above since. If this version is more than +1 minor version apart it will also show the latest CLI of the next minor version, and suggest a way to download it. Note that if there are still e.g. microservice updates needed with the current CLI, we need to promt the user to first install those before continuing with the next minor release.
+Lastly, the CLI checks if a newer CLI version is available via the update API (see: rfc/update-api.json). If this is the case, it will print the latest CLI version instead of the output described above.
+If the current version and latest version diverge more than one minor version, it will also show the latest CLI of the next minor version, and suggest a way to download it.
+An example:
+current CLI version: `2.3.2`  
+available CLI version with minor version `2.4.0`: `2.4.1`, `2.4.2`, `2.4.3`  
+latest CLI version: `2.5.0`  
+`constellation upgrade check` will show `2.5.0` as latest, and suggest that the next step in the upgrade process is `2.4.3`.  
+Since any CLI can only upgrade from one minor version below to its own version, we need to perform the upgrade to `2.4.3` before upgrading to `2.5.0`.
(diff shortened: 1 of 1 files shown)


malt3 (approved):
LGTM

rfc/updates.md
```diff
@@ -194,6 +194,51 @@ Note that:
...
   Deployment variations could be introduced into the Helm charts if they become necessary in the future.
 
+### CLI commands
+
+`constellation upgrade` has 2 sub commands:
+
+* `constellation upgrade check`
+* `constellation upgrade apply`
+
+When `constellation upgrade check` is called it checks if the current CLI embeds helm charts and kubernetes components which are newer than the ones defines in the `constellation-config.json`. If this is the case, it prints a detailed list of all the components that will be updated. Moreover, it checks via the update API (see: rfc/update-api.json) for new image patch versions as they are always compatible with inside a minor version.
```
derpsteb, line 204:
```suggestion
When `constellation upgrade check` is called it checks if the current CLI includes helm charts and kubernetes components that are newer than the ones configured in `constellation-config.json`. If this is the case, the CLI prints a list of all components that will be updated. Moreover, it checks for new image patch versions via the update API (see: rfc/update-api.json). Image patch versions are forward compatible within one minor version.
```

rfc/updates.md
```diff
@@ -194,6 +194,51 @@ Note that:
...
+
+Lastly, it uses the update API (see: rfc/update-api.json) if there is a newer CLI version available. If this is the case, it will print the latest version and omit ony of the points above since. If this version is more than +1 minor version apart it will also show the latest CLI of the next minor version, and suggest a way to download it. Note that if there are still e.g. microservice updates needed with the current CLI, we need to promt the user to first install those before continuing with the next minor release.
```
derpsteb, line 206:
```suggestion
Lastly, the CLI checks if a newer CLI version is available via the update API (see: rfc/update-api.json). If this is the case, it will print the latest CLI version instead of the output described above. 
If the difference between `current_version` and `latest_version` is greater than one it will also show the latest CLI of the next minor version, and suggest a way to download it. 
An example:

current CLI version: `2.3.2`  
available CLI version with minor version `2.4`: `2.4.1`, `2.4.2`, `2.4.3`  
latest CLI version: `2.5.0`  
`constellation upgrade check` will show `2.5.0` as latest, and suggest that the next step in the upgrade process is `2.4.3`.  

The need to download CLI version `2.4.3` first comes from the fact that each minor version of the CLI is only able to upgrade from `current_version-1` to `current_version`.
If there are still microservice updates needed with the current CLI, we need to prompt the user to first install those before continuing with the next minor release.
```

rfc/updates.md
```diff
@@ -194,6 +194,51 @@ Note that:
...
+
+We also print `In newer CLI versions there are even newer versions available.` if e.g. there is a newer patch version of Kubernetes available in one of the proposed minor versions.
```
derpsteb, line 208: Wouldn't this required quite a complicated lookup, or at least an extension of the current update API, to resolve which versions will be available with another CLI version?

rfc/updates.md
```diff
@@ -194,6 +194,51 @@ Note that:
...
+
+Then executing `constellation upgrade check --write-config` all the new version values are automatically written to the `constellation-conf.json` so that the user can directly
```
derpsteb, line 210:
```suggestion
Executing `constellation upgrade check --write-config` writes all new version values to `constellation-conf.json`.
This allows the user to execute `constellation upgrade apply` without manually modifying `constellation-conf.json`.
```

rfc/updates.md
```diff
@@ -194,6 +194,51 @@ Note that:
...
+  CloudNodeManager: 1.24.1 --> 1.24.2
+
+Possible Constellation microservices upgrade to 2.2.0:
+  KMS: 2.1.3 --> 2.2.0
+  joinService: 2.1.3 --> 2.2.0
+  nodeOperator: 2.1.3 --> 2.2.0
+
+There is a new CLI available: v2.6.0
+
+Your current CLI version is more than more minor version apart from the latest release.
```
derpsteb, line 233:
```suggestion
Your current CLI version is more than one minor version apart from the latest release.
```

derpsteb:
Thanks for writing this down! I really like these RFCs..

Something I haven't done in the past myself, but maybe we should start doing when writing text in markdown: one line per sentence. Makes it easier to add suggestions for single sentences :D.

rfc/updates.md
3u13r: I don't think so. For the microservices we should just be able to get the latest version from the upgrade API and print this notification if the latest does not match the one embedded in the CLI.

derpsteb: That makes sense :).

PR #648, opened 2022-11-25
Add helm's quote function to various fields

derpsteb:
The constellationUID is sometimes interpreted as integer if it contains 0e, as the yaml parsing interprets that as scientific notation. Since it is a [best practices](https://helm.sh/docs/chart_best_practices/values/#make-types-clear) to quote string fields anyways this patch also quotes other fields where an actual string is required.

Manually tested on Azure & GCP.
e2e test runs: [azure](https://github.com/edgelesssys/constellation/actions/runs/3546643517), [gcp](https://github.com/edgelesssys/constellation/actions/runs/3546644802), [aws](https://github.com/edgelesssys/constellation/actions/runs/3546650017).

### DIFF
--- a/cli/internal/helm/charts/edgeless/constellation-services/charts/autoscaler/templates/aws-deployment.yaml
+++ b/cli/internal/helm/charts/edgeless/constellation-services/charts/autoscaler/templates/aws-deployment.yaml
@@ -22,7 +22,7 @@ spec:
     spec:
       containers:
         - name: cluster-autoscaler
-          image: {{ .Values.image }}
+          image: {{ .Values.image | quote }}
           imagePullPolicy: IfNotPresent
           livenessProbe:
             httpGet:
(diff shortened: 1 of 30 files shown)


daniel-weisse:
Might be a use case for fuzz testing :wink: 
While this won't catch all potential errors, can you add a test where we set `.Values.constellationUID` to a problematic value?

Nirusu:
It's not just `0e` by the way, more like a clean integer *and* `0e` numbers ;) 

Fix generally looks fine for me on a quick glance, environment variables should always be strings.

derpsteb:
The [existing tests](https://github.com/edgelesssys/constellation/blob/main/cli/internal/helm/loader_test.go#L172) already used an int as constellation UID. But we can't really test this behavior as the problem only arises when the API server tries to parse the structs, as far as I can tell.

I guess we could build a test case that imports the api server's parsing code. But I am unsure if we really want to do this.

derpsteb:
I think the long duration on azure was some hickup at azure? I had an init take 20mins on azure with a cluster started at the same time. But subsequent clusters were fine. Starting another run: [azure2](https://github.com/edgelesssys/constellation/actions/runs/3547274877)

Nirusu (approved):
I am not super familiar with all the Helm charts we deploy, but LGTM.

derpsteb:
The manual e2e tests on azure are taking very long since a few days ago. The phenomenon can also be seen on main. Creating clusters with this branch using the CLI works as expected. If the changes from this PR introduced the slow manual e2e tests, it should break init and not functionality within the cluster.

PR #649, opened 2022-11-25
Constellation Operator: Add image version field

malt3:
### Proposed change(s)

- Constellation Operator: Add image version field

### Additional info
- The CLI still needs to be extended to use this new field for OS image version selection during upgrades

### Checklist

- [x] Link to Milestone


daniel-weisse (approved):
I like it

PR #653, opened 2022-11-25
AB#2644 Fetch measurements from CDN and perform metadata validation

daniel-weisse:
### Proposed change(s)
- Only build Constellation CLI with release cosign public key if we build a release CLI
  - Use dev cosign public key otherwise
- Adapt `constellation config fetch-measurements` command to work with new JSON format and our artifact CDN
- Enforce metadata validation on `fetch-measurement`
  - This only works for images/measurements signed using our updated build pipeline from: https://github.com/edgelesssys/constellation/pull/641
  - For fetching measurements of older images, older CLI have to be used. This is not a breaking change, since old CLIs should not be used to upgrade to newer images. New CLIs don't need to fetch measurements for old images, since downgrading is not intended or supported
- Enforce metadata validation for measurements
- Add YAML marshal and unmarshal methods to `cloudprovider.Provider`
- Replace some instances where tests were using signatures generated using the release key

### Additional info
There are some transitional changes to tge `upgrade plan` command so unit test pass.
Changes will be dropped/refactored with [AB#2654](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/2654)
 

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -148,9 +148,6 @@ const (
 	// Releases.
 	//
 
-	// S3PublicBucket contains measurements & releases.
-	S3PublicBucket = "https://public-edgeless-constellation.s3.us-east-2.amazonaws.com/"
-
 	// CDNRepositoryURL is the base URL of the Constellation CDN artifact repository.
 	CDNRepositoryURL = "https://cdn.confidential.cloud"
 	// CDNImagePath is the default path to image references in the CDN repository.
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -150,16 +150,13 @@ const (
 
 	// S3PublicBucket contains measurements & releases.
 	S3PublicBucket = "https://public-edgeless-constellation.s3.us-east-2.amazonaws.com/"
-	// CosignPublicKey signs all our releases.
-	CosignPublicKey = `-----BEGIN PUBLIC KEY-----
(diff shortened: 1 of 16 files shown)


internal/constants/constants.go
```diff
@@ -150,16 +150,13 @@ const (
 
 	// S3PublicBucket contains measurements & releases.
 	S3PublicBucket = "https://public-edgeless-constellation.s3.us-east-2.amazonaws.com/"
```
Nirusu, line 152:
```suggestion
```
You can probably remove this now, should be unused...?

internal/constants/keys_enterprise.go
```diff
@@ -0,0 +1,16 @@
+//go:build enterprise
```
datosh, line 1: nit: Should we have another build flag here? Maybe `release` or `!development`.
I can imagine having more `enterprise` features in the future, and the need to test them before a release. Which could lead to conflicts reusing the build flag.
I am also fine with keeping it, and changing it once it becomes necessary.

datosh (approved):
LGTM! Love the adoption of adding and removing `TODO(AB#xxxx)` keeps the code super clear while it is being worked on 👍

malt3 (approved):
LGTM

PR #654, opened 2022-11-25
config: sort measurements numerically

thomasten:
Previously, writing the config file caused the measurements to be randomly ordered. This PR adds a bit of custom marshaling to sort the measurements numerically.

### DIFF
--- a/internal/attestation/measurements/measurements.go
+++ b/internal/attestation/measurements/measurements.go
@@ -364,6 +364,8 @@ func (c mYamlContent) Less(i, j int) bool {
 }
 
 func (c mYamlContent) Swap(i, j int) {
+	// The slice is filled like {key1, value1, key2, value2, ...}.
+	// We need to swap both key and value.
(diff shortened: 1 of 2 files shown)


internal/attestation/measurements/measurements.go
```diff
@@ -325,3 +342,28 @@ type encodedMeasurement struct {
...
+	}
+	rhs, err := strconv.Atoi(c[2*j].Value)
+	if err != nil {
+		panic(err)
+	}
+	return lhs < rhs
+}
+
+func (c mYamlContent) Swap(i, j int) {
+	c[2*i], c[2*j] = c[2*j], c[2*i]
```
datosh, line 367: We should repeat the comment from `mYamlContent` about the `key1, value1, key2, value2, ...` structure to explain the `2*` everywhere. 
ngl.. I was a bit confused whats going on here😆

daniel-weisse (approved):

PR #655, opened 2022-11-26
initserver: add client verification

3u13r:
### Proposed change(s)

- add client verification to `initserver.Init()`
- pass initSecretHash as Tags (AWS, Azure) or Metadata (GCP) 

### Checklist

- [x] Link to Milestone


cli/internal/terraform/terraform.go
```diff
@@ -74,30 +74,39 @@ func (c *Client) PrepareWorkspace(provider cloudprovider.Provider, vars Variable
...
 	}
 
-	return ip, nil
+	secretOutput, ok := tfState.Values.Outputs["initSecret"]
+	if !ok {
+		return "", "", errors.New("no initSecret output found")
+	}
+	secret, ok := secretOutput.Value.(string)
+	if !ok {
+		return "", "", errors.New("invalid type in IP output: not a string")
```
daniel-weisse, line 106:
```suggestion
		return "", "", errors.New("invalid type in initSecret output: not a string")
```

cli/internal/terraform/terraform/aws/modules/instance_group/main.tf
```diff
@@ -57,26 +57,35 @@ resource "aws_autoscaling_group" "autoscaling_group" {
...
+  //  key                 = "constellation-uid"
+  //  value               = var.uid
+  //  propagate_at_launch = true
+  //}
+  //
+  //tag {
+  //  key                 = "KubernetesCluster"
+  //  value               = "Constellation-${var.uid}"
+  //  propagate_at_launch = true
+  //}
```
daniel-weisse, line 80: remove?

internal/cloud/cloud.go
```diff
@@ -11,4 +11,6 @@ const (
 	TagRole = "constellation-role"
 	// TagUID is the tag/label key used to identify the UID of a cluster.
 	TagUID = "constellation-uid"
+	// TagInitSecretHash is the tag/label key used to identify the hash of the init secret.
+	TagInitSecretHash = "init-secret-hash"
```
daniel-weisse, line 15: Can we prefix this with `constellation-`, or would we risk running into name length limitations?

internal/cloud/metadata/metadata.go
```diff
@@ -30,6 +30,8 @@ type InstanceMetadata struct {
 	// AliasIPRanges is a list of IP ranges that are attached.
 	// May be empty on certain CSPs.
 	AliasIPRanges []string
+	// InitSecretHash is the hash of the init secret.
+	InitSecretHash string
```
daniel-weisse, line 34: I would prefer to have a function to retrieve `InitSecretHash`, instead of having it as part of `InstanceMetadata`.
The current code already retrieves all the hash information in separate functions to the main `Self` method, so there shouldn't be much to refactor

daniel-weisse (changes requested):
We also need a way to make this work for QEMU.
We can either implement something using the metadata API, or use a hardcoded value for now

bootstrapper/initproto/init.pb.go
```diff
@@ -40,6 +40,7 @@ type InitRequest struct {
 	EnforceIdkeydigest   bool                   `protobuf:"varint,13,opt,name=enforce_idkeydigest,json=enforceIdkeydigest,proto3" json:"enforce_idkeydigest,omitempty"`
 	ConformanceMode      bool                   `protobuf:"varint,14,opt,name=conformance_mode,json=conformanceMode,proto3" json:"conformance_mode,omitempty"`
 	KubernetesComponents []*KubernetesComponent `protobuf:"bytes,15,rep,name=kubernetes_components,json=kubernetesComponents,proto3" json:"kubernetes_components,omitempty"`
+	InitSecret           []byte                 `protobuf:"bytes,16,opt,name=initSecret,proto3" json:"initSecret,omitempty"`
```
datosh, line 43: We are `breakingWith` our `naming_convention` for protobuf fields.

cli/internal/terraform/terraform/aws/main.tf
```diff
@@ -34,6 +35,12 @@ resource "random_id" "uid" {
   byte_length = 4
 }
 
+resource "random_password" "initSecret" {
```
datosh, line 38: Since this password is generated in Terraform it will be stored in Terraform's statefile. Is this an attack vector we might be concerned about? I think we should mention this in our docs.
See Terraform docs regarding secrets in statefile: https://developer.hashicorp.com/terraform/language/state/sensitive-data

3u13r: I think we should mention this when we finally expose Terraform to the user. Currently the statefile is managed by us and not mentioned in the docs.

internal/cloud/aws/cloud.go
```diff
@@ -115,6 +109,11 @@ func (c *Cloud) UID(ctx context.Context) (string, error) {
 	return readInstanceTag(ctx, c.imds, cloud.TagUID)
 }
 
+// InitSecretHash returns the InitSecretHash of the current instance.
+func (c *Cloud) InitSecretHash(ctx context.Context) (string, error) {
```
daniel-weisse, line 113: This method should probably return `[]byte` directly

daniel-weisse (approved):

datosh (approved):

PR #659, opened 2022-11-28
Fix components diagram

m1ghtym0:
### Proposed change(s)

- Components has a dangling reference `D` that referenced the AccessManager before. It is currently [shown as `D` in the Diagram](https://docs.edgeless.systems/constellation/architecture/components)
- Fixes the diagram by updating the references


datosh (approved):
Good catch! Thanks for changing it 👍

PR #668, opened 2022-11-29
CSI driver fixes

daniel-weisse:
### Proposed change(s)

* Fix an issue where CSI resize operations were using a disks volume ID instead of the disk's UID as key ID
* Add a udev rule to label disks with empty usage label

### Additional info
- ~~We might want to split this PR, since I would like the CSI resize fixes in main before we can release a new CSI driver version~~ Dropped the CSI chart changes for now, will update the chart once we have a new release

### Checklist

- [ ] Link to Milestone


malt3 (approved):
LGTM

PR #669, opened 2022-11-29
Debugd: collect systemd logs to Opensearch

katexochen:
### Proposed change(s)

This change lays the fundamentals of collecting logs to Opensearch. You can activate the collection when deploying with cdbg:

```
./cdbg deploy --info logcollect=true,logcollect.admin=yourname
```

For QEMU, the credentials for Opensearch can be passed via the info as well:

```
./cdbg deploy --info logcollect=true,logcollect.admin=yourname,qemu.opensearch-pw=xxx
```


3u13r:
Works on QEMU, see logstash with `metadata.provider: QEMU`.

malt3:
Testing works great for me too! Was able to perform simple queries. Maybe we have to warn developers to increase default memory on QEMU. I had issues with the default (2GiB).

malt3 (approved):
LGTM

PR #673, opened 2022-11-29
Image update API

malt3:
### Proposed change(s)

- Implement CLI side of update API (only for os images for now
- Extend `constellation upgrade plan` to use update API
- Extend `constellation upgrade execute` to fetch image references from API
- Add compatibility checks for `upgrade plan`
  - Minor version of CLI cannot be older than os images
  - Only allow updates to current minor version in cluster or next minor version (if CLI is also on that minor version)
  - Only allow upgrades (downgrade protection)
- Add automatic release trigger to github actions to add constellation os image versions in the update API 

### Checklist

- [x] Link to Milestone


hack/add-version/main.go
```diff
@@ -0,0 +1,322 @@
...
+		return "minor", nil
+	case semver.Canonical(version) == version:
+		return "patch", nil
+	default:
+		return "", fmt.Errorf("invalid version %q", version)
+	}
+}
+
+func versionJSONPath(granularity, base string) string {
+	return path.Join("constellation/v1/updates", stream, granularity, base, imageKind+".json")
```
daniel-weisse, line 317: Add a `CDNUpdatesPath` constant for this?

daniel-weisse (approved):

PR #682, opened 2022-11-30
image: add version and debug field to lookup table

katexochen:

### Proposed change(s)

- Add version and debug info to image lookup-table.


malt3:
I'm running the pipeline on this change: https://github.com/edgelesssys/constellation/actions/runs/3585907905

I'll also test to see if the CLI still works (just a safeguard, I do expect this to work).

malt3:
I did encounter an issue: [Image lookup table is currently encoded as `map[string]map[string]string`](https://github.com/edgelesssys/constellation/blob/841eba9bf936fa1444c18100389741ec3b487b5c/internal/image/image.go#L45). This breaks if we add flat fields (version and debug). We could make the type slightly more complex to counter this (leading to different json format):

```go
type imageLookupTable map[string]map[string]string

type imageInfo struct {
    LUT imageLookupTable `json:"lookupTable"`
    Debug bool  `json:"debug"`
    Version string  `json:"version"`
}
```

Or we could make the CSPs concrete fields:

```go
type imageInfo struct {
    AWS map[string]string `json:"aws"`
    Azure map[string]string `json:"azure"`
    GCP map[string]string `json:"gcp"`
    QEMU map[string]string `json:"qemu"`
    Debug bool  `json:"debug"`
    Version string  `json:"version"`
}
```

Below is the issue I encountered:

```
./constellation create -c 1 -w 1 --name malte -y 
WARNING: Creating a debug cluster. This cluster is not secure and should only be used for debugging purposes.
DO NOT USE THIS CLUSTER IN PRODUCTION.

Creating  
Error: fetching image reference: decoding image reference: json: cannot unmarshal bool into Go value of type map[string]string
```

malt3:
In any case we should also extend [the RFC](https://github.com/edgelesssys/constellation/blob/main/rfc/image-discoverability.md) to include the new fields / new format.

malt3 (approved):
LGTM and works as expected! Thanks 🐳

PR #691, opened 2022-11-30
Feat/join service component configmap

3u13r:
### Proposed change(s)

- JoinService uses components config map to hand out URLs where to download the K8s components
- Bootstrapper uses components from JoinService if available
- JoinService fallback migration

For testing you can use the following `JoinImage`: `ghcr.io/edgelesssys/constellation/join-service:feat-joinServiceComponentConfigmap`

Upgradability: The Bootstrapper works with the current main JoinService as well as with the new one. If the Bootstrapper receives no components it uses the ones included in the bootstrapper. 
You can also just update the joinservice by itself. It will do a fallback migration (i.e. make the necessary configmap changes which are now done during the first bootstrapper init). Note that besides updating the image, you must also update the permissions of the clusterrole `join-service` to include permissions to `create` and `update` `configmaps` in the `""` apigroup. Of course, while it then responds with components, they are ignored by the old bootstrapper. 
### Checklist

- [x] Link to Milestone


joinservice/internal/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,85 @@
...
+	"k8s.io/client-go/rest"
+)
+
+// Client is a kubernetes client.
+type Client struct {
+	client *kubernetes.Clientset
+}
+
+// NewClient creates a new kubernetes client.
+func NewClient() *Client {
```
daniel-weisse, line 28:
```suggestion
// New creates a new kubernetes client.
func New() *Client {
```

joinservice/internal/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,85 @@
...
+	// creates the in-cluster config
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		panic(err.Error())
+	}
+	// creates the clientset
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		panic(err.Error())
+	}
```
daniel-weisse, line 38: Should not panic

joinservice/internal/server/server.go
```diff
@@ -106,12 +116,35 @@ func (s *Server) IssueJoinTicket(ctx context.Context, req *joinproto.IssueJoinTi
...
 
+	log.Infof("Querying K8sVersion ConfigMap for components ConfigMap name")
+	componentsConfigMapName, err := s.getK8sComponentsConfigMapName()
+	if errors.Is(err, fs.ErrNotExist) {
+		// If the file does not exist, the Constellation was initialized with an version before 2.3.0
+		// As a migration step, the join service will create the ConfigMap with the K8s components which
+		// match the K8s minor version of the cluster.
+		log.Warnf("Reference to K8sVersion ConfigMap does not exist, creating fallback Components ConfigMap and referencing it in K8sVersion ConfigMap")
+		log.Warnf("This is expected if the Constellation was initialized with an CLI before version 2.3.0")
+		log.Warnf("DEPRECATION WARNING: This is a migration step and will be removed in a future release")
```
daniel-weisse, line 133:
```suggestion
		// If the file does not exist, the Constellation was initialized with a version before 2.3.0
		// As a migration step, the join service will create the ConfigMap with the K8s components which
		// match the K8s minor version of the cluster.
		log.Warnf("Reference to K8sVersion ConfigMap does not exist, creating fallback Components ConfigMap and referencing it in K8sVersion ConfigMap")
		log.Warnf("This is expected if the Constellation was initialized with a CLI before version 2.3.0")
		log.Warnf("DEPRECATION WARNING: This is a migration step and will be removed in a future release")
```
https://owl.purdue.edu/owl/general_writing/grammar/articles_a_versus_an.html

daniel-weisse (approved):
Haven't run any in cluster tests, but code looks good to me

malt3 (approved):
I tested the feature. Upgrading the join service in-place works as expected. Code LGTM

PR #708, opened 2022-12-02
Add debug stream to versions API

katexochen:
### Proposed change(s)

- rename pkg updates => versionsapi
- allow debug stream
- enable add-version hack tool to add versions to different streams

### DIFF
--- a/hack/add-version/main.go
+++ b/hack/add-version/main.go
@@ -28,7 +28,7 @@ import (
 	s3types "github.com/aws/aws-sdk-go-v2/service/s3/types"
 	"github.com/edgelesssys/constellation/v2/internal/constants"
 	"github.com/edgelesssys/constellation/v2/internal/logger"
-	"github.com/edgelesssys/constellation/v2/internal/update"
+	"github.com/edgelesssys/constellation/v2/internal/versionsapi"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 	"golang.org/x/mod/semver"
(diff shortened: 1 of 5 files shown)


hack/add-version/main.go
```diff
@@ -203,120 +278,120 @@ func newVersionManager(ctx context.Context, region, bucket, distributionID strin
...
 }
 
-func (m *versionManager) addVersion(ctx context.Context, baseVersion, version string, list *update.VersionsList) error {
-	granularity, err := granularityFromVersion(baseVersion)
-	if err != nil {
-		return err
-	}
-	list.Versions = append(list.Versions, version)
+func (m *versionManager) updateVersionList(ctx context.Context, list *versionsapi.List) error {
 	sort.Strings(list.Versions)
```
malt3, line 304: Should be [semver sort](https://pkg.go.dev/golang.org/x/mod/semver#Sort). My bad :)

```suggestion
	semver.Sort(list.Versions)
```

hack/add-version/main.go
```diff
@@ -203,120 +278,120 @@ func newVersionManager(ctx context.Context, region, bucket, distributionID strin
...
 }
 
-func versionURL(granularity, base string) string {
-	return constants.CDNRepositoryURL + "/" + versionJSONPath(granularity, base)
+type granularity int
+
+const (
+	granularityMajor granularity = iota
+	granularityMinor
+	granularityPatch
```
malt3, line 383:
```suggestion
```

This is currently unused as we  only map from 

- major -> minor
- minor -> patch

So i would probably want to remove it. But this is really just a  nit.

malt3 (approved):
Tiny remarks. Otherwise LGTM

PR #709, opened 2022-12-02
images: use unified naming for all images

katexochen:
Discontinue prefixing image names with 'debug-' or 'branchname-'.


malt3:
We should also update the rfc (rfc/image-discoverability.md) and ensure we can still find debug images (hack/find-image/find-image.sh)

PR #711, opened 2022-12-05
AB#2685 Add documentation for automatic IAM creation

stdoutput:
### Proposed change(s)

- Add the docs for `constellation iam create <provider>`
- Remove the current bash scripts for automatic IAM setup using the specific provider CLI as creation via the script should be preferred.

### Checklist

- [ ] Link to Milestone


stdoutput:
Currently, formatting the hyperlinks of the cloud provider CLIs does not work properly. All other hyperlinks work. I thought this would be related to line breaks in the hyperlink, but that was not the case. If anyone is familiar with docusaurus and/or our docs specifically, feel free to check this.

docs/docs/getting-started/first-steps.md
```diff
@@ -40,45 +40,61 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
...
-    az role assignment create --assignee-principal-type ServicePrincipal --assignee-object-id "${identityID}" --role 'Virtual Machine Contributor' --scope "/subscriptions/${SUBSCRIPTION_ID}"
-    az role assignment create --assignee-principal-type ServicePrincipal --assignee-object-id "${identityID}" --role 'Application Insights Component Contributor' --scope "/subscriptions/${SUBSCRIPTION_ID}"
-    echo "subscription: ${SUBSCRIPTION_ID}
-    tenant: $(az account show --query tenantId -o tsv)
-    location: ${LOCATION}
-    resourceGroup: ${RESOURCE_GROUP}
-    userAssignedIdentity: $(az identity show -n "${SERVICE_PRINCIPAL_NAME}" -g "${RESOURCE_GROUP}-identity" --query id --out tsv)
-    appClientID: $(jq -r '.appId' azureServiceAccountKey.json)
-    clientSecretValue: $(jq -r '.password' azureServiceAccountKey.json)"
+    constellation iam create azure --region=westus --resourceGroup=constellTest --servicePrincipal=spTest
```
datosh, line 49: Even though the reference contains detailed information and is linked, I think we could state briefly what is expected of the user.

When reading the command for the first time, I asked myself the following question: Is `resourceGroup` a name for something that is to be generated, or a reference to something that already exists. The same for `servicePrincipal`.

docs/README.md
```diff
@@ -13,7 +13,7 @@ npm run build
 npm run serve
 ```
 
-Browse to <http://localhost:3000/constellation>
+Browse to <http://localhost:3000/constellation> and choose the "Next" version in the top right.
```
datosh, line 16: Thanks for improving this 🙏 ran into the same issue and forgot to extend this 🙂

docs/docs/getting-started/first-steps.md
```diff
@@ -40,45 +40,61 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
 2. Fill in your cloud provider specific information.
 
     <tabs groupId="csp">
```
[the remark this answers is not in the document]
thomasten: Yes, I'd also suggest to have the three new (and recommended) ways in sync with the tabs in step 1. So rename the value `azure-command` to `azure` and so on. You then need to rename the existing `aws` to something else.

docs/docs/getting-started/first-steps.md
thomasten: I wonder whether we should have a workflow page for IAM creation and link to this instead. But it might also be too redundant and superfluous. Opinions?

docs/docs/getting-started/first-steps.md
```diff
@@ -40,45 +40,61 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
...
-    location: ${LOCATION}
-    resourceGroup: ${RESOURCE_GROUP}
-    userAssignedIdentity: $(az identity show -n "${SERVICE_PRINCIPAL_NAME}" -g "${RESOURCE_GROUP}-identity" --query id --out tsv)
-    appClientID: $(jq -r '.appId' azureServiceAccountKey.json)
-    clientSecretValue: $(jq -r '.password' azureServiceAccountKey.json)"
+    constellation iam create azure --region=westus --resourceGroup=constellTest --servicePrincipal=spTest
     ```
 
-    Fill the values produced by the script into your configuration file.
+    Find an exact command description in our [CLI reference](https://docs.edgeless.systems/constellation/reference/cli).
```
thomasten, line 52:
```suggestion
    Find an exact command description in the [CLI reference](../reference/cli.md).
```

docs/docs/getting-started/first-steps.md
```diff
@@ -40,45 +40,61 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
...
 
-    :::tip
+    By default, Constellation uses `Standard_DC4as_v5` CVMs (4 vCPUs, 16 GB RAM) to create your cluster. Optionally, you can switch to a different VM type by modifying **instanceType** in the configuration file. For CVMs, any VM type with a minimum of 4 vCPUs from the [DCasv5 & DCadsv5](https://docs.microsoft.com/en-us/azure/virtual-machines/dcasv5-dcadsv5-series) or [ECasv5 & ECadsv5](https://docs.microsoft.com/en-us/azure/virtual-machines/ecasv5-ecadsv5-series) families is supported.
 
-    Alternatively, you can leave `clientSecretValue` empty and provide the secret via the `CONSTELL_AZURE_CLIENT_SECRET_VALUE` environment variable.
+    You can also run `constellation config instance-types` to get the list of all supported options.
 
-    :::
+    Paste the output into the corresponding fields of the `constellation-conf.yaml` file created in the previous step.
+    </tabItem>
```
thomasten, line 59: I think this sentence belongs just below the bash command

docs/docs/getting-started/first-steps.md
```diff
@@ -40,45 +40,61 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
...
+    Create the IAM configuration on AWS with the following command (You must be authenticated with the [AWS CLI](https://aws.amazon.com/en/cli/) in the shell session):
+
+    ```bash
+    constellation iam create aws --zone=eu-central-1a --prefix=constellTest
+    ```
 
-    Run `constellation config instance-types` to get the list of all supported options.
+    Find an exact command description in our [CLI reference](https://docs.edgeless.systems/constellation/reference/cli).
 
+    Please note that currently, Constellation OS images are currently replicated to the following regions:
```
thomasten, line 86:
```suggestion
    Constellation OS images are currently replicated to the following regions:
```

docs/docs/getting-started/first-steps.md
```diff
@@ -184,7 +176,7 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
       Supported are all machines from the N2D family. It defaults to `n2d-standard-4` (4 vCPUs, 16 GB RAM), but you can use any other VMs from the same family. Refer to [N2D machine series](https://cloud.google.com/compute/docs/general-purpose-machines#n2d_machines) or run `constellation config instance-types` to get the list of all supported options.
 
     </tabItem>
-    <tabItem value="aws" label="AWS">
+    <tabItem value="aws" label="AWS (Console)">
```
thomasten, line 179: This doesn't use the AWS Console. We should name it differently, maybe "(manually)".

thomasten (changes requested):

docs/docs/getting-started/first-steps.md
```diff
@@ -40,45 +40,61 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
...
-    Fill the values produced by the script into your configuration file.
+    Find an exact command description in our [CLI reference](https://docs.edgeless.systems/constellation/reference/cli).
 
-    :::tip
+    By default, Constellation uses `Standard_DC4as_v5` CVMs (4 vCPUs, 16 GB RAM) to create your cluster. Optionally, you can switch to a different VM type by modifying **instanceType** in the configuration file. For CVMs, any VM type with a minimum of 4 vCPUs from the [DCasv5 & DCadsv5](https://docs.microsoft.com/en-us/azure/virtual-machines/dcasv5-dcadsv5-series) or [ECasv5 & ECadsv5](https://docs.microsoft.com/en-us/azure/virtual-machines/ecasv5-ecadsv5-series) families is supported.
 
-    Alternatively, you can leave `clientSecretValue` empty and provide the secret via the `CONSTELL_AZURE_CLIENT_SECRET_VALUE` environment variable.
+    You can also run `constellation config instance-types` to get the list of all supported options.
 
-    :::
```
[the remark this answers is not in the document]
stdoutput: The IAM create command is not going to set this environment variable, should I rework the command to do that?

docs/docs/getting-started/first-steps.md
stdoutput: I would prefer to only briefly state the command here for users who want to set up really quickly here and moving detailed descriptions to a workflow page. There, we could also explain how the IAM resources can be re-used or deleted. Let me know your opinions.

docs/docs/getting-started/first-steps.md
stdoutput: I changed that. I also noticed we didn't mention `constellation config instance-types` in every CP's tab. Should we do so? On that, I also noticed we don't mention supported AWS machines anywhere. I think we should add them somewhere, right?

docs/docs/getting-started/first-steps.md
thomasten: > moving detailed descriptions to a workflow page. There, we could also explain how the IAM resources can be re-used or deleted

A workflow page seems a good solution if we actually have additional infos there like you propose 👍

docs/docs/getting-started/first-steps.md
thomasten: Yes, we should do both. AWS types are probably too much to list them all, but we should at least link to https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enable-nitrotpm-prerequisites.html

docs/docs/getting-started/first-steps.md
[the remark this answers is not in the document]
stdoutput: I think it sounds like something which would be set automatically, if we say "**the** `CONSTELL_AZURE_CLIENT_SECRET_VALUE`". 
Also i think we should explain why an environment variable is preferred here.

I would phrase it like "Since this is a sensitive value, alternatively you can leave `clientSecretValue` empty and provide the secret via an environment variable such as `CONSTELL_AZURE_CLIENT_SECRET_VALUE`."

Probably you can improve that, but generally I would provide this information to the user, since i personally didn't get the grasp of it at first when reading it myself.

docs/docs/getting-started/first-steps.md
stdoutput: Added that [here](https://github.com/edgelesssys/constellation/pull/711/commits/d71ccef4ffd52869affcaa61a24af3a80ee9d0f0). Could you verify it for technical correctness?

docs/docs/getting-started/first-steps.md
stdoutput: Should we also link to the workflow pages in the getting-started guide? Currently, i think we only do this for the verification section (which is removed in v2.3 anyways?) and cluster recovery. However, i think for a user it might be nice to get a link to detailed information about the step he is currently performing, such as IAM or cluster creation. What do you think?

docs/docs/getting-started/first-steps.md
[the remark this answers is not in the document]
thomasten: Stating the reason makes sense!

"an environment variable such as" would be wrong because it's about exactly this variable:
https://github.com/edgelesssys/constellation/blob/be01cf7129d6f5ebcc87244aec8f3a94dc6c2d89/internal/constants/constants.go#L133-L135

Feel free to rephrase if you can make this more clear.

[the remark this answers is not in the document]
stdoutput: Got it. Makes sense.

"Since this is a sensitive value, alternatively you can set the `CONSTELL_AZURE_CLIENT_SECRET_VALUE` environmental variable, leave `clientSecretValue` empty in the `constellation-conf.yml` and provide the secret via the created environment variable."

Does that sound good?

docs/docs/getting-started/first-steps.md
datosh: If we have a dedicated workflow page for IAM it is probably better to link that than the CLI reference.

I think links to workflows are helpful if they do not interrupt the flow of the guide. I checked our quick start for this:

I like: "This will allow you to [recover your cluster](https://constellation-docs.edgeless.systems/constellation/workflows/recovery) in case of a disaster."

Less so: "For details, see the [verification section](https://constellation-docs.edgeless.systems/constellation/workflows/verify-cluster)."

thomasten: completely agree with Fabian

docs/docs/getting-started/first-steps.md
thomasten: That's fine 👍

docs/docs/getting-started/first-steps.md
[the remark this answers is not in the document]
thomasten: IMO still sounds a bit confusing. Maybe this is clearer (though a bit verbose)?

"Since this is a sensitive value, alternatively you can leave `clientSecretValue` empty in the configuration file and pass it via an environment variable instead. To this end, create the environment variable `CONSTELL_AZURE_CLIENT_SECRET_VALUE` and set it to the secret value."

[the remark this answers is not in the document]
stdoutput: Sounds good!

stdoutput:
Not yet 100% happy with the current structure & phrasing, but couldn't really find out what bothers me. @thomasten @datosh can you read through it from a users perspective and let me know what you think?

docs/docs/workflows/iam.md
```diff
@@ -0,0 +1,210 @@
+# IAM Configuration
```
datosh, line 1: The workflow title and the content do not fit super well.

The page talks a lot about IAM, but it also explains how to generate resources and set values to the config that are not generated by `iam create`.

I think we could do two things:
1. Move all non-IAM related settings to a dedicated workflow
2. Rename this section to something more general "Cluster config and IAM"

I lean more towards 2. to keep it simple, although 1. has its merits if we think about a different persona (admin with extended permissions) creating IAM resources and another persona (DevOps engineer) creating the actual cluster.

docs/docs/workflows/iam.md
```diff
@@ -0,0 +1,210 @@
...
+If you require the OS image to be available in another region, [let us know](https://github.com/edgelesssys/constellation/issues/new?assignees=&labels=&template=feature_request.md&title=Support+new+AWS+image+region:+xx-xxxx-x).
+
+You can find a list of all [regions in AWS's documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions).
+
+By default, Constellation uses `m6a.xlarge` VMs (4 vCPUs, 16 GB RAM) to create your cluster. Optionally, you can switch to a different VM type by modifying **instanceType** in the configuration file. Supported are all nitroTPM-enabled machines. Refer to the [list of nitroTPM-enabled instance types](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enable-nitrotpm-prerequisites.html) or run `constellation config instance-types` to get the list of all supported options.
+
+</tabItem>
+</tabs>
+
+Alternatively, you can manually create the IAM configuration on your CSP.
```
datosh, line 77: We could also merge the manual boxes with the automated boxes to achieve a better "flow" in the docs. Now I need to notice that this is not interesting for me, and skip some amount of content. 

Another option would be to have something like: "If you created these resources automatically continue with <link to create cluster>" as we have further below.

docs/docs/getting-started/first-steps.md
```diff
@@ -40,181 +40,51 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
...
     ```
 
-    Fill the values produced by the script into your configuration file.
-
-    :::tip
-
-    Alternatively, you can leave `clientSecretValue` empty and provide the secret via the `CONSTELL_AZURE_CLIENT_SECRET_VALUE` environment variable.
-
+    :::info
+    This command creates IAM configuration on the Azure region `westus` using a new resource group `constellTest` and a new service principal `spTest`.
```
datosh, line 51: using would imply to me that I am responsible for creating the SP so that `iam create` can use it.
```suggestion
    This command creates IAM configuration on the Azure region `westus` creating a new resource group `constellTest` and a new service principal `spTest`.
```

docs/docs/getting-started/first-steps.md
```diff
@@ -40,181 +40,51 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
...
-    echo "project: ${PROJECT_ID}
-    serviceAccountKeyPath: $(realpath gcpServiceAccountKey.json)"
+    constellation iam create gcp --projectID=yourproject-12345 --zone=europe-west1-a --serviceAccountID=constell-test
     ```
 
-    Fill the values produced by the script into your configuration file.
-
-    By default, Constellation uses `n2d-standard-4` VMs (4 vCPUs, 16 GB RAM) to create your cluster. Optionally, you can switch to a different VM type by modifying **instanceType** in the configuration file. Supported are all machines from the N2D family. Refer to [N2D machine series](https://cloud.google.com/compute/docs/general-purpose-machines#n2d_machines) or run `constellation config instance-types` to get the list of all supported options.
+    :::info
+    This command creates IAM configuration in the GCP project `yourproject-12345` on the GCP zone `europe-west1-a` using a new service account `constell-test`.
```
datosh, line 63:
```suggestion
    This command creates IAM configuration in the GCP project `yourproject-12345` on the GCP zone `europe-west1-a` creating a new service account `constell-test`.
```

docs/docs/reference/cli.md
```diff
@@ -27,6 +27,11 @@ Commands:
 * [recover](#constellation-recover): Recover a completely stopped Constellation cluster
 * [terminate](#constellation-terminate): Terminate a Constellation cluster
 * [version](#constellation-version): Display version of this CLI
+* [iam](#constellation-iam): Work with the IAM configuration on your cloud provider
```
datosh, line 30: #743 also updated the docs reference

datosh (commented):
Overall I like the proposed structure!

I think we need to be deliberate in finding good section titles to guide the user on their journey.

docs/docs/workflows/iam.md
stdoutput: I also thought about a "drawer" which you could open up and close, but didn't find a suitable option in Docusaurus. I think the manual creation boxes might be off-putting to a user because of their size. Ideally I would "hide" them or give the user a big hint to skip them if they used the command (which is recommended either way).

docs/docs/workflows/iam.md
stdoutput: I also prefer option 2, as this generally refers to the same step in the setup (filling the `constellation-conf.yml`)

I think a DevOps engineer, setting up the cluster, will know if his access rights are insufficient for creating the IAM so he can forward this to an Administrator, on the other side and Administrator will know if the DevOps engineer can choose the machine types, etc.

Maybe this is my naive perspective - let me know what you think

datosh: Makes sense, let's go with 2!

docs/docs/workflows/iam.md
stdoutput: That seems perfect! Don't know why i didn't find it before

docs/docs/workflows/config.md
```diff
@@ -0,0 +1,228 @@
+# Configure your cluster
+
+Before your cluster can be created, you need to configure the identity and access management (IAM) for your CSP and choose machine types for your cluster.
```
thomasten, line 3:
```suggestion
Before you can create your cluster, you need to configure the identity and access management (IAM) for your cloud service provider (CSP) and choose machine types for the nodes.
```
prefer active voice
spell out acronyms on first use on a page (except for those that are more well known than the spelled-out term)

docs/docs/workflows/config.md
```diff
@@ -0,0 +1,228 @@
...
+
+## Choosing VM types
+
+Constellation supports the following VM types:
+<tabs groupID="csp">
```
thomasten, line 8:
```suggestion
<tabs groupId="csp">
```
Docusaurus seems to be case sensitive with this. At least this change fixes the tabs not being in sync.

docs/docs/workflows/config.md
```diff
@@ -0,0 +1,228 @@
...
+constellation iam create azure --region=westus --resourceGroup=constellTest --servicePrincipal=spTest
+```
+
+:::info
+This command creates IAM configuration on the Azure region `westus` creating a new resource group `constellTest` and a new service principal `spTest`.
+:::
+
+Note that CVMs are currently only supported in a few regions, check [Azure's products available by region](https://azure.microsoft.com/en-us/global-infrastructure/services/?products=virtual-machines&regions=all). These are:
+  * `westus`
+  * `eastus`
```
thomasten, line 48: IMO the regions are essential enough to also mention them in the *first steps*. We should prevent users from making a mistake there.

docs/docs/workflows/config.md
```diff
@@ -0,0 +1,228 @@
...
+:::info
+This command creates IAM configuration for the AWS zone `eu-central-1a` using the prefix `constellTest` for all named resources being created.
+:::
+
+Paste the output into the corresponding fields of the `constellation-conf.yaml` file.
+
+Constellation OS images are currently replicated to the following regions:
+    * `eu-central-1`
+    * `us-east-2`
+    * `ap-south-1`
```
thomasten, line 93:
```suggestion
* `eu-central-1`
* `us-east-2`
* `ap-south-1`
```
otherwise it's not rendered correctly

docs/docs/workflows/config.md
```diff
@@ -0,0 +1,228 @@
...
+
+If you require the OS image to be available in another region, [let us know](https://github.com/edgelesssys/constellation/issues/new?assignees=&labels=&template=feature_request.md&title=Support+new+AWS+image+region:+xx-xxxx-x).
+
+You can find a list of all [regions in AWS's documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions).
+
+</tabItem>
+</tabs>
+
+<details>
+<summary>Alternatively, you can manually create the IAM configuration on your CSP.</summary>
```
thomasten, line 103:
```suggestion
<summary>Alternatively, you can manually create the IAM configuration on your CSP.</summary>
The following describes the configuration fields and how you obtain the required information or create the required resources.
```
This introductory sentence shall make clear why the following is a list of config field names.

docs/docs/workflows/config.md
```diff
@@ -0,0 +1,228 @@
...
+<tabs groupId="csp">
+<tabItem value="azure" label="Azure (manually)">
```
thomasten, line 105:
```suggestion
<tabItem value="azure" label="Azure">
```
as this is now separated, we can omit it

docs/docs/workflows/config.md
```diff
@@ -0,0 +1,228 @@
...
+</tabs>
+</details>
+
+Now that you've configured your CSP, you can [create your cluster](./create.md).
+
+## Deleting an IAM configuration
+
+Created IAM configurations can be kept and reused for new clusters. Alternatively, you can also delete them if they aren't being used anymore.
+
+### Prerequisites
```
thomasten, line 217:
```suggestion
**Prerequisites:**
```
otherwise it creates an entry in the sidebar on the right and then we would need another heding for the actual commands

docs/docs/workflows/config.md
```diff
@@ -0,0 +1,228 @@
...
+* [HashiCorp's Terraform](https://developer.hashicorp.com/terraform/downloads?product_intent=terraform) is installed on your machine.
```
thomasten, line 218:
```suggestion
* [Terraform](https://developer.hashicorp.com/terraform/downloads) is installed on your machine.
```
nit/optional: IMO the product name is descriptive enough

docs/docs/workflows/config.md
```diff
@@ -0,0 +1,228 @@
...
+* Access to the `terraform.tfstate` file created by the `constellation iam create` command.
+
+You can delete the IAM configuration using the following commands:
+```bash
+# Navigate to the directory containing the terraform.tfstate file
+cd constellation-iam-terraform
+# Destroy the IAM configuration via Terraform
+terraform destroy
+# Confirm deletion via typing "yes"
```
thomasten, line 227:
```suggestion
# Confirm deletion by typing "yes"
```

thomasten (changes requested):
Great, I like the restructuring!

docs/docs/getting-started/first-steps.md
```diff
@@ -40,181 +40,55 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
...
-    Fill the values produced by the script into your configuration file.
+    :::info
+    This command creates IAM configuration on the Azure region `westus` creating a new resource group `constellTest` and a new service principal `spTest`.
+    :::
 
     :::tip
-
-    Alternatively, you can leave `clientSecretValue` empty and provide the secret via the `CONSTELL_AZURE_CLIENT_SECRET_VALUE` environment variable.
-
+    Since `clientSecretValue` is a sensitive value, you can leave it empty in the configuration file and pass it via an environment variable instead. To this end, create the environment variable `CONSTELL_AZURE_CLIENT_SECRET_VALUE` and set it to the secret value.
```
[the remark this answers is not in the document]
stdoutput: You mean to duplicate "Fill in your cloud provider specific information" into all of the tabs?

docs/docs/workflows/config.md
```diff
@@ -0,0 +1,228 @@
+# Configure your cluster
```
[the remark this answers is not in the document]
stdoutput: You definitely have a point there. The only thing that seems counter-intuitive to me with that approach is the path for a user setting up his cluster. He would go to create -> config -> create. Should we maybe add the config-generation step to the *config* workflow page and add a box to the create-step which indicates that before creating a cluster, a filled config file is needed (while pointing to the *config*-page from there)? Imo, grouping these 2 would make more sense.

docs/docs/getting-started/first-steps.md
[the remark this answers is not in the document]
thomasten: Yes. Or alternatively, maybe remove the tip and have it only in the workflow? Or something else?

docs/docs/workflows/config.md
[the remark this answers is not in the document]
thomasten: sounds good!

docs/docs/getting-started/first-steps.md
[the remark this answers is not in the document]
stdoutput: I think removing it here is the way to go. I was unsure because it was a tip with security-impact, but i assume people who take enough care to not put the bare secret into the config file will also read the workflow page

docs/docs/getting-started/first-steps.md
```diff
@@ -39,182 +39,48 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
 
 2. Fill in your cloud provider specific information.
 
+    First you need to create an [IAM configuration](../workflows/config.md#creating-an-iam-configuration). The easiest way to do this is the following CLI command:
+
     <tabs groupId="csp">
-    <tabItem value="azure" label="Azure (CLI)">
 
-    You need several resources for the cluster. You can use the following `az` script to create them:
```
thomasten, line 45: There's a tip at the end of this page regarding this az script, which doesn't make sense anymore and should be removed.

docs/docs/workflows/config.md
```diff
@@ -0,0 +1,261 @@
...
+
+You must be authenticated with the [Azure CLI](https://learn.microsoft.com/en-us/cli/azure/install-azure-cli) in the shell session.
+
+```bash
+constellation iam create azure --region=westus --resourceGroup=constellTest --servicePrincipal=spTest
+```
+
+:::info
+This command creates IAM configuration on the Azure region `westus` creating a new resource group `constellTest` and a new service principal `spTest`.
+:::
```
thomasten, line 74: we should "unbox" the three infos on this page, too

thomasten (commented):
Almost there ;)

Also take a look at the 3 hidden comments from my previous review again.

thomasten (approved):
I was thinking about having the region notices on both pages. But I leave it up to you.

stdoutput:
> I was thinking about having the region notices on both pages. But I leave it up to you.

Makes sense!

PR #713, opened 2022-12-05
AB#2651 Compatibility warning for MiniConstellation

stdoutput:
### Proposed change(s)

- Adding a `warning` box to the documentation indicating the importance of meeting the requirements for running a MiniConstellation.

### Checklist

- [ ] Link to Milestone


docs/docs/getting-started/first-steps-local.md
```diff
@@ -4,19 +4,25 @@
...
 <!-- vale on -->
 
-MiniConstellation uses virtualization to create a local cluster with one control-plane node and one worker node. 
+MiniConstellation uses virtualization to create a local cluster with one control-plane node and one worker node.
 
 :::info
 
-MiniConstellation **doesn't** require hardware with Confidential VM (CVM) support. 
+MiniConstellation **doesn't** require hardware with Confidential VM (CVM) support.
 For attestation, MiniConstellation currently uses the software-based vTPM provided by KVM/QEMU.
```
thomasten, line 12: Now there are three boxes in a row. I think we can "unbox" the first one and append it to the paragraph above (replacing *MiniConstellation* with *it*). What do you think?

docs/docs/getting-started/first-steps-local.md
```diff
@@ -4,19 +4,25 @@
...
 
 :::
 
+:::warning
```
thomasten, line 16: docusaurus doesn't know *warning*. Choose one of these https://docusaurus.io/docs/markdown-features/admonitions

docs/docs/getting-started/first-steps-local.md
```diff
@@ -4,19 +4,25 @@
...
+
+MiniConstellation has specific soft- and hardware requirements such as a linux OS running on an x86-64 CPU. Please pay close attention to all of the [Prerequisites](#prerequisites) when setting up.
```
thomasten, line 18:
```suggestion
MiniConstellation has specific soft- and hardware requirements such as a Linux OS running on an x86-64 CPU. Pay attention to all [prerequisites](#prerequisites) when setting up.
```
[prune every excess word](https://learn.microsoft.com/en-us/style-guide/top-10-tips-style-voice#be-brief)

docs/docs/getting-started/first-steps-local.md
```diff
@@ -4,19 +4,25 @@
 With the `constellation mini` command, you can deploy and test Constellation locally without a cloud subscription. This mode is called MiniConstellation. Conceptually, MiniConstellation is similar to [MicroK8s](https://microk8s.io/), [K3s](https://k3s.io/), and [minikube](https://minikube.sigs.k8s.io/docs/).
```
thomasten, line 4: regarding
https://github.com/edgelesssys/constellation/blob/c52dfb09b903fd1b9e241580c7b0f2e57357b2e3/cli/internal/cmd/miniup.go#L101-L102
how about extending the message such that users get to know the platform which works without needing to consult the docs?

thomasten (changes requested):

docs/docs/getting-started/first-steps-local.md
stdoutput: I also think this box can be merged into the top paragraph. However I think we should not replace the current top paragraph with it completely, as we would then not inform the user about the virtualization aspect. My suggestion would be something like:
```
MiniConstellation uses virtualization to create a local cluster with one control-plane node and one worker node. It **does not** require hardware with Confidential VM support. For Attestation, it uses software-based vTPM provided by KVM/QEMU.
```

docs/docs/getting-started/first-steps-local.md
stdoutput: Should we directly refer to the linux/amd64 platform here? Is a change of the supported platforms in the near future likely?

docs/docs/getting-started/first-steps-local.md
thomasten: Exactly what I had in mind 👍

But note that
* the linter won't allow you to change *doesn't* in *does not*
* it should be "For **a**ttestation, it uses **a** software-based"

docs/docs/getting-started/first-steps-local.md
thomasten: Not sure. @daniel-weisse do you have an opinion on this?

daniel-weisse: Change in support is unlikely, since the miniConstellation requires libvirt with QEMU/KVM, which is not available on Windows, and has poor performance and does not support the features we need on OSX

datosh (approved):
LGTM

docs/docs/getting-started/first-steps-local.md
```diff
@@ -4,19 +4,25 @@
...
 
-MiniConstellation **doesn't** require hardware with Confidential VM (CVM) support. 
+MiniConstellation **doesn't** require hardware with Confidential VM (CVM) support.
 For attestation, MiniConstellation currently uses the software-based vTPM provided by KVM/QEMU.
 
 :::
 
+:::caution
+
+MiniConstellation has specific soft- and hardware requirements such as a Linux OS running on an x86-64 CPU. Pay attention to all [prerequisites](#prerequisites) when setting up.
```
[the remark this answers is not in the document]
thomasten: IMO this is not a reference to a heading (like "see Prerequisites"), but a normal sentence where a link has been inserted. Just commenting on this because I suggested to change it the other way round earlier. Not a too strong opinion though in case you both disagree.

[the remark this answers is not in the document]
datosh: Ah I didn't see that comment of yours. I also don't have strong opnions here. 

Your argumentation makes perfect sense to me.

docs/docs/getting-started/first-steps-local.md
stdoutput: https://github.com/edgelesssys/constellation/blob/46a13eb4f76af7fcbeaa19685ad08184ec3e3571/cli/internal/cmd/miniup.go#L101-L103

I did it like that now, what do you think? Was kind of unsure on how to phrase it

thomasten (approved):

daniel-weisse (approved):

PR #714, opened 2022-12-05
docs: disable e2e badges

3u13r:
### Proposed change(s)

- temporarily disable e2e badges in the README.md 

Since the E2E test are flaky, which is at least partly due to external factors, we should remove them for now.


thomasten:
I'd remove them permanently. This will happen every now and again and I don't think we should switch them on and off all the time.

PR #716, opened 2022-12-05
rfc: use hash annotation during upgrades

3u13r:
### Proposed change(s)

- labels do not work use annotations instead

+ minor updates and corrections


malt3:
As discussed: Needs a little rework to fit in with operator pattern

malt3 (approved):
LGTM

PR #717, opened 2022-12-05
Full sonobuoy logs

derpsteb:
### Proposed change(s)

- Upload full sonobuoy logs (includes pod logs) until we have improved logs via OpenSearch. These changes can be reverted once that feature lands.
- Also upload logs with csp name included in artifact name so that there is one artifact per CSP.

### DIFF
--- a/.github/actions/e2e_sonobuoy/action.yml
+++ b/.github/actions/e2e_sonobuoy/action.yml
@@ -2,6 +2,9 @@ name: sonobuoy
 description: "Execute the e2e test framework sonobuoy."
 
 inputs:
+  cloudProvider:
+    description: "CSP name necessary for artifact naming."
+    required: true
   sonobuoyVersion:
     description: "Version of sonobuoy test CLI to use."
     default: "0.56.12"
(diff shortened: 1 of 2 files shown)


katexochen:
Do you have a test run for this change?

.github/actions/e2e_sonobuoy/action.yml
```diff
@@ -2,6 +2,9 @@ name: sonobuoy
 description: "Execute the e2e test framework sonobuoy."
 
 inputs:
+  cloudProvider:
+    description: "Necessary for log naming."
```
malt3, line 6: nit: I would add the subject to the description. But maybe that's subjective so feel free to ignore.

malt3 (commented):
Changeset looks fine. I would also like to see a test run.

malt3:
> Now we do: https://github.com/edgelesssys/constellation/actions/runs/3620680468

The logs seem to be encoded twice (tar.gz from sonobuoy, zip from gh actions). That's not ideal but fine for me as we want to move towards centralized log collection anyways.

Issue #631, opened 2022-11-23
Microservice upgrades via upgrade command


m1ghtym0:
Constellation microservices should be upgradable in an existing cluster using the `upgrade` command.

Predecessor: #589 

- [x] Move/Refactor bootstrapper helm client to internal
- [x] Implement Helm chart upgrade for CLI
- [x] Ensure CLI upgrade does not overwrite newer releases
- [x] Document helm upgrade cli cmd

PR #729, opened 2022-12-06
MVP for microservice upgrades

derpsteb:
### Proposed change(s)

Introduce functionality to upgrade microservices in a running cluster as per [update.rfc](https://github.com/edgelesssys/constellation/blob/main/rfc/updates.md). Currently an upgrade can be triggered by running `constellation upgrade execute --helm`. This is only temporary until we restructure the upgrade command according to the RFC. I am also open to change the temporary solution. 

`--keepValues` is disabled for cert-manager and cilium as their docs explicitly warn against using the flag. Reason being that newly introduced values are not picked up when using `--keepValues`. Therefore I went with the recommended way of fetching the existing values from the cluster and merging them with the loaded values from disk. `MergeMaps` will prioritize the values from the second argument, see the respective unit test. Calls to `MergeMaps` should always have the cluster values in the second argument.

Helm does not support updating CRDs as outlined [here](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/). There is some discussion on the topic. In the end we will have to decide how to handle CRDs on a case by case basis. We must take extra care of updating CRDs when Constellation users could rely on the installed CRDs.
CRDs updates are managed like the following:
- cert-manager: the upstream chart uses helm's templates folder to manager them. They intentionally side-step the `crds` folder in order to have full control of the CRDs. We follow that approach. Constellation users will have to take care that their cert-manager installation is on the same version as Constellation's. This will be described in the docs (ticket is created).
- cilium: cilium lets it's operator handle CRD installation. Therefore we don't need to worry about those. The cilium upgrade docs also don't mention a need to handle CRDs manually.
- constellation-operators: during installation helm will automatically install the CRDs included in the `crds` folder. CRD updates have to be done manually. The current implementation is effectively running `kubectl apply -f ./crds/` on the respective folders. 
- constellation-services: no CRDs included.

We may want to include an upgrade rollback command, but won't implement it now. If helm declares an update as successful, but release does not work as expected the docs will help by suggesting relevant helm-rollback commands.

### Related issue
- #631 

### Checklist

- [ ] Link to Milestone


cli/internal/helm/README.md
```diff
@@ -0,0 +1,17 @@
+# cilium
+- Collect values using `helm get values`, merge with values from chart and apply. Needs to be done to pick up any newly introduced values.
+- `--reuse-values` is turned off.
```
daniel-weisse, line 3: Could you add some context to this readme?
I imagine this to be a bit confusing if not read in the context of this PR

cli/internal/helm/client.go
```diff
@@ -7,28 +7,50 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+const (
+	// timeout is the maximum time given to the helm client.
+	upgradeTimeout = 1 * time.Minute
```
daniel-weisse, line 33: I think a larger timeout could be beneficial here.
Especially when considering upgrades to a large cluster, 1 minute may not be enough.

Maybe we should consider adding `timeout` flags to commands that perform potentially long running operations (Topic for next dev sync?)

cli/internal/helm/client.go
```diff
@@ -53,3 +75,163 @@ func (c *Client) CurrentVersion(release string) (string, error) {
...
+		return fmt.Errorf("loading operators: %w", err)
+	}
+	conServicesChart, err := loadChartsDir(helmFS, conServicesPath)
+	if err != nil {
+		return fmt.Errorf("loading constellation-services chart: %w", err)
+	}
+
+	// Prevent half-finished upgrades
+	// action.Lock.Lock()
+	// defer action.Lock.Unlock()
```
daniel-weisse, line 108: Is this block still relevant?

cli/internal/helm/client.go
```diff
@@ -53,3 +75,163 @@ func (c *Client) CurrentVersion(release string) (string, error) {
...
+	}
+	tagVals[config.GetProvider().String()] = true
+	if _, err := action.RunWithContext(ctx, conOperatorsReleaseName, conOperatorChart, conOperatorChart.Values); err != nil {
+		return fmt.Errorf("upgrading services: %w", err)
+	}
+
+	tagVals, ok = conServicesChart.Values["tags"].(map[string]any)
+	if !ok {
+		return err
+	}
```
daniel-weisse, line 145: Same here

cli/internal/helm/client.go
```diff
@@ -53,3 +75,163 @@ func (c *Client) CurrentVersion(release string) (string, error) {
...
+
+	return nil, errors.New("parsed []byte, but did not find a CRD")
+}
+
+func (c *Client) updateOperatorCRDs(ctx context.Context, conOperatorChart *chart.Chart) error {
+	for _, dep := range conOperatorChart.Dependencies() {
+		for _, crdFile := range dep.Files {
+			if strings.HasPrefix(crdFile.Name, "crds/") {
+				log.Printf("updating crd: %s", crdFile.Name)
```
daniel-weisse, line 228:
```suggestion
```
Or replace with CLI debug logging

cli/internal/helm/client_test.go
```diff
@@ -0,0 +1,21 @@
+package helm
+
+import (
+	"fmt"
+	"io/ioutil"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/tj/assert"
```
daniel-weisse, line 9:
```suggestion
	"github.com/stretchr/testify/require"
	"github.com/stretchr/testify/assert"
```
I really wish there was some way to black list certain repositories...
For some reason VS Code likes to add tj/assert instead of stretchr/testitfy/assert....

internal/deploy/helm/helm_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+package helm
+
+import (
+	"testing"
+
+	"github.com/tj/assert"
```
daniel-weisse, line 12:
```suggestion
	"github.com/stretchr/testify/assert"
```

daniel-weisse (commented):
I think the place for helm upgrades is fine for now.
The whole upgrade structure _should_ be implemented with v2.4.
If we don't manage to get it done until then we can always hide/remove the flag for the release or think of a more user friendly solution

cli/internal/helm/client.go
derpsteb: From looking at the helm code I was under the impression that the timeout is per upgrade action. So cilium, cert-manager, operators and services each have their own 1min timeout. 

Increasing to 3min (?).
Added to dev sync agenda.

cli/internal/helm/client.go
```diff
@@ -7,28 +7,50 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+	crdClient *apiextensionsclient.Clientset
 }
 
 // NewClient returns a new initializes client for the namespace Client.
-func NewClient(kubeConfigPath, helmNamespace string) (*Client, error) {
+func NewClient(kubeConfigPath, helmNamespace string, client *apiextensionsclient.Clientset) (*Client, error) {
 	settings := cli.New()
 	settings.KubeConfig = kubeConfigPath // constants.AdminConfFilename
 
+	// TODO: Replace log.Printf with actual CLILogger during refactoring of upgrade cmd.
```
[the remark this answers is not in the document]
derpsteb: Changed to CLI logger.

derpsteb:
This is currently missing automatic creation of CRD and CR backups. However, to keep PRs smaller I would merge this without backups. 

Only merge after 2.3 release.

derpsteb:
This is currently missing automatic creation of CRD and CR backups. However, to keep PRs smaller I would merge this without backups. 

Only merge after 2.3 release.

cli/internal/cloudcmd/upgrade.go
```diff
@@ -232,4 +245,13 @@ func (u *stableClient) kubernetesVersion() (string, error) {
 
 type helmInterface interface {
 	CurrentVersion(release string) (string, error)
+	Upgrade(ctx context.Context, config *config.Config, timeout time.Duration) error
+}
+
+// Using the type from cli/internal/cmd would introduce the following import cycle:
+// cli/internal/cmd (upgradeexecute.go) -> cli/internal/cloudcmd (upgrade.go) ->
+// -> cli/internal/cloudcmd/helm (client.go) -> cli/internal/cmd (log.go).
```
daniel-weisse, line 253:
```suggestion
```
Defining the interface where it is needed is the correct approach.
We also explicitly define the interface in `cli/internal/log.go` so we can return a type that only allows calling `Debugf` and `Sync`

cli/internal/cmd/upgradeexecute.go
```diff
@@ -27,13 +29,23 @@ func newUpgradeExecuteCmd() *cobra.Command {
 		RunE:  runUpgradeExecute,
 	}
 
+	// TODO: remove before merging to main. For testing only.
+	cmd.Flags().Bool("helm", false, "DEV ONLY: execute helm upgrade")
+	cmd.Flags().Duration("timeout", 3*time.Minute, "DEV ONLY: change upgrade timeout. Might be useful for slow connections or big clusters.")
```
daniel-weisse, line 34: I think we want to keep the flags for now? 
Not sure if the rest of the upgrade operations will be done before v2.4, so we should merge with the flags available.
Maybe remove the `DEV ONLY` and expand on the flag help text.
We could also mark the flags as hidden until we decide on a more permanent solution

cli/internal/helm/client.go
```diff
@@ -7,28 +7,45 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+	log       debugLog
 }
 
 // NewClient returns a new initializes client for the namespace Client.
-func NewClient(kubeConfigPath, helmNamespace string) (*Client, error) {
+func NewClient(kubeConfigPath, helmNamespace string, client *apiextensionsclient.Clientset, log debugLog) (*Client, error) {
 	settings := cli.New()
 	settings.KubeConfig = kubeConfigPath // constants.AdminConfFilename
 
+	// TODO: Replace log.Printf with actual CLILogger during refactoring of upgrade cmd.
```
daniel-weisse, line 42:
```suggestion
```

cli/internal/helm/client.go
```diff
@@ -53,3 +70,156 @@ func (c *Client) CurrentVersion(release string) (string, error) {
...
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// Using the type from cli/internal/cmd would introduce the following import cycle:
+// cli/internal/cmd (upgradeexecute.go) -> cli/internal/cloudcmd (upgrade.go) ->
+// -> cli/internal/cloudcmd/helm (client.go) -> cli/internal/cmd (log.go).
```
daniel-weisse, line 221:
```suggestion
```

cli/internal/helm/client.go
derpsteb: Added timeout flag, increased default timeout to 3min.

daniel-weisse (approved):

PR #737, opened 2022-12-06
docs: update dependencies

katexochen:

### Proposed change(s)

- I'm not an expert in node. I tested this and it seems to work. Not sure if there are any downsides. Also unsure why renovate isn't suggesting this change but keeps opening PRs that break everything, like #681.


m1ghtym0 (approved):
Tested it as well: https://deploy-preview-86--edgeless-docs.netlify.app/constellation
LGTM

PR #738, opened 2022-12-06
Annotate nodes with their Kubernetes components hash

3u13r:
### Proposed change(s)

- The first bootstrapper labels itself
- The joinservice adds each node as JoiningNode CRD with node name and components hash once it issues a join ticket
- Node Operator consumes the JoiningNode CRD and attaches the label once it sees the node in K8s

Updatability:

- Since the JoinService must create a JoiningNode CRD this must be registered before.
- Operator can be updated independently
- Also note the permission changes needed for the Operator and JoinService

### Checklist

- [x] Link to Milestone


operators/constellation-node-operator/controllers/joiningnode_controller.go
```diff
@@ -0,0 +1,127 @@
...
+		logr.Error(err, "unable to delete JoiningNode")
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *JoiningNodesReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// index pending nodes by nodename.
```
malt3, line 97:
```suggestion
	// index joining nodes by nodename.
```

joinservice/internal/kubernetes/kubernetes.go
```diff
@@ -70,6 +80,30 @@ func (c *Client) CreateConfigMap(ctx context.Context, configMap corev1.ConfigMap
...
+		},
+	})
+
+	_, err := c.dynClient.Resource(joiningNodeResource).Create(ctx, joiningNode, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to create joining node: %w", err)
+	}
+	return nil
+}
+
```
malt3, line 106: I think the logic here should be adapted to update the resource if it already exists to ensure we don't get stuck here (could happen if we accidentally issue tickets twice).

malt3 (commented):
Some minor changes are needed

malt3 (approved):
LGTM

derpsteb:
[GCP](https://github.com/edgelesssys/constellation/actions/runs/3646068397) testrun with 2+3 nodes.
[AWS](https://github.com/edgelesssys/constellation/actions/runs/3646075658) testrun with 2+3 nodes.

derpsteb (approved):

PR #745, opened 2022-12-07
AB#2544 Add upgrade agent for automatic version updates

stdoutput:
### Proposed change(s)

- Add an upgrade agent for updating Kubernetes versions while providing a strict interface to keep minimal access into the machine.

### Additional info
- Upgrade agent should only be deployed to the control plane nodes.
- Server is being exposed via a [UDS](https://www.google.com/search?q=unix+domain+socket) which limits it to only be accessible locally.

### Checklist

- [ ] Link to Milestone


upgrade-agent/install.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+	if err != nil {
+		return fmt.Errorf("installing from %q: copying to destination %q: %w", kubernetesComponent.URL, kubernetesComponent.InstallPath, err)
+	}
+
+	return nil
+}
+
+// extractArchive extracts tar gz archives to a prefixed destination.
+func (i *osInstaller) extractArchive(archivePath, prefix string, perm fs.FileMode) error {
```
3u13r, line 91: I think we can omit this for now as the kubeadm (currently) is not an archive.

upgrade-agent/upgradeproto/upgrade.proto
```diff
@@ -0,0 +1,18 @@
...
+option go_package = "github.com/edgelesssys/constellation/v2/upgrade-agent/upgradeproto";
+
+service Update {
+    rpc ExecuteUpdate(ExecuteUpdateRequest) returns (ExecuteUpdateResponse);
+}
+
+message ExecuteUpdateRequest {
+    string kubeadm_URL = 1;
+    string kubeadm_Hash = 2;
+    string wanted_Kubernetes_Version = 3;
```
3u13r, line 14:
```suggestion
    string kubeadm_url = 1;
    string kubeadm_hash = 2;
    string wanted_kubernetes_version = 3;
```

Sorry, this is my fault. I haven't merged the correction yet, but it's part of https://github.com/edgelesssys/constellation/pull/716.

upgrade-agent/install.go
```diff
@@ -0,0 +1,299 @@
...
+	fs      *afero.Afero
+	hClient httpClient
+	// clock is needed for testing purposes
+	clock clock.WithTicker
+	// retriable is the function used to check if an error is retriable. Needed for testing.
+	retriable func(error) bool
+}
+
+// newOSInstaller creates a new osInstaller.
+func newOSInstaller() *osInstaller {
```
3u13r, line 45: This is more code than I remembered we had for this. I think when we remove then `.tar` part (see comment below), this should be a lot less. Still, I'm on the fence whether this is still considered ["a little copying"](https://go-proverbs.github.io/). What are your opinions @katexochen @stdoutput: should we move the installer from `bootstrapper/internal` to `internal` and share the code?

katexochen: Yes, I agree.

stdoutput: IMO, yes (move it)

upgrade-agent/server.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return &pb.ExecuteUpdateResponse{}, err
+	}
+
+	upgradeCmd := exec.CommandContext(ctx, "kubeadm", "upgrade plan")
+	if err := upgradeCmd.Run(); err != nil {
+		return &pb.ExecuteUpdateResponse{}, err
+	}
+
+	applyCmd := exec.CommandContext(ctx, "kubeadm", "upgrade apply", updateRequest.Wanted_Kubernetes_Version)
```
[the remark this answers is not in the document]
stdoutput: So I would check via `kubectl version --short`? Or is there a preffered way here?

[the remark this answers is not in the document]
3u13r: What I thought of was just to add some input validation for the `wanted_kubernetes_version`.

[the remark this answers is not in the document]
stdoutput: [That's in already ](https://github.com/edgelesssys/constellation/blob/3766bee3265242f4a8f0b6b092311fdceab97b08/upgrade-agent/server.go#L124-L130)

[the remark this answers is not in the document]
3u13r: Ah, sorry. I didn't see that. Maybe we shoud add `^` and `$` to the regex.

stdoutput:
Also, i wondered about whether the agent needs to be secured / use TLS for the connection?

3u13r:
Overall architecture looks good! Currently, I don't see anything large missing. I think we can already add this in the `CMakeLists.txt`. We can integrate this in the CI and Image in a separate PR.

upgrade-agent/server.go
```diff
@@ -0,0 +1,151 @@
...
+	grpcServer := grpc.NewServer()
+
+	upgradeproto.RegisterUpdateServer(grpcServer, s)
+
+	cleanup := func() error {
+		if _, err := os.Stat(sockAddr); err == nil {
+			if err := os.RemoveAll(sockAddr); err != nil {
+				return err
+			}
+		}
```
3u13r, line 65:
```suggestion
	err := os.RemoveAll(sockAddr)
	if errors.Is(err, fs.ErrNotExist) {
		return nil
	} else if err != nil {
		return err
	}
```
Using `os.Stat` when followed by anything like `os.Open`, `os.Remove`, or similar is an antipattern since it leads to [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issues. Those file methods are small wrappers around their syscall counterparts like [open](https://man7.org/linux/man-pages/man2/open.2.html), [close](https://man7.org/linux/man-pages/man2/close.2.html), and so on. Those syscalls have defined error cases with which one can avoid a prior stat usage. Those errors are also exposed as Go error types. I.e. in 90% of the cases a `os.Stat` can be replaced with a `errors.Is(err, fs.ErrNotExist)` :smile:

upgrade-agent/server.go
```diff
@@ -0,0 +1,151 @@
...
+
+	s.log.Infof("Stopped")
+}
+
+// ExecuteUpdate installs & verifies the provided kubeadm, then executes `kubeadm upgrade plan` & `kubeadm upgrade apply {wanted_Kubernetes_Version}` to upgrade to the specified version.
+func (s *Server) ExecuteUpdate(ctx context.Context, updateRequest *upgradeproto.ExecuteUpdateRequest) (*upgradeproto.ExecuteUpdateResponse, error) {
+	s.log.Infof("Upgrade to Kubernetes version started: %s", updateRequest.WantedKubernetesVersion)
+	installer := installer.NewOSInstaller()
+	if err := prepareUpdate(ctx, installer, updateRequest); err != nil {
+		return &upgradeproto.ExecuteUpdateResponse{}, err
```
3u13r, line 95: I think we should wrap the errors in grpc errors similar to e.g. the [JoinService ](https://github.com/edgelesssys/constellation/blob/main/joinservice/internal/server/server.go#L121)

3u13r (changes requested):

upgrade-agent/server.go
```diff
@@ -0,0 +1,140 @@
...
+	return &upgradeproto.ExecuteUpdateResponse{}, nil
+}
+
+// verifyVersion verifies the provided Kubernetes version.
+func verifyVersion(version string) error {
+	if !versionRegexp.MatchString(version) {
+		return fmt.Errorf("invalid kubernetes version: %s", version)
+	}
+	return nil
+}
```
daniel-weisse, line 133: I think the [semver package](https://pkg.go.dev/golang.org/x/mod/semver#IsValid) provides this functionality.

daniel-weisse (commented):
:+1:

daniel-weisse (approved):
looks good to me

3u13r (approved):
One last typo. Please also fix the CI linter checks. Otherwise LGTM.

PR #746, opened 2022-12-07
OS Images: adopt new API

malt3:
### Proposed change(s)

- use "ref", "stream", "version" to refer to images
- use new API paths for images and versions
- Adopt changes for CI and CLI

### Additional info

- [OS build pipeline run](https://github.com/edgelesssys/constellation/actions/runs/3639326620)
- [Generate measurements run](https://github.com/edgelesssys/constellation/actions/runs/3650337117)
  - this one also uses constellation create with the new naming scheme

### Checklist

- [x] Link to Milestone


go.mod
```diff
@@ -84,6 +84,7 @@ require (
 	github.com/spf13/cobra v1.6.1
 	github.com/stretchr/testify v1.8.1
 	github.com/talos-systems/talos/pkg/machinery v1.2.7
+	github.com/tj/assert v0.0.0-20171129193455-018094318fb0
```
katexochen: Please use stretchr/testify/assert :wink:

.github/workflows/build-os-image.yml
```diff
@@ -79,41 +93,65 @@ jobs:
         id: version
         uses: ./.github/actions/pseudo_version
 
+      - name: Determine ref
+        id: ref
+        run: |
+          if [ "${{ inputs.isRelease }}" = "true" ]; then
```
katexochen, line 99:
```suggestion
          if [[ "${{ inputs.isRelease }}" = "true" ]]; then
```
In general, try using `[[` instead of `[`

.github/workflows/build-os-image.yml
```diff
@@ -28,13 +38,13 @@ jobs:
           ref: ${{ github.head_ref }}
 
       - name: Build bootstrapper
-        if: ${{ inputs.debug == false }}
+        if: ${{ inputs.stream != 'debug' }}
```
katexochen, line 41:
```suggestion
        if: inputs.stream != 'debug'
```
No need for `${{` in if

internal/config/validation.go
```diff
@@ -19,19 +19,6 @@ import (
 	"github.com/go-playground/validator/v10"
 )
 
-func validateImage(fl validator.FieldLevel) bool {
-	image := fl.Field().String()
-	switch {
-	case image == "":
-		return false
-	case image == "..":
```
katexochen, line 27: Shouldn't this be a `strings.Contains` as well?

katexochen (changes requested):
Reviewed up to including 00d80454ab999fa7a932f531d3bfc06031650344

internal/config/validation.go
katexochen: So there's no need for this function anymore?

malt3: Yes I can remove it!

.github/actions/shortname/action.yaml
```diff
@@ -0,0 +1,50 @@
+name: Determine ref, stream and version for a shortname
```
katexochen, line 1: Name should be short an precise.

.github/actions/shortname/action.yaml
```diff
@@ -0,0 +1,50 @@
...
+        IFS="/" read -r -a string_array <<< "${{ inputs.shortname }}"
+        for ((i=0; i<${#string_array[@]}; i++)); do
+        echo "${string_array[i]}"
+        if [[ ${string_array[i]} == "ref"* ]]; then
+        ref=${string_array[i+1]}
+        elif [[ ${string_array[i]} == "stream"* ]]; then
+        stream=${string_array[i+1]}
+        else
+        version=${string_array[i]}
+        fi
```
katexochen, line 39: Please indent. Also why do we need to match ref*/stream*?

.github/actions/shortname/action.yaml
```diff
@@ -0,0 +1,50 @@
...
+        {
+          echo "ref=$ref"
+          echo "stream=$stream"
+          echo "version=$version"
+        }  >> "$GITHUB_OUTPUT"
+        {
+          echo "ref=$ref"
+          echo "stream=$stream"
+          echo "version=$version"
+        }  >> "$GITHUB_ENV"
```
katexochen, line 50:
```suggestion
        {
          echo "ref=$ref"
          echo "stream=$stream"
          echo "version=$version"
        }  | tee "$GITHUB_OUTPUT" "$GITHUB_ENV"
```

katexochen (changes requested):

.github/actions/shortname/action.yaml
malt3: I indented and chaned ref* -> ref and stream* -> stream

katexochen (approved):

malt3:
CI is green (except for the link checker). Merging!

PR #752, opened 2022-12-08
Document cert-manager installation.

derpsteb:
### Proposed change(s)

- Explain that cert-manager is preinstalled and what limitations this brings in regards to user-managed cert-manager.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/docs/docs/architecture/cert-manager.md
+++ b/docs/docs/architecture/cert-manager.md
@@ -1,0 +1,8 @@
+# Cert-manager
+
+Constellation ships with cert-manager preinstalled.
+The default installation is part of the `kube-system` namespace, as all other Constellation-managed components.
+While you are free to install more instances of `cert-manager` into other namespaces, please be aware that any new installation need to use the same version as the one installed with Constellation.
+Or rely on the same CRD versions as Constellation's installation.
+Also remeber to set the `installCRDs` value to `false` when installing new `cert-manager` instances.
+Because CRDs are cluster-wide resources and `cert-manager` uses CRDs to expose it's functionality it will create problems if you have two installations of cert-manager depending on different versions of those CRDs.
--- a/docs/docs/architecture/cert-manager.md
+++ b/docs/docs/architecture/cert-manager.md
@@ -2,7 +2,9 @@
 
 Constellation ships with cert-manager preinstalled.
 The default installation is part of the `kube-system` namespace, as all other Constellation-managed components.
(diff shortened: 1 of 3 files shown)


docs/docs/architecture/cert-manager.md
```diff
@@ -0,0 +1,10 @@
+# Cert-manager
+
+Constellation ships with cert-manager preinstalled.
+The default installation is part of the `kube-system` namespace, as all other Constellation-managed components.
+You are free to install more instances of `cert-manager` into other namespaces.
+However, please be aware that any new installation need to use the same version as the one installed with Constellation.
```
datosh, line 6: We could use a `:::caution` or `:::info` box to highlight this information.

datosh (commented):
This is currently not showing up on the docs and needs to be added to `constellation/docs/sidebars.js`

m1ghtym0 (commented):
I'm unsure if `/architecture` is the best place to put it.
Makes sense ofc, but not sure if people would look into that.
We could also place it into `Workflows` like `Install cert-manager` or smth.
Or we have another top-level category, for Kubernetes operations for example.

docs/docs/architecture/cert-manager.md
derpsteb: Couldn't find any information on this in the official docs. But I think there is nothing to be done apart from setting the namespace.

derpsteb:
Moved to Workflows.

docs/docs/workflows/installing-cert-manager.md
```diff
@@ -0,0 +1,12 @@
...
+Please read this section before installing cert-manager.
+:::
+Constellation ships with cert-manager preinstalled.
+The default installation is part of the `kube-system` namespace, as all other Constellation-managed components.
+You are free to install more instances of `cert-manager` into other namespaces.
+However, please be aware that any new installation need to use the same version as the one installed with Constellation.
+Or rely on the same CRD versions as Constellation's installation.
+Also remember to set the `installCRDs` value to `false` when installing new `cert-manager` instances.
+It will create problems if you have two installations of cert-manager depending on different versions of the installed CRDs.
+CRDs are cluster-wide resources and `cert-manager` uses depends on specific versions of those CRDs for each release.
```
datosh, line 12:
```suggestion
CRDs are cluster-wide resources and `cert-manager` depends on specific versions of those CRDs for each release.
```

datosh (approved):
I think the word is extra, but other than that this LGTM

PR #758, opened 2022-12-08
image: fix script for PKI generation

katexochen:

- Moved previous line 33 to new line 24, as the script would silently fail if the `pki` dir did not exist
- Fixed the check for set variable, this would now fail due to our bash options, using default values instead
- Only using capital var names for external things


malt3 (approved):
LGTM. Thanks for the fix!

PR #776, opened 2022-12-09
join: synchronize control plane joining

3u13r:
### Proposed change(s)

- Deploy our node controller on every CSP (added QEMU, AWS) 
- Only enable working controllers inside the operator
- let control plane nodes join only one-by-one to avoid a race in etcd

### Checklist

- [x] Link to Milestone


joinservice/internal/kubernetes/kubernetes.go
```diff
@@ -81,22 +82,46 @@ func (c *Client) CreateConfigMap(ctx context.Context, configMap corev1.ConfigMap
...
 
 // AddNodeToJoiningNodes adds the provided node as a joining node CRD.
-func (c *Client) AddNodeToJoiningNodes(ctx context.Context, nodeName string, componentsHash string) error {
-	joiningNodeResource := schema.GroupVersionResource{Group: "update.edgeless.systems", Version: "v1alpha1", Resource: "joiningnodes"}
-
+func (c *Client) AddNodeToJoiningNodes(ctx context.Context, nodeName string, componentsHash string, isControlPlane bool) error {
 	joiningNode := &unstructured.Unstructured{}
+
+	objectMetadataName := nodeName
+	deadline := metav1.NewTime(time.Now().Add(48 * time.Hour))
```
derpsteb, line 89: what?! :D

48h seems a bit long doesn't it?

joinservice/internal/kubernetes/kubernetes.go
```diff
@@ -81,22 +82,46 @@ func (c *Client) CreateConfigMap(ctx context.Context, configMap corev1.ConfigMap
...
+		return fmt.Errorf("failed to create joining node, maybe another control-plane joining is already in progress: %w", err)
+	}
+	return nil
+}
+
+func (c *Client) addWorkerToJoiningNodes(ctx context.Context, joiningNode *unstructured.Unstructured) error {
+	joiningNodeResource := schema.GroupVersionResource{Group: "update.edgeless.systems", Version: "v1alpha1", Resource: "joiningnodes"}
 	_, err := c.dynClient.Resource(joiningNodeResource).Apply(ctx, joiningNode.GetName(), joiningNode, metav1.ApplyOptions{FieldManager: "join-service"})
 	if err != nil {
 		return fmt.Errorf("failed to create joining node: %w", err)
```
derpsteb, line 127: use same error message as above?

operators/constellation-node-operator/config/manager/kustomization.yaml
```diff
@@ -5,6 +5,12 @@ generatorOptions:
...
-  files:
+- files:
   - controller_manager_config.yaml
+  name: manager-config
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+images:
+- name: controller
+  newName: ghcr.io/edgelesssys/constellation/node-operator
+  newTag: v0.0.1
```
malt3, line 16: This change can be reverted. I think this happens automatically on some make targets but does not make a whole lot of sense (we don't deploy a tag v0.0.1).

operators/constellation-node-operator/controllers/joiningnode_controller.go
```diff
@@ -73,23 +80,35 @@ func (r *JoiningNodesReconciler) Reconcile(ctx context.Context, req ctrl.Request
 		return r.Update(ctx, &node)
 	})
 	if err != nil {
-		logr.Error(err, "unable to update Node")
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		// check if the deadline has been reached
+		// requeue if not
+		if joiningNode.Spec.Deadline == nil || r.Now().Before(joiningNode.Spec.Deadline.Time) {
```
derpsteb, line 85: `r.Now().Before(joiningNode.Spec.Deadline.Time)` doesn't this evaluate to `true` in case the deadline is still in the future?

3u13r: Yes. If the deadline is in the future, we re-queue the request to have a look at it again.

joinservice/internal/kubernetes/kubernetes.go
3u13r: Note the difference between `.Apply()` and `Create()`. We don't expect the first one to error at all. Every worker is create under a new name, and even if names collide, the resource is updated. For control-planes we always use the same name to provoke a collision, when 2 control-plane try to join simultaneously.

malt3 (approved):
LGTM

joinservice/internal/kubernetes/kubernetes.go
3u13r: Yes. If we were to insert an empty string for `deadline` we could even disable it completely, which was my original idea. But the code path for this in the JoinService is a bit cumbersome. Note that this only applies to worker nodes. Superfluous entries do not create any issues here. So this is basically garbage collection.

PR #782, opened 2022-12-12
cli: automatically add iam values to config

stdoutput:
### Proposed change(s)

- Add a `generate-config` flag to the `constellation iam create` command which enables automatic generation of a config file (path specified by the `config` flag. By default, `constellation-conf.yaml`) **and** automatic filling of the resulting IAM values into said file.

### Checklist

- [ ] Link to Milestone


stdoutput:
**RFC**
Currently, `config.New()` has no option to customize or disable config validation, which would be needed to use it to load an incomplete config (all IAM values & measurements will be missing at the point where i load it). Therefore i made a temporary function `config.NewWithoutValidation()` which skips the validation step. This is not a clean solution. I think we should allow an option to provide a custom validation function to `config.New()`. What do you think?

3u13r:
Short summary from today's stand-up discussion: 
- We don't want to write the IAM credentials/information directly into the `constellation-conf.yaml` but we want to reference a file (similar to what GCP already does)
- This way we you can safely share even Azure `constellation-conf.yaml` which currently include secrets
- This should side-step the requested change for moving the validation
- The field should be provider specific (aka. be .provider.azure.IAMConfig and not .IAMConfig) 
- The new file will probably be one struct for which includes all fields for all CSPs which is then marshalled into the file.

datosh:
> Therefore i made a temporary function `config.NewWithoutValidation()` which skips the validation step. This is not a clean solution. I think we should allow an option to provide a custom validation function to `config.New()`. What do you think?

I like the solution. The goal of `config.New` was to make it hard to use it wrong, e.g., for the to validate the config. `NewWithoutValidation` is explicit about what it is skipping. I think the API is good 🙂 

You could share more code between `New` and `NewWithoutValidation` so maintenance is easier. Maybe have a private `new` that implement most of the new logic.

stdoutput:
> What do you think is not clean about the solution or should be improved?
I think it might be better to keep the "working (at that point in time)" validation steps (such as checking if all fields are there and if the file is not corrupted generally) could be kept to prevent errors when the values are later on inserted into the file. But with the proposed rework, that would not be necessary i guess ;-)

stdoutput:
> We don't want to write the IAM credentials/information directly into the `constellation-conf.yaml` but we want to reference a file (similar to what GCP already does)
Since the values created by the `iam create` step are also used in other steps (e.g. `create`), we would also have to change all accesses to these values in the rest of the code. Since only 1 value of 1 cloud provider is sensitive here, i think we should rather look for another way. I might be missing a big point here, but from my thinking there are other ways to prevent the secret being exposed in the file.

> This way we you can safely share even Azure `constellation-conf.yaml` which currently include secrets
We could also tackle this by using the environment variable `CONSTELL_AZURE_CLIENT_SECRET_VALUE`

stdoutput:
Last consensus was (iirc):

- Leave `constellation iam create` as is, do not add config values automatically from there but only print them and ask the user to paste them in the config
- Create an opt-in for creating iam resources and filling iam values into the config automatically at the `constellation config generate` command

@katexochen are you OK with that?

katexochen:
> @katexochen are you OK with that?

Yes, sounds good! :)

stdoutput:
Any1 got an idea on how to implement this in a "clean" manner?

I think adding all IAM-flags for every CP to config generate would be ugly and redundant as we have them on iam create already. But how could we get around it? Have subcommands for config generate? Seems ugly as well.

Any ideas are welcome 😄

katexochen:
> Create an opt-in for creating iam resources and filling iam values into the config automatically at the constellation config generate command

Iirc Thomas and I were also okay with adding an opt-in option to iam command to generate config instead, but you didn't like it this way? Would solve the problem imo...

stdoutput:
> > Create an opt-in for creating iam resources and filling iam values into the config automatically at the constellation config generate command
> 
> Iirc Thomas and I were also okay with adding an opt-in option to iam command to generate config instead, but you didn't like it this way? Would solve the problem imo...

Sounds good. Maybe I mixed something up, I thought you were against doing it that way. Then I'll go work on that! Thank you :)

katexochen:
> I thought you were against doing it that way

I was just against doing that opt-out.

stdoutput:
Implemented an idea of how it could look for Azure now. Would appreciate if someone would review this so I can go on with the other providers.

daniel-weisse (commented):
Overall structure looks good to me and makes sense.
Probably not the correct person to review this, but we have code owners now to fix that problem :smile:

datosh (commented):
Yes, the chosen approach works for me as well!

I have built & tested the current state locally for Azure. I have two comments, but they are unrelated to this PR (so we can choose to fix this here or in another PR):

The output when creating resources for Azure is misaligned. I think it would be more readable to have a consistent indentation for region, resource group & service principal:

```
./constellation iam create azure --region westus --resourceGroup fabian-delete-me-iam --servicePrincipal fabian-delete-me-iam --generate-config
The following IAM configuration will be created:
Region: westus
Resource Group: fabian-delete-me-iam
Service Principal:      fabian-delete-me-iam
The configuration file constellation-conf.yaml will be automatically generated and populated with the IAM values.
Do you want to create the configuration? [y/n]: y
Creating  
Your IAM configuration was created and filled into constellation-conf.yaml successfully.
```

The help text for resource group is not correct:

```
--resourceGroup string      Name of the resource group your IAM resources will be created in.
```

When I run `iam create` two resource groups seem to be created. The one I provided and an additional with suffix `-identity`. The suffixed one contains the IAM resources mentioned in helptext. I think we should mention both in helptext.

cli/internal/cmd/iamcreateaws.go
```diff
@@ -65,22 +72,44 @@ func iamCreateAWS(cmd *cobra.Command, spinner spinnerInterf, creator iamCreator)
...
+		conf.Provider.AWS.Region = awsFlags.region
+		conf.Provider.AWS.Zone = awsFlags.zone
+		conf.Provider.AWS.IAMProfileControlPlane = iamFile.AWSOutput.ControlPlaneInstanceProfile
+		conf.Provider.AWS.IAMProfileWorkerNodes = iamFile.AWSOutput.WorkerNodeInstanceProfile
+
+		if err := fileHandler.WriteYAML(awsFlags.configPath, conf, file.OptOverwrite); err != nil {
+			return err
+		}
+		cmd.Printf("Your IAM configuration was created and filled into %s successfully.\n", awsFlags.configPath)
+	} else {
```
katexochen, line 106: Use early return instread of else.
```suggestion
        return nil
	}
```

cli/internal/cmd/iamcreateaws.go
```diff
@@ -65,22 +72,44 @@ func iamCreateAWS(cmd *cobra.Command, spinner spinnerInterf, creator iamCreator)
...
-	cmd.Println("Your IAM configuration was created successfully. Please fill the above values into your configuration file.")
+	if awsFlags.generateConfig {
+		conf.Provider.AWS.Region = awsFlags.region
+		conf.Provider.AWS.Zone = awsFlags.zone
+		conf.Provider.AWS.IAMProfileControlPlane = iamFile.AWSOutput.ControlPlaneInstanceProfile
+		conf.Provider.AWS.IAMProfileWorkerNodes = iamFile.AWSOutput.WorkerNodeInstanceProfile
+
+		if err := fileHandler.WriteYAML(awsFlags.configPath, conf, file.OptOverwrite); err != nil {
+			return err
+		}
```
katexochen, line 104: This line isn't covered by the unit test. You could add a test case with a read-only fs. (This case is also the reason to have a setupFS function, that you removed.)

cli/internal/cmd/iamcreateaws_test.go
```diff
@@ -99,16 +155,37 @@ func TestIAMCreateAWS(t *testing.T) {
...
 
-			err := iamCreateAWS(cmd, nopSpinner{}, tc.creator)
+			err := iamCreateAWS(cmd, nopSpinner{}, tc.creator, fileHandler)
 
 			if tc.wantErr {
 				assert.Error(err)
 			} else {
 				if tc.wantAbort {
 					assert.False(tc.creator.createCalled)
 				} else {
```
katexochen, line 178: This deep nesting can be simplified with early returns

cli/internal/cmd/configgenerate.go
```diff
@@ -51,6 +51,22 @@ func configGenerate(cmd *cobra.Command, fileHandler file.Handler, provider cloud
 		return err
 	}
 
+	_, err = createConfig(cmd, fileHandler, provider, flags.file)
+	if err != nil {
```
katexochen, line 55:
```suggestion
	if _, err = createConfig(cmd, fileHandler, provider, flags.file); err != nil {
```

cli/internal/cmd/configgenerate.go
```diff
@@ -51,6 +51,22 @@ func configGenerate(cmd *cobra.Command, fileHandler file.Handler, provider cloud
...
+		cmd.Println("Config file written to", flags.file)
+		cmd.Println("Please fill in your CSP-specific configuration before proceeding.")
+		cmd.Println("For more information refer to the documentation:")
+		cmd.Println("\thttps://docs.edgeless.systems/constellation/getting-started/first-steps")
+	}
+	return nil
+}
+
+// createConfig creates a config file for the given provider and writes it to the given outputFile. If the outputFile is "-", the config is written to stdout.
+func createConfig(cmd *cobra.Command, fileHandler file.Handler, provider cloudprovider.Provider, outputFile string) (*config.Config, error) {
```
katexochen, line 69: Imo it would be a bit cleaner to just generate the config in `createConfig` and do the output in `configGenerate`. This would also make the overwrite in iamcreate unnecessary.

internal/config/config.go
```diff
@@ -320,6 +320,28 @@ func New(fileHandler file.Handler, name string) (*Config, error) {
 	return c, c.Validate()
 }
 
+// TODO: Create a better way of validating config values,
+// maybe provide an option to pass a custom validation function
+
+// NewWithoutValidation creates a new config by:
+// 1. Reading config file via provided fileHandler from file with name.
+// 2. Read secrets from environment variables.
+func NewWithoutValidation(fileHandler file.Handler, name string) (*Config, error) {
```
katexochen, line 329: Func is currently unused, is this obsolete?

katexochen (changes requested):
Only some minor architecture and language comments, idea looks good so far.

katexochen (approved):
lgtm, well done. :whale: 

Some minor follow ups that shouldn't be done in this PR:

- [ ] The IAM commands currently lack debug logging, we should add some lines, at least the parsed flags etc
- [ ] Look if we can reducing redundancy in the IAM commands, maybe by introducing an interface that the the flags fullfill, so that that the core logic doesn't have to be duplicated

stdoutput:
>  The IAM commands currently lack debug logging, we should add some lines, at least the parsed flags etc

I forgot to mention this in our conversation today. I only discovered this today, think we had a bad timing between merging IAM and Debug logging so that they "missed" each other. 

> Look if we can reducing redundancy in the IAM commands, maybe by introducing an interface that the the flags fullfill, so that that the core logic doesn't have to be duplicated

Agreed.

I will bring up both points in the next backlog refinement.

katexochen:
> I will bring up both points in the next backlog refinement.

Feel free to already create tickets for this and start implementation in a separate PR as you like.

PR #780, opened 2022-12-12
Release docs for v2.3

malt3:
### Proposed change(s)

- Release docs for v2.3

PR #783, opened 2022-12-12
docs: few fixes and rewording

thomasten:
Some fixes and rewording in view of the upcoming release.


katexochen:
Notice that the docs for the release are already generated in #780. Might be a good idea to rebase on that branch...

thomasten:
I think it should be the other way round: merge this one and then regenerate v2.3.

m1ghtym0 (approved):
Changes LGTM!

PR #788, opened 2022-12-12
rfc: update documentation of new versions/image API

katexochen:
Changes to include: https://gist.github.com/katexochen/4f8a51117b8d68c0dcd880ac51c8a41b

### DIFF
--- a/rfc/apis.md
+++ b/rfc/apis.md
@@ -34,7 +34,9 @@ There may be more API groups in the future (e.g. `cli`)
 - [`/constellation/v1/ref/<ref>/stream/<stream>/image/<version>/csp/<csp>/measurements.image.json`](image-api.md)
 - [`/constellation/v1/ref/<ref>/stream/<stream>/image/<version>/csp/<csp>/image.raw`](image-api.md)
 
-## Meaning of `ref`, `stream` and `version` for resource names
+## API path identifiers  `ref`, `stream` and `version`
+
+### Meaning of `ref`, `stream` and `version` for resource names
 
 Components in this API are identified by `ref`, `stream` and `version`.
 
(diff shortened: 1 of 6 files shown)


rfc/apis.md
```diff
@@ -0,0 +1,122 @@
...
+- `ref/-/stream/<stream>/<version>` is equivalent to the short form `stream/<stream>/<version>`
+
+
+## Resource names
+
+Resources can be identified by `ref/<ref>/stream/<stream>/<version>` and allow for short form encoding as explained above.
+Resources using this encoding include:
+
+- `version`: version information for Constellation components
+- `image`: metadata for individual Constellation OS images
```
katexochen, line 71: This part isn't really clear yet. Also, it doesn't fit well the API paths...

rfc/apis.md
```diff
@@ -0,0 +1,122 @@
...
+  - `sbom`: SBOMs for Constellation OS images
+
+There may be more API groups in the future (e.g. `cli`)
+
+## API paths overview
+
+- [`/constellation/v1/ref/<ref>/stream/<stream>/versions/latest/<kind>.json`](version-api.md#latest)
+- [`/constellation/v1/ref/<ref>/stream/<stream>/versions/major/<base>/<kind>.json`](version-api.md#major-to-minor-version-list)
+- [`/constellation/v1/ref/<ref>/stream/<stream>/versions/minor/<base>/<kind>.json`](version-api.md#minor-to-patch-version-list)
+- [`/constellation/v1/ref/<ref>/stream/<stream>/image/<version>/info.json`](image-api.md#image-lookup-table)
```
katexochen, line 30:
```suggestion
- [`/constellation/v1/ref/<ref>/stream/<stream>/<version>/image/info.json`](image-api.md#image-lookup-table)
```
Might be better the other way around, especially regarding the meaning of a short path.

katexochen:
We should also define how to discover major versions in the versions API. Currently we can just assume it is v2, but be should consider that we might want to bump the major version and might also have tooling that in this case would need to deal with both major vers (i.e., tooling that couldn't use a constant string anymore).

rfc/apis.md
katexochen: This would also make deleting a single version much simpler.

rfc/apis.md
```diff
@@ -0,0 +1,132 @@
...
+- `debug`: Image containing the debugd, allows access to the serial console
+
+## Short paths for resource identification
+
+Short paths allow an easier version handling, e.g., in end user configuration files or CLI tools.
+When using the default values, short paths omit either the `ref` (if set to `-`) or the `ref` and `stream`
+parts (if set to `-` and `stable`):
+
+- `ref/-/stream/stable/<version>` is equivalent to the short path `<version>`
+- `ref/-/stream/<stream>/<version>` is equivalent to the short path `stream/<stream>/<version>`
```
katexochen, line 92: How would a short path for API version v2 look like? If the API is versioned but users are using an unversioned short path, much of the benefits of the versioning seems to be lost.

malt3: The API version should probably be hardcoded in the API client and not be part of the short name. Instead the short name should stay forwards compatible with new API versions.

rfc/apis.md
katexochen: As discussed with @malt3 , I will implement this change before the next release.

rfc/version-api.md
```diff
@@ -108,13 +121,18 @@ The CLI can query the `latest` endpoint to inform users about new releases of th
...
 
-- Explicit compatbility information between different resource kinds
+- Explicit compatibility information between different resource kinds
     - Example: OS image `v2.3.4` requires microservice version > `v2.3.1`
     - Allows fine grained control in case of unexpected incompatibility
 - Version metadata
     - Changelogs
     - Deprecation warnings
-    - Security information
+    - Security informatio
```
malt3, line 138:
```suggestion
    - Security information
```

malt3 (approved):
otherwise lgtm

PR #794, opened 2022-12-13
List versions with find-image script

katexochen:
### Proposed change(s)

- Adding the ability to list version with the find image script
- Adding JSON output option
- Adding help option


malt3 (approved):
LGTM and works great

PR #730, opened 2022-12-06
E2E Test Mini Constellation

datosh:
### Proposed change(s)

- This PR depends on permissions and users configured in https://github.com/edgelesssys/edgeless-iam/pull/1
- Terraform is used to create an environment (VM with nested virt. support) in Azure.
  - A long lived resource group is used to minimize permissions and avoid IAM replication delay.
  - All resources are suffixed with a random ID to prevent naming collisions
- Currently only positive tests are executed on MiniConstellation
  - Does the CLI command work `constellation mini up`
  - Are two nodes available?
  - Are the expected deployments available?
- These tests are only to be executed when a release is prepared to verify MiniConstellation functionality
  - An additional (nightly/weekly) test for QEMU will be implemented in a separate PR
- All parts can also be run easily from local machine to aid in development:
```sh
cd .github/actions/e2e_mini
terraform init
terraform apply -auto-approve
export AZURE_VM_IP=$(terraform output -raw public_ip)
scp -i id_rsa ../../../build/constellation adminuser@${AZURE_VM_IP}:.
ssh -i id_rsa adminuser@${AZURE_VM_IP} 'bash -s' < run-e2e.sh
ssh -i id_rsa adminuser@${AZURE_VM_IP} # for manual inspection
```

Successful manual trigger: https://github.com/edgelesssys/constellation/actions/runs/3629545881

### Checklist

- [x] Update [CHANGELOG.md](https://github.com/edgelesssys/constellation/blob/main/CHANGELOG.md)
- [x] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)
- [x] Link to Milestone
- [ ] Revert temporary code changes

PR #796, opened 2022-12-13
E2E Test Mini Constellation

datosh:
### Proposed change(s)

- This PR depends on permissions and users configured in https://github.com/edgelesssys/edgeless-iam/pull/1
- Terraform is used to create an environment (VM with nested virt. support) in Azure.
  - A long lived resource group is used to minimize permissions and avoid IAM replication delay.
  - All resources are suffixed with a random ID to prevent naming collisions
- Currently only positive tests are executed on MiniConstellation
  - Does the CLI command work `constellation mini up`
  - Are two nodes available?
  - Are the expected deployments available?
- These tests are only to be executed when a release is prepared to verify MiniConstellation functionality
  - An additional (nightly/weekly) test for QEMU will be implemented in a separate PR

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/actions/e2e_mini/run-e2e.sh
+++ b/.github/actions/e2e_mini/run-e2e.sh
@@ -1,0 +1,58 @@
+#!/usr/bin/env bash
+#
+# This script installs all dependencies for MiniConstellation to be run on a
+# fresh Ubuntu 22.04 LTS installation.
+# It expects to find the to be used Constellation CLI to be available at
+# $HOME/constellation
+#
+set -euxo pipefail
+
+cloud-init status --wait
+
+export DEBIAN_FRONTEND=noninteractive
+sudo apt update -y
+
+sudo apt install -y bridge-utils cpu-checker libvirt-clients libvirt-daemon libvirt-daemon-system qemu qemu-kvm virtinst xsltproc ca-certificates curl gnupg lsb-release
+
+sudo mkdir -p /etc/apt/keyrings
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
+echo \
+  "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/ubuntu \
+  $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+sudo apt update -y
+sudo apt install -y docker-ce docker-ce-cli containerd.io docker-compose-plugin
+sudo usermod -aG docker "$USER"
+newgrp docker
+
+mkdir constellation_workspace
+cd constellation_workspace
+mv "$HOME"/constellation .
+chmod u+x constellation
+
+sudo sh -c 'echo "127.0.0.1 license.confidential.cloud" >> /etc/hosts'
+
+mkdir -p "$HOME"/.docker
+touch "$HOME"/.docker/config.json
+
+./constellation mini up
+
+curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+sudo install kubectl /usr/local/bin/kubectl
+
+export KUBECONFIG="$PWD/constellation-admin.conf"
+
+# Wait for nodes to actually show up in K8s control plane
+sleep 10
+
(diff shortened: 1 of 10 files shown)


datosh:
PTAL @katexochen 
I have created a news PR, because there was a lot of noise in the old one #730 
I have tested this as it would have been used in a release running with v2.3.0 image: https://github.com/edgelesssys/constellation/actions/runs/3687194316/jobs/6240458553

.github/actions/e2e_mini/run-e2e.sh
```diff
@@ -0,0 +1,58 @@
...
+# $HOME/constellation
+#
+set -euxo pipefail
+
+cloud-init status --wait
+
+export DEBIAN_FRONTEND=noninteractive
+sudo apt update -y
+
+sudo apt install -y bridge-utils cpu-checker libvirt-clients libvirt-daemon libvirt-daemon-system qemu qemu-kvm virtinst xsltproc ca-certificates curl gnupg lsb-release
```
katexochen, line 15: nit: Would suggest to split line for better readability.

.github/actions/e2e_mini/run-e2e.sh
```diff
@@ -0,0 +1,58 @@
...
+kubectl wait --for=condition=Ready --timeout=600s nodes worker-0
+# Wait for deployments
+kubectl -n kube-system wait --for=condition=Available=True --timeout=180s deployment coredns
+kubectl -n kube-system wait --for=condition=Available=True --timeout=180s deployment cilium-operator
+# Wait for daemon sets
+kubectl -n kube-system rollout status --timeout 180s daemonset cilium
+kubectl -n kube-system rollout status --timeout 180s daemonset join-service
+kubectl -n kube-system rollout status --timeout 180s daemonset kms
+kubectl -n kube-system rollout status --timeout 180s daemonset konnectivity-agent
+kubectl -n kube-system rollout status --timeout 180s daemonset verification-service
```
katexochen, line 58: nit: The verification-service has now much more time get ready than the join-service... is this desired?

.github/actions/e2e_mini/run-e2e.sh
```diff
@@ -0,0 +1,58 @@
...
+
+sudo apt install -y bridge-utils cpu-checker libvirt-clients libvirt-daemon libvirt-daemon-system qemu qemu-kvm virtinst xsltproc ca-certificates curl gnupg lsb-release
+
+sudo mkdir -p /etc/apt/keyrings
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
+echo \
+  "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/ubuntu \
+  $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+sudo apt update -y
+sudo apt install -y docker-ce docker-ce-cli containerd.io docker-compose-plugin
```
katexochen, line 23: nit: Maybe put a output group around the apt operations, makes the test output easier to read imo

katexochen (approved):
lgtm, only some nits...

.github/actions/e2e_mini/run-e2e.sh
datosh: I know what you mean, but I was not able to find a more elegant solution.

If I change it to: 180s for Cilium and 10s for all subsequent, the join-service only has little to come online if Cilium is quick to come online. 

Only idea I had was to fire them all in parallel..? But I am not sure the additional complexity would benefit us. In the end, I am fine that the test checks that all services are online, no matter how long they took, since we do not have hard requirements on startup times. 

Let me know if you have any concrete ideas to improve this code 🙂

katexochen: As you said, cilium is expected to be up first, then the other services need some time to get up, but should be ready at nearly the same time. This is something you can't understand based on the commands. Maybe add an empty line/ comment after the cilium line, followed by 180s for join-service and then only a few seconds for the other sets? But it is also ok for me to leave this as it is right now, I was just wondering...

datosh: The 180s are not guaranteed to block for that time, but the command finishes as soon as deployment is online. This might not leave enough time for a subsequent deployment if its timeout is a lot lower.

I will keeep it as is for now.

PR #805, opened 2022-12-15
Azure disclaimer

datosh:
### Proposed change(s)

- add disclaimer for potentially long inits on Azure
- add @flxflx as codeowner for docs (you were not automatically added as reviewer, let's see if it works once it landed in `main`)

### DIFF
--- a/docs/docs/getting-started/first-steps.md
+++ b/docs/docs/getting-started/first-steps.md
@@ -128,6 +128,12 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
 
 4. Initialize the cluster
 
+    :::caution
+
+    Initialization on **Azure** might be slow and might take up to 60 minutes to initialize all K8s nodes. We are already working with Microsoft to resolve the issue. The stability of already initialized clusters is not affected.
+
+    :::
+
     ```bash
     constellation init
     ```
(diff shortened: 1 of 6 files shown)


flxflx (approved):

docs/docs/getting-started/first-steps.md
```diff
@@ -128,6 +128,12 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
 
 4. Initialize the cluster
 
+    :::caution
+
+    Initialization on **Azure** might be slow and might take up to 60 minutes to initialize all K8s nodes. We are already working with Microsoft to resolve the issue. The stability of already initialized clusters is not affected.
```
katexochen, line 133:
```suggestion
    Initialization on **Azure** might be slow and might take up to 60 minutes to initialize all Kubernetes nodes. We're already working with Microsoft to resolve the issue. The stability of already initialized clusters isn't affected.
```
Imo we shouldn't use the 'k8s' short form in our docs. And is there any reason to ignore the vale language recommendations?

flxflx: Agreed. Shouldn't use K8s.

flxflx: Also agree to that we should follow Vale recommendations except for the "we", which is OK in most cases.

datosh: I've added #820 to fix those.

PR #806, opened 2022-12-16
gcp: support internal load balancers

3u13r:
### Proposed change(s)

- The Terraform is a translation of https://cloud.google.com/load-balancing/docs/tcp/set-up-int-tcp-proxy-migs

Note that this PR breaks the global load balancing support and ONLY supports internal LBs on GCP.

### DIFF
--- a/cli/internal/terraform/terraform/gcp/modules/internal_loadbalancer/variables.tf
+++ b/cli/internal/terraform/terraform/gcp/modules/internal_loadbalancer/variables.tf
@@ -1,0 +1,50 @@
+variable "name" {
+  type        = string
+  description = "Base name of the load balancer."
+}
+
+variable "region" {
+  type        = string
+  description = "The region where the load balancer will be created."
+}
+
+variable "network" {
+  type        = string
+  description = "The network to which all network resources will be attached."
+}
+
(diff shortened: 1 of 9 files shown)


cli/internal/terraform/terraform/gcp/modules/internal_loadbalancer/variables.tf
```diff
@@ -0,0 +1,50 @@
...
+}
+
+variable "backend_subnet" {
+  type        = string
+  description = "The subnet to which all backend network resources will be attached."
+}
+
+variable "health_check" {
+  type        = string
+  description = "The type of the health check. 'HTTPS' or 'TCP'."
```
datosh, line 23: We could use an additional `validation.condition` check using `contains` if we want to restrict it to HTTPS & TCP.

See: 
* [validation](https://developer.hashicorp.com/terraform/language/expressions/custom-conditions#input-variable-validation)
* [contains](https://developer.hashicorp.com/terraform/language/functions/contains)

3u13r:
This was merged in https://github.com/edgelesssys/constellation/pull/826. Keeping the branch for a few more weeks as we gave out links pointing to it.

PR #807, opened 2022-12-16
join: make Azure instance names k8s compliant

3u13r:
### Proposed change(s)

- Azure machine names contain upper case letters. K8s doesn't allow them for any resource. We already normalize them in the bootstrapper. With 2.3.0 we introduced the joiningnode CRD which is created by the JoinService. For worker nodes the entries are named after the node name from inside the CSR, which is the non-normalized Azure machine name. 

### Checklist

- [x] Check if the name also needs to be normalized in the Spec, since our node operator consumes this field. Yes.
- [x] Link to Milestone

### DIFF
--- a/joinservice/internal/kubernetes/kubernetes.go
+++ b/joinservice/internal/kubernetes/kubernetes.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/edgelesssys/constellation/v2/internal/constants"
@@ -85,7 +86,11 @@ func (c *Client) CreateConfigMap(ctx context.Context, configMap corev1.ConfigMap
 func (c *Client) AddNodeToJoiningNodes(ctx context.Context, nodeName string, componentsHash string, isControlPlane bool) error {
 	joiningNode := &unstructured.Unstructured{}
 
-	objectMetadataName := nodeName
+	compliantNodeName := k8sCompliantHostname(nodeName)
+
+	// JoiningNodes referencing a worker node are named after the worker node.
+	// JoiningNodes referencing the control-plane node are named "control-plane".
+	objectMetadataName := compliantNodeName
 	deadline := metav1.NewTime(time.Now().Add(48 * time.Hour))
 	if isControlPlane {
 		objectMetadataName = "control-plane"
@@ -99,7 +104,7 @@ func (c *Client) AddNodeToJoiningNodes(ctx context.Context, nodeName string, com
 			"name": objectMetadataName,
 		},
 		"spec": map[string]any{
-			"name":           nodeName,
+			"name":           compliantNodeName,
 			"componentshash": componentsHash,
 			"iscontrolplane": isControlPlane,
 			"deadline":       deadline,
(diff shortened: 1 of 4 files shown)


joinservice/internal/kubernetes/kubernetes.go
```diff
@@ -142,3 +147,12 @@ func (c *Client) AddReferenceToK8sVersionConfigMap(ctx context.Context, k8sVersi
...
 }
+
+// k8sCompliantHostname transforms a hostname to an RFC 1123 compliant, lowercase subdomain as required by Kubernetes node names.
+// The following regex is used by k8s for validation: /^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$/ .
+// Only a simple heuristic is used for now (to lowercase, replace underscores).
+func k8sCompliantHostname(in string) string {
+	hostname := strings.ToLower(in)
+	hostname = strings.ReplaceAll(hostname, "_", "-")
+	return hostname
+}
```
katexochen, line 158: As you have the regex for the validation already, I'd suggest to use it to validate the hostname in a last step of the function:

```suggestion
var validHostnameRegex := regexp.MustCompile("^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$")

// k8sCompliantHostname transforms a hostname to an RFC 1123 compliant, lowercase subdomain as required by Kubernetes node names.
// Only a simple heuristic is used for now (to lowercase, replace underscores).
func k8sCompliantHostname(in string) (string, error) {
	hostname := strings.ToLower(in)
	hostname = strings.ReplaceAll(hostname, "_", "-")
	if !validHostnameRegexp.MatchString(hostname) {
	    return "", fmt.Errorf("failed to generate a Kubernetes compliant hostname for %s", in)
	}
	return hostname, nil
}
```

joinservice/internal/kubernetes/kubernetes.go
```diff
@@ -85,7 +86,11 @@ func (c *Client) CreateConfigMap(ctx context.Context, configMap corev1.ConfigMap
 func (c *Client) AddNodeToJoiningNodes(ctx context.Context, nodeName string, componentsHash string, isControlPlane bool) error {
 	joiningNode := &unstructured.Unstructured{}
 
-	objectMetadataName := nodeName
+	compliantNodeName := k8sCompliantHostname(nodeName)
+
+	// JoiningNodes referencing a worker node are named after the worker node.
+	// JoiningNodes referencing the control-plane node are named "control-plane".
```
katexochen, line 92: I don't understand what this comment refers to. The line below seems somewhat unrelated, we always assign the same compliant name, no matter if this is a control plane or not.

joinservice/internal/kubernetes/kubernetes.go
katexochen: Also add a small unit test for this function, as it is easy to test.

katexochen (changes requested):

joinservice/internal/kubernetes/kubernetes.go
3u13r: The comment refers to the whole code block. The first line explains the first two lines and the second line refers to the `if`. I wanted to keep the comment together as to highlight the difference.

katexochen (approved):
Lgtsanta... :santa:

PR #818, opened 2022-12-21
Only upgrade helm releases if versions changed

daniel-weisse:
### Proposed change(s)

- Adapts the `helm.Upgrade` method to only run upgrades for a chart if the local chart version is greater than the one currently installed in Constellation

### Checklist

- ~[ ] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)~
- [x] Link to Milestone

### DIFF
--- a/cli/internal/helm/client.go
+++ b/cli/internal/helm/client.go
@@ -16,6 +16,7 @@ import (
 	"github.com/edgelesssys/constellation/v2/internal/constants"
 	"github.com/edgelesssys/constellation/v2/internal/deploy/helm"
 	"github.com/pkg/errors"
+	"golang.org/x/mod/semver"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/cli"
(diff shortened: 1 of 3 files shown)


cli/internal/helm/client.go
```diff
@@ -215,6 +215,29 @@ func (c *Client) updateOperatorCRDs(ctx context.Context, chart *chart.Chart) err
...
+	}
+	if !strings.HasPrefix(newVersion, "v") {
+		newVersion = "v" + newVersion
+	}
+
+	// If the current version is not a valid semver,
+	// we cant compare it to the new version.
+	// -> We can't upgrade.
+	if !semver.IsValid(currentVersion) {
+		return false
```
derpsteb, line 232: Maybe treat this as an error so it's easier to spot?

derpsteb (approved):

cli/internal/helm/client.go
daniel-weisse: I agree. Adjusted the debug messages to be a bit clearer.

Ideally I would like the user to see what exactly is happening when an upgrade is performed, e.g. what updates are actually applied. 
But I guess that will be implemented once we have the complete Constellation updates ready to go (as its already defined in the rfc).
Imo we should not hack in any extra logging (apart from debug messages) at the current time, and leave this open for the future.

PR #819, opened 2022-12-21
AB#2581: Move Konnectivity socket to non-persistent /run

Nirusu:
### Proposed change(s)
- Move Konnectivity socket from default /etc/kubernetes/konnectivity-server/konnectivity-server.socket (persistent) to /run/konnectivity-server/konnectivity-server.socket (non-persistent)

### Related issue
- https://github.com/kubernetes-sigs/apiserver-network-proxy/issues/443

While this is not a fix for the cause of this issue (which is still unknown), it cam serve as a mitigation as the Konnectivity server socket should not persist anymore across reboots of a node.


3u13r (approved):
LGTM

PR #826, opened 2022-12-24
gcp: unofficial support for internal LBs

3u13r:
### Proposed change(s)
- Scan global and now also regional forwarding rules for our Constellation endpoint. 
- Add unofficial example Terraform

I didn't want to change #806 since we gave out a link pointing to the Terraform on this branch. 
In the best scenario be could deploy the "internal" infrastructure and for normal deployments simply deploy some sort of gateway. I haven't found this possibility for internal *proxy* load balancers yet. Without such a solution the infrastructure is a bit different. How should this be tested in the future?  

### Checklist

- [x] Link to Milestone


datosh (approved):
LGTM

PR #839, opened 2022-12-29
upgrade: support Kubernetes components

3u13r:
### Proposed change(s)

- Rename `NodeImage` and friends to `NodeVersion` 
- Add Kubernetes components to `NodeVersion`
- Let the JoinService look up the current components ConfigMap in the `NodeVersion` custom resource instead of having a separate `k8s-version` ConfigMap

Discussion: Should we announce this feature in the release notes already? 
My opinion: No, since it is part of the complete upgrade feature which needs to be fully implemented, documented, and tested.

How to test:
1. Build the JoinService and node operator from this branch and use it in `versions.go`
2. Create  a valid `k8s-components-sha256-ABC` ConfigMap (Download links and hashes need to match+work)
3. Reference new ConfigMap in `NodeVersion` named `node-versioin`
4. Observe the update process (see. `NodeVersion` CRD, `PendingNode` CRD, and `Noes`)

What works best for me is to downgrade the patch version of the Kubernetes components in the `versions.go`. Then create a Constellation and upgrade to the latest patch version. This version should be visible via `kubectl get nodes -A -o wide`.

### Checklist

- [x] Link to Milestone


malt3:
Breaking change: Renaming the custom resource from `NodeImage` to `NodeVersion` could break a cluster during upgrade. While this might work fine by correctly patching the CRDs and restarting the operator to trigger a recreation of the resources, this is something we should test.

operators/constellation-node-operator/internal/deploy/deploy.go
```diff
@@ -110,22 +112,72 @@ func createAutoscalingStrategy(ctx context.Context, k8sClient client.Writer, pro
...
+			names = append(names, configMap.Name)
+		}
+	}
+	if len(names) == 1 {
+		return names[0], nil
+	} else if len(names) == 0 {
+		return "", fmt.Errorf("no configmaps found")
+	}
+
+	return "", fmt.Errorf("multiple configmaps found")
```
malt3, line 180:
```suggestion
	if len(names) == 1 {
		return names[0], nil
	} else if len(names) == 0 {
		return "", fmt.Errorf("no configmaps found")
	}

	return "", fmt.Errorf("multiple configmaps found")
```

nit: convert to switch

operators/constellation-node-operator/internal/deploy/deploy.go
```diff
@@ -110,22 +112,72 @@ func createAutoscalingStrategy(ctx context.Context, k8sClient client.Writer, pro
...
+		return nil
+	}); err != nil {
+		return fmt.Errorf("updating nodeversion: %w", err)
+	}
+
+	return nil
+}
+
+// findK8sComponentsConfigMap finds the k8s-components configmap in the kube-system namespace.
+// It returns an error if there is no or multiple configmaps matching the prefix "k8s-components".
```
malt3, line 161: I'm unsure if the error case of multiple configmaps is acceptable. Let's say we rename the `NodeVersion` CRD or it gets deleted for some other reason in a running cluster. If multiple instances of the components config map exist at this point, we will have a broken cluster.
Maybe we should instead simply assume the latest?

malt3:
Testing works (even between two minor versions).

malt3 (approved):
LGTM

PR #845, opened 2023-01-02
ci: add e2e test for constellation recover

stdoutput:
### Proposed change(s)

- Add an end-to-end test for `constellation recover`
- Add cluster UID to Terraform outputs

### How the test works:
- Restart one worker node and wait for it to rejoin the cluster
- Restart all control plane nodes 
- Execute `constellation recover`

### Checklist

- [x] Link to Milestone


.github/actions/constellation_recover/action.yml
```diff
@@ -0,0 +1,102 @@
...
+  using: "composite"
+  steps:
+    # Get Constellation UID
+    - name: Get Constellation cluster UID
+      shell: bash
+      run: |
+        CONSTELLATION_UID=$(yq -r ".uid" constellation-id.json)
+        echo "CONSTELLATION_UID=$CONSTELLATION_UID" >> $GITHUB_ENV
+      env:
+        KUBECONFIG: ${{ inputs.kubeconfig }}
```
datosh, line 31: Kubeconfig is not used in this step

```suggestion
```

.github/actions/constellation_recover/action.yml
```diff
@@ -0,0 +1,102 @@
...
+    # Restart worker node machine and wait for it to rejoin the cluster
+    - name: Restart worker node on GCP
```
datosh, line 33: Have you explored some options for cloud provider independent restarts?

My first idea would be to [use `kubectl debug`](https://constellation-docs.edgeless.systems/constellation/workflows/troubleshooting#connect-to-nodes) and execute a reboot from the node itself. Do you think this could work and remove CSP specific implementation?

.github/actions/constellation_recover/action.yml
```diff
@@ -0,0 +1,102 @@
...
+        kubectl wait --for=condition=ready --all nodes --timeout=10m
+      env:
+        KUBECONFIG: ${{ inputs.kubeconfig }}
+    - name: Restart worker node on Azure
+      if: ${{ inputs.cloudProvider == 'azure' }}
+      shell: bash
+      run: |
+        echo Shutting down worker instance: e2e-test-${CONSTELLATION_UID}-worker
+        az vmss restart --instance-ids 0 --name e2e-test-${CONSTELLATION_UID}-worker --resource-group ${{ inputs.resourceGroup }}
+        kubectl wait --for=condition=ready --all nodes --timeout=10m
```
datosh, line 50: I am not sure, but is the K8s API quick enough to see the node as restarting (i.e. unhealthy) for this to actually wait for the restart to finish? I could image this check going right through without ever seeing the node go unhealthy. 

I think a more robust assert should wait for a node to show up as unhealthy, and then healthy again.

cli/internal/terraform/terraform.go
```diff
@@ -74,39 +74,59 @@ func (c *Client) PrepareWorkspace(path string, vars Variables) error {
...
+
+	return CreateOutput{
+		IP:     ip,
+		Secret: secret,
+		UID:    uid,
+	}, nil
+}
+
+// CreateOutput contains the Terraform output values of a cluster creation.
+type CreateOutput struct {
```
datosh, line 126: 👍 for refactoring this code right away instead of adding another returned `string`

datosh (changes requested):
We should also consider adding this test to the [weekly e2e test](https://github.com/edgelesssys/constellation/blob/main/.github/workflows/e2e-test-weekly.yml). Only running this on latest supported K8s version should be fine.

.github/actions/constellation_recover/action.yml
stdoutput: @daniel-weisse what do you think? Is there any particular reason for not doing so on the worker node restart?

.github/actions/constellation_recover/action.yml
stdoutput: It did so on my local tests, but I get the point. I think about doing something like:
```
kubectl wait --for=condition=Ready=false --timeout=10m e2e-test-${CONSTELLATION_UID}-worker
kubectl wait --for=condition=Ready=true --timeout=10m --all nodes```

File: .github/docs/README.md [hunk 5]
```diff
@@ -104,3 +105,48 @@ The following secrets need to be defined:
...
+| Branch  | Constellation_Testing | ref-cli          | 2022.0912.123456 | `az sig image-version list --resource-group constellation-images --gallery-name Constellation_Testing_CVM --gallery-image-definition $(go run $(git rev-parse --show-toplevel)/hack/pseudo-version/pseudo-version.go -print-branch) --query "sort_by([], &publishingProfile.publishedDate)[].id" -o table` | `az sig image-version list-community --public-gallery-name ConstellationCVM-d1905bb0-a66c-497e-a9e6-4410ca7e3701 --gallery-image-definition $(go run $(git rev-parse --show-toplevel)/hack/pseudo-version/pseudo-version.go -print-branch) --location northeurope` |
```
katexochen, line 152: Note that this is documented in the rfc in #788. Imo we shouldn't duplicate this information here.

.github/actions/constellation_recover/action.yml
stdoutput: I couldn't get the reboot of the host machine to work from inside a pod. Since the time left until 2.4.0 release is running out, I think we should leave this as is for now, while I will create a separate ticket to find a solution for this after the release. I think we should prefer having the test with csp-specific code over not having a test, but please let me know what you think here

katexochen:
@stdoutput If this PR is related to a feature like the one you linked, it should be labeled with `feature`, not with `no-changelog`. Also note that the linked issue will be closed as soon as this PR is merged, which is likely not what you want...

stdoutput:
> @stdoutput If this PR is related to a feature like the one you linked, it should be labeled with `feature`, not with `no-changelog`. Also note that the linked issue will be closed as soon as this PR is merged, which is likely not what you want...

I thought non-user-facing things should not be included in the changelog? Also, shouldn't the issue only close once all linked PRs are closed?

katexochen:
> I thought non-user-facing things should not be included in the changelog? 

Hm, good point. The issue is tagged with feature... not sure what that means. xD Maybe don't label it at all, so it will be listed under "other changes"?

> Also, shouldn't the issue only close once all linked PRs are closed?

Sadly, that isn't the case as far as I know. In the development section it states

> Successfully merging this pull request may close these issues.

stdoutput:
> Sadly, that isn't the case as far as I know. In the development section it states

True.. It says "May be fixed by 875 **or** 845". I will just remove the linking on both PRs

katexochen:
> True.. It says "May be fixed by 875 or 845". I will just remove the linking on both PRs

You can just mention the issue in the PR description, that generates at least a reference in the issue to the PR.

stdoutput:
https://github.com/edgelesssys/constellation/actions/runs/3902853617/jobs/6666472986

Successful run of the test

.github/actions/constellation_recover/action.yml
```diff
@@ -0,0 +1,74 @@
+name: Constellation recover
```
katexochen, line 1: Could you move this file to `e2e_recover` ? All payloads for `e2e_test` should start with `e2e_`

.github/actions/constellation_recover/action.yml
```diff
@@ -0,0 +1,74 @@
...
+      env:
+        KUBECONFIG: ${{ inputs.kubeconfig }}
+    - name: Restart all control plane nodes
+      shell: bash
+      run: |
+        CONTROL_PLANE_NODES=$(kubectl get nodes --selector='node-role.kubernetes.io/control-plane' -o json | jq '.items[].metadata.name' -r)
+        for CONTROL_PLANE_NODE in ${CONTROL_PLANE_NODES}
+        do
+        kubectl debug node/$CONTROL_PLANE_NODE --image=ubuntu -- bash -c "echo reboot > temp.txt && chroot /host < temp.txt"
+        done
```
katexochen, line 43:
```suggestion
        for CONTROL_PLANE_NODE in ${CONTROL_PLANE_NODES}; do
            kubectl debug node/$CONTROL_PLANE_NODE --image=ubuntu -- bash -c "echo reboot > temp.txt && chroot /host < temp.txt"
        done
```

.github/actions/constellation_recover/action.yml
```diff
@@ -0,0 +1,74 @@
...
+      run: |
+        timeout=600
+        start_time=$(date +%s)
+        count=0
+        while true; do
+            output=$(constellation recover --master-secret=${{ inputs.masterSecret }})
+            if echo "$output" | grep -q "Pushed recovery key."; then
+                echo "$output"
+                i=$(echo "$output" | grep -o "Pushed recovery key." | wc -l | sed 's/ //g')
+                count=$(($count+$i))
```
katexochen, line 57:
```suggestion
                count=$((count+i))
```

.github/actions/constellation_recover/action.yml
```diff
@@ -0,0 +1,74 @@
...
+                if [ $count -eq ${{ inputs.controlNodesCount }} ]; then
```
katexochen, line 58:
```suggestion
                if [[ $count -eq ${{ inputs.controlNodesCount }} ]]; then
```

.github/actions/constellation_recover/action.yml
```diff
@@ -0,0 +1,74 @@
...
+                    exit 0
+                fi
+            fi
+
+            current_time=$(date +%s)
+            if ((current_time - start_time > timeout)); then
+                echo "Timeout hit"
```
katexochen, line 65: Add information on what timed out.

.github/actions/constellation_recover/action.yml
```diff
@@ -0,0 +1,74 @@
...
+        kubectl debug node/$CONTROL_PLANE_NODE --image=ubuntu -- bash -c "echo reboot > temp.txt && chroot /host < temp.txt"
+        done
+      env:
+        KUBECONFIG: ${{ inputs.kubeconfig }}
+    - name: Constellation recover
+      shell: bash
+      run: |
+        timeout=600
+        start_time=$(date +%s)
+        count=0
```
katexochen, line 51:
```suggestion
        recovered=0
```

.github/actions/e2e_test/action.yml
```diff
@@ -212,3 +212,15 @@ runs:
       with:
         cloudProvider: ${{ inputs.cloudProvider }}
         osImage: ${{ inputs.osImage }}
+
+    - name: Run recover test
+      if: inputs.test == 'recover'
+      id: constellation-recover
```
katexochen, line 218:
```suggestion
```
id isn't needed

katexochen (changes requested):

stdoutput:
[Successful run of the test](https://github.com/edgelesssys/constellation/actions/runs/3937931747/jobs/6736041724)

katexochen (approved):

datosh (approved):
Oh my - I never actually submitted my review - sorry for that! 

Thanks for all the improvements & changes you did over the last days. 
I think the only originial comment from my review that still applies is that the name of the file could be a bit more 'speaking'. 

LGTM!

stdoutput:
last [successful run](https://github.com/edgelesssys/constellation/actions/runs/3956909096/jobs/6776705160)

merging this

PR #847, opened 2023-01-02
cli: create local backups before helm-upgrade

derpsteb:
### Proposed change(s)
- Backup all CRDs and CRs before executing a helm-upgrade. 

### Related issue
- #631 

### Additional info
- These backups can be used if an unintended/broken change to the CRDs was made. For all standard resources a helm-rollback can fix the problem. However, when changing CRDs, all resources of that type are implicitly deleted. They can not be recovered with helm-rollback. 
- [GCP 1+2, sonobuoy quick](https://github.com/edgelesssys/constellation/actions/runs/3823689899)
- [AWS, 1+2, sonobuoy quick](https://github.com/edgelesssys/constellation/actions/runs/3823691466)

### Checklist

- [x] Link to Milestone


cli/internal/helm/client.go
```diff
@@ -9,34 +9,33 @@ package helm
...
 // Client handles interaction with helm.
 type Client struct {
-	config    *action.Configuration
-	crdClient *apiextensionsclient.Clientset
-	log       debugLog
+	config  *action.Configuration
+	kubectl crdClient
+	fs      afero.Fs
+	log     debugLog
 }
```
daniel-weisse, line 35: Any specific reason for not using `file.Handler` for file interactions in the `Client`?
As far as I see, this code only makes use of `MkdirAll` and `WriteFile`, which are both supported by `file.Handler`.

cli/internal/helm/backup_test.go
```diff
@@ -0,0 +1,140 @@
...
+	testCases := map[string]struct {
+		crd          apiextensionsv1.CustomResourceDefinition
+		resource     unstructured.Unstructured
+		expectedFile string
+		getCRsError  error
+		wantError    bool
+	}{
+		"success": {
+			crd:          apiextensionsv1.CustomResourceDefinition{Spec: apiextensionsv1.CustomResourceDefinitionSpec{Names: apiextensionsv1.CustomResourceDefinitionNames{Plural: "foobars"}, Group: "some.group", Versions: []apiextensionsv1.CustomResourceDefinitionVersion{{Name: "versionZero"}}}},
+			resource:     unstructured.Unstructured{Object: map[string]any{"metadata": map[string]any{"name": "foobar"}}},
```
daniel-weisse, line 76:
```suggestion
			crd: apiextensionsv1.CustomResourceDefinition{
				Spec: apiextensionsv1.CustomResourceDefinitionSpec{
					Names: apiextensionsv1.CustomResourceDefinitionNames{Plural: "foobars"},
					Group: "some.group",
					Versions: []apiextensionsv1.CustomResourceDefinitionVersion{{Name: "versionZero"}},
				},
			},
			resource:     unstructured.Unstructured{Object: map[string]any{"metadata": map[string]any{"name": "foobar"}}},
```
Consider splitting on multiple lines to improve readability

cli/internal/helm/backup_test.go
```diff
@@ -0,0 +1,140 @@
...
+	getCRsError  error
+}
+
+func (c stubCrdClient) Initialize(kubeconfig []byte) error {
+	return nil
+}
+
+func (c stubCrdClient) ApplyCRD(ctx context.Context, rawCRD []byte) error {
+	return nil
+}
```
daniel-weisse, line 126:
```suggestion
type stubCrdClient struct {
	crds         []apiextensionsv1.CustomResourceDefinition
	getCRDsError error
	crs          []unstructured.Unstructured
	getCRsError  error
	crdClient
}
```
You can embed the `crdClient` interface to save on defining unused functions for the testing struct.
Note that this will panic if one of the functions is called (which might be useful when the code is updated in the future to make use of them).

cli/internal/helm/client.go
```diff
@@ -45,7 +44,16 @@ func NewClient(kubeConfigPath, helmNamespace string, client *apiextensionsclient
 		return nil, fmt.Errorf("initializing config: %w", err)
 	}
 
-	return &Client{config: actionConfig, crdClient: client, log: log}, nil
+	kubeconfig, err := os.ReadFile(kubeConfigPath)
+	if err != nil {
+		return nil, fmt.Errorf("reading gce config: %w", err)
+	}
```
daniel-weisse, line 50: nit: call `afero.NewOsFs()` before this block. This allows us to use `fs.ReadFile()` instead of having to use the `os` package to do so.

cli/internal/helm/client.go
```diff
@@ -168,47 +176,15 @@ func (c *Client) GetValues(release string) (map[string]any, error) {
...
-	}
-	if groupVersionKind.Kind == "CustomResourceDefinition" {
-		return obj.(*v1.CustomResourceDefinition), nil
-	}
-
-	return nil, errors.New("parsed []byte, but did not find a CRD")
-}
-
-// updateCRDs walks through the dependencies of the given chart and applies
+// updateOperatorCRDs walks through the dependencies of the given chart and applies
```
daniel-weisse, line 179:
```suggestion
// updateCRDs walks through the dependencies of the given chart and applies
```

cli/internal/helm/client.go
```diff
@@ -168,47 +176,15 @@ func (c *Client) GetValues(release string) (map[string]any, error) {
...
 // the files in the dependencie's 'crds' folder.
 // This function is NOT recursive!
 func (c *Client) updateCRDs(ctx context.Context, chart *chart.Chart) error {
 	for _, dep := range chart.Dependencies() {
 		for _, crdFile := range dep.Files {
 			if strings.HasPrefix(crdFile.Name, "crds/") {
-				c.log.Debugf("Updating crd: %s", crdFile.Name)
-				err := c.ApplyCRD(ctx, crdFile.Data)
+				c.log.Debugf("updating crd: %s", crdFile.Name)
```
daniel-weisse, line 186:
```suggestion
				c.log.Debugf("Updating crd: %s", crdFile.Name)
```
I'm not sure if we have defined logging guidelines somewhere, but generally log messages start Uppercase and end without a `.`

daniel-weisse (changes requested):

cli/internal/helm/client.go
derpsteb: No, wasn't aware of it. Changed.

cli/internal/helm/backup.go
```diff
@@ -0,0 +1,80 @@
...
+				if err != nil {
+					return err
+				}
+				if err := c.fs.Write(path, yamlBytes); err != nil {
+					return err
+				}
+			}
+		}
+
+		c.log.Debugf("created backups for resource type: %s", crd.Name)
```
daniel-weisse, line 77:
```suggestion
		c.log.Debugf("Created backups for resource type: %s", crd.Name)
```

daniel-weisse (approved):

PR #853, opened 2023-01-03
cli: ask user to confirm cert-manager upgrades

derpsteb:
### Proposed change(s)
- Since cert-manager upgrades change CRDs that the user might depend on we now ask for confirmation of the step. An unintended version upgrade will delete resources (e.g. cert-manager issuer) cluster users created.

### Related issue
- #631 

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/upgradeexecute.go
+++ b/cli/internal/cmd/upgradeexecute.go
@@ -8,10 +8,12 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
 	"github.com/edgelesssys/constellation/v2/cli/internal/cloudcmd"
+	"github.com/edgelesssys/constellation/v2/cli/internal/helm"
 	"github.com/edgelesssys/constellation/v2/cli/internal/image"
 	"github.com/edgelesssys/constellation/v2/internal/attestation/measurements"
 	"github.com/edgelesssys/constellation/v2/internal/config"
@@ -30,6 +32,7 @@ func newUpgradeExecuteCmd() *cobra.Command {
 	}
 
 	cmd.Flags().Bool("helm", false, "Execute helm upgrade. This feature is still in development an may change without anounncement. Upgrades all helm charts deployed during constellation-init.")
+	cmd.Flags().BoolP("yes", "y", false, "Run upgrades without further confirmation. WARNING: might delete your resources in case you are using cert-manager in your cluster. Please read the docs.")
 	cmd.Flags().Duration("timeout", 3*time.Minute, "Change helm upgrade timeout. This feature is still in development an may change without anounncement. Might be useful for slow connections or big clusters.")
 	if err := cmd.Flags().MarkHidden("helm"); err != nil {
 		panic(err)
@@ -59,27 +62,35 @@ func runUpgradeExecute(cmd *cobra.Command, args []string) error {
 }
 
 func upgradeExecute(cmd *cobra.Command, imageFetcher imageFetcher, upgrader cloudUpgrader, fileHandler file.Handler) error {
-	configPath, err := cmd.Flags().GetString("config")
+	flags, err := parseUpgradeExecuteFlags(cmd)
 	if err != nil {
-		return err
+		return fmt.Errorf("parsing flags: %w", err)
 	}
-	conf, err := config.New(fileHandler, configPath)
+	conf, err := config.New(fileHandler, flags.configPath)
 	if err != nil {
 		return displayConfigValidationErrors(cmd.ErrOrStderr(), err)
 	}
 
-	helm, err := cmd.Flags().GetBool("helm")
-	if err != nil {
-		return err
-	}
-	if helm {
-		timeout, err := cmd.Flags().GetDuration("timeout")
-		if err != nil {
-			return err
+	if flags.helmFlag {
+		err = upgrader.UpgradeHelmServices(cmd.Context(), conf, flags.upgradeTimeout, helm.DenyDestructive)
+		if errors.Is(err, helm.ErrConfirmationMissing) {
+			if !flags.yes {
+				cmd.PrintErrln("WARNING: Upgrading cert-manager will destroy all custom resources you have manually created that are based on the current version of cert-manager.")
+				ok, askErr := askToConfirm(cmd, "Do you want to upgrade cert-manager anyway?")
+				if askErr != nil {
+					return fmt.Errorf("asking for confirmation: %w", err)
+				}
+				if !ok {
+					cmd.Println("Aborting upgrade.")
+					return nil
+				}
+			}
+			err = upgrader.UpgradeHelmServices(cmd.Context(), conf, flags.upgradeTimeout, helm.AllowDestructive)
 		}
-		if err := upgrader.UpgradeHelmServices(cmd.Context(), conf, timeout); err != nil {
+		if err != nil {
 			return fmt.Errorf("upgrading helm: %w", err)
 		}
+
 		return nil
 	}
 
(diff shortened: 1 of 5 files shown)


katexochen:
Imo, user interaction should happen in the cmd package and not somewhere in imports. Also I'd say it isn't a good idea to pass the cobra.Command out of the package, that seems like an anti-pattern. 

Instead, you could split the upgrade function in cloudcmd into a function that returns the needed information to do user interaction and a function that does the execution.

Or, even better, give upgrade a bool parameter to do x (what is it that needs to be confirmed? destruction of cert manager?), and pass false as default. Then return a typed error in case you would need to do x. You then ask the user for confirmation an rerun the function with the boolean set to true.

derpsteb:
Applied @katexochen last suggestion of using a custom error type.

cli/internal/cmd/upgradeexecute.go
```diff
@@ -68,18 +70,32 @@ func upgradeExecute(cmd *cobra.Command, imageFetcher imageFetcher, upgrader clou
...
+		if errors.Is(err, helm.ErrConfirmationMissing) {
+			cmd.PrintErrln("WARNING: Upgrading cert-manager will destroy all custom resources you have manually created that are based on the current version of cert-manager.")
+			ok, askErr := askToConfirm(cmd, "Do you want to upgrade cert-manager anyways?")
+			if askErr != nil {
+				return fmt.Errorf("asking for confirmation: %w", err)
+			}
+			if !ok {
+				cmd.Println("Aborting upgrade.")
+				return nil
+			}
```
katexochen, line 92: This should only happen if `yes` flag isn't set, compare create command.

katexochen: As this will require parsing another flag, I'd also propose to add a parseUpgradeExecFlags function in the same style as many other commands have.

cli/internal/cmd/upgradeexecute.go
```diff
@@ -68,18 +70,32 @@ func upgradeExecute(cmd *cobra.Command, imageFetcher imageFetcher, upgrader clou
...
+	if helmFlag {
 		timeout, err := cmd.Flags().GetDuration("timeout")
 		if err != nil {
 			return err
 		}
-		if err := upgrader.UpgradeHelmServices(cmd.Context(), conf, timeout); err != nil {
+		err = upgrader.UpgradeHelmServices(cmd.Context(), conf, timeout, false)
+		if errors.Is(err, helm.ErrConfirmationMissing) {
+			cmd.PrintErrln("WARNING: Upgrading cert-manager will destroy all custom resources you have manually created that are based on the current version of cert-manager.")
+			ok, askErr := askToConfirm(cmd, "Do you want to upgrade cert-manager anyways?")
```
katexochen, line 85:
```suggestion
			ok, askErr := askToConfirm(cmd, "Do you want to upgrade cert-manager anyway?")
```

cli/internal/helm/client.go
```diff
@@ -94,8 +94,11 @@ func (c *Client) currentVersion(release string) (string, error) {
 	return rel[0].Chart.Metadata.Version, nil
 }
 
+// ErrConfirmationMissing signals that an action requires user confirmation.
+var ErrConfirmationMissing = errors.New("action requires user confirmation")
+
 func (c *Client) upgradeRelease(
-	ctx context.Context, timeout time.Duration, chartPath, releaseName string, hasCRDs bool,
+	ctx context.Context, timeout time.Duration, chartPath, releaseName string, hasCRDs bool, confirmation bool,
```
katexochen, line 101: Calling the field confirmation might be a bit unspecific, but I'm currently missing a better term. Maybe `destructive`? Note that it is also possible to add named constant values for true and false that make reading the function call easier. 
```
// in helm.go
const AllowDestructive = true

// in caller.go
helmclient.Upgrade(ctx, config, timeout, helm.AllowDestructive)
// is much clearer to the reader than
helmclient.Upgrade(ctx, config, timeout, true)
```

cli/internal/helm/client.go
```diff
@@ -116,6 +119,10 @@ func (c *Client) upgradeRelease(
 		return nil
 	}
 
+	if releaseName == certManagerReleaseName && !confirmation {
+		return ErrConfirmationMissing
+	}
```
katexochen, line 124: Maybe add a unit test to cover this case and check the correct error is returned.

katexochen (changes requested):
Structure looks good so far.

cli/internal/helm/client.go
derpsteb: Nice. `AllowDestructive` seems pretty clear to me.

daniel-weisse:
Question about the general workflow of asking for user confirmation:
Why do we pass the confirmation flag to `helm.UpgradeHelmServices`?
Wouldn't it suffice to print the confirmation warning (and ask the user to confirm) if the CLI flag was set to `false` before calling `helm.UpgradeHelmServices()`?
That way the confirmation handling is contained to the main CLI function, and the helm package does not need to know about anything else.

katexochen:
> before calling helm.UpgradeHelmServices()

As I understood the problem, we can only know in the helm package if the action is destructive or not...

derpsteb:
It seemed like the nicer UX to me to only ask for confirmation if it is actually necessary. 
1. Present user with upgrades that are going to happen
2. If cert-manager will be upgraded: warn and ask for confirmation
3. upgrade

It would also be possible to do:
1. Warn about potential deletion of resources, ask for confirmation
2. Upgrade / Abort

There will be multiple instances where cert-manager will not be upgraded, but other components will be. For those cases we do not need to ask for confirmation. Also: maybe we want to allow users to skip cert-manager upgrades(?).

cli/internal/helm/client_test.go
```diff
@@ -110,3 +115,47 @@ func TestIsUpgrade(t *testing.T) {
...
+	}
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+
+			client := Client{crdClient: nil, actions: &stubActionWrapper{}, log: logger.NewTest(t)}
+			err := client.upgradeRelease(context.Background(), 0, certManagerPath, certManagerReleaseName, false, tc.allowDestructive)
+			if tc.wantError {
+				assert.Error(err)
```
katexochen, line 140: you can assert.ErrorIs()

cli/internal/helm/client.go
```diff
@@ -45,26 +53,26 @@ func NewClient(kubeConfigPath, helmNamespace string, client *apiextensionsclient
...
 // Canceling the context will prompt helm to abort and roll back the ongoing upgrade.
-func (c *Client) Upgrade(ctx context.Context, config *config.Config, timeout time.Duration) error {
-	if err := c.upgradeRelease(ctx, timeout, ciliumPath, ciliumReleaseName, false); err != nil {
+func (c *Client) Upgrade(ctx context.Context, config *config.Config, timeout time.Duration, confirmation bool) error {
+	if err := c.upgradeRelease(ctx, timeout, ciliumPath, ciliumReleaseName, false, confirmation); err != nil {
 		return fmt.Errorf("upgrading cilium: %w", err)
 	}
 
-	if err := c.upgradeRelease(ctx, timeout, certManagerPath, certManagerReleaseName, false); err != nil {
+	if err := c.upgradeRelease(ctx, timeout, certManagerPath, certManagerReleaseName, false, confirmation); err != nil {
```
katexochen, line 67: Should we rename all parameter to `allowDestructive`?

cli/internal/cmd/upgradeexecute.go
```diff
@@ -59,27 +62,35 @@ func runUpgradeExecute(cmd *cobra.Command, args []string) error {
...
+				cmd.PrintErrln("WARNING: Upgrading cert-manager will destroy all custom resources you have manually created that are based on the current version of cert-manager.")
+				ok, askErr := askToConfirm(cmd, "Do you want to upgrade cert-manager anyway?")
+				if askErr != nil {
+					return fmt.Errorf("asking for confirmation: %w", err)
+				}
+				if !ok {
+					cmd.Println("Aborting upgrade.")
+					return nil
+				}
+			}
```
katexochen, line 87: As far as I can tell the UpgradeHelmServices function is currently called twice in case the yes flag isn't set.

katexochen (changes requested):
Nice, only some minor things left :)

daniel-weisse (approved):

katexochen (approved):

PR #856, opened 2023-01-04
versionsapi: change image path

katexochen:

### Proposed change(s)

- As discussed in https://github.com/edgelesssys/constellation/pull/788#discussion_r1050526782, this changes the API path to an image of a version.
- Also fixing a bug where deletion of a single version wouldn't delete the artifacts in s3 bucket for that versions. This is now much easier after the API change.
- Also fixing a bug where the workflow call to add version to versions API would fail due to wrong syntax.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/build-os-image.yml
+++ b/.github/workflows/build-os-image.yml
@@ -151,7 +151,7 @@ jobs:
           {
             echo "imageVersion=${IMAGE_VERSION}"
             echo "imageName=ref/${REF}/stream/${STREAM}/${IMAGE_VERSION}"
-            echo "imageApiBasePath=constellation/v1/ref/${REF}/stream/${STREAM}/image/${IMAGE_VERSION}"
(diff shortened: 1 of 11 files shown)


katexochen:
OS image build for this branch: https://github.com/edgelesssys/constellation/actions/runs/3837499082

katexochen:
I've copied the last release to the a path using the new path schema, so the CLI for v2.4.0 can still find v2.3.0.
Old path: https://s3.console.aws.amazon.com/s3/buckets/cdn-constellation-backend?region=eu-central-1&prefix=constellation/v1/ref/-/stream/stable/image/v2.3.0/&showversions=false
New path: https://s3.console.aws.amazon.com/s3/buckets/cdn-constellation-backend?region=eu-central-1&prefix=constellation/v1/ref/-/stream/stable/v2.3.0/image/&showversions=false

malt3 (approved):
LGTM

PR #857, opened 2023-01-04
ci: ignore ntia.gov and twitter checking links

katexochen:

### Proposed change(s)

- Add failing but working links to ignore list.

### Checklist


Nirusu:
ntia.gov has the same issue as linux-kvm.org btw (they don't serve the full TLS chain)

katexochen:
@Nirusu grouped them so it is clear that this is the same kind of error

Nirusu (approved):
I am a bit wary about disabling link checking for twitter.com, but I guess we check our own Twitter account often enough if it's still there... My personal guess is that that they might have restricted some IP ranges due to some bot traffic... but we would have to investigate this more in depth.

LGTM.

PR #865, opened 2023-01-04
operator: add kubernetes cluster version to constellation-version

3u13r:
### Proposed change(s)

- change operator module to refer to the constellation/v2 module. This allows imports from this module.
- Let the CLI send the full Kubernetes semver from its `versions.go` to the Bootstrapper, which adds it to the K8s components ConfigMap. This is then read by the operator and included in the `constellation-version`.

I'm open for suggestion regarding an import alias other than `mainconstants`.

The cluster version in `constellation-version` is not yet used as part of the reconsolidation. It is also _somewhat_ foreign to be part of the **node**version  CRD, but I think keeping the K8s components reference and this version in the same place is the better trade off.

### Checklist

- [x] Link to Milestone


malt3 (approved):
LGTM

PR #871, opened 2023-01-05
ci: reproducible builds ko (no gcp)

leongross:
### Proposed change(s)

- Reproducible container images using `ko` for our microservices:
  - Node Operator
  - Join Service
  - KMS
  - Verification Service
- Base Image building using `apko`
  - Generate and upload `apko` images and signatues
  - Generate and upload SBOMS and signatures
  

### Checklist

- [x] Link to Milestone


leongross:
E2E: https://github.com/edgelesssys/constellation/commit/0f00716f6f8c804276c5b15d100a6bc2cce9d842

.github/actions/build_apko/action.yml
```diff
@@ -0,0 +1,61 @@
...
+    required: false
+  registry:
+    description: "Container registry to use"
+    default: "ghcr.io"
+    required: true
+  githubToken:
+    description: "GitHub authorization token"
+    required: true
+
+# Linux runner only (podman required)
```
datosh, line 16: Maybe this limitation should be stated in the description? It's the first thing I read when I use an Action I have not used before.

.github/workflows/build-apko-image.yml
```diff
@@ -0,0 +1,28 @@
+name: Build and upload all apko container images
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - "release/**"
+      - "feat/reproducible-builds-ko-nogcp"
```
malt3, line 9: I think this could be allowed on any branch. Some devs build microservices on their feature branch during testing. The same applied to all workflows.

.github/workflows/build-constellation-node-operator-ko.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Check out repository
+        id: checkout
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # tag=v3.1.0
+        with:
+          ref: ${{ github.head_ref }}
+
+      - name: Install operator-sdk
+        uses: ./.github/actions/install_operator_sdk
+        with:
+          version: v1.22.2
```
malt3, line 27: This can be skipped when building with ko

malt3:
General question: Should those workflows replace the existing workflows? Having duplicate workflows sounds hard to maintain.

.github/actions/build_apko/action.yml
```diff
@@ -0,0 +1,61 @@
...
+  steps:
+    - name: Install podman
+      shell: bash
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y podman
+
+    - name: Log in to the Container registry
+      id: podman-login
+      uses: redhat-actions/podman-login@v1
```
datosh, line 28: Dependent actions should be referenced by hash with version as comment.

```suggestion
      uses: redhat-actions/podman-login@7283c5d4ffb37e87a455332f0d68b144536d842f # v1.5
```

.github/actions/build_micro_service_ko/action.yml
```diff
@@ -0,0 +1,78 @@
...
+    description: "Password for Cosign private key"
+    required: false
+
+# Linux runner only
+runs:
+  using: "composite"
+  steps:
+    - name: Determine pseudo version
+      #if: ${{ inputs.pseudoVersion == 'true' }}
+      if: ${{ failure() }}
```
datosh, line 44: Probably left over debugging code? 

```suggestion
      if: ${{ inputs.pseudoVersion == 'true' }}
```

.github/workflows/build-apko-image.yml
datosh, line 9: before merging, same for other workflow files
```suggestion
```

.github/actions/build_apko/action.yml
```diff
@@ -0,0 +1,61 @@
...
+        else
+          echo "Building image for ${{ inputs.apkoConfig }}"
+
+          imageName=$(basename ${{ inputs.apkoConfig }} | cut -d. -f1 )
+          registry=${{ inputs.registry }}/edgelesssys/apko-${imageName}
+          outTar=$imageName.tar
+
+          podman run -v "$PWD":/work cgr.dev/chainguard/apko build ${{ inputs.apkoConfig }} $registry $outTar
+          podman load < $outTar
+          podman push $registry
```
datosh, line 60: We should also use cosign for the `apko` generated images.

`libcryptsetup` will be the final image, right? So users should be able to check it.

In case where we use `apko` images as a base image for further steps, we should check the signature before proceeding to use it.

datosh (commented):
Looks good to me so far!

Is there already a plan how to switch from the current build code to this one, since this PR removes none of the build code for containers we have right now? I don't think we should have / maintain two build paths on `main`, or do we require more testing before we are comfortable removing the current one?

.github/workflows/build-apko-image.yml
leongross: The initial ideas was to prevent unnecessary re-building without changes to the image definitions.
Since all the `apko` image definitions should reside in `image/apko` anyways, we could maybe remove the branch filtering and add path filtering for the said folder?

This could look like this:
```suggestion
          paths:
            - 'image/apko/*.yaml'

```

.github/actions/build_micro_service_ko/action.yml
leongross: Yes, this is a debug relic and should be removed.

.github/workflows/build-apko-image.yml
leongross: Yes, I added this branch filter to run the tests on this branch, this will be removed before merge.

malt3: I slightly misread the code and thought the branch filter was on the `workflow_dispatch`. Adding an additional path filter for the apko yamls on this workflow sounds like a good idea though :)

.github/workflows/build-joinservice-image-ko.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Check out repository
+        id: checkout
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # tag=v3.1.0
+        with:
+          ref: ${{ github.head_ref }}
+
+      - name: Setup Go environment
+        uses: actions/setup-go@c4a742cab115ed795e34d4513e2cf7d472deb55f # tag=v3.3.1
+        with:
+          go-version: "1.19.3"
```
katexochen, line 29:
```suggestion
          go-version: "1.19.4"
```

.github/workflows/build-kms-image-ko.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Check out repository
+        id: checkout
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # tag=v3.1.0
+        with:
+          ref: ${{ github.head_ref }}
+
+      - name: Setup Go environment
+        uses: actions/setup-go@c4a742cab115ed795e34d4513e2cf7d472deb55f # tag=v3.3.1
+        with:
+          go-version: "1.19.3"
```
katexochen, line 29:
```suggestion
          go-version: "1.19.4"
```

.github/workflows/build-kms-image-ko.yml
```diff
@@ -0,0 +1,44 @@
...
+
+      - name: Set up ko
+        uses: imjasonh/setup-ko@v0.6
```
katexochen, line 32: Pin hash.

.github/workflows/build-verification-service-ko.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Check out repository
+        id: checkout
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # tag=v3.1.0
+        with:
+          ref: ${{ github.head_ref }}
+
+      - name: Setup Go environment
+        uses: actions/setup-go@c4a742cab115ed795e34d4513e2cf7d472deb55f # tag=v3.3.1
+        with:
+          go-version: "1.19.3"
```
katexochen, line 27:
```suggestion
          go-version: "1.19.4"
```

.ko.yaml
```diff
@@ -0,0 +1,74 @@
...
+baseImageOverrides:
+  github.com/edgelesssys/constellation/operators/constellation-node-operator/v2: ghcr.io/edgelesssys/apko-alpine-base-user-65532
+  github.com/edgelesssys/constellation/v2/hack/qemu-metadata-api: ghcr.io/edgelesssys/apko-alpine-qemu-metadata-api
+
+builds:
+- id: kms
+  dir: .
+  main: ./kms/cmd
+  env:
+  - CGO_ENABLED=0
```
katexochen, line 12:
```suggestion
  - CGO_ENABLED=0
  - GOOS=linux
  - GOARCH=amd64
```
Is there a reason why kms and joinservice lack GOOS and GOARCH? Same for missing `-a `flag.

image/apko/alpine-base-user-65532.yaml
```diff
@@ -0,0 +1,25 @@
+contents:
```
katexochen, line 1: nit: Imo image dir should contain OS image only, maybe create a separate top lvl dir named `container`, `apko`, etc.

.github/actions/build_apko/action.yml
```diff
@@ -0,0 +1,159 @@
...
+          echo "Building image for ${{ inputs.apkoConfig }}"
+
+          imageName=$(basename ${{ inputs.apkoConfig }} | cut -d. -f1 )
+          registry=${{ inputs.registry }}/edgelesssys/apko-${imageName}
+          outTar=$imageName.tar
+
+          mkdir -p sboms/$imageName
+
+          # build the image
+            podman run \
```
daniel-weisse, line 108: indentation is off for the `podman run` part

.github/actions/build_apko/action.yml
leongross: Currently, `libcryptsetup` is only part of the [alpine-libcryptsetup.yaml](./image/apko/alpine-libcryptsetup.yaml) image that can be used as a base image for an `apko` container, if the application running in this container needs this libraray.

> I was trying to motivate adding cosign signatures to the `apko` images so that "users of the libcryptsetup image can check the signature on it via cosign verify".

I'm not sure if I understand what you mean. Which signatures should be stored in which images and in when should these signatures be verified?

.github/workflows/build-kms-image-ko.yml
katexochen:
```suggestion
          go-version: "1.19.5"
```

.github/actions/build_apko/action.yml
leongross: Due to issues with the `ghcr.io` registry I decided to roll back to docker, so I think this is obsolete.

.ko.yaml
leongross: No there is no reason for that. I think your idea is good to make it more explicit.

image/apko/alpine-base-user-65532.yaml
leongross: maybe we could have opinion of the code owner of the `images/` folder (@malt3) on that?

malt3: Probably the right call. I'd add a separate folder for this since they are used for very different purposes.

.github/actions/build_micro_service_ko/action.yml
```diff
@@ -0,0 +1,80 @@
...
+  cosignPassword:
+    description: "Password for Cosign private key"
+    required: false
+
+# Linux runner only
+runs:
+  using: "composite"
+  steps:
+    - name: Determine pseudo version
+      #if: ${{ inputs.pseudoVersion == 'true' }}
```
malt3, line 43:
```suggestion
```

.github/actions/build_ko/action.yml
```diff
@@ -0,0 +1,111 @@
...
+        cat <<EOF > container_data_ko.json
+        {
+          "container_full": "${container_full}",
+          "container_image": "${container_image}",
+          "container_sha256": "${container_sha256}"
+        }
+        EOF
+
+    - name: Upload Container Data # since github censors hashes that may share data with secrets, we need to upload the data as an artifact
+      uses: actions/upload-artifact@v2
```
malt3, line 108:
```suggestion
      uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
```

malt3 (approved):
LGTM. Slight changes will be needed to fully switch over the release pipeline but I would be in favor of merging as it is.

.github/actions/build_apko/action.yml
```diff
@@ -0,0 +1,167 @@
...
+        COSIGN_PASSWORD: ${{ inputs.cosignPassword }}
+      if: ${{ inputs.cosignPublicKey != '' && inputs.cosignPrivateKey != '' && inputs.cosignPassword != '' }}
+
+    - name: Upload SBOMs
+      uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
+      with:
+        name: sboms
+        path: sboms.zip
+      if: always()
+      continue-on-error: true
```
katexochen, line 167: Why does this need to run even on error?

leongross: This seems to be a debugging artifact, it should be safe to remove.

PR #879, opened 2023-01-05
operator: reconcile Kubernetes cluster version

3u13r:
### Proposed change(s)

- Operator now reconciles the Kubernetes cluster version
- Add additional `AwaitingAnnotation` Group if the JoiningNode controller hasn't annotated the node yet. This ensures that we don't overwrite annotation once we copy them over from their heir. 
- Add missing `PATH` to `upgrade-agent`. This is used for and inside `kubeadm`.

I tested this on Azure with 1-1 and 3-1 (C-W) nodes.

### Checklist

- [x] Link to Milestone


operators/constellation-node-operator/controllers/nodeversion_controller.go
```diff
@@ -85,6 +94,17 @@ func (r *NodeVersionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+
+	// Check if we need to upgrade the cluster version.
+	serverVer, err := r.ServerVersion()
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	// GitVersion is the semantic version of the Kubernetes server e.g. "v1.24.9"
+	if serverVer.GitVersion != desiredNodeVersion.Spec.KubernetesClusterVersion {
+		r.tryStartClusterVersionUpgrade(ctx, req.NamespacedName)
+	}
```
malt3, line 106: Not entirely convinced if this is necessary, but maybe this comparison should be less strict by either using [strings.EqualFold](https://pkg.go.dev/strings#EqualFold) or semver compare?
I'm concerned about cases where the `GitVersion` field uses slightly different representation of the same semantic version which would prevent this from ever reaching the desired state.

malt3 (approved):
Just one possible improvement. Otherwise LGTM

Issue #724, opened 2022-12-05
Document Terraform support


m1ghtym0:
Document how Constellation can be used with resource provisioning via Terraform instead of `constellation create`.

3u13r:
merged in https://github.com/edgelesssys/constellation/pull/1037

PR #897, opened 2023-01-06
hack: move terraform readmes into cli

katexochen:

### Proposed change(s)

- IAM creation is now part of the official CLI. The documentation in hack is replaced by the official docs.
- Moving developer documentation about iamlive to the devdoc directory.

### Checklist


thomasten:
Note that there's a reference from the docs to the AWS IAM Readme:
https://github.com/edgelesssys/constellation/blob/49534d463d912f0c28b4a27efaaed7de3f761cc3/docs/docs/workflows/config.md?plain=1#L221-L231

katexochen:
> Note that there's a reference from the docs to the AWS IAM Readme

Hm.. should we delete the link or keep the readmes but move them back to the terraform code?

thomasten:
> > Note that there's a reference from the docs to the AWS IAM Readme
> 
> Hm.. should we delete the link or keep the readmes but move them back to the terraform code?

Oh, I haven't noticed yet that the terraform code has been moved.

I don't have a strong opinion. For now, the links still work because they are pinned to v2.2. I guess #724 will change that part of the docs anyway and make that readme file finally obsolete? Then I would be fine with either your suggestion or keep your PR as is.

katexochen:
I'll move it back to the terraform code, might be a good starter for documenting the thing.

katexochen:
I didn't update the links, as there is no newer stable link to update to. In v2.3.0 the readme is in another folder as the terraform. Imo we should rethink how we can update references like this in the docs, or if we want them at all.

thomasten (approved):

PR #912, opened 2023-01-09
docs: remove outdated info on disk integrity

derpsteb:
### Proposed change(s)

- Just read this and seemed outdated to me. Please correct me if that's wrong

### Additional info
- The CSI drivers we are currently deploying offer integrity
protected disks. The state disk is also mounted with dm-integrity.


daniel-weisse:
>primitives for integrity protection in the CVM attacker model

`dm-integrity`, while providing integrity protection for disks, is not perfect, especially in the very strict CVM attacker model.
This could be fixed by using a more complex approach, e.g. dm-x. However, this is currently not supported by the Linux Kernel (or by us).

This hint is simply there to inform users, that even though we have integrity protection, the integrity protection we would like to use is not available.

PR #913, opened 2023-01-09
ci: add purge branch workflow

katexochen:

### Proposed change(s)

- Delete images on branch deletion
- Note that workflows with delete trigger cannot be tested on a branch

### Checklist

### DIFF
--- a/.github/workflows/purge-branch.yml
+++ b/.github/workflows/purge-branch.yml
@@ -1,0 +1,14 @@
+name: Purge branch
+
+on:
+  delete:
+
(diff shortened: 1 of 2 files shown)


malt3 (commented):
I think adding a branch filter here would be a good idea. I know the go code has one but I feel like we should signal our intent clearly.

katexochen:
> adding a branch filter here would be a good idea

This isn't possible for a delete event, according to actionlint.

malt3 (approved):

PR #919, opened 2023-01-09
rfc: add recovery section to eKMS rfc

derpsteb:
### Proposed change(s)
- Describe what needs to be changed to support eKMS during recovery

### DIFF
--- a/rfc/external-kms.md
+++ b/rfc/external-kms.md
@@ -30,6 +30,28 @@ It's designed to support different (cloud) KMS services as backend.
 Currently, only the ClusterKMS backend is used, which uses the master secret from the Kubernetes secrets to derive the DEKs.
 The clients of the CKMS are the join-service and the CSI drivers.
 
+### Recovery
+
+When a node boots, the disk-mapper will setup the encrypted state disk during initramfs.
+To decrypt an already existing disks, the disk-mapper will ask all available joinservice instances for a decryption key for it's current disk UUID.
+In case there is no join-service instance that can provide the decryption key, manual recovery becomes necessary.
+
+To enable manual recovery the disk-mapper starts a recovery server.
+The recovery server waits for a `recover` gRPC call.
+During a `recover` call the CLI will attest the measurements of the node.
+After successful attestation the CLI will provide a disk decryption key and measurement secret for the given disk UUID.
+The measurement secret, together with a measurement salt (not secret) is used to derive the clusterID.
+
+*Changes for eKMS; regarding disk decryption:*
+* Recovery server additionally accepts one KMS URI and one storage URI instead of a masterSecret.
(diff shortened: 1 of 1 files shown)


rfc/external-kms.md
```diff
@@ -30,6 +30,28 @@ It's designed to support different (cloud) KMS services as backend.
...
+
+*Changes for eKMS; regarding disk decryption:*
+* Recovery server additionally accepts one KMS URI and one storage URI instead of a masterSecret.
+* The current flow (provide masterSecret directly) will still be supported in case no eKMS is used.
+* If recovery server receives KMS+storage URIs it will query the storage URI for a DEK that matches it's current disk UUID.
+* Upon retrieval of the disk DEK it will ask the KMS URI for decryption of the DEK.
+
+*Changes for eKMS; regarding clusterID:*
+* Additionally to the DEK to decrypt the state disk, the recovery server will ask for a second DEK that is used as measurementSecret.
+* The measurementSecret DEK can be used like the existing value.
```
daniel-weisse, line 53: This is already implemented by the Joinservice, I don't think we need to make any changes for this.

rfc/external-kms.md
```diff
@@ -30,6 +30,28 @@ It's designed to support different (cloud) KMS services as backend.
...
+
+To enable manual recovery the disk-mapper starts a recovery server.
+The recovery server waits for a `recover` gRPC call.
+During a `recover` call the CLI will attest the measurements of the node.
+After successful attestation the CLI will provide a disk decryption key and measurement secret for the given disk UUID.
+The measurement secret, together with a measurement salt (not secret) is used to derive the clusterID.
+
+*Changes for eKMS; regarding disk decryption:*
+* Recovery server additionally accepts one KMS URI and one storage URI instead of a masterSecret.
+* The current flow (provide masterSecret directly) will still be supported in case no eKMS is used.
```
daniel-weisse, line 47: To make things a bit more simple I would opt for a unified systems, i.e. not keep the current recovery flow and the eKMS recovery flow separate.
This could be achieved by embedding the master secret in a cKMS storage/kms URI.

For exmaple:
Our current implementation defines the following cKMS URI `kms://cluster-kms`.
We could adapt this URI to `kms://cluster-kms?masterSecret=<masterSecret>`, which directly contains the master secret.
If we don't want to directly provide restarting nodes with the master secret, we could instead provide just the DEK used for state disk decryption.
Storage URI in this case would be the same as during init with cKMS: `storage://no-store`.
This way the proto message is independent of the used KMS backend. To recover, a node receives KMS and storage URI, uses this to set up a KMS connection (or simply extracts the key, in case of cKMS), requests a decryption key, and decrypts its state disk.

daniel-weisse (approved):

PR #924, opened 2023-01-10
dev-docs: move into top-level dir

m1ghtym0:
### Proposed change(s)

- Not sure about `dev-docs` as the folder name. I'm open for suggestions
- Do we want more sorting / sub-folders?


katexochen:
> Do we want more sorting / sub-folders?

Imo not necessary, can be done if the number of documents grows...

katexochen (approved):

PR #932, opened 2023-01-10
versions: repair hash generation

katexochen:
### Proposed change(s)

- Hash generator was broken, as the Go file/ast changed
- Moving the generator into an own package and removing the build tag, so it can be linted

### Checklist

### DIFF
--- a/internal/versions/hash-generator/generate.go
+++ b/internal/versions/hash-generator/generate.go
@@ -1,5 +1,3 @@
-//go:build ignore
-
 /*
 Copyright (c) Edgeless Systems GmbH
 
@@ -170,5 +168,5 @@ func main() {
 		log.Fatalf("no components lists found")
 	}
 
-	fmt.Printf("Successfully generated hashes for %d components in %d component lists.", componentCtr, componentListsCtr)
+	fmt.Printf("Successfully generated hashes for %d components in %d component lists.\n", componentCtr, componentListsCtr)
 }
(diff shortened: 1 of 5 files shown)


internal/versions/hash-generator/generate.go
```diff
@@ -159,5 +164,9 @@ func main() {
 	if err := os.WriteFile(filePath, buf.Bytes(), 0o644); err != nil {
 		log.Fatalf("error writing file %s: %s", filePath, err)
 	}
-	fmt.Println("Successfully generated hashes.")
+	if componentCtr == 0 {
+		log.Fatalf("no components lists found")
+	}
+
+	fmt.Printf("Successfully generated hashes for %d components in %d component lists.", componentCtr, componentListsCtr)
```
3u13r, line 171:
```suggestion
	fmt.Printf("Successfully generated hashes for %d components in %d component lists.\n", componentCtr, componentListsCtr)
```

internal/versions/hash-generator/generate_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+package main
+
+import (
+	"testing"
+
+	"go.uber.org/goleak"
+)
+
+func TestMain(m *testing.M) {
```
3u13r, line 15: What is the use of this file currently? Does the TestMain do anything if there are no test?

katexochen: Iirc yes, the code is compiled into the test.

PR #946, opened 2023-01-11
cli: iam destroy

miampf:
### Proposed change(s)

- Add the new ```iam destroy``` subcommand to destroy IAM configuration with the Constellation CLI
- Rename the ```DestroyCluster``` function in [terraform.go](https://github.com/edgelesssys/constellation/blob/feat/iam-destroy-user/cli/internal/terraform/terraform.go) to ```Destroy```

This change will allow users to delete their IAM configurations using the constellation CLI instead of the terraform CLI.

### Checklist

- [x] Update [docs](https://github.com/edgelesssys/constellation/blob/feat/iam-destroy-user/docs/docs/workflows/config.md)
- [x] Link to Milestone

### DIFF
--- a/cli/internal/cloudcmd/iam.go
+++ b/cli/internal/cloudcmd/iam.go
@@ -21,6 +21,7 @@ import (
 	"github.com/edgelesssys/constellation/v2/internal/cloud/gcpshared"
 	"github.com/edgelesssys/constellation/v2/internal/constants"
 	"github.com/edgelesssys/constellation/v2/internal/file"
+	"github.com/spf13/afero"
 )
 
 type IAMDestroyer struct {
@@ -35,21 +36,25 @@ func NewIAMDestroyer(ctx context.Context) *IAMDestroyer {
 		},
 	}
 }
-
-// DeleteGCPServiceAccountKeyFile deletes gcpServiceAccountKey.json if the IAM users in TerraformIAMWorkingDir and the file match up
-func (d *IAMDestroyer) DeleteGCPServiceAccountKeyFile(ctx context.Context, fsHandler file.Handler) (bool, error) {
+func (d *IAMDestroyer) RunDeleteGCPKeyFile(ctx context.Context) (bool, error) {
+	fsHandler := file.NewHandler(afero.NewOsFs())
 	cl, err := d.newTerraformClient(ctx)
 	if err != nil {
 		return false, err
 	}
 
+	return d.deleteGCPKeyFile(ctx, fsHandler, cl)
+}
+
+// DeleteGCPKeyFile deletes gcpServiceAccountKey.json if the IAM users in TerraformIAMWorkingDir and the file match up
+func (d *IAMDestroyer) deleteGCPKeyFile(ctx context.Context, fsHandler file.Handler, cl terraformClient) (bool, error) {
 	tfState, err := cl.Show(ctx)
 	if err != nil {
 		return false, err
 	}
 
 	if tfState.Values == nil {
-		return false, errors.New("No Values field in terraform state")
+		return false, errors.New("no Values field in terraform state")
 	}
 
 	saKeyJSON, containsKey := tfState.Values.Outputs["sa_key"]
--- a/cli/internal/cloudcmd/iam.go
+++ b/cli/internal/cloudcmd/iam.go
@@ -93,8 +93,8 @@ func (d *IAMDestroyer) deleteGCPKeyFile(ctx context.Context, fsHandler file.Hand
 	return true, nil
 }
 
-// DestroyIAMUser destroys the previously created IAM User and deletes the local IAM terraform files.
-func (d *IAMDestroyer) DestroyIAMUser(ctx context.Context) error {
+// DestroyIAMConfiguration destroys the previously created IAM User and deletes the local IAM terraform files.
+func (d *IAMDestroyer) DestroyIAMConfiguration(ctx context.Context) error {
 	cl, err := d.newTerraformClient(ctx)
 	if err != nil {
 		return err
--- a/cli/internal/cloudcmd/iam.go
+++ b/cli/internal/cloudcmd/iam.go
@@ -24,7 +24,8 @@ func DestroyIAMUser(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	return c.DestroyCluster(ctx)
+	c.Destroy(ctx)
+	return c.CleanUpWorkspace()
 }
 
 // IAMCreator creates the IAM configuration on the cloud provider.
--- a/cli/internal/cloudcmd/iam.go
+++ b/cli/internal/cloudcmd/iam.go
@@ -7,6 +7,8 @@ package cloudcmd
 
 import (
 	"context"
+	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"io"
 	"path"
@@ -15,7 +17,10 @@ import (
 	"github.com/edgelesssys/constellation/v2/cli/internal/iamid"
 	"github.com/edgelesssys/constellation/v2/cli/internal/terraform"
 	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
+	"github.com/edgelesssys/constellation/v2/internal/cloud/gcpshared"
 	"github.com/edgelesssys/constellation/v2/internal/constants"
+	"github.com/edgelesssys/constellation/v2/internal/file"
+	"github.com/spf13/afero"
 )
 
 type IAMDestroyer struct {
@@ -29,6 +34,47 @@ func NewIAMDestroyer(ctx context.Context) *IAMDestroyer {
 			return terraform.New(ctx, constants.TerraformIAMWorkingDir)
 		},
 	}
+}
+
+// DeleteGCPServiceAccountKeyFile deletes gcpServiceAccountKey.json if the IAM users in TerraformIAMWorkingDir and the file match up
+func (d *IAMDestroyer) DeleteGCPServiceAccountKeyFile(ctx context.Context) (bool, error) {
+	cl, err := d.newTerraformClient(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	tfState, err := cl.Show(ctx)
+	if err != nil {
+		return false, err
+	}
+	saKeyJSON, containsKey := tfState.Values.Outputs["sa_key"]
+	if !containsKey {
+		return false, nil
+	}
+	saKey, err := base64.StdEncoding.DecodeString(fmt.Sprintf("%v", saKeyJSON.Value))
+	if err != nil {
+		return false, err
+	}
+
+	var tfSaKey gcpshared.ServiceAccountKey
+	var fileSaKey gcpshared.ServiceAccountKey
+	fsHandler := file.NewHandler(afero.NewOsFs())
+
+	if err := json.Unmarshal(saKey, &tfSaKey); err != nil {
+		return false, err
+	}
+	if err := fsHandler.ReadJSON("gcpServiceAccountKey.json", &fileSaKey); err != nil {
+		return false, err
+	}
+
+	if tfSaKey != fileSaKey {
+		return false, nil
+	}
+
+	if err := fsHandler.Remove("gcpServiceAccountKey.json"); err != nil {
+		return false, err
+	}
+	return true, nil
 }
 
 // DestroyIAMUser destroys the previously created IAM User and deletes the local IAM terraform files.
--- a/cli/internal/cloudcmd/iam.go
+++ b/cli/internal/cloudcmd/iam.go
@@ -66,7 +66,7 @@ func (d *IAMDestroyer) deleteGCPKeyFile(ctx context.Context, fsHandler file.Hand
 
 	saKeyString, ok := saKeyJSON.Value.(string)
 	if !ok {
-		return false, nil
+		return false, errors.New("sa_key field in terraform state is not a string")
 	}
 	saKey, err := base64.StdEncoding.DecodeString(saKeyString)
 	if err != nil {
--- a/cli/internal/cloudcmd/iam.go
+++ b/cli/internal/cloudcmd/iam.go
@@ -20,7 +20,6 @@ import (
 	"github.com/edgelesssys/constellation/v2/internal/cloud/gcpshared"
 	"github.com/edgelesssys/constellation/v2/internal/constants"
 	"github.com/edgelesssys/constellation/v2/internal/file"
-	"github.com/spf13/afero"
 )
 
 type IAMDestroyer struct {
@@ -37,7 +36,7 @@ func NewIAMDestroyer(ctx context.Context) *IAMDestroyer {
 }
 
 // DeleteGCPServiceAccountKeyFile deletes gcpServiceAccountKey.json if the IAM users in TerraformIAMWorkingDir and the file match up
-func (d *IAMDestroyer) DeleteGCPServiceAccountKeyFile(ctx context.Context) (bool, error) {
+func (d *IAMDestroyer) DeleteGCPServiceAccountKeyFile(ctx context.Context, fsHandler file.Handler) (bool, error) {
 	cl, err := d.newTerraformClient(ctx)
 	if err != nil {
 		return false, err
@@ -58,12 +57,11 @@ func (d *IAMDestroyer) DeleteGCPServiceAccountKeyFile(ctx context.Context) (bool
 
 	var tfSaKey gcpshared.ServiceAccountKey
 	var fileSaKey gcpshared.ServiceAccountKey
-	fsHandler := file.NewHandler(afero.NewOsFs())
 
 	if err := json.Unmarshal(saKey, &tfSaKey); err != nil {
 		return false, err
 	}
-	if err := fsHandler.ReadJSON("gcpServiceAccountKey.json", &fileSaKey); err != nil {
+	if err := fsHandler.ReadJSON(constants.GCPServiceAccountKeyFile, &fileSaKey); err != nil {
 		return false, err
 	}
 
@@ -71,7 +69,7 @@ func (d *IAMDestroyer) DeleteGCPServiceAccountKeyFile(ctx context.Context) (bool
 		return false, nil
 	}
 
-	if err := fsHandler.Remove("gcpServiceAccountKey.json"); err != nil {
+	if err := fsHandler.Remove(constants.GCPServiceAccountKeyFile); err != nil {
 		return false, err
 	}
 	return true, nil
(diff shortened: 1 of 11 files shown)


cli/internal/cmd/iamdestroy.go
```diff
@@ -0,0 +1,91 @@
...
+
+	if proceed, err := deleteGCPServiceAccountKeyFile(cmd, destroyer, fsHandler); err != nil && !proceed {
+		return err
+	}
+
+	spinner.Start("Destroying IAM User", false)
+
+	if err := destroyer.DestroyIAMUser(cmd.Context()); err != nil {
+		return fmt.Errorf("couldn't destroy IAM User: %w", err)
+	}
```
daniel-weisse, line 51: Is it OK to delete the key file before we run `terraform destroy`.
IMO we should do it after to ensure we only delete the file if everything else went according to plan.
@katexochen any opinion on this?

cli/internal/cmd/iamdestroy.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+		if !ok {
+			fmt.Println("The destruction of the IAM user was aborted")
+			return nil
+		}
+	}
+
+	if proceed, err := deleteGCPServiceAccountKeyFile(cmd, destroyer, fsHandler); err != nil && !proceed {
+		return err
+	}
```
daniel-weisse, line 45: If `proceed == false` and `err == nil`, the function will not return here.

Maybe the confirm/abort flow from the iam-create command would work here?
```suggestion
 	proceed, err := deleteGCPServiceAccountKeyFile(cmd, destroyer, fsHandler)
	if err != nil {
		return err
	}
	if !proceed {
		cmd.Println("Destruction was aborted.")
		return nil
	}
```

Please note that this will probably require some changes in the `deleteGCPServiceAccountKeyFile` function.
Mainly the call to `fsHandler.Stat()` has to check if the received error was of type `os.ErrNotExists`, and if so, return `false, nil`

cli/internal/terraform/terraform.go
```diff
@@ -272,8 +277,8 @@ func (c *Client) CreateIAMConfig(ctx context.Context, provider cloudprovider.Pro
 	}
 }
 
-// DestroyCluster destroys a Constellation cluster using Terraform.
-func (c *Client) DestroyCluster(ctx context.Context) error {
+// Destroy destroys a data like a Constellation cluster using Terraform.
```
daniel-weisse, line 280:
```suggestion
// Destroy runs terraform destroy on the configured workspace.
```

cli/internal/cmd/iam.go
```diff
@@ -40,3 +41,18 @@ func newIAMCreateCmd() *cobra.Command {
 
 	return cmd
 }
+
+// NewIAMDestroyCmd returns a new cobra.Command for the iam destroy subcommand.
+func newIAMDestroyCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "destroy",
+		Short: "Destroy an IAM user and delete local terraform files for the IAM user",
+		Long:  "Destroy an IAM user and delete local terraform files for the IAM user",
```
thomasten, line 50:
```suggestion
		Long:  "Destroy an IAM user and delete local terraform files for the IAM user.",
```
make sure to be consistent with https://github.com/edgelesssys/constellation/blob/main/dev-docs/conventions.md#cli-reference

cli/internal/cmd/iamcreateaws.go
```diff
@@ -32,7 +32,7 @@ func newIAMCreateAWSCmd() *cobra.Command {
 		"Find available zones here: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-availability-zones. "+
 		"Note that we do not support every zone / region. You can find a list of all supported regions in our docs.")
 	must(cobra.MarkFlagRequired(cmd.Flags(), "zone"))
-	cmd.Flags().Bool("yes", false, "create the IAM configuration without further confirmation")
+	cmd.Flags().BoolP("yes", "y", false, "Create the IAM configuration without further confirmation")
```
thomasten, line 35:
```suggestion
	cmd.Flags().BoolP("yes", "y", false, "create the IAM configuration without further confirmation")
```
same here

cli/internal/cmd/iamcreateazure.go
```diff
@@ -31,7 +31,7 @@ func newIAMCreateAzureCmd() *cobra.Command {
 	must(cobra.MarkFlagRequired(cmd.Flags(), "region"))
 	cmd.Flags().String("servicePrincipal", "", "name of the service principal that will be created (required)")
 	must(cobra.MarkFlagRequired(cmd.Flags(), "servicePrincipal"))
-	cmd.Flags().Bool("yes", false, "create the IAM configuration without further confirmation")
+	cmd.Flags().BoolP("yes", "y", false, "Create the IAM configuration without further confirmation")
```
thomasten, line 34:
```suggestion
	cmd.Flags().BoolP("yes", "y", false, "create the IAM configuration without further confirmation")
```

cli/internal/cmd/iamdestroy.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/edgelesssys/constellation/v2/cli/internal/cloudcmd"
+	"github.com/edgelesssys/constellation/v2/internal/constants"
+	"github.com/edgelesssys/constellation/v2/internal/file"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+)
+
+func runDestroyIAMUser(cmd *cobra.Command, _args []string) error {
+	spinner := newSpinner(cmd.ErrOrStderr())
```
thomasten, line 18:
```suggestion
	spinner := newSpinner(cmd.ErrOrStderr())
	defer spinner.Stop()
```
I'm not familiar with the spinner, but from other code I see that it seems to need a deferred call to Stop()

katexochen (changes requested):
Some things left to do:

- [ ] Check if there is a running cluster. We should ensure that we call `constellation terminate` before we allow a IAM destroy. You can use the files that are deleted in `cmd/terminate.go` as indicator.
- [ ] Only ask user once for confirmation. We should check for the gcpServiceAccountKey before asking the user for confirmation, and state that the file will be deleted there.
- [ ] We should add some debug output like the other commands have.
- [ ] Deletion of the gcpKey file can be done directly in the `cmd` package, no need to have a method for that in `cloudcmd`. Maybe make the cloudcmd func return the gcpKey directly

cli/internal/cloudcmd/iam.go
```diff
@@ -15,9 +18,88 @@ import (
...
+	}
+
+	if tfState.Values == nil {
+		return false, errors.New("no Values field in terraform state")
+	}
+
+	saKeyJSON, containsKey := tfState.Values.Outputs["sa_key"]
+	if !containsKey {
+		return false, nil
+	}
```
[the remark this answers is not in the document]
miampf: I've added your proposed changes, but I only check if ```saKeyJSON``` is ```nil```

cli/internal/cmd/iamdestroy.go
miampf: This order is kind of needed since the key file is compared with the `sa_key` from the terraform state file to ensure the files belong to the same IAM group. I didn't encounter problems with using this order of deleting files, however, if it seems safer, I could rework this so that the `sa_key` is saved in a variable before the terraform files get deleted.

cli/internal/cmd/iamdestroy.go
miampf: Returning `true, nil` makes more sense to me since a user can be destroyed if the file doesn't exist. This means that the function could be used in Azure or AWS environments without problems. I've added an additional check to catch errors beside `os.ErrNotExist` and added  a check in `destroyIAMUser`, do you think that this works as an alternative?

cli/internal/cmd/iamdestroy.go
daniel-weisse: The problem I see is if something goes wrong during `terraform destroy` the key file is already gone.
But since we already check if the key file to be deleted matches the one from the terraform state I guess this is fine.

cli/internal/cmd/iamdestroy.go
```diff
@@ -0,0 +1,105 @@
...
+	fmt.Println("Successfully destroyed IAM configuration")
+	return nil
+}
+
+func deleteGCPServiceAccountKeyFile(cmd *cobra.Command, destroyer iamDestroyer, fsHandler file.Handler) (bool, error) {
+	if _, err := fsHandler.Stat(constants.GCPServiceAccountKeyFile); err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return false, err
+		}
+		return true, err // file just doesn't exist
```
daniel-weisse, line 88:
```suggestion
		return true, nil // file just doesn't exist
```
Should be `true, nil`, correct?
There should be no difference for the calling function if the file was successfully deleted or if it doesnt even exist in the first place

cli/internal/cmd/iamdestroy.go
katexochen: I'd say its fine, too. If something goes wrong during termination we are likely in a state where we cannot continue to use the IAM creds anyway. The only thing that should be possible is to repeat the termination until it succeeds.
Maybe it is even safer this way, I could image a scenario where terraform succeeds to delete the cloud resource of the gcpKey, but the deletion of another resource fails. In this case, the terraform state would be updated and wouldn't contain the gcpKey any longer, but we wouldn't have deleted the local gcpKey file and couldn't match it the next time we run terminate.

cli/internal/cmd/iamdestroy.go
miampf: Yes, sorry for that, I've corrected it

daniel-weisse (approved):
Looks good to me

Just a heads up: `cli/internal/cmd/iam.go` has undergone some changes (mainly the removal of `createiam<CSP>.go` files so there will be merge conflicts

cli/internal/cmd/iamdestroy.go
```diff
@@ -0,0 +1,158 @@
...
+	c.log.Debugf("Checking if %s exists", constants.GCPServiceAccountKeyFile)
+	_, err = fsHandler.Stat(constants.GCPServiceAccountKeyFile)
+	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return err
+		}
+	} else {
+		c.log.Debugf("%s exists", constants.GCPServiceAccountKeyFile)
+		gcpFileExists = true
+	}
```
katexochen, line 75:
```suggestion
	if errors.Is(err, os.ErrNotExist) {
		c.log.Debugf("%s exists", constants.GCPServiceAccountKeyFile)
		gcpFileExists = true
	} else if err != nil {
		return err
	} 
```

cli/internal/cmd/iamdestroy.go
```diff
@@ -0,0 +1,158 @@
...
+
+	if !yes {
+		// Confirmation
+		confirmString := "Do you really want to destroy your IAM configuration?"
+		if gcpFileExists {
+			confirmString += " (This will also delete " + constants.GCPServiceAccountKeyFile + ")"
```
katexochen, line 81:
```suggestion
			confirmString += fmt.Sprintf(" (This will also delete %q)", constants.GCPServiceAccountKeyFile)
```

cli/internal/cmd/iamdestroy.go
```diff
@@ -0,0 +1,158 @@
...
+			return err
+		}
+		if !ok {
+			cmd.Println("The destruction of the IAM configuration was aborted")
+			return nil
+		}
+	}
+
+	if gcpFileExists {
+		c.log.Debugf("Starting to delete %s", constants.GCPServiceAccountKeyFile)
```
katexochen, line 94:
```suggestion
		c.log.Debugf("Deleting %q", constants.GCPServiceAccountKeyFile)
```

cli/internal/cmd/iamdestroy.go
```diff
@@ -0,0 +1,158 @@
...
+
+func (c *destroyCmd) deleteGCPServiceAccountKeyFile(cmd *cobra.Command, destroyer iamDestroyer, fsHandler file.Handler) (bool, error) {
+	c.log.Debugf("Checking if %s exists", constants.GCPServiceAccountKeyFile)
+	if _, err := fsHandler.Stat(constants.GCPServiceAccountKeyFile); err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return false, err
+		}
+		c.log.Debugf("File %s doesn't exist", constants.GCPServiceAccountKeyFile)
+		return true, nil // file just doesn't exist
+	}
```
katexochen, line 126: This was already done in iamDestroy()

cli/internal/cmd/iamdestroy.go
```diff
@@ -0,0 +1,158 @@
...
+
+	if err := fsHandler.Remove(constants.GCPServiceAccountKeyFile); err != nil {
+		ok, err := askToConfirm(cmd, "The file gcpServiceAccountKey.json could not be deleted. Either it does not exist or the file belongs to another IAM configuration. Do you want to proceed anyway?")
+		if err != nil {
+			return false, err
+		}
+		if !ok {
+			return false, nil
+		}
+	}
```
katexochen, line 154: We shouldn't do a second user interaction here. Please decide for one way: either abort with an error, or print the error and continue anyway.

cli/internal/cmd/iamdestroy.go
```diff
@@ -0,0 +1,158 @@
...
+	c.log.Debugf("Getting service account key from the tfstate")
+	tfSaKey, err := destroyer.RunGetTfstateSaKey(cmd.Context())
+	if err != nil {
+		return false, err
+	}
+
+	c.log.Debugf("Checking if keys are the same")
+	if tfSaKey != fileSaKey {
+		return false, nil
+	}
```
katexochen, line 144: Can't we just not delete the file, inform the user and continue?

cli/internal/cmd/iamdestroy_test.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	newFsMissing := func() file.Handler {
+		fh := file.NewHandler(afero.NewMemMapFs())
+		return fh
+	}
+
+	fhWithAdminConf := file.NewHandler(afero.NewMemMapFs())
+	require.NoError(fhWithAdminConf.Write(constants.AdminConfFilename, []byte("")))
+	fhWithClusterIDFile := file.NewHandler(afero.NewMemMapFs())
+	require.NoError(fhWithClusterIDFile.Write(constants.ClusterIDsFileName, []byte("")))
```
katexochen, line 38: This is still not providing the needed isolation between test cases

cli/internal/cmd/iamdestroy_test.go
```diff
@@ -0,0 +1,231 @@
...
+		return fs
+	}
+	newFsExist := func() file.Handler {
+		fs := file.NewHandler(afero.NewMemMapFs())
+		require.NoError(fs.Write(constants.GCPServiceAccountKeyFile, []byte(gcpFile)))
+		return fs
+	}
+
+	fsInvalidJson := file.NewHandler(afero.NewMemMapFs())
+	require.NoError(fsInvalidJson.Write(constants.GCPServiceAccountKeyFile, []byte("asdf")))
```
katexochen, line 159: And again, this is global state between test cases

cli/internal/cmd/iamdestroy_test.go
```diff
@@ -0,0 +1,231 @@
...
+		"auth_uri": "",
+		"client_email": "",
+		"client_id": "",
+		"client_x509_cert_url": "",
+		"private_key": "",
+		"private_key_id": "",
+		"project_id": "",
+		"token_uri": "",
+		"type": ""
+	}
```
katexochen, line 145: Same as with the cloudcmd test, this can be stubbed.

katexochen (changes requested):

cli/internal/cloudcmd/iam_test.go
```diff
@@ -148,3 +153,175 @@ func TestIAMCreator(t *testing.T) {
...
+			} else {
+				assert.NoError(err)
+			}
+
+			if tc.wantValidSaKey {
+				var saKeyComp gcpshared.ServiceAccountKey
+				require.NoError(t, json.Unmarshal([]byte(gcpFile), &saKeyComp))
+
+				assert.Equal(saKey, saKeyComp)
+			}
```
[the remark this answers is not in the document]
miampf: This way we wouldn't check if we really don't have an error when we don't want an error. I would suggest adding `assert.NoError(err)` after `if tc.wantErr{...}`.

cli/internal/cmd/iamdestroy.go
miampf: If I'm not mistaken your proposal wouldn't work since it explicitly sets `gcpFileExists` to `true` if the file doesn't exist. I believe even with a logical not in front of `errors.Is(...)` this code wouldn't quite work out without having other if cases before it and then we are really similar to the code I wrote again. What do you think?

cli/internal/cmd/iamdestroy.go
miampf: If the keys don't match up, the configuration in the `iam-terraform` folder and the gcp IAM account are probably not the same. I don't think we should just continue since we might destroy the wrong IAM configuration. IMO we should either alert and ask the user if they want to continue or just abort.

katexochen: Why should we destroy the wrong one? We'd destroy the iam config we created, leaving other files we didn't create but the user created otherwise untouched. That sounds quite reasonable to me.

cli/internal/cloudcmd/iam_test.go
```diff
@@ -148,3 +153,198 @@ func TestIAMCreator(t *testing.T) {
...
+			if tc.wantDestroyCalled {
+				assert.True(destroyer.client.(*stubTerraformClient).destroyCalled)
+			} else {
+				assert.False(destroyer.client.(*stubTerraformClient).destroyCalled)
+			}
+			if tc.wantCleanupWorkspaceCalled {
+				assert.True(destroyer.client.(*stubTerraformClient).cleanUpWorkspaceCalled)
+			} else {
+				assert.False(destroyer.client.(*stubTerraformClient).cleanUpWorkspaceCalled)
+			}
```
katexochen, line 207:
```suggestion
			if tc.wantErr {
				assert.Error(err)
			} else {
				assert.NoError(err)
			}
			assert.Equal(tc.wantDestroyCalled, tc.tfClient.destroyCalled)
			assert.Equal(tc.wantCleanupWorkspaceCalled, tc.tfClient.cleanUpWorkspaceCalled)
```
You shouldn't just cast things, as this can panic at runtime. I'd suggest to just change the type of tc.tfClient to the concrete type.

katexochen (approved):
lgtm! :whale:

PR #958, opened 2023-01-12
dev-docs: Go package docs

daniel-weisse:
### Proposed change(s)

Add doc comments to Go packages to improve developer experience when working with new code

- To package docs are written using go doc comments, meaning they are directly embedded in the Go code
  - For a package named `pkgA`, the documentation will be in `pkgA/pkgA.go`. Where this file did not exist previously I have renamed or added it.
  - In some cases we use a simple `README.md` file instead. This is done in two cases:
    - when there is no actual Go package at that path, for example for the root folder of a Constellation component like `joinservice` or `bootstrapper`
    - the documentation includes mermaid graphs

### Additional info
`internal/deploy/user` was removed since it was not used by any other package.

So far I have only added documentation to most packages in `internal/` and packages I am codeowner of. (Excluding `kms/` since there is a larger rework inprogress, and hack files)
I would ask of the other codeowners of other files to add pkg docs where missing.


datosh:
> * `/e2e`, not sure if we require doc comments, since these are just small programs used in e2e tests
>   
>   * Maybe these should be moved to `/hack` ?

My understanding of `hack` is things we do not support to keep stable. I think `e2e` does not fit that description. If we want to move `e2e` from the root level I'd say it could go to `.github/e2e` since it is only used from actions? `e2e` will also grow in future with the addition of upgrade related tests.

katexochen:
> Maybe these should be moved to /hack ?

Imo it is good where it is, but needs a package doc comment, too.

datosh:
> > Maybe these should be moved to /hack ?
> 
> Imo it is good where it is, but needs a package doc comment, too.

I've added some package level docs for `e2e`. Let me know what you think 🙂

datosh (approved):

PR #959, opened 2023-01-12
operator: make test more resilient

3u13r:
### Proposed change(s)

- Tests are now successful 5 times in a row inside the CI: https://github.com/edgelesssys/constellation/actions/runs/3903687514


operators/constellation-node-operator/controllers/scalinggroup_controller.go
```diff
@@ -61,6 +62,8 @@ func NewScalingGroupReconciler(scalingGroupUpdater scalingGroupUpdater, client c
 func (r *ScalingGroupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	logr := log.FromContext(ctx)
 
+	time.Sleep(time.Second)
```
malt3, line 65:
```suggestion
```

I think this is an artifact from your tests

malt3 (approved):
Thanks for improving test reliability!

PR #962, opened 2023-01-13
cli: update default idkeydigest value

derpsteb:
### Proposed change(s)

We started experiencing new idkeydigest values yesterday evening. Azure recently (2 days ago) stated the fact that they want to support the current key for one year [1]. This morning we are only seeing the new idkeydigest on newly started machines. Therefore we are updating the expected key.

[1] https://github.com/Azure/confidential-computing-cvm-guest-attestation/blob/main/CVM-attestation-guidance.pdf

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -255,7 +255,7 @@ func Default() *Config {
 				InstanceType:         "Standard_DC4as_v5",
 				StateDiskType:        "Premium_LRS",
 				DeployCSIDriver:      func() *bool { b := true; return &b }(),
-				IDKeyDigest:          "57486a447ec0f1958002a22a06b7673b9fd27d11e1c6527498056054c5fa92d23c50f9de44072760fe2b6fb89740b696",
+				IDKeyDigest:          "0356215882a825279a85b300b0b742931d113bf7e32dde2e50ffde7ec743ca491ecdd7f336dc28a6e0b2bb57af7a44a3",
 				EnforceIDKeyDigest:   func() *bool { b := true; return &b }(),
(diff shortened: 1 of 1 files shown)


derpsteb:
Unittests are failing because the Azure mirror is down. Can reproduce locally.

internal/config/config.go
```diff
@@ -256,7 +256,7 @@ func Default() *Config {
 				StateDiskType:        "Premium_LRS",
 				DeployCSIDriver:      func() *bool { b := true; return &b }(),
 				IDKeyDigest:          "57486a447ec0f1958002a22a06b7673b9fd27d11e1c6527498056054c5fa92d23c50f9de44072760fe2b6fb89740b696",
```
malt3, line 258:
```suggestion
				IDKeyDigest:          "0356215882a825279a85b300b0b742931d113bf7e32dde2e50ffde7ec743ca491ecdd7f336dc28a6e0b2bb57af7a44a3",
```

Should we update this already?

derpsteb: I initially didn't because I thought it might change again. But I guess that's the only new value we saw for now.

Changed.

Nirusu:
Should be a candidate for 2.4.1 backport I suppose.

Also I would suggest removing the no-changelog label, it’s relevant for a new release if we keep it this way for now.

derpsteb:
Wasn't aware we want to keep it this way. Removing..

Nirusu:
Not sure if long-term, but if we currently want to do a patch release then I guess it makes sense for now (unless someone wants to veto on that?)

derpsteb:
Made obsolete by https://github.com/edgelesssys/constellation/pull/991

PR #970, opened 2023-01-13
docs: add missing asterisk reference

m1ghtym0:
### Proposed change(s)

- Add missing asterisk

### Checklist

- [ ] Link to Milestone


thomasten (approved):
I'm not sure if the asterisk in the table was intended to be a footnote reference, so I believe it's not strictly neeeded. But I don't have an opinion whether this is better or not, so it's fine with me.

PR #971, opened 2023-01-16
ci: delete old images from main ref

katexochen:
### Proposed change(s)

- This fixes the currently broken versionsapi action output.
- Add missing implementation of `versionsapi list --json` flag.
- Delete old images from main ref. Last 10 images are kept. Only a single image is deleted at a time, this keeps the action simple and the critical sequence isn't locked for long, so the likelihood of multiple other invocations during that period, which would cancel each other, is reduced.

### Checklist

### Next steps

- [ ] Frequently build OS images on main ref
- [ ] Delete other image streams on main (console, nightly)


.github/workflows/purge-main.yml
```diff
@@ -0,0 +1,69 @@
+name: Purge old images from main branch
+
+on:
+  workflow_dispatch:
+  #   branches:
+  #     - main
```
malt3, line 6:
```suggestion
```

malt3 (approved):

PR #980, opened 2023-01-16
versions: rename KmsImage to KeyServiceImage

derpsteb:
### Proposed change(s)

- Presumably because of the missed ko PR.


katexochen (approved):
Damn, you were just a second faster :laughing:  #981.

katexochen:
> Presumably because of the missed ko PR.

I've made some mistake during rebase.

PR #987, opened 2023-01-16
Refactor init & recovery to use KMS URI instead of mastersecret

derpsteb:
### Proposed change(s)

- Move code related to communication with KMS backends to `internal/kms`
- Use `internal/kms` in all places where we need to generate new secrets. 

### Additional info
- This refactoring is happening to improve our external KMS support. After streamlining the code that relies on KMS interaction in this PR, an upcoming PR will expose eKMS configs in the config and improve useability of the existing eKMS backends. 
- See commit messages for more background

### Checklist

- [x] Link to Milestone


cli/internal/cmd/recover.go
```diff
@@ -231,9 +191,9 @@ func (d *recoverDoer) setDialer(dialer grpcDialer, endpoint string) {
 	d.endpoint = endpoint
 }
 
-func (d *recoverDoer) setSecrets(getDiskKey func(string) ([]byte, error), measurementSecret []byte) {
-	d.getDiskKey = getDiskKey
-	d.measurementSecret = measurementSecret
+func (d *recoverDoer) setSecrets(masterSecret keyservice.MasterSecret) {
+	d.kmsURI = fmt.Sprintf(keyservice.ClusterKMSURI, masterSecret.EncodeKey(), masterSecret.EncodeSalt())
+	d.storageURI = keyservice.NoStoreURI
 }
```
daniel-weisse, line 197: Not a fan of having a dedicated function just to parse the master secret into KMS URIs, especially considering that this will most likely be removed once the option for eKMS is implemented.
Maybe try to refactor this to be more general, or move the URI parsing logic into a switch/case block, so its easily extended for different KMS backends

disk-mapper/internal/recoveryserver/server_test.go
```diff
@@ -71,59 +77,13 @@ func TestServe(t *testing.T) {
 
 func TestRecover(t *testing.T) {
 	testCases := map[string]struct {
-		initialMsg message
-		keyMsg     message
-		wantErr    bool
+		masterKey string
+		salt      string
+		wantErr   bool
```
daniel-weisse, line 82: Would probably make sense to refactor this function now to work on KMS URIS, so we don't have to refactor again once eKMS gets impemented

disk-mapper/recoverproto/recover.proto
```diff
@@ -5,16 +5,16 @@ package recoverproto;
 option go_package = "github.com/edgelesssys/constellation/v2/disk-mapper/recoverproto";
 
 service API {
-    rpc Recover(stream RecoverMessage) returns (stream RecoverResponse) {}
+    rpc Recover(RecoverMessage) returns (RecoverResponse) {}
```
daniel-weisse, line 8: Can you add some doc comments to this proto file?
You can follow [this example](https://github.com/edgelesssys/constellation/blob/83adbc621006671dc71ea60d3a58f19a4d642692/joinservice/joinproto/join.proto) from the dev-docs PR

internal/kms/kms/cluster/cluster.go
```diff
@@ -20,8 +20,8 @@ type KMS struct {
 }
 
 // New creates a new ClusterKMS.
-func New(salt []byte) *KMS {
-	return &KMS{salt: salt}
+func New(key []byte, salt []byte) *KMS {
+	return &KMS{masterKey: key, salt: salt}
 }
```
daniel-weisse, line 25:
```suggestion
func New(key []byte, salt []byte) (*KMS, error) {
	if len(key) = 0 || len(salt) == 0 {
		return errors.New("missing salt / master key")
	}
	return &KMS{masterKey: key, salt: salt}, nil
}
```
With the future removal of `CreateKEK`, we should ensure `New` creates a valid configuration

internal/kms/setup/setup.go
```diff
@@ -28,13 +28,29 @@ const (
...
+
+// EncodeKey returns the url-encoded key of the MasterSecret. This encoding is necessary when sending the mastersecret as part of the KMS URI.
+func (m *MasterSecret) EncodeKey() string {
+	return base64.URLEncoding.EncodeToString(m.Key)
+}
+
+// EncodeSalt returns the url-encoded salt of the MasterSecret. This encoding is necessary when sending the mastersecret as part of the KMS URI.
+func (m *MasterSecret) EncodeSalt() string {
+	return base64.URLEncoding.EncodeToString(m.Salt)
+}
```
daniel-weisse, line 52:
```suggestion
// MasterSecret holds the master key and salt for deriving keys.
type MasterSecret struct {
	Key  []byte `json:"key"`
	Salt []byte `json:"salt"`
}

// EncodeToURI returns an URI encoding the master secret.
func (m *MasterSecret) EncodeToURI() string {
	return fmt.Sprintf(
		ClusterKMSURI,
		base64.URLEncoding.EncodeToString(m.Key),
		base64.URLEncoding.EncodeToString(m.Salt),
	}
}
```

nit: Instead of having two separate encoding methods, whose output is then manually parsed into the `ClusterKMS` format string, use one method that returns a prepared URI.

internal/kms/setup/setup.go
```diff
@@ -199,12 +215,26 @@ func getGCPStorageConfig(uri *url.URL) (string, string, error) {
...
+	r, err := getConfig(uri.Query(), []string{"key", "salt"})
 	if err != nil {
-		return nil, err
+		return MasterSecret{}, err
 	}
-	return base64.URLEncoding.DecodeString(r[0])
+
+	if len(r) != 2 {
+		return MasterSecret{}, fmt.Errorf("did not receive ")
+	}
```
daniel-weisse, line 226: Error message is incomplete

daniel-weisse (changes requested):

cli/internal/cmd/init.go
```diff
@@ -30,10 +30,10 @@ import (
 	"github.com/edgelesssys/constellation/v2/internal/file"
 	"github.com/edgelesssys/constellation/v2/internal/grpc/dialer"
 	grpcRetry "github.com/edgelesssys/constellation/v2/internal/grpc/retry"
+	keyservice "github.com/edgelesssys/constellation/v2/internal/kms/setup"
```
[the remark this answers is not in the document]
derpsteb: Agree.

internal/kms/setup/setup.go
derpsteb: Nice. Didn't realize that this function will only be called from code that only uses clusterKMS.

cli/internal/cmd/recover.go
derpsteb: Are you fine with moving this to the PR that actually makes eKMS user facing? As you say it will change again very soon.

daniel-weisse: Looked at the complete code of recover again:

The `setSecrets` method was used for two reasons:
1. We are working on an interface, so we should use setter functions to add values to it
2. We wanted an easy way to stub the dynamic key derivation function

With the switch from sending individual keys over a streaming rpc, to just sending the required information so the nodes can handle setup themselves, 2. is no longer necessary.
Instead the `setSecrets` method should be renamed to something like `setURIs` and take care of setting KMS and storage URI for the interface.

```Go
func (d *recoverDoer) setURIs(kmsURI, storageURI string) {
	d.kmsURI = kmsURI
	d.storageURI = storageURI
}
```

bootstrapper/internal/initserver/initserver.go
```diff
@@ -185,7 +192,7 @@ func (s *Server) setupDisk(masterSecret, salt []byte) error {
 	}
 	uuid = strings.ToLower(uuid)
 
-	diskKey, err := crypto.DeriveKey(masterSecret, salt, []byte(crypto.HKDFInfoPrefix+uuid), crypto.DerivedKeyLengthDefault)
+	diskKey, err := cloudKms.GetDEK(ctx, crypto.HKDFInfoPrefix+uuid, crypto.StateDiskKeyLength)
```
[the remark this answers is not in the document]
derpsteb: Discussed on call: GetDataKey can't know the purpose of the key. Therefore we stay with one prefix. Changed name to `DEKPrefix`.

disk-mapper/internal/recoveryserver/server_test.go
derpsteb: Refactored and introduced a KMS stub.

daniel-weisse (approved):

derpsteb:
Successful e2e test on azure: https://github.com/edgelesssys/constellation/actions/runs/3957539412/jobs/6778033767

PR #991, opened 2023-01-16
azure: allow a set of idkeydigest values

3u13r:
### Proposed change(s)

- Add a new `idkeydigest` package which handles the marshalling and unmarshalling for YAML and JSON 
- Allow the validator to use a list of digests
- Newly created constellation configs have a list of values, "old" v2 configs are still fully compatible 

Since `idkeydigest` is a leaf package I used it liberally throughout the code. We can decide to uphold some of the package interfaces and do a conversion to `[][]byte` before passing it down. Happy to take suggestions for this.

Note that you must use the joinservice of this branch.

### Checklist

- [x] Link to Milestone


internal/watcher/validator.go
```diff
@@ -143,16 +144,15 @@ func (u *Updatable) Update() error {
...
 			return err
 		}
-		idkeydigest, err = hex.DecodeString(string(idkeydigestRaw))
-		if err != nil {
-			return fmt.Errorf("parsing hexstring: %s: %w", idkeydigestRaw, err)
+		if err = json.Unmarshal(idkeydigestRaw, &digest); err != nil {
+			return fmt.Errorf("unmarshaling content of IDKeyDigestFilename: %s: %w", idkeydigestRaw, err)
 		}
-		u.log.Debugf("New idkeydigest: %x", idkeydigest)
+		u.log.Debugf("New idkeydigest: %v", digest)
```
daniel-weisse, line 150: Until v2.6.0 this function, in case the `idkeydigests` file is not found, should try to read from the older `idkeydigest` file, and use it to create the `idekeydigests` file in the new format.

3u13r: I don't think that we need to build a migration for this, as there is no upgrade path currently between minor versions in Constellation. Therefore, creating a new ConfigMap seems a bit much. Maybe it is enough for the joinservice to also consume the old file?

internal/attestation/idkeydigest/idkeydigest.go
```diff
@@ -0,0 +1,117 @@
+/*
+Copyright (c) Edgeless Systems GmbH
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+package idkeydigest
```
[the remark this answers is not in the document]
3u13r: Do you have Pro/Cons for this? My arguments why I created a new package were: 1. the digest is not a measurement and 2. if we choose to implement MAA in the future, we already have the place for this. Also idkeydigest is currently a leaf package and measurements is not, making importing more complex.

internal/watcher/validator.go
daniel-weisse: Mostly I want this so we don't start neglecting upgrade compatibility just because "we are not quite there yet".
Thinking about this again, just being able to read the old file should be sufficient.
Maybe this won't even be required, since helm should recreate the relevant configmap during an upgrade, but I'm not 100% sure on this.

internal/attestation/idkeydigest/idkeydigest.go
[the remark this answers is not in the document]
daniel-weisse: See my comment below:
Makes it easier if we want to use automatic generation of this value.
However, its also possible to just define the default value in the `measurements` package as `[][]byte`, and then have another function in `idkeydigest` that converts it to the aliased type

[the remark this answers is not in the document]
3u13r: If its only about the interface with which you can get the default values then my suggestion would be to an indirection, where we don't export `MustGetDefaultIDKeyDigests()` rename and use it inside `idkeydigest.DefaultsFor(csp)`.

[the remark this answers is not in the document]
daniel-weisse: Maybe instead of having the values as hex encoded strings that need to be parsed with potential panics, resulting in a global variable declaration in `config`, just having them as `[][]byte` inside the code would be enough.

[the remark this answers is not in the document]
daniel-weisse: The idea for merging was mainly for automatic source code generation

internal/watcher/validator.go
3u13r: I was thinking only about image updates performed from within K8s. If we upgrade via helm, then I don't even know what it does to the configmap and without a way to test, we should simply skip this.

daniel-weisse (approved):
lgtm

internal/attestation/idkeydigest/idkeydigest.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// DefaultsFor returns the default IDKeyDigests for the given cloud provider.
+func DefaultsFor(csp cloudprovider.Provider) IDKeyDigests {
+	switch csp {
+	case cloudprovider.Azure:
+		return IDKeyDigests{
+			{0x57, 0x48, 0x6a, 0x44, 0x7e, 0xc0, 0xf1, 0x95, 0x80, 0x02, 0xa2, 0x2a, 0x06, 0xb7, 0x67, 0x3b, 0x9f, 0xd2, 0x7d, 0x11, 0xe1, 0xc6, 0x52, 0x74, 0x98, 0x05, 0x60, 0x54, 0xc5, 0xfa, 0x92, 0xd2, 0x3c, 0x50, 0xf9, 0xde, 0x44, 0x07, 0x27, 0x60, 0xfe, 0x2b, 0x6f, 0xb8, 0x97, 0x40, 0xb6, 0x96},
+			{0x03, 0x56, 0x21, 0x58, 0x82, 0xa8, 0x25, 0x27, 0x9a, 0x85, 0xb3, 0x00, 0xb0, 0xb7, 0x42, 0x93, 0x1d, 0x11, 0x3b, 0xf7, 0xe3, 0x2d, 0xde, 0x2e, 0x50, 0xff, 0xde, 0x7e, 0xc7, 0x43, 0xca, 0x49, 0x1e, 0xcd, 0xd7, 0xf3, 0x36, 0xdc, 0x28, 0xa6, 0xe0, 0xb2, 0xbb, 0x57, 0xaf, 0x7a, 0x44, 0xa3},
+		}
```
derpsteb, line 37: Could we keep the values as strings and return the result of `hex.EncodeToString(..)`? That would be much easier to update imo.

daniel-weisse: See [my earlier comment](https://github.com/edgelesssys/constellation/pull/991#discussion_r1072377533) on the issue.
The underlying code works on `[]byte` anyways, so I don't see any benefit from keeping it as hex encoded string until we get to the point where it is actually used.

If our checking tool generates the output as we expect it in the source code, there should be no difference in ease of updates.

internal/config/config.go
```diff
@@ -161,8 +166,8 @@ type AzureConfig struct {
 	//   Enable secure boot for VMs. If enabled, the OS image has to include a virtual machine guest state (VMGS) blob.
 	SecureBoot *bool `yaml:"secureBoot" validate:"required"`
 	// description: |
-	//   Expected value for the field 'idkeydigest' in the AMD SEV-SNP attestation report. Only usable with ConfidentialVMs. See 4.6 and 7.3 in: https://www.amd.com/system/files/TechDocs/56860.pdf
-	IDKeyDigest string `yaml:"idKeyDigest" validate:"required_if=EnforceIdKeyDigest true,omitempty,hexadecimal,len=96"`
+	//   List of accepted values for the field 'idkeydigest' in the AMD SEV-SNP attestation report. Only usable with ConfidentialVMs. See 4.6 and 7.3 in: https://www.amd.com/system/files/TechDocs/56860.pdf
+	IDKeyDigests Digests `yaml:"idKeyDigests" validate:"required_if=EnforceIdKeyDigest true,omitempty"`
```
derpsteb, line 170: There is not validation happening when parsing the config now, or is there? I think its a nice UX to catch malformed values.

bootstrapper/internal/kubernetes/kubernetes.go
```diff
@@ -89,7 +89,7 @@ func New(cloudProvider string, clusterUtil clusterUtil, configProvider configura
 // InitCluster initializes a new Kubernetes cluster and applies pod network provider.
 func (k *KubeWrapper) InitCluster(
 	ctx context.Context, cloudServiceAccountURI, versionString string, measurementSalt []byte, enforcedPCRs []uint32,
-	enforceIDKeyDigest bool, idKeyDigest []byte, azureCVM bool,
+	enforceIDKeyDigest bool, idKeyDigest idkeydigest.IDKeyDigests, azureCVM bool,
```
derpsteb, line 92: nit:
```suggestion
	enforceIDKeyDigest bool, idKeyDigests idkeydigest.IDKeyDigests, azureCVM bool,
```

bootstrapper/initproto/init.proto
```diff
@@ -25,6 +25,7 @@ message InitRequest {
   bool conformance_mode = 14;
   repeated KubernetesComponent kubernetes_components = 15;
   bytes init_secret = 16;
+  repeated bytes idkeydigests = 17;
```
derpsteb, line 28: Could we send this value through the helm charts instead? Could be set in [loader.go](https://github.com/edgelesssys/constellation/blob/749acdd55d0266d6ba5bdd5b5d20d038009a956e/cli/internal/helm/loader.go#L406). This would prevent changing the init API and would be more in line with our current approach of transporting as many values as possible through the helm release values.

cli/internal/helm/charts/edgeless/constellation-services/charts/join-service/values.schema.json
```diff
@@ -14,10 +14,10 @@
             "description": "Whether or not idkeydigest should be enforced during attestation on azure.",
             "type": "boolean"
         },
-        "idkeydigest": {
-            "description": "Expected idkeydigest value for Azure SNP attestation.",
+        "idkeydigests": {
+            "description": "List of expected idkeydigest values for Azure SNP attestation.",
             "type": "string",
```
derpsteb, line 19: Not sure if you tested the code in a live cluster, but does this work with an array?
I would expect the type needs to become `array`.

bootstrapper/internal/initserver/initserver.go
```diff
@@ -140,14 +141,17 @@ func (s *Server) Init(ctx context.Context, req *initproto.InitRequest) (*initpro
 		return nil, status.Errorf(codes.Internal, "persisting node state: %s", err)
 	}
 
+	// Check if we are running on a CVM
+	_, isCVM := s.issuer.(*snp.Issuer)
```
derpsteb, line 145: That's one nice trick.

cli/internal/helm/charts/edgeless/constellation-services/charts/join-service/values.schema.json
3u13r: Yes I tested it. I think only the example is wrong. The field itself holds a string with the marshalled JSON array. Will adapt the example.

bootstrapper/initproto/init.proto
3u13r: Good idea!

internal/config/config.go
3u13r: No there us not. I guess the earliest point where a malformed value now shows up is `constellation init`. This is still somewhat early and can be retried without terminating the cluster (just the error message currently says otherwise).

cli/internal/helm/charts/edgeless/constellation-services/charts/join-service/values.schema.json
derpsteb: Ah right, that's why. Thanks!

internal/config/config.go
derpsteb: A user won't know that the error message is not correct in that case :D

derpsteb (approved):

internal/config/config.go
3u13r: Found a good place to do validation. Re-implemented it.

3u13r:
Since I touched the snp/validator.go file again to improve the error message, I re-requested the review, since this part is security relevant.

derpsteb (approved):

bootstrapper/initproto/init.proto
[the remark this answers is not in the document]
3u13r: removed from proto thanks to @derpsteb suggestion

PR #995, opened 2023-01-17
cli: debug: various improvements

Nirusu:
### Proposed change(s)
- Minor style changes (fix typos, fill values, _config_ everywhere instead of switching between _config_ and _configuration_)
- Remove ambiguous log messages
- Don't print service account uri (while it could be useful for debugging, it is full of secrets)
- Print parsed arguments only if parsing did not fail (they will be the default value anyways otherwise)
- Add debug logs for errors encountered in retry loops
- Disable spinner for other debug logged commands than just `create` (`mini up`, `create`)

Nothing truly groundbreaking, not sure if change log worthy but I haven't tagged it for now.

(Also more of a refactor than a fix so ignore the branch naming)

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/cli/internal/cmd/configfetchmeasurements.go
+++ b/cli/internal/cmd/configfetchmeasurements.go
@@ -133,9 +133,9 @@ func (cfm *configFetchMeasurementsCmd) parseURLFlag(cmd *cobra.Command, flag str
 	if err != nil {
 		return nil, fmt.Errorf("parsing config generate flags '%s': %w", flag, err)
 	}
-	cfm.log.Debugf("Flag %s has raw URL %s", flag, rawURL)
+	cfm.log.Debugf("Flag %s has raw URL %q", flag, rawURL)
 	if rawURL != "" {
-		cfm.log.Debugf("Parsing raw URL %s", rawURL)
+		cfm.log.Debugf("Parsing raw URL")
 		return url.Parse(rawURL)
 	}
 	return nil, nil
@@ -146,13 +146,13 @@ func (cfm *configFetchMeasurementsCmd) parseFetchMeasurementsFlags(cmd *cobra.Co
 	if err != nil {
 		return &fetchMeasurementsFlags{}, err
 	}
-	cfm.log.Debugf("Parsed measurements URL as %q", measurementsURL.String())
+	cfm.log.Debugf("Parsed measurements URL as %v", measurementsURL)
 
 	measurementsSignatureURL, err := cfm.parseURLFlag(cmd, "signature-url")
 	if err != nil {
 		return &fetchMeasurementsFlags{}, err
 	}
-	cfm.log.Debugf("Parsed measurements signature URL as %q", measurementsSignatureURL.String())
+	cfm.log.Debugf("Parsed measurements signature URL as %v", measurementsSignatureURL)
 
 	config, err := cmd.Flags().GetString("config")
 	if err != nil {
--- a/cli/internal/cmd/configfetchmeasurements.go
+++ b/cli/internal/cmd/configfetchmeasurements.go
@@ -76,7 +76,7 @@ func (cfm *configFetchMeasurementsCmd) configFetchMeasurements(
 	}
 	cfm.log.Debugf("Using flags %v", flags)
 
-	cfm.log.Debugf("Loading config file from %s", flags.configPath)
+	cfm.log.Debugf("Loading configuration file from %s", flags.configPath)
 	conf, err := config.New(fileHandler, flags.configPath)
 	if err != nil {
 		return displayConfigValidationErrors(cmd.ErrOrStderr(), err)
@@ -117,12 +117,12 @@ func (cfm *configFetchMeasurementsCmd) configFetchMeasurements(
 		cmd.PrintErrln("Make sure the downloaded measurements are trustworthy!")
 	}
 
-	cfm.log.Debugf("Verified measurements with Rekor, updating measurements in config")
+	cfm.log.Debugf("Verified measurements with Rekor, updating measurements in configuration")
 	conf.UpdateMeasurements(fetchedMeasurements)
 	if err := fileHandler.WriteYAML(flags.configPath, conf, file.OptOverwrite); err != nil {
 		return err
 	}
-	cfm.log.Debugf("Wrote config to YAML")
+	cfm.log.Debugf("Wrote configuration to YAML")
 	return nil
 }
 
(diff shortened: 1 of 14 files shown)


katexochen (changes requested):
Will you squash this PR? If not I'd suggest to at least squash some of the commits together...

cli/internal/cmd/init.go
```diff
@@ -125,16 +128,14 @@ func (i *initCmd) initialize(cmd *cobra.Command, newDialer func(validator *cloud
 	if err != nil {
 		return err
 	}
-	i.log.Debugf("Got service account uri %s", serviceAccURI)
-	i.log.Debugf("Loading master secret file from %s", flags.masterSecretPath)
```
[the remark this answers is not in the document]
Nirusu: Removed the first one completely since "Successfully" messages are not really needed (if you know the entry point and get to the next message, you know that it passed without an error). Can add it back though.

I removed the master secret log message because that one does not make sense. If the path is "", it will not load any file; it will generate one. readOrGenerateMasterSecret does already plenty of logging in the beginning of both code paths so no information is gone (since masterSecretPath has already been printed earlier two times).

cli/internal/cmd/init.go
```diff
@@ -256,20 +265,24 @@ func writeRow(wr io.Writer, col1 string, col2 string) {
...
 	masterSecretPath, err := cmd.Flags().GetString("master-secret")
-	i.log.Debugf("Master secret path flag value is %s", masterSecretPath)
 	if err != nil {
 		return initFlags{}, fmt.Errorf("parsing master-secret path flag: %w", err)
 	}
+	debugPrintMasterSecretArgument := masterSecretPath
+	if debugPrintMasterSecretArgument == "" {
+		debugPrintMasterSecretArgument = "not specified"
+	}
+	i.log.Debugf("Master secret path flag value is %s", debugPrintMasterSecretArgument)
```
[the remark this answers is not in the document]
Nirusu: Can also change it that way, just looked cleaner to me to make this a full sentence to know that this can be intended.

cli/internal/cmd/init.go
```diff
@@ -294,7 +307,7 @@ type masterSecret struct {
 // readOrGenerateMasterSecret reads a base64 encoded master secret from file or generates a new 32 byte secret.
 func (i *initCmd) readOrGenerateMasterSecret(outWriter io.Writer, fileHandler file.Handler, filename string) (masterSecret, error) {
 	if filename != "" {
-		i.log.Debugf("Reading master secret from file")
+		i.log.Debugf("Reading master secret from file %s", filename)
```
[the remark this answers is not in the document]
Nirusu: %s is consistent with other filenames printed (this code path cannot have an empty file name).

Would have to change it everywhere.

cli/internal/cmd/init.go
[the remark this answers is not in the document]
katexochen: Okay, then I'm fine with it as is. :+1:

cli/internal/cmd/init.go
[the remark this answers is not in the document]
katexochen: Imo it makes the code harder to read. Logging shouldn't increase the complexity of the code.

cli/internal/cmd/verify.go
```diff
@@ -124,20 +124,20 @@ func (v *verifyCmd) parseVerifyFlags(cmd *cobra.Command, fileHandler file.Handle
...
 		return verifyFlags{}, fmt.Errorf("parsing cluster-id argument: %w", err)
 	}
-	v.log.Debugf("cluster-id: %s", clusterID)
+	v.log.Debugf("Cluster ID flag is %s", clusterID)
 
 	endpoint, err := cmd.Flags().GetString("node-endpoint")
 	if err != nil {
 		return verifyFlags{}, fmt.Errorf("parsing node-endpoint argument: %w", err)
 	}
-	v.log.Debugf("node-endpoint: %s", endpoint)
```
katexochen, line 140: Imo this isn't a good idea, as it is not as precise as before. Node endpoint flag could also be the endpoint flag of the node. Maybe something like "Flage 'node-endpoint' is %q"?

Nirusu: Let's just go with "'node-endpoint' flag is %q" so every sentence has flag at the end?

Nirusu:
Alright, I went a bit loco with the %s -> %q replacement but I think it looks good for the most part and gives a distinction whether a string is hard-coded or a variable in a debug message. This also went into other components so double-check if that's fine, please. I kinda went hardcore with manual search & replace.

cli/internal/cmd/verify.go
katexochen: Fine for me, but I'd also apply this to the other flags you've changed above.

katexochen: ok, then let's just leave it like this.

katexochen:
> I kinda went hardcore with manual search & replace.

Yes, that was a little to much maybe. There are indeed situations where it isn't needed/wanted to use %q. The versionsapi uses the %q where need. I'd suggest to only replace %s with %q when handling things like user input for path etc. Also please don't replace %s with %v. You can also print errors with %s without calling err.Error().

Nirusu:
I'll just limit this to flag parsing and user entry from the config then. That should hopefully be a clear enough distinction.

Nirusu:
Another attempt I guess. Revert commit is not 100% clean since other changes were already on top but I'll squash everything later anyway.

cli/internal/cmd/configfetchmeasurements.go
```diff
@@ -146,19 +146,19 @@ func (cfm *configFetchMeasurementsCmd) parseFetchMeasurementsFlags(cmd *cobra.Co
 	if err != nil {
 		return &fetchMeasurementsFlags{}, err
 	}
-	cfm.log.Debugf("Parsed measurements URL")
+	cfm.log.Debugf("Parsed measurements URL as %q", measurementsURL.String())
```
katexochen, line 149: measurementsURL can be nil, that's why the unit test fails. We also already log this in parseURLFlag(), so it think we can just delete these two lines.

cli/internal/cmd/configfetchmeasurements.go
```diff
@@ -146,19 +146,19 @@ func (cfm *configFetchMeasurementsCmd) parseFetchMeasurementsFlags(cmd *cobra.Co
...
 
 	measurementsSignatureURL, err := cfm.parseURLFlag(cmd, "signature-url")
 	if err != nil {
 		return &fetchMeasurementsFlags{}, err
 	}
-	cfm.log.Debugf("Parsed measurements signature URL")
+	cfm.log.Debugf("Parsed measurements signature URL as %q", measurementsSignatureURL.String())
```
katexochen, line 155: measurementsURL can be nil, that's why the unit test fails. We also already log this in parseURLFlag(), so it think we can just delete these two lines.

katexochen (commented):
Rest LGTM

cli/internal/cmd/configfetchmeasurements.go
Nirusu: Changed this to %v. Would opt to keep this generally since it is a parsed output, thus can be useful for debugging.

katexochen (approved):

Nirusu:
Haven't seen the last review comment. Yes, I'll squash the whole PR. Most changes are not notable enough for single commits.

PR #999, opened 2023-01-17
ci: adapt to new measurements format

stdoutput:
### Proposed change(s)

- The formatting of the measurements of debug images on the versions API has changed ( from [this](https://cdn.confidential.cloud/constellation/v1/ref/main/stream/debug/image/v2.4.0-pre.0.20221229165011-473e16feb2cc/csp/azure/measurements.image.json) to [this](https://cdn.confidential.cloud/constellation/v1/ref/main/stream/debug/v2.5.0-pre.0.20230112150103-bcd8aa9acc25/image/csp/gcp/measurements.image.json) ) which needs to be taken into account when fetching the measurements in the test. I adapted to said changes. A successful run of the test can be seen [here](https://github.com/edgelesssys/constellation/actions/runs/3938516358/jobs/6737304444)


katexochen (approved):
lgtm

PR #1001, opened 2023-01-17
rfc: extend updates rfc with constraints section

derpsteb:
### Proposed change(s)

- In order to properly reason about the `constellation-upgrade` cmd UX it's helpful to define the constraints that should apply during an upgrade. This extension to the RFC describes our initial constraints.

### Checklist

### DIFF
--- a/rfc/updates.md
+++ b/rfc/updates.md
@@ -267,3 +267,28 @@ Upgrading Constellation microservices to 2.2.0:
 Warning: Please backup any important components before upgrading Kubernetes
 Apply change [yes/No]?
 ```
+
+# Compatibility
+`constellation-upgrade` has to handle the version of four components: CLI, image, microservices and Kubernetes.
+To do this correctly and keep the cluster in a working condition some constraints are required.
+
+- Microservice, image and CLI are versioned in lockstep, i.e. each release a new version of all componentes is released.
+  Each of these version numbers is one Constellation version.
+- Each Constellation version is compatible to three or four Kubernetes versions.
+  When a new Kubernetes version is released Constellation will support four versions for one release cycle; before phasing out the oldest Kubernetes version.
+  To learn if microserviceVersion A.B.C is compatible with Kubernetes version X.Y, one has to check whether the Constellation version is compatible with the Kubernetes version X.Y.
+  The Constellation version will be the same as the microserviceVersion: A.B.C. Same goes for the image and CLI.
+- Each Constellation version X.Y.Z is compatible with all patch versions of another Constellation version X.Y+1.
+- The individual versions of Microservice, image and CLI all have to be compatible with the targeted Kubernetes version.
+  Without loss of generality, this means that during a Kubernetes upgrade the oldest Constellation component has to be compatible with the new Kubernetes version.
+  If not, this oldest component has to be updated first.
+- For Kubernetes versions the version-target of an upgrade has to be supported by the current Constellation version.
+  The currently running Kubernetes version is not relevant for the Kubernetes upgrade.
+
+These constraints are enforced by the CLI.
+If users decide to change specific versions on their own nothing is stopping them.
+
+The compatibility information should be separated from the enforcement code.
+This way a minimal implemention can be created where the compatibility information is embedded into the CLI.
+As a next step the information can be served through the [Constellation API](./apis.md).
+By serving the compatibility information dynamically, faulty versions can be excluded from upgrade paths even after they have been released.
--- a/rfc/updates.md
+++ b/rfc/updates.md
@@ -198,9 +198,10 @@ Image patch versions are forward compatible within one minor version.
 Lastly, the CLI checks if a newer CLI version is available via the update API (see: rfc/update-api.json). If this is the case, it will print the latest CLI version instead of the output described above.
 If the current version and latest version diverge more than one minor version, it will also show the latest CLI of the next minor version, and suggest a way to download it.
 An example:
-current CLI version: `2.3.2`
-available CLI version with minor version `2.4.0`: `2.4.1`, `2.4.2`, `2.4.3`
-latest CLI version: `2.5.0`
+- current CLI version: `2.3.2`
+- available CLI version with minor version `2.4.0`: `2.4.1`, `2.4.2`, `2.4.3`
+- latest CLI version: `2.5.0`
+
 `constellation upgrade check` will show `2.5.0` as latest, and suggest that the next step in the upgrade process is `2.4.3`.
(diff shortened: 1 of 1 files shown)


rfc/updates.md
```diff
@@ -267,3 +267,28 @@ Upgrading Constellation microservices to 2.2.0:
 Warning: Please backup any important components before upgrading Kubernetes
 Apply change [yes/No]?
 ```
+
+# Compatibility
+`constellation-upgrade` has to handle the version of four components: CLI, image, microservices and Kubernetes.
+To do this correctly and keep the cluster in a working condition some constraints are required.
+
+- Microservice, image and CLI are versioned in lockstep, i.e. each release a new version of all componentes is released.
```
3u13r, line 275:
```suggestion
- Constellation microservices, VM image and the CLI are versioned in lockstep, i.e. each release a new version of all components is released.
```

rfc/updates.md
```diff
@@ -267,3 +267,28 @@ Upgrading Constellation microservices to 2.2.0:
...
+  Each of these version numbers is one Constellation version.
```
3u13r, line 276: I think this is somewhat hard to understand. Maybe more clear: 
`The Constellation version references all of these components`.
Still, if we want to say that each component has the same semver attached to it, we already do this in the sentence before.

rfc/updates.md
```diff
@@ -267,3 +267,28 @@ Upgrading Constellation microservices to 2.2.0:
...
+- Each Constellation version is compatible to three or four Kubernetes versions.
```
3u13r, line 277:
```suggestion
- Each Constellation version is compatible with three or four Kubernetes versions.
```

rfc/updates.md
```diff
@@ -267,3 +267,28 @@ Upgrading Constellation microservices to 2.2.0:
...
+  When a new Kubernetes version is released Constellation will support four versions for one release cycle; before phasing out the oldest Kubernetes version.
```
3u13r, line 278:
```suggestion
  When a new Kubernetes version is released, Constellation will support four versions for one release cycle; before phasing out the oldest Kubernetes version.
```

rfc/updates.md
```diff
@@ -267,3 +267,28 @@ Upgrading Constellation microservices to 2.2.0:
...
+  To learn if microserviceVersion A.B.C is compatible with Kubernetes version X.Y, one has to check whether the Constellation version is compatible with the Kubernetes version X.Y.
+  The Constellation version will be the same as the microserviceVersion: A.B.C. Same goes for the image and CLI.
```
3u13r, line 280: We've already established this a few points before.

rfc/updates.md
```diff
@@ -267,3 +267,28 @@ Upgrading Constellation microservices to 2.2.0:
...
+- Each Constellation version X.Y.Z is compatible with all patch versions of another Constellation version X.Y+1.
```
3u13r, line 281:
```suggestion
- Each Constellation version X.Y.Z is compatible with all patch versions of the next Constellation minor version X.Y+1.
```

rfc/updates.md
```diff
@@ -267,3 +267,28 @@ Upgrading Constellation microservices to 2.2.0:
...
+- The individual versions of Microservice, image and CLI all have to be compatible with the targeted Kubernetes version.
+  Without loss of generality, this means that during a Kubernetes upgrade the oldest Constellation component has to be compatible with the new Kubernetes version.
```
3u13r, line 283:
```suggestion
This means that during a Kubernetes upgrade the oldest Constellation component has to be compatible with the new Kubernetes version.
```
We should probably explicitly establish that one can upgrade e.g. only the microservices. Then it is clear what "**oldest** Constellation component" means. If this is what you are going for here.

rfc/updates.md
```diff
@@ -267,3 +267,28 @@ Upgrading Constellation microservices to 2.2.0:
...
+  If not, this oldest component has to be updated first.
+- For Kubernetes versions the version-target of an upgrade has to be supported by the current Constellation version.
+  The currently running Kubernetes version is not relevant for the Kubernetes upgrade.
+
+These constraints are enforced by the CLI.
+If users decide to change specific versions on their own nothing is stopping them.
```
3u13r, line 289:
```suggestion
If users decide to change specific versions by changing the Kubernetes resources itself, nothing is stopping them.
```

rfc/updates.md
```diff
@@ -267,3 +267,28 @@ Upgrading Constellation microservices to 2.2.0:
 Warning: Please backup any important components before upgrading Kubernetes
 Apply change [yes/No]?
 ```
+
+# Compatibility
+`constellation-upgrade` has to handle the version of four components: CLI, image, microservices and Kubernetes.
```
3u13r, line 272:
```suggestion
`constellation upgrade` has to handle the version of four components: CLI, image, microservices and Kubernetes.
```
This is not a new CLI binary, right?

3u13r (changes requested):
Some wording suggestion. Overall great summary! I think this handles most (all?) of the points.

malt3 (approved):
This correctly describes what we want to enforce!

rfc/updates.md
derpsteb: Can't find where we explicitly state it like that. Doesn't hurt to repeat it?

3u13r: I think we already somewhat mention this in the first point:
> Constellation microservices, VM image and the CLI are versioned in lockstep, i.e. each release a new version of all components is released.
  The Constellation version references all of these components.

If we want to also mention the semver, I think we should do this there. If needed we can then say something like `since the microserviceVersion is the same as the CLI version ...`.

rfc/updates.md
```diff
@@ -266,4 +267,30 @@ Upgrading Constellation microservices to 2.2.0:
...
+  This means that during a Kubernetes upgrade the oldest Constellation component has to be compatible with the new Kubernetes version.
+  If not, this oldest component has to be updated first.
+- For Kubernetes versions the version-target of an upgrade has to be supported by the current Constellation version.
+  The currently running Kubernetes version is not relevant for the Kubernetes upgrade.
+
+These constraints are enforced by the CLI.
+If users decide to change specific versions by changing the Kubernetes resources, nothing is stopping them.
+
+The compatibility information should be separated from the enforcement code.
+This way a minimal implemention can be created where the compatibility information is embedded into the CLI.
```
3u13r, line 294:
```suggestion
This way a minimal implementation can be created where the compatibility information is embedded into the CLI.
```

3u13r (approved):
last nit, otherwise LGTM!

PR #1002, opened 2023-01-17
verify: use fixed user data

thomasten:
### Proposed change(s)

- use fixed user data in the VerifyService protocol

### Checklist

- [x] Link to Milestone


daniel-weisse (approved):
`/hack/pcr-reader/main.go` needs a fix to remove `UserData` field, otherwise looks good to me

PR #1005, opened 2023-01-17
ci: fix CLI SBOM generation

Nirusu:
### Proposed change(s)
- Update Syft to 0.66.1
- Update Grype to 0.55.0
- Install the two tools to `/usr/local/bin` to avoid polluting the git tree
- Build linux/amd64 CLI as reference for Syft to collect the used modules

Syft's `go-module` did not collect the package list correctly before ever since the build was separated from the SBOM generation. The alternative would be to use `go-mod-file`, which would collect every package this monorepo imports - even if it's not inside the CLI. Since we have separate SBOMs for the microservices (which look good to me), that would be somewhat undesirable.

Thus we build the linux/amd64 CLI as reference to collect all the modules. Testing between what we currently ship (Linux/macOS & amd64/arm64), the generated SBOMs hold the same packages for the CLI, so for the moment we can do that and just release one SBOM file for the CLI instead of four.

However, in case we add more platform-specific code in the future to the CLI, it might become necessary to ship multiple SBOMs per platform or architecture to represent accurate package information for each released CLI. 

That should be kept in mind.

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -107,14 +107,15 @@ jobs:
 
       - name: Download Syft & Grype
         shell: bash
+        working-directory: /tmp
         run: |
           SYFT_VERSION=0.66.1
           GRYPE_VERSION=0.55.0
-          curl -LO https://github.com/anchore/syft/releases/download/v${SYFT_VERSION}/syft_${SYFT_VERSION}_linux_amd64.tar.gz
+          curl -Lo syft_${SYFT_VERSION}_linux_amd64.tar.gz https://github.com/anchore/syft/releases/download/v${SYFT_VERSION}/syft_${SYFT_VERSION}_linux_amd64.tar.gz
           tar -xzf syft_${SYFT_VERSION}_linux_amd64.tar.gz
           sudo install syft /usr/bin/syft
           syft version
-          curl -LO https://github.com/anchore/grype/releases/download/v${GRYPE_VERSION}/grype_${GRYPE_VERSION}_linux_amd64.tar.gz
+          curl -Lo grype_${GRYPE_VERSION}_linux_amd64.tar.gz https://github.com/anchore/grype/releases/download/v${GRYPE_VERSION}/grype_${GRYPE_VERSION}_linux_amd64.tar.gz
           tar -xzf grype_${GRYPE_VERSION}_linux_amd64.tar.gz
           sudo install grype /usr/bin/grype
           grype version
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -100,38 +100,56 @@ jobs:
   signed-sbom:
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-cli.yml
```diff
@@ -92,30 +92,46 @@ jobs:
...
+          go-version: "1.19.5"
+      - name: Download Syft & Grype
+        shell: bash
         run: |
-          SYFT_VERSION=0.62.1
-          GRYPE_VERSION=0.53.1
+          SYFT_VERSION=0.66.1
+          GRYPE_VERSION=0.55.0
           curl -LO https://github.com/anchore/syft/releases/download/v${SYFT_VERSION}/syft_${SYFT_VERSION}_linux_amd64.tar.gz
           tar -xzf syft_${SYFT_VERSION}_linux_amd64.tar.gz
```
datosh, line 105: Just to be safe, we could run the download from `/tmp` or manually delete the `tar.gz` and its contents after extracting them.

datosh (approved):

.github/workflows/release-cli.yml
Nirusu: Since the step is now before the Checkout action it should be good (PWD is some temp dir).

Nirusu: Done, working-directory is now explicitly `/tmp`.

.github/workflows/release-cli.yml
```diff
@@ -92,30 +92,47 @@ jobs:
...
+          curl -Lo grype_${GRYPE_VERSION}_linux_amd64.tar.gz https://github.com/anchore/grype/releases/download/v${GRYPE_VERSION}/grype_${GRYPE_VERSION}_linux_amd64.tar.gz
           tar -xzf grype_${GRYPE_VERSION}_linux_amd64.tar.gz
-          ./grype version
-          pwd >> "$GITHUB_PATH"
-        shell: bash
+          sudo install grype /usr/bin/grype
+          grype version
       - name: Install Cosign
         uses: sigstore/cosign-installer@9becc617647dfa20ae7b1151972e9b3a2c338a2b # tag=v2.8.1
+      - name: Checkout
```
katexochen, line 115: There should be newlines between all steps. See #1008

Nirusu: That should probably be dealt with in your PR. It will need a rebase anyway.

katexochen: Already did, just wanted to mention so that you remember it when doing the rebase :wink:

Nirusu: This should go in before #1008.

Nirusu: Establishing new conventions and creating a new PR created *after* this PR was opened and already approved by one person, introducing new merge conflicts, ignoring my suggestion to merge this in before _and_ marking this as resolved without notice is **very** petty behaviour. 

I am sorry, but this is pretty rude.

PR #1011, opened 2023-01-18
ci: move Syft & Grype installation into an action

Nirusu:
### Proposed change(s)

- Update Syft to v0.66.2 (CLI + container SBOMs) (was released yesterday after I initially opened my previous PR: https://github.com/anchore/syft/releases/tag/v0.66.2)
- Update Grype to v0.55.0 (container SBOMs, CLI was already updated with #1005)
- Use same "download to /tmp and install to /usr/bin" install method for Syft and Grype for the container SBOMs as for the release CLI action
- Move the install into an action

This should make it easier to keep the version for the CLI and containers in sync.
Small disadvantage is that now the Checkout needs to be before the install again. The action does download to /tmp though, so it should keep the git workspace clean.

### DIFF
--- a/.github/actions/install_syft_grype/action.yml
+++ b/.github/actions/install_syft_grype/action.yml
@@ -1,0 +1,36 @@
+name: Install Syft & Grype
+description: |
+  Installs Syft & Grype.
+inputs:
+  versionSyft:
+    description: "Version of the operator-sdk to install"
+    default: "0.66.2"
+    required: false
+  versionGrype:
+    description: "Version of Grype to install"
+    default: "0.55.0"
+    required: false
+  os:
+    description: "Download build of Syft & Grype for operating system [linux, darwin]"
+    default: "linux"
+    required: false
+  arch:
+    description: "Download build of Syft & Grype for architecture [amd64, arm64]"
+    default: "amd64"
+    required: false
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install Syft & Grype
+      shell: bash
+      working-directory: /tmp
+      run: |
+        curl -Lo syft_${{ inputs.versionSyft }}_${{ inputs.os }}_${{ inputs.arch }}.tar.gz https://github.com/anchore/syft/releases/download/v${{ inputs.versionSyft }}/syft_${{ inputs.versionSyft }}_${{ inputs.os }}_${{ inputs.arch }}.tar.gz
+        tar -xzf syft_${{ inputs.versionSyft }}_${{ inputs.os }}_${{ inputs.arch }}.tar.gz
+        sudo install syft /usr/bin/syft
+        syft version
+        curl -Lo grype_${{ inputs.versionGrype }}_${{ inputs.os }}_${{ inputs.arch }}.tar.gz https://github.com/anchore/grype/releases/download/v${{ inputs.versionGrype }}/grype_${{ inputs.versionGrype }}_${{ inputs.os }}_${{ inputs.arch }}.tar.gz
+        tar -xzf grype_${{ inputs.versionGrype }}_${{ inputs.os }}_${{ inputs.arch }}.tar.gz
+        sudo install grype /usr/bin/grype
+        grype version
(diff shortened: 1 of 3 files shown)


.github/actions/install_syft_grype/action.yml
```diff
@@ -0,0 +1,36 @@
+name: Install Syft & Grype
+description: |
+  Installs Syft & Grype.
+inputs:
+  versionSyft:
+    description: "Version of the operator-sdk to install"
```
katexochen, line 6:
```suggestion
    description: "Version of Syft to install"
```

.github/actions/install_syft_grype/action.yml
```diff
@@ -0,0 +1,36 @@
...
+    default: "0.66.2"
+    required: false
+  versionGrype:
+    description: "Version of Grype to install"
+    default: "0.55.0"
+    required: false
```
katexochen, line 12: Is it needed that these can be configured? Shouldn't we just pin them?

katexochen (changes requested):
Nice, thanks!

.github/actions/install_syft_grype/action.yml
```diff
@@ -0,0 +1,36 @@
...
+  os:
+    description: "Download build of Syft & Grype for operating system [linux, darwin]"
+    default: "linux"
+    required: false
+  arch:
+    description: "Download build of Syft & Grype for architecture [amd64, arm64]"
+    default: "amd64"
+    required: false
```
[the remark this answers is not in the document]
Nirusu: Doesn't work for Darwin (RUNNER_OS returns `macOS`).

.github/actions/install_syft_grype/action.yml
Nirusu: Not really, we can also pin them. Just wrote it as a generic function.

.github/actions/install_syft_grype/action.yml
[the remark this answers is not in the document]
Nirusu: Needs a translation layer (x64 -> amd64, macOS -> Darwin) but I'll add this quickly. Not really needed since Linux callers only use it currently (and there's not ARM runners), but uh, let's go for it I guess.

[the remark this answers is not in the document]
Nirusu: Alright, new version should support Linux/macOS and amd64/arm64.

.github/actions/install_syft_grype/action.yml
```diff
@@ -1,36 +1,38 @@
...
+        
+        echo "Downloading for ${OS}/${ARCH}"
+        
+        curl -Lo syft_${SYFT_VERSION}_${OS}_${ARCH}.tar.gz https://github.com/anchore/syft/releases/download/v${SYFT_VERSION}/syft_${SYFT_VERSION}_${OS}_${ARCH}.tar.gz
+        tar -xzf syft_${SYFT_VERSION}_${OS}_${ARCH}.tar.gz
         sudo install syft /usr/bin/syft
         syft version
-        curl -Lo grype_${{ inputs.versionGrype }}_${{ inputs.os }}_${{ inputs.arch }}.tar.gz https://github.com/anchore/grype/releases/download/v${{ inputs.versionGrype }}/grype_${{ inputs.versionGrype }}_${{ inputs.os }}_${{ inputs.arch }}.tar.gz
-        tar -xzf grype_${{ inputs.versionGrype }}_${{ inputs.os }}_${{ inputs.arch }}.tar.gz
+        curl -Lo grype_${GRYPE_VERSION}_${OS}_${ARCH}.tar.gz https://github.com/anchore/grype/releases/download/v${GRYPE_VERSION}/grype_${GRYPE_VERSION}_${OS}_${ARCH}.tar.gz
```
katexochen, line 35:
```suggestion
        curl -fsSLo grype_${GRYPE_VERSION}_${OS}_${ARCH}.tar.gz https://github.com/anchore/grype/releases/download/v${GRYPE_VERSION}/grype_${GRYPE_VERSION}_${OS}_${ARCH}.tar.gz
```

katexochen (approved):
Sorry, missed those two the first time. But LGTM otherwise.

katexochen (approved):

PR #1012, opened 2023-01-18
internal: use go-kms-wrapping for KMS backends

daniel-weisse:
### Proposed change(s)

- Use hashicorp's [go-kms-wrapping](https://pkg.go.dev/github.com/hashicorp/go-kms-wrapping/v2) for our KMS backends
  - Implemented and tested for AWS KMS, Azure Key Vault, Azure managed HSM, and GCP KMS
- Update credential usage, by default a configuration is applied that allows a user to set credentials for the KMS:
    - AWS uses a Access Key and Access Key ID
    - Azure uses Tenant ID, Client ID, Client Secret (same as managed identities for Constellation on Azure)
    - GCP uses a service account credentials file (same as service accounts for Constellation GCP)
- Add the `uri` package to parse and create KMS/Storage URIs

### Additional info
- ~The goal is to replace all supported KMS backends with go-kms-wrapping, I would imagine a straightforward implementation that only differs in the setup code between CSPs is possible.~
  - Done, worked out nicely :smile: 

- ~As part of this PR I will also remove the `CreateKEK` methods, as they are no longer used in our implementation.~
  - Done, KMS setup is left to the admin managing the installation

- I have set up a key for the integration test. Run it using the following parameters (and proper credentials):
  ```bash
   go test -v -tags=integration -gcpKms -gcpProjectID=constellation-331613 --gcpKeyRing=test-ring --gcpLocation=global -kekID=integration-test-kek
  ```
  - There are also keys that can be used for AWS and Azure, you should be able to find them in our cloud subscriptions
  - If you want to test Azure HSM, you will need to setup a managed HSM yourself (Don't forget to clean up resources once you are done)

### Checklist

- [ ] Link to Milestone


internal/kms/kms/internal/internal.go
```diff
@@ -0,0 +1,82 @@
...
+
+	dek, err := c.Wrapper.Decrypt(ctx, wrappedKey)
+	if err != nil {
+		return nil, fmt.Errorf("decrypting DEK: %w", err)
+	}
+
+	return dek, nil
+}
+
+// putDEK saves an encrypted Data Encryption Key to storage.
```
derpsteb, line 69:
```suggestion
// putDEK encrypts a given plaintext Data Encryption Key and saves it to storage.
```

internal/kms/kms/azure/azure.go
```diff
@@ -9,154 +9,51 @@ package azure
...
+	return &KMSClient{
+		kms: &internal.KMSClient{
+			Storage: store,
+			Wrapper: wrapper,
+		},
+	}, nil
 }
 
-// GetDEK decrypts a DEK from storage.
+// GetDEK fetches an encrypted Data Encryption Key from storage and decrypts it using a KEK stored in Google's KMS.
```
derpsteb, line 53:
```suggestion
// GetDEK fetches an encrypted Data Encryption Key from storage and decrypts it using a KEK stored in Azure's KMS.
```

derpsteb:
So little code. So simple. Nice. :rainbow:

internal/kms/kms/gcp/gcp.go
```diff
@@ -8,328 +8,63 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+	kms    *internal.KMSClient
+	client io.Closer
 }
 
 // New initializes a KMS client for Google Cloud Platform.
-func New(ctx context.Context, projectID, locationID, keyRingID string, store kmsInterface.Storage, protectionLvl kmspb.ProtectionLevel, kekID string, opts ...gax.CallOption) (*KMSClient, error) {
+func New(ctx context.Context, store kmsInterface.Storage, cfg uri.GCPConfig) (*KMSClient, error) {
 	if store == nil {
 		store = storage.NewMemMapStorage()
 	}
```
derpsteb, line 39: Maybe this would be better put into the "no-store" case of setup.go? And then rename "no-store" to "local-store" or "mem-store". Would make the behaviour a bit more explicit imo.

internal/kms/test/gcp_test.go
```diff
@@ -0,0 +1,81 @@
...
+	if !*runGcpKms {
+		t.Skip("Skipping Google KMS test")
+	}
+	if *gcpProjectID == "" || *gcpLocation == "" || *gcpKeyRing == "" || *kekID == "" {
+		flag.Usage()
+		t.Fatal("Required flags not set: --gcp-project, --gcp-location, --gcp-keyring, --kek-id")
+	}
+	require := require.New(t)
+
+	store := storage.NewMemMapStorage()
```
derpsteb, line 34: Why use MemMapStorage here, but cloud storage for AWS/Azure?

internal/kms/kms/gcp/gcp.go
```diff
@@ -8,328 +8,63 @@ SPDX-License-Identifier: AGPL-3.0-only
...
-	locationID       string
-	keyRingID        string
-	newClient        func(ctx context.Context, opts ...option.ClientOption) (clientAPI, error)
-	waitBackoffLimit int
-	storage          kmsInterface.Storage
-	protectionLevel  kmspb.ProtectionLevel
-	kekID            string
-	opts             []gax.CallOption
+	kms    *internal.KMSClient
+	client io.Closer
```
derpsteb, line 32: Why have this separate property? Couldn't the Close be defined on the KMSClient directly? Especially since it modifies the `kms` property.

derpsteb:
Ran the following commands to grant my service account required permissions on GCP:
`gcloud projects add-iam-policy-binding PROJECTID --member=serviceAccount:SERVICEACCOUNTEMAIL --role=roles/cloudkms.viewer`
`gcloud projects add-iam-policy-binding PROJECTID --member=serviceAccount:SERVICEACCOUNTEMAIL --role=roles/cloudkms.signerVerifier`
`gcloud projects add-iam-policy-binding PROJECTID --member=serviceAccount:SERVICEACCOUNTEMAIL --role=roles/cloudkms.cryptoKeyEncrypterDecrypter`

I think only the first and last role is required? 
Used this command to run the tests:
`go test -v -tags=integration -gcp-kms -gcp-project=constellation-331613 --gcp-keyring=test-ring --gcp-location=global -kek-id=integration-test-kek -gcp-credentials-path=/home/xcv/bin/constellation/constellation-gcp-serviceacc-key.json`


On Azure needed to:
- create key vault
- add key to vault under "Keys" in the sidebar
- add app client as Application under "Access policies" 
- grant permissions for: keys get, wrap, unwrap. This is the minimal set required for the integration test.
Ran tests with:
`go test -v -tags=integration -az-kms --az-tenant-id=TENANTID --az-client-id=CLIENTID --az-client-secret=SECRET --az-vault-name=test-kms-implementation --kek-id=somekey`

On AWS needed to:
- create own key to have the correct permissions on the key to encrypt/decrypt with it
Ran tests with:
`go test -v -tags=integration -aws-kms --aws-access-key-id=KEYID --aws-access-key=SECRET --aws-region=us-east-1 --kek-id=KEYID`
Credentials are stored in `~/.aws/credentials`.

Looks great to me. Muuch cleaner and easier to understand!

internal/kms/kms/gcp/gcp.go
daniel-weisse: no-store != in-memory storage

The first case is just used as a placeholder, since cKMS does not use any kind of storage, the later is used to simply store keys in a map (everything is lost on restart). This is only used for unit testing, so I guess we could remove it from the default initialization functions and simply throw an error if no store is set

internal/kms/test/gcp_test.go
daniel-weisse: What do you mean?
AWS and Azure also use MemMapStorage for the KMS tests, don't they?

internal/kms/kms/gcp/gcp.go
daniel-weisse: `Close` in this case, is a method on the GCP API client used by the kms wrapper. The wrapper interface itself does not have a `Close` method
For `Close` to be a method of the internal KMS Client we would need to run a type assertion on the underlying API client of the wrapper and then call `Close` on it if it has such a method.

Not sure if this would be cleaner

internal/kms/test/gcp_test.go
derpsteb: Whoops I was confused by the order of the tests in the different files I guess. Nvm.

internal/kms/kms/gcp/gcp.go
derpsteb: Yes. But because of this check "no-store" will create a MemMapStorage in `setup.go:KMS(..., "no-store", ..)`, which seems like an odd behavior to me. I forgot that we need `no-store` for cKMS. Throwing an error seems like a good solution to me.

derpsteb (approved):
Feature to track test automation is created.

PR #1017, opened 2023-01-18
ci: downgrade Syft to v0.65.0

Nirusu:
### Proposed change(s)

- Downgrade Syft to v0.65.0

### Related issue
- https://github.com/anchore/syft/issues/1465
> We will have an updated README and a new way to specify a private key in the next release of Syft. 

> A second PR will follow that implements custom PKI provided by the user.

### DIFF
--- a/.github/actions/install_syft_grype/action.yml
+++ b/.github/actions/install_syft_grype/action.yml
@@ -7,7 +7,7 @@ runs:
       shell: bash
       working-directory: /tmp
       env:
-        SYFT_VERSION: "0.66.2"
+        SYFT_VERSION: "0.65.0" # Before upgrading, check if this has been fixed: https://github.com/anchore/syft/issues/1465
         GRYPE_VERSION: "0.55.0"
         OS: ${{ runner.os }}
         ARCH: ${{ runner.arch }}


.github/actions/container_sbom/action.yml
```diff
@@ -21,7 +21,7 @@ runs:
       id: checkout
       uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
       with:
-        ref: ${{ inputs.ref || github.head_ref }}
```
malt3, line 24: This is not a correct solution. The release pipeline (that calls this indirectly) depends on the ref used during checkout being different from github.head_ref

malt3: You would have to pass inputs.ref all the way down here or perform the checkout in each workflow before calling the action

Nirusu: I think the checkout is actually not needed here, since the parents should already do the checkout before (the action is buried few layers deep). They probably should do this in any case, I missed the levels of depth the action is called here.

Removed it.

malt3 (approved):

derpsteb:
Do it now! :pray:

Nirusu:
@derpsteb Merged, fingers crossed!

PR #1018, opened 2023-01-18
config: detailed validation errors for k8s version

derpsteb:
### Proposed change(s)

These extended error messages help users in understanding what is wrong with the current configuration and how to remediate the issue. 

Target version too new:
```
../constellation create -c 1 -w 2 -y
Problems validating config file:                                                                                                             
        KubernetesVersion specifies an unsupported Kubernetes version. The configured version 1.77 is newer than the newest version supported by this CLI: 1.26                                                                                                                           
Fix the invalid entries or generate a new configuration using `constellation config generate`                                                
Error: invalid configuration
```

Target version too old:
```
../constellation create -c 1 -w 2 -y
Problems validating config file:                                                                                                             
        KubernetesVersion specifies an unsupported Kubernetes version. You can upgrade to these versions with your current Constellation CLI: [1.23 1.24 1.25 1.26]                                                                                                                        
Fix the invalid entries or generate a new configuration using `constellation config generate`                                          
Error: invalid configuration 
```

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/config/validation.go
+++ b/internal/config/validation.go
@@ -9,6 +9,7 @@ package config
 import (
 	"bytes"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/edgelesssys/constellation/v2/internal/attestation/measurements"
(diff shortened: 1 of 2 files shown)


internal/config/validation.go
```diff
@@ -17,12 +18,42 @@ import (
...
+	i := 0
+	for k := range versions.VersionConfigs {
+		validVersions[i] = string(k)
+		i++
+	}
+	validVersionsSorted := semver.ByVersion(validVersions)
+	sort.Sort(validVersionsSorted)
+
+	var errorMsg string
+	configured := fe.Value().(string)
```
3u13r, line 43:
```suggestion
	configured, ok := fe.Value().(string)
	if !ok {
		errorMsg = "The configured version is not a valid string."
	}
```
Not sure about how to handle an invalid cast, but I think we should handle it somehow.

internal/config/validation.go
```diff
@@ -17,12 +18,42 @@ import (
...
+	maxVersion := validVersionsSorted[len(validVersionsSorted)-1]
+	if configured < validVersionsSorted[0] {
+		errorMsg = fmt.Sprintf("You can upgrade to these versions with your current Constellation CLI: %s.", validVersionsSorted)
```
3u13r, line 46:
```suggestion
		errorMsg = fmt.Sprintf("Upgrade to one of these versions with this CLI: %s.", validVersionsSorted)
```
I'd omit "Constellation" since it is probably clear that it is the Constellation CLI and to align with the message below. Also shortened it a bit and increased the nudging to emphasize the needed action to resolve the error.

3u13r (changes requested):

internal/config/validation.go
daniel-weisse: I think we should leave out the "You can upgrade" part, since this error will occur even when you try to create a new Constellation.
Maybe just align it with the version-too-new error, and tell the user that the configured version is not supported?

derpsteb: Tbh I wouldn't want to leave out the upgrade part because it is supposed to support users when they upgraded their CLI while upgrading a cluster. 
Imagine your first CLI tells you to upgrade the CLI. Now the second CLI tells you sth like "please select a supported version". I would feel like there is a gap between both CLIs in that case. (?)

daniel-weisse: That would be a user error, wouldn't it?
A user is expected to updated their config when they want to upgrade the cluster.
Failure to do so can result in invalid configurations (hence the "This version is not supported").

I also feel like reporting a "You can upgrade to these version with your CLI" when a user tries to create a completely new cluster may lead to some questions that would be avoided by simply giving a more general response.

derpsteb: After another live discussion we decided to extend the versionapi so it can server information on cli-k8s compatibility. This removes the original motivation for the initial error message. Changed both error messages to generic messages.

3u13r (approved):
LGTM

datosh (approved):

PR #1025, opened 2023-01-19
ci: run CodeQL only on Go/Python changes

katexochen:

### Checklist

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -4,8 +4,22 @@ on:
   push:
     branches:
       - main
-      - release/v*
+      - "release/**"
+    paths:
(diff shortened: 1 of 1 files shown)


stdoutput:
LGTM🐳

Maybe, at some point, we could even split up the matrix to run the go check on go file changes and the python check on python file changes, but since the python check is very fast, that doesn't seem necessary for now

stdoutput (approved):
LGTM🐳

Maybe, at some point, we could even split up the matrix to run the go check on go file changes and the python check on python file changes, but since the python check is very fast, that doesn't seem necessary for now

datosh (approved):

PR #1033, opened 2023-01-19
ci: downgrade Syft back to 0.59.0

Nirusu:
### Proposed change(s)
- Downgrade Syft back to 0.59.0

### Related issue
- Can't use ~0.60.0+ (?)~ due to: https://github.com/anchore/syft/issues/1483 (actually the old version we had before yesterday also seems to be affected)
- Can't use 0.65.0 due to: https://github.com/anchore/syft/issues/1465

When upgrading at a later point, please check if both issues are fixed.

### DIFF
--- a/.github/actions/install_syft_grype/action.yml
+++ b/.github/actions/install_syft_grype/action.yml
@@ -7,7 +7,7 @@ runs:
       shell: bash
       working-directory: /tmp
       env:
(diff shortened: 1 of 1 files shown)


Nirusu:
Actually seems to have existed in 0.59.0, too. So it's a coincidence running into this now shortly after the upgrade. Huh, wonder what changed.

Nirusu:
Downgrading won't fix the issue. The change happened in our (ap)ko tooling / images.
Closing this for now and trying to fix this some other way.

PR #1037, opened 2023-01-19
docs: document terraform support

3u13r:
### Proposed change(s)

- Minimal Terraform documentation

### Checklist

- [x] Link to Milestone

### DIFF
--- a/docs/docs/getting-started/first-steps.md
+++ b/docs/docs/getting-started/first-steps.md
@@ -9,7 +9,7 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
 
 ## Create a cluster
 
-1.  Create the configuration file and IAM resources for your selected cloud provider
+1. Create the configuration file and IAM resources for your selected cloud provider
 
     First, you need to create a [configuration file](../workflows/config.md) and an [IAM configuration](../workflows/config.md#creating-an-iam-configuration). The easiest way to do this is the following CLI command:
 
@@ -76,6 +76,7 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
 -->
 
 2. Create the cluster with one control-plane node and two worker nodes. `constellation create` uses options set in `constellation-conf.yaml`.
+If you want to use Terraform for managing the cloud resources instead, follow our [Terraform guide](../workflows/create.md).
 
     :::tip
 
--- a/docs/docs/getting-started/first-steps.md
+++ b/docs/docs/getting-started/first-steps.md
@@ -76,7 +76,7 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
 -->
 
 2. Create the cluster with one control-plane node and two worker nodes. `constellation create` uses options set in `constellation-conf.yaml`.
-If you want to use Terraform for managing the cloud resources instead, follow our [Terraform guide](../workflows/create.md).
+    If you want to use Terraform for managing the cloud resources instead, follow the corresponding instructions in the [Create workflow](../workflows/create.md).
 
     :::tip
 
--- a/docs/docs/getting-started/first-steps.md
+++ b/docs/docs/getting-started/first-steps.md
@@ -151,6 +151,8 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
 
 ## Terminate your cluster
 
+Use the CLI to terminate your cluster. If you used Terraform to manage your cloud resources, follow the corresponding instructions in the [Terminate workflow](../workflows/terminate.md).
+
 ```bash
 constellation terminate
 ```
--- a/docs/docs/workflows/create.md
+++ b/docs/docs/workflows/create.md
@@ -18,6 +18,9 @@ Before you create the cluster, make sure to have a [valid configuration file](./
 
 ### Create
 
+<tabs groupId="provider">
+<tabItem value="cli" label="CLI">
+
 Choose the initial size of your cluster.
 The following command creates a cluster with one control-plane and two worker nodes:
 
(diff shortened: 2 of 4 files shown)


docs/docs/getting-started/first-steps.md
```diff
@@ -76,6 +76,7 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
 -->
 
 2. Create the cluster with one control-plane node and two worker nodes. `constellation create` uses options set in `constellation-conf.yaml`.
+If you want to use Terraform for managing the cloud resources instead, follow our [Terraform guide](../workflows/create.md).
```
thomasten, line 79:
```suggestion
    If you want to use Terraform for managing the cloud resources instead, follow the corresponding instructions in the [Create workflow](../workflows/create.md).
```
indentation is needed for docusaurus to render this correctly

docs/docs/workflows/create.md
```diff
@@ -29,6 +32,51 @@ For details on the flags, consult the command help via `constellation create -h`
 
 *create* stores your cluster's state into a [`terraform.tfstate`](../architecture/orchestration.md#cluster-creation-process) file in your workspace.
 
+</tabItem>
+<tabItem value="terraform" label="Terraform">
+
+:::caution
+
+There is no extensive input validation with Terraform. If unsure, please use the Constellation CLI instead.
```
thomasten, line 40: how much effort would it be to add input validation?

docs/docs/workflows/create.md
```diff
@@ -29,6 +32,51 @@ For details on the flags, consult the command help via `constellation create -h`
...
+
+:::
+
+Constellation also supports managing the infrastructure via Terraform. This allows for an easier GitOps integration as well as meeting regulatory requirements.
+Since the Constellation CLI also uses Terraform under the hood, the same Terraform files can be reused.
```
thomasten, line 45:
```suggestion
Constellation supports managing the infrastructure via Terraform. This allows for an easier GitOps integration as well as meeting regulatory requirements.
Since the Constellation CLI also uses Terraform under the hood, you can reuse the same Terraform files.
```

docs/docs/workflows/create.md
```diff
@@ -29,6 +32,51 @@ For details on the flags, consult the command help via `constellation create -h`
...
+For now, please try to refrain from changing the Terraform resource definitions, as Constellation depends on certain components.
```
thomasten, line 46:
```suggestion
For now, please refrain from changing the Terraform resource definitions, as Constellation is tightly coupled to them.
```
maybe this is more clear?

docs/docs/workflows/create.md
```diff
@@ -29,6 +32,51 @@ For details on the flags, consult the command help via `constellation create -h`
...
+
+Download the Terraform files for the selected CSP from our [GitHub repository](https://github.com/edgelesssys/constellation/tree/main/cli/internal/terraform/terraform).
```
thomasten, line 48: Do I have to download all files including subdirectories? This is tedious on the GitHub UI. Any idea for a better UX here? (I once used [curl for this](https://github.com/edgelesssys/constellation/tree/main/cli/internal/terraform/terraform/iam/aws), but with subdirectories it's more complicated.)

docs/docs/workflows/create.md
```diff
@@ -29,6 +32,51 @@ For details on the flags, consult the command help via `constellation create -h`
...
+There, define all needed variables found in `variables.tf` using the values from the `constellation-config.yaml`.
+
+To find the image reference for your CSP and region, execute:
+
+```bash
+CONSTELL_VER=vX.Y.Z
+curl -s https://cdn.confidential.cloud/constellation/v1/ref/-/stream/stable/$CONSTELL_VER/image/info.json | jq
+```
+
+Now, initialize and apply Terraform to create the configured infrastructure:
```
thomasten, line 60:
```suggestion
Initialize and apply Terraform to create the configured infrastructure:
```

docs/docs/workflows/create.md
```diff
@@ -29,6 +32,51 @@ For details on the flags, consult the command help via `constellation create -h`
...
+
+```bash
+terraform init
+terraform apply
+```
+
+The initialization depends on the already created `constellation-config.yaml` and the `constellation-id.json`.
+Therefore, create the `constellation-id.json` using the output from the Terraform state and the `constellation-conf.yaml`:
```
thomasten, line 68:
```suggestion
The Constellation [init step](#the-init-step) requires the already created `constellation-config.yaml` and the `constellation-id.json`.
Create the latter using the output from the Terraform state and the `constellation-conf.yaml`:
```
I was confused because I first related "the initialization" to `terraform init`, so I suggest to clarify this a bit

docs/docs/workflows/create.md
```diff
@@ -29,6 +32,51 @@ For details on the flags, consult the command help via `constellation create -h`
...
+
+```bash
+CONSTELL_IP=$(terraform output ip)
+CONSTELL_INIT_SECRET=$(terraform output initSecret | jq -r | tr -d '\n' | base64)
+CONSTELL_CSP=$(cat constellation-conf.yaml | yq ".provider | keys | .[0]")
+jq --null-input --arg cloudprovider "$CONSTELL_CSP" --arg ip "$CONSTELL_IP" --arg initsecret "$CONSTELL_INIT_SECRET" '{"cloudprovider":$cloudprovider,"ip":$ip,"initsecret":$initsecret}' > constellation-id.json
+```
+
+</tabItem>
+</tabs>
```
thomasten, line 78: If we create a cluster with these steps, do we have the same files in the workspace as if we used the create command? In particular, will the terminate command work?

thomasten, line 78: And regarding terminate, what's the plan for this? Do this later?

thomasten (changes requested):

docs/docs/workflows/create.md
3u13r: No, I just forgot, that we have a separate `terminate.md`. Will add this now.

docs/docs/workflows/create.md
3u13r: Changed `latter` to `constellation-id.json` so one does not need to backtrack to the sentence before.

docs/docs/workflows/create.md
3u13r: Yes, basically every time I had the problem I resorted to cloning the entire repository instead. 

We could show those steps. From experience: In the one case we had, the user was able to do this step on their own.

docs/docs/workflows/create.md
3u13r: This mostly depends on how lax we want to be in regards to the instance type aka if we were to allow only a small subset of (C)VM types, then we probably can add validation for this. 
Apart from that, I think the validation is actually quite similar. I also just added some more defaults for AWS.

thomasten: So I guess with the additions you've made, you still feel this warning is needed and helpful for the reader, right?

docs/docs/workflows/create.md
thomasten: Ok, probably not worth to extend this then.

docs/docs/workflows/create.md
thomasten: Then we need to refer to the Terminate workflow from the "Terminate your cluster" section on the First steps page.

docs/docs/workflows/create.md
```diff
@@ -29,6 +32,51 @@ For details on the flags, consult the command help via `constellation create -h`
...
+
+There is no extensive input validation with Terraform. If unsure, please use the Constellation CLI instead.
+
+:::
+
+Constellation supports managing the infrastructure via Terraform. This allows for an easier GitOps integration as well as meeting regulatory requirements.
+Since the Constellation CLI also uses Terraform under the hood, you can reuse the same Terraform files.
+For now, please refrain from changing the Terraform resource definitions, as Constellation is tightly coupled to them.
+
+Download the Terraform files for the selected CSP from our [GitHub repository](https://github.com/edgelesssys/constellation/tree/main/cli/internal/terraform/terraform).
```
thomasten, line 48:
```suggestion
Download the Terraform files for the selected CSP from the [GitHub repository](https://github.com/edgelesssys/constellation/tree/main/cli/internal/terraform/terraform).
```

docs/docs/workflows/terminate.md
```diff
@@ -29,3 +31,22 @@ Termination can fail if additional resources have been created that depend on th
...
+
+</tabItem>
+<tabItem value="terraform" label="Terraform">
+Terminate the cluster by running:
+
+```bash
+terraform destroy
+```
+
+Delete all files which are no longer needed:
```
thomasten, line 43:
```suggestion
Delete all files that are no longer needed:
```
restrictive clause requires *that* in American English

3u13r: Interesting, I did not know that.

docs/docs/workflows/create.md
3u13r: I'm not 100% sure, but since Terraform is the "expert" way, I removed the warning for now. If we see common errors we can add it again.

thomasten (approved):
Thanks! Good to finally have basic "official" Terraform support now :)

PR #1038, opened 2023-01-19
ci: fix GCP cloud controller manager build

datosh:
### Proposed change(s)

- Fixed broken build as `"${GITHUB_WORKSPACE}/bin/bazel"` didnt exist anymore after https://github.com/edgelesssys/constellation/commit/8aa84fd7597b69dbd7f4c59c17410eae570b2b7d, but bazelisk is available on all runners.
- As described in linked issue, the bazel build is currently broken for older version, so I have removed them from CI for now.
- Build for latest version works so I have upgraded image in versions.go

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/build-ccm-gcp.yml
+++ b/.github/workflows/build-ccm-gcp.yml
@@ -11,9 +11,12 @@ jobs:
       packages: write
     strategy:
       matrix:
-        version: [v25.2.0, v24.0.0, v23.0.0]
+        version: [v26.0.1]
+        # TODO: Once issue is closed upgrade older versions, rebuild, and update versions.go
+        # https://github.com/kubernetes/cloud-provider-gcp/issues/451
+        # version: [v26.0.1, v25.5.0, v24.0.0]
         include:
-          - version: v25.2.0
+          - version: v26.0.1
             latest: true
     steps:
       - name: Checkout kubernetes/cloud-provider-gcp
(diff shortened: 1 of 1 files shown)


malt3 (approved):

internal/versions/versions.go
```diff
@@ -284,7 +284,7 @@ var VersionConfigs = map[ValidK8sVersion]KubernetesVersion{
 		CloudControllerManagerImageAWS: "registry.k8s.io/provider-aws/cloud-controller-manager:v1.26.0@sha256:fdeb61e3e42ecd9cca868d550ebdb88dd6341d9e91fcfa9a37e227dab2ad22cb", // renovate:container
 		// CloudControllerManagerImageGCP is the CCM image used on GCP.
 		// TODO: use newer "cloud-provider-gcp" from https://github.com/kubernetes/cloud-provider-gcp when newer releases are available.
-		CloudControllerManagerImageGCP: "ghcr.io/edgelesssys/cloud-provider-gcp:v25.2.0@sha256:86fa9d31ed0b3d0d8806f13d6e7debd3471028b2cb7cca3a876d8a31612a7ba5", // renovate:container
```
katexochen, line 287: This should be build on main and updated through renovate.

datosh: I was not aware that our Renovate config is able to do so. Can you point me to the config part where it matches the K8s version to the image versions?

katexochen: https://github.com/edgelesssys/constellation/blob/main/renovate.json#L192 That's why there is the `// renovate:container` in the back. :wink:

datosh: Quick sync with @katexochen: I have removed bump of version, so that we can use the image build from `main`. Will create a new PR once image build ran on `main`.

PR #1055, opened 2023-01-23
ci: add workflow for proto code generation check

katexochen:

### Proposed change(s)

- Currently the proto generation isn't checked for being up to date.

### Checklist

### DIFF
--- a/.github/workflows/test-proto.yml
+++ b/.github/workflows/test-proto.yml
@@ -1,0 +1,35 @@
+name: Proto generate check
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - "release/**"
+    paths:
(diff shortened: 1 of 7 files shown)


daniel-weisse (approved):
Since proto the doc comments in proto files are parsed to the Go output, should we update them to have "correct" Go doc comments syntax?
E.g. instead of `kms_uri` write `KmsUri?

malt3 (approved):
Thanks!

katexochen:
> Since proto the doc comments in proto files are parsed to the Go output, should we update them to have "correct" Go doc comments syntax?

Thought about that, too. But not sure, Imo this is something the proto compiler should do, as there is no manual way respecting both protobuf and Go conventions.

PR #1067, opened 2023-01-24
ci: use peter-evans' action to create prs

katexochen:

### Proposed change(s)

- We used 3 different actions before. Peter Evans action has the most stars, so only using this one from now on.
- Refactored some legacy code of the CLI reference update workflow, which was left from when the docs were in a separate repo. It is now much simpler.

### Checklist


malt3 (approved):
Looks very clean!

PR #1070, opened 2023-01-25
debugd: keep mutex locked while receiving files

malt3:
### Proposed change(s)

- debugd: keep mutex locked while receiving files

### Checklist

- [x] Add labels (e.g., for changelog category)

PR #1073, opened 2023-01-25
operator: make test more resilient

3u13r:
### Proposed change(s)

- This should fix https://github.com/edgelesssys/constellation/actions/runs/4008623269/attempts/1


malt3:
Do we maybe want to use `git cherry-pick -x` to reference the original commit? Just a nit (if at all). Feel free to resolve as you feel is best.

malt3:
Unit test issue will be resolved by https://github.com/edgelesssys/constellation/pull/1070

malt3 (approved):

PR #1081, opened 2023-01-26
ci: fix Constellation recover e2e test

msanft:
### Proposed change(s)

- Wait for the control plane nodes to get back up after pushing the recovery keys to ensure control plane availability after recovering the cluster

### Additional info
- [Successful run with 1 CP node](https://github.com/edgelesssys/constellation/actions/runs/4015145256/jobs/6896527567)
- [Successful run with 3 CP nodes](https://github.com/edgelesssys/constellation/actions/runs/4015146173/jobs/6896526073)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/actions/e2e_recover/action.yml
+++ b/.github/actions/e2e_recover/action.yml
@@ -11,15 +11,6 @@ inputs:
   masterSecret:
     description: "The master-secret for the cluster."
     required: true
-  cloudProvider:
-    description: "Which cloud provider to use."
-    required: true
-  gcpProject:
-    description: "The GCP project Constellation is deployed in."
-    required: false
-  resourceGroup:
-    description: "The Azure resource group Constellation is deployed in."
-    required: false
 
 runs:
   using: "composite"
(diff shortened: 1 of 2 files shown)


.github/actions/e2e_recover/action.yml
```diff
@@ -68,6 +68,26 @@ runs:
...
+              exit 0
+          fi
+
+          current_time=$(date +%s)
+          if ((current_time - start_time > timeout)); then
+              echo "Waiting for control plane to get back up timed out after $timeout seconds."
+              exit 1
+          fi
+
+          echo "Cannot reach control plane: $output, retrying in 10 seconds"
```
daniel-weisse, line 89: `$output` doesn't contain any value in this case, I assume because the error during `kubectl wait` is printed to stderr.

Maybe just remove `$output` from the message here?

daniel-weisse (commented):
Code looks good to me.
There are currently 3 unused variables in this action:
`cloudProvider`, `gcpProject`, and `resourceGroup` (which should be named `azureResourceGroup` to keep in line with other actions)

I think they can be removed?

daniel-weisse (approved):

msanft:
[Successful run](https://github.com/edgelesssys/constellation/actions/runs/4025193237), merging

PR #1087, opened 2023-01-26
terraform: make control-planes stateful on gcp

3u13r:
### Proposed change(s)

- Make the control-plane instance group stateful by making the disk and internal ip stateful.

Then stopping an instance on GCP, it is automatically re-created from the template. Per default, a new state disk and IP is used. This is problematic when stopping all VMs at the same time.

To stop an instance on GCP do:
1. Go to the details view of the control-plane instance
2. Go back to the instance overview
3. Select all the instances to be stopped
4. Click the "Stop" button.

Steps 1-2 are needed since GCP normally does not really allow you to stop the instance (see greyed out button in the details view of the instance). 

Does anyone have opinions regarding if we should backport this?

### Checklist

- [x] Link to Milestone


katexochen (approved):
lgtm

PR #1109, opened 2023-01-30
cli: add upgrade-check command

derpsteb:
### Proposed change(s)

- Add `constellation upgrade check` as a command and remove `constellation upgrade plan`. Upgrade check shows the smallest upgrade a user could do to the current cluster with their current CLI. Smallest since we have to make some decision on which upgrade to show in order to support the `--write-config` flag. We could add a `--all` flag to show all possible upgrades. 
  - Includes a `--write-config` flag can be used to persist the suggested update into the config.
  - Includes a `--ref` and `--stream` flag to select the ref and stream which should be checked for image upgrades. 
- Extend config validation with supported versions validation. This will check that the versions specified in the config adhere to our defined version drift of 1. I.e. the CLI will always have to be up to one minor version newer than the configured version.
- Add `--force` flag that disables config validation.
- Add config value `MicroserviceVersion`. This specifies the version of the `constellation-services` chart and (implicitly) the version of the `constellation-operators` chart. I decided to make the operator version implicit as we would otherwise have to have another config value for it. We could think about unifying the operators and services charts. I don't think there is a technical reason for keeping them separate. 

### Checklist

- [x] Link to Milestone


internal/constants/constants.go
```diff
@@ -175,4 +175,4 @@ const (
 )
 
 // VersionInfo is the version of a binary. Left as a separate variable to allow override during build.
-var VersionInfo = "0.0.0"
+var VersionInfo = "v0.0.0"
```
daniel-weisse, line 178: Our build process injects this value without the `v` prefix, so changing the default value to something else might have unexpected differences between unit tests and actual performance.
Also `vX.Y.Z` is [not valid semver](https://semver.org/#is-v123-a-semantic-version).

We should define a set way to label our binary builds, helm charts and container images to avoid confusion in the future.
I think currently we use `vX.Y.Z` for container images, and `X.Y.Z` for binary builds and helm charts.
Helm charts are a bit annoying since not all companies use the same scheme for version labeling. For example cert-manager uses `vX.Y.Z`.

internal/versions/versions.go
```diff
@@ -85,11 +101,11 @@ const (
...
 	// currently supported versions.
 	//nolint:revive
-	V1_24 ValidK8sVersion = "1.24"
+	V1_24 ValidK8sVersion = "1.24.9"
 	//nolint:revive
-	V1_25 ValidK8sVersion = "1.25"
+	V1_25 ValidK8sVersion = "v1.25.6"
 	//nolint:revive
-	V1_26 ValidK8sVersion = "1.26"
+	V1_26 ValidK8sVersion = "v1.26.1"
```
daniel-weisse, line 108: Same here, I think we could/will run into some problems if we just start prefixing the version with `v`

daniel-weisse (commented):
Not a real review, just some thoughts after going over the changes

cli/internal/cmd/upgradecheck.go
```diff
@@ -0,0 +1,530 @@
...
+	newKubernetes     []string
+	currentServices   string
+	currentImage      string
+	currentKubernetes string
+}
+
+// TODO: does this need more cleanup?
+func (v *versionUpgrade) buildString() (string, error) {
+	upgradeMsg := bytes.Buffer{}
+
```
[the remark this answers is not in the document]
derpsteb: Considered and implemented.

internal/config/config.go
```diff
@@ -58,14 +59,17 @@ type Config struct {
...
+	Image string `yaml:"image" validate:"required,version_compatibility"`
 	// description: |
 	//   Size (in GB) of a node's disk to store the non-volatile state.
 	StateDiskSizeGB int `yaml:"stateDiskSizeGB" validate:"min=0"`
 	// description: |
-	//   Kubernetes version to be installed in the cluster.
+	//   Kubernetes version to be installed into the cluster.
 	KubernetesVersion string `yaml:"kubernetesVersion" validate:"supported_k8s_version"`
 	// description: |
+	//	 Microservice version to be installed into the cluster. Optional until v2.7. Defaults to the version of the CLI.
```
datosh, line 70: Should we use this string to inform about allowed version ranges? Or should `upgrade check` only ever modify this value?

datosh, line 70: Should we use this string to inform about allowed version ranges? Or should `upgrade check` only ever modify this value?

internal/config/config.go
```diff
@@ -317,9 +322,9 @@ func FromFile(fileHandler file.Handler, name string) (*Config, error) {
 // 1. Reading config file via provided fileHandler from file with name.
 // 2. Read secrets from environment variables.
 // 3. Validate config.
```
datosh, line 324: Should we update doc string to explain what `force` does during validation?

cli/internal/cmd/configfetchmeasurements.go
```diff
@@ -78,9 +79,9 @@ func (cfm *configFetchMeasurementsCmd) configFetchMeasurements(
 	cfm.log.Debugf("Using flags %v", flags)
 
 	cfm.log.Debugf("Loading configuration file from %q", flags.configPath)
-	conf, err := config.New(fileHandler, flags.configPath)
+	conf, err := config.New(fileHandler, flags.configPath, flags.force)
 	if err != nil {
-		return displayConfigValidationErrors(cmd.ErrOrStderr(), err)
+		return config.DisplayValidationErrors(cmd.ErrOrStderr(), err)
```
datosh, line 84: Have you considered to make the `--force` flag more broad? That is, it ignores any errors returned from config validation, keep print them as warnings to user, and then continue with execution.

I think most `--force` flags are a lot broader in what they ignore. I would expect this to be useful in situations where the user knows what they are doing, e.g., during testing, debugging, ...

internal/compatibility/compatibility.go
```diff
@@ -0,0 +1,139 @@
...
+func NextMinorVersion(version string) (string, error) {
+	major, minor, err := parseCanonicalSemver(EnsurePrefixV(version))
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("v%d.%d", major, minor+1), nil
+}
+
+func parseCanonicalSemver(version string) (major int, minor int, err error) {
+	version = semver.Canonical(version) // ensure version is in canonical form (vX.Y.Z)
```
datosh, line 129: There is also semver.MajorMinor. Would this function be better suited here?

internal/compatibility/compatibility.go
```diff
@@ -0,0 +1,139 @@
...
+	num, err := fmt.Sscanf(version, "v%d.%d", &major, &minor)
+	if err != nil {
+		return 0, 0, fmt.Errorf("parsing version: %w", err)
+	}
+	if num != 2 {
+		return 0, 0, fmt.Errorf("parsing version: expected 3 numbers, got %d", num)
+	}
```
datosh, line 136: I don't really understand how we can make this error claim here. 
Could you add a few unit tests for this function so I better understand expected inputs and outputs/errors?

internal/config/config.go
```diff
@@ -58,14 +59,17 @@ type Config struct {
...
 	//   Machine image used to create Constellation nodes.
-	Image string `yaml:"image" validate:"required"`
+	Image string `yaml:"image" validate:"required,version_compatibility"`
 	// description: |
 	//   Size (in GB) of a node's disk to store the non-volatile state.
 	StateDiskSizeGB int `yaml:"stateDiskSizeGB" validate:"min=0"`
 	// description: |
-	//   Kubernetes version to be installed in the cluster.
+	//   Kubernetes version to be installed into the cluster.
 	KubernetesVersion string `yaml:"kubernetesVersion" validate:"supported_k8s_version"`
```
datosh, line 68: Would it make sense to have a `Version` type here (and for microservice version etc..)? We have a lot of code now that accesses major, minor, patch, .. fields.
We could also have consistent de-/encoding with or without the `v` prefix.
This could make our version handling more consistent and maybe help with some of the feedback from @daniel-weisse?

.github/actions/constellation_create/action.yml
```diff
@@ -173,14 +173,15 @@ runs:
           --info logcollect.github.run-attempt="${{ github.run_attempt }}" \
           --info logcollect.github.ref-name="${{ github.ref_name }}" \
           --info logcollect.github.sha="${{ github.sha }}" \
-          --info logcollect.github.runner-os="${{ runner.os }}"
+          --info logcollect.github.runner-os="${{ runner.os }}" \
+          --force
```
datosh, line 177: Is `--force` here just as a precaution or are we actually doing invalid configs in our tests?

cli/internal/helm/client.go
```diff
@@ -91,6 +91,16 @@ func (c *Client) Upgrade(ctx context.Context, config *config.Config, timeout tim
 	return nil
 }
 
+// Versions queries the cluster for running versions and returns a map of releaseName -> version.
+func (c *Client) Versions() (string, error) {
+	serviceVersion, err := c.currentVersion(conServicesReleaseName)
+	if err != nil {
+		return "", fmt.Errorf("getting con-services version: %w", err)
```
3u13r, line 98: For user-facing errors, we should probably write out "constellation".

cli/internal/helm/client.go
```diff
@@ -91,6 +91,16 @@ func (c *Client) Upgrade(ctx context.Context, config *config.Config, timeout tim
...
+	}
+
+	return "v" + serviceVersion, nil
```
3u13r, line 101: Is this a use-case for `compatibility.EnsurePrefixV()` ?

cli/internal/helm/loader.go
```diff
@@ -90,6 +90,16 @@ func NewLoader(csp cloudprovider.Provider, k8sVersion versions.ValidK8sVersion)
...
 }
 
+// AvailableServiceVersions returns the chart version number of the bundled service versions.
+func AvailableServiceVersions() (string, error) {
+	servicesChart, err := loadChartsDir(helmFS, conServicesPath)
+	if err != nil {
+		return "", fmt.Errorf("loading constellation-services chart: %w", err)
+	}
+
+	return "v" + servicesChart.Metadata.Version, nil
```
3u13r, line 100: Is this a use-case for `compatibility.EnsurePrefixV()` ?

3u13r (changes requested):

internal/config/config.go
derpsteb: I think I don't understand what you mean. This comment ends up in the default config as a comment. And the version number should hint at the fact that setting the `microserviceVersion` value in the config is optional for v2.6. For compatibility reasons. Starting with 2.7 the values _has_ to be set. I extended the comment a bit.

The value is a bit redundant as a given CLI in version X is only able to upgrade the service versions to the one version included in CLI X. So its more informational to the user than anything else. 

I don't think this comment should mention available versions as we would have to update the comment manually.
`upgrade check` displays the available versions and writes them to the config if `--write-config` is set.

cli/internal/cmd/configfetchmeasurements.go
derpsteb: I shortly thought about it but came to the conclusion that it only really makes sense for version validation right now.

E.g. setting a VM type that is not supported is pretty likely to break things. On the other hand, setting weird version combinations is more likely to work. 

But I am happy to hear other opinions and change the flag to disable validation altogether.

.github/actions/constellation_create/action.yml
derpsteb: Currently the CLI always has the version of the last release. Right now `2.5.0`. (Should be 2.5.1?) However, the debug images are all on v2.6.0-pre. Our compatibility constraints say that the CLI is not compatible with newer images. Thus the validation fails. With bazel we will hopefully be able to give the CLI a valid pseudoversion, move it to a prerelease version as well and the `--force` flag can be removed from the action.

Let me put some version on this as a comment into the action.

cli/internal/cmd/upgradecheck.go
```diff
@@ -0,0 +1,527 @@
...
+	log.Debugf("Current image minor version is %s", currentImageMinorVer)
+	log.Debugf("Current CLI minor version is %s", currentCLIMinorVer)
+	log.Debugf("Next image minor version is %s", nextImageMinorVer)
+
+	cliImageCompare := semver.Compare(currentCLIMinorVer, currentImageMinorVer)
+
+	allowedMinorVersions := []string{currentImageMinorVer, nextImageMinorVer}
+	switch {
+	case cliImageCompare < 0:
+		cmd.PrintErrln("Warning: CLI version is older than cluster image version. This is not supported.")
```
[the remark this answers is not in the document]
derpsteb: The version compared here is the one coming from the cluster. That means it is not caught by the config validator.

I agree that we should error in this case. However, in case the force flag is set we need to continue. I left the warning for that case but can also remove it altogether if you think someone using `--force` should not see warnings. We need to continue in case `force` is set as disabling the config validator without disabling this check is useless.

internal/config/config.go
datosh: I was thinking along the lines of: `This version should not surpess the current cluster version by more than one. See <link to our upgrade docs>` 

In case the user wants to manually set a version, instead of using `upgrade check` they get a hint about accepted values and further documentation.

internal/config/config.go
derpsteb: I guess this Version lib could make sense in streamlining version handling: https://pkg.go.dev/github.com/hashicorp/go-version

datosh: Yeah looks like this could work great for us. 
Since this PR already has a lot of changes we can also move this to a seperate one.

internal/config/config.go
derpsteb: In case you put values into the field that don't adhere to the compatibility constraints you will get pretty precise error messages during config validation [1].

I will add links to the docs once I wrote them for the respective fields. 

[1] https://github.com/edgelesssys/constellation/pull/1109/files#diff-da15a1a7d9e1c206009e1cec96dc5f52b1ad2fc84ca481a477bc73716866c3c1R315-R319

go.mod
```diff
@@ -86,6 +85,7 @@ require (
 	github.com/spf13/afero v1.9.3
 	github.com/spf13/cobra v1.6.1
 	github.com/stretchr/testify v1.8.1
+	github.com/tj/assert v0.0.0-20171129193455-018094318fb0
```
daniel-weisse, line 88: I think you introduced the wrong assert import somewhere

internal/compatibility/compatibility.go
```diff
@@ -0,0 +1,139 @@
...
+	"golang.org/x/mod/semver"
+)
+
+var (
+	// ErrMajorMismatch signals that the major version of two compared versions don't match.
+	ErrMajorMismatch = errors.New("missmatching major version")
+	// ErrMinorDrift signals that the minor version of two compared versions are further apart than one.
+	ErrMinorDrift = errors.New("target version needs to be equal or up to one minor version higher")
+	// ErrSemVer signals that a given version does not adhere to the Semver syntax.
+	ErrSemVer = errors.New("invalid semver")
```
daniel-weisse, line 27:
```suggestion
	ErrSemVer = errors.New("invalid semantic version")
```
nit: not using abbreviations in error descriptions seems like a good idea

internal/compatibility/compatibility_test.go
```diff
@@ -0,0 +1,265 @@
...
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+package compatibility
+
+import (
+	"testing"
+
+	"github.com/edgelesssys/constellation/v2/internal/constants"
+	"github.com/tj/assert"
```
daniel-weisse, line 13:
```suggestion
	"github.com/stretchr/testify/assert"
```

internal/config/config.go
```diff
@@ -502,6 +519,14 @@ func (c *Config) Validate() error {
 		return err
 	}
 
+	versionCompatibilityValidator := validateVersionCompatibility
+	if force {
+		versionCompatibilityValidator = returnsTrue
+	}
+	if err := validate.RegisterValidation("version_compatibility", versionCompatibilityValidator); err != nil {
+		return err
+	}
```
daniel-weisse, line 528:
```suggestion
	if force {
		if err := validate.RegisterValidation("version_compatibility", validateVersionCompatibility); err != nil {
			return err
		}
	}
```
I think this is a bit simpler.
We can gate all "optional" validation behind the `force` flag, and just not register them if it is unset.
Needs to be tested if this actually works, the validator might throw an error for undefined validations if we specify one but never register it, not sure.

internal/config/validation.go
```diff
@@ -281,3 +297,61 @@ func getPlaceholderEntries(m Measurements) []uint32 {
...
+	err := validateVersionCompatibilityHelper(fe.Field(), fe.Value().(string))
+	var msg string
+
+	switch err {
+	case compatibility.ErrSemVer:
+		msg = fmt.Sprintf("configured version (%s) does not adhere to SemVer syntax", fe.Value().(string))
+	case compatibility.ErrMajorMismatch:
+		msg = fmt.Sprintf("the CLI's major version (%s) has to match your configured major version (%s)", constants.VersionInfo, fe.Value().(string))
+	case compatibility.ErrMinorDrift:
+		msg = fmt.Sprintf("only the CLI (%s) can be up to one minor version newer than the configured version (%s)", constants.VersionInfo, fe.Value().(string))
```
daniel-weisse, line 319:
```suggestion
	switch {
	case errors.Is(compatibility.ErrSemVer):
		msg = fmt.Sprintf("configured version (%s) does not adhere to SemVer syntax", fe.Value().(string))
	case errors.Is(compatibility.ErrMajorMismatch):
		msg = fmt.Sprintf("the CLI's major version (%s) has to match your configured major version (%s)", constants.VersionInfo, fe.Value().(string))
	case errors.Is(compatibility.ErrMinorDrift):
		msg = fmt.Sprintf("only the CLI (%s) can be up to one minor version newer than the configured version (%s)", constants.VersionInfo, fe.Value().(string))
```

Using `switch err{ case ErrType: }` does not catch wrapped errors, since its basically the same as `if err == ErrType`

internal/versions/versions.go
```diff
@@ -85,11 +101,11 @@ const (
...
 	// currently supported versions.
 	//nolint:revive
-	V1_24 ValidK8sVersion = "1.24"
+	V1_24 ValidK8sVersion = "v1.24.9"
 	//nolint:revive
-	V1_25 ValidK8sVersion = "1.25"
+	V1_25 ValidK8sVersion = "v1.25.6"
 	//nolint:revive
-	V1_26 ValidK8sVersion = "1.26"
+	V1_26 ValidK8sVersion = "v1.26.1"
```
daniel-weisse, line 109: Checked our code again, if I understand it correctly the `V1_X ValidK8sVersion` variables are only used to keep track of the minor Kubernetes version supported by Constellation (We don't support choosing patch versions).

I think we should keep them as only Major.Minor, not sure about the `v` prefix.

derpsteb: The reason for switching to exposing the patch version is that we might want to offer a patch-version upgrade to the user. I think its cleaner to show users that the version will change from e.g. `1.25.6` to `1.25.7` instead of doing this implicitly. Users will still not be able to chose arbitrary patch versions, but it will be obvious which one is running/upgraded.

cli/internal/cmd/upgradecheck.go
```diff
@@ -493,13 +492,13 @@ func getCompatibleImageMeasurements(cmd *cobra.Command, client *http.Client, rek
 			},
 		)
 		if err != nil {
-			cmd.PrintErrf("Skipping compatible image %q: %s\n", shortPath, err)
+			writer.Write([]byte(fmt.Sprintf("Skipping compatible image %q: %s\n", shortPath, err)))
```
daniel-weisse:
```suggestion
			fmt.Fprintf(writer, "Skipping compatible image %q: %s\n", shortPath, err)
```

cli/internal/cmd/upgradecheck.go
```diff
@@ -493,13 +492,13 @@ func getCompatibleImageMeasurements(cmd *cobra.Command, client *http.Client, rek
...
 			continue
 		}
 
-		if err = verifyWithRekor(cmd.Context(), rekor, hash); err != nil {
-			cmd.PrintErrf("Warning: Unable to verify '%s' in Rekor.\n", hash)
-			cmd.PrintErrf("Make sure measurements are correct.\n")
+		if err = verifyWithRekor(ctx, rekor, hash); err != nil {
+			writer.Write([]byte(fmt.Sprintf("Warning: Unable to verify '%s' in Rekor.\n", hash)))
+			writer.Write([]byte(fmt.Sprintf("Make sure measurements are correct.\n")))
```
daniel-weisse:
```suggestion
			fmt.Fprintf(writer, "Warning: Unable to verify '%s' in Rekor.\n", hash)
			fmt.Fprintln(writer, "Make sure measurements are correct.")
```

daniel-weisse (approved):

cli/internal/cmd/upgradecheck.go
```diff
@@ -0,0 +1,535 @@
...
+	}
+
+	supportedServices, supportedImages, supportedK8s, err := u.collect.supportedVersions(cmd.Context(), currentImage, csp)
+	if err != nil {
+		return err
+	}
+	u.log.Debugf("current service version: %s", currentServices)
+	u.log.Debugf("supported service version: %s", supportedServices)
+	u.log.Debugf("current k8s version: %s", currentK8s)
+	u.log.Debugf("supported k8s version: %s", supportedK8s)
```
daniel-weisse, line 151:
```suggestion
	u.log.Debugf("Current service version: %s", currentServices)
	u.log.Debugf("Supported service version: %s", supportedServices)
	u.log.Debugf("Current k8s version: %s", currentK8s)
	u.log.Debugf("Supported k8s version: %s", supportedK8s)
```
Log Message should start with upper case.
Haven't checked every new log message, there might be other cases where they start in lowercase
https://github.com/edgelesssys/constellation/blob/main/dev-docs/conventions.md#logging

derpsteb: Updated messages from the PR and some other places.

cli/internal/cmd/configfetchmeasurements.go
datosh: It is fine for me to keep the scope of `force` as is.

datosh (approved):

malt3 (approved):
Nice work!

3u13r (approved):

PR #1117, opened 2023-01-31
cli: use /manager as binary path

derpsteb:
### Proposed change(s)

- The change to `/ko-app/v2` is incorrect as we are currently not building ko images for this operator.

### Checklist

- [ ] Link to Milestone


malt3 (approved):
Good catch. Slipped through my review.

PR #1118, opened 2023-01-31
versions: update constellation operator image

derpsteb:
### Proposed change(s)

- Update operator image. Necessary because helm charts expect ko image, but current image is not a ko image. 

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/versions/versions.go
+++ b/internal/versions/versions.go
@@ -69,7 +69,7 @@ const (
 	// Check for new versions at https://github.com/GoogleCloudPlatform/guest-agent/releases and update in /.github/workflows/build-gcp-guest-agent.yml.
 	GcpGuestImage = "ghcr.io/edgelesssys/gcp-guest-agent:20220927.00@sha256:3dea1ae3f162d2353e6584b325f0e325a39cda5f380f41e5a0ee43c6641d3905" // renovate:container
 	// ConstellationOperatorImage is the image for the constellation node operator.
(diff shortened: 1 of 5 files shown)


malt3 (approved):

internal/versions/versions.go
```diff
@@ -69,7 +69,7 @@ const (
 	// Check for new versions at https://github.com/GoogleCloudPlatform/guest-agent/releases and update in /.github/workflows/build-gcp-guest-agent.yml.
 	GcpGuestImage = "ghcr.io/edgelesssys/gcp-guest-agent:20220927.00@sha256:3dea1ae3f162d2353e6584b325f0e325a39cda5f380f41e5a0ee43c6641d3905" // renovate:container
 	// ConstellationOperatorImage is the image for the constellation node operator.
-	ConstellationOperatorImage = "ghcr.io/edgelesssys/constellation/node-operator:v2.6.0-pre.0.20230127131021-e174146e0c93@sha256:2c2f5f4e14b620c5a504534a399b845b8d8b5c7dafe9f7e98dd2b6981699f9de" // renovate:container
+	ConstellationOperatorImage = "ghcr.io/edgelesssys/constellation/node-operator:v2.6.0-pre.0.20230131091620-88e3da750e24@sha256:6502f9021c30385f166990af75af2d8da46a9cdeecb9851e2a47aa802d84cccb" // renovate:container
```
katexochen, line 72: Was this image built on main? Why wasn't it build automatically and updated through renovate?

derpsteb: It is built on main: https://github.com/edgelesssys/constellation/actions/runs/4052544118.
Wasn't aware I am not allowed to update the image manually.

katexochen: You are allowed. But the self reference in the workflow file is missing, so this will happen again.

PR #1120, opened 2023-01-31
ci: update Syft to 0.72.0 and Grype to 0.57.1

Nirusu:
### Proposed change(s)
- Upgrade Syft to 0.69.0
- Upgrade Grype to 0.56.0

Syft fixed the APK parsing issues in 0.68.1 caused by apko adding empty fields and finally added private key attestation back with 0.69.0. Let's give it a shot, since it should unblock updating the Alpine base images with newer versions of apko again.

Also updating Grype just for the sake of it. It's based on Syft 0.68.1 which should have any parsing issues fixed hopefully. Not the attestation part, but we don't care about that for Grype.

### Related issues
- #1017 

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/.github/actions/container_sbom/action.yml
+++ b/.github/actions/container_sbom/action.yml
@@ -34,8 +34,7 @@ runs:
         SYFT_ATTEST_PASSWORD: ${{ inputs.cosignPassword }} # Required for Syft 0.69.0+ as they overwrite COSIGN_PASSWORD
       run: |
         set -ex
-        echo "$COSIGN_PRIVATE_KEY" > cosign.key
-        syft attest --key cosign.key ${{ inputs.containerReference }} -o cyclonedx-json > container-image.att.json
+        syft attest --key env://COSIGN_PRIVATE_KEY ${{ inputs.containerReference }} -o cyclonedx-json > container-image.att.json
         cosign attach attestation ${{ inputs.containerReference }} --attestation container-image.att.json
         # TODO: type should be auto-discovered after issue is resolved:
         # https://github.com/sigstore/cosign/issues/2264
--- a/.github/actions/container_sbom/action.yml
+++ b/.github/actions/container_sbom/action.yml
@@ -31,6 +31,7 @@ runs:
         COSIGN_PUBLIC_KEY: ${{ inputs.cosignPublicKey }}
         COSIGN_PRIVATE_KEY: ${{ inputs.cosignPrivateKey }}
         COSIGN_PASSWORD: ${{ inputs.cosignPassword }}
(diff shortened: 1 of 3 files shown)


Nirusu:
```bash
syft attest --key cosign.key ghcr.io/edgelesssys/constellation/joinservice:v2.6.0-pre.0.20230131140552-27cae81bd7bc@sha256:cd33aacb5733f6cb7c9d9694a8d6c5337c78e019274ea1d531e25114f86b537b -o cyclonedx-json
2023/01/31 14:14:16 error during command execution: 1 error occurred:
	* unable to attest SBOM: exit status 1
```

🤦🏻 

Back to draft...

Nirusu:
0.69.0 / 0.69.1 is not fully suitable since it hardcodes `custom` predicates while we use CycloneDX.

This will be fixed in an upcoming release.

Personally I am still not super happy with the error messages generated by the new updates (just exit codes when something goes wrong) so I hope this will be fixed in some way, but I guess that's better than being blocked on old apko tooling / Alpine base images.

Nirusu:
Alright, this should be unblocked now. 0.70.0 has all the required fixes.

Only left issue is that, in case it fails, the error messages are not really useful, but it's not super hard to debug this locally.
I hope they fix this, though. Otherwise I can try fixing it myself (since PRs get faster responses since issues ;)), but given they are using a bus for the output, not really sure what a good fix would look like.

On the other hand, upgrading here means we can unblock upgrading Alpine base images again :)

Nirusu:
(ping @katexochen)

katexochen:
> we can unblock upgrading our Alpine base images again since we finally have the fixed APK parser in there

To what issue are you referring exactly? And which APK parser do you mean? The package hasher for apk?

katexochen:
Ah sorry, you were referring to the apk parser of Syft, right?

Nirusu:
Yes.

katexochen:
Not sure, but this still seems to fail due to syft? https://github.com/edgelesssys/constellation/actions/runs/4122737823

Nirusu:
Good catch, I used another image for testing but for some reason this fails again on another image. No idea why... Will investigate :/

katexochen:
Is this issue again blocked by upstream?

Nirusu:
It's blocked by the fact that some images seem(ed) to work but others don't - the output does not seem to be valid CycloneDX for some of them. Have not investigated yet why this is.

Nirusu:
@katexochen 
Alright, I just bypassed using "syft attest" and instead do it manually so I can manually specify the predicate type.
This now seems to work, see if you can break it ;)

katexochen (approved):
lgtm

katexochen:
Still not there :( https://github.com/edgelesssys/constellation/actions/runs/4243339556/jobs/7375956447

Nirusu:
Yes, on it :(

PR #1131, opened 2023-02-01
disable lb e2e test on aws until upstream issue is fixed

datosh:

### Proposed change(s)

- Disable e2e lb test on aws until https://github.com/kubernetes/cloud-provider-aws/issues/566 is fixed

### Additional info
- I was not able to reproduce any failings of the tests itself or hanging. The only problem is that the security group created for LB is not correctly cleaned up by CCM and therefore `constellation terminate` fails.
  - I have tested: v1.25 locally, v1.26 locally, [v1.25 manual pipeline ](https://github.com/edgelesssys/constellation/actions/runs/4063031347)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/e2e-test-weekly.yml
+++ b/.github/workflows/e2e-test-weekly.yml
@@ -75,6 +75,9 @@ jobs:
             version: "1.24"
           - test: "lb"
             version: "1.25"
+          # disable on AWS until cleanup is fixed: https://github.com/kubernetes/cloud-provider-aws/issues/566
+          - test: "lb"
+            provider: "aws"
           # Currently not supported on AWS.
           - test: "autoscaling"
             provider: "aws"


katexochen:
reproduction 1: https://github.com/edgelesssys/constellation/actions/runs/4065385901
reproduction 2: https://github.com/edgelesssys/constellation/actions/runs/4065392893

3u13r (approved):
LGTM

katexochen:
> reproduction 1: https://github.com/edgelesssys/constellation/actions/runs/4065385901
reproduction 2: https://github.com/edgelesssys/constellation/actions/runs/4065392893

We should investigate why the test is failing and fix the cause of it instead of disabling it. The resource leakage is annoying (thanks for opening the ticket!) but imo no reason to disable the test, as the leaking security group doesn't cost much and can easily be cleaned up once in a while.

datosh:
> > reproduction 1: https://github.com/edgelesssys/constellation/actions/runs/4065385901
> > reproduction 2: https://github.com/edgelesssys/constellation/actions/runs/4065392893
> 
> We should investigate why the test is failing and fix the cause of it instead of disabling it. The resource leakage is annoying (thanks for opening the ticket!) but imo no reason to disable the test, as the leaking security group doesn't cost much and can easily be cleaned up once in a while.

I will have another look. Thanks for providing these data points!

datosh:
Closed in favor of #1282

PR #1133, opened 2023-02-01
init: create kubeconfig file with unique user/cluster name

daniel-weisse:
### Proposed change(s)

- Move `--name` flag from `constellation create` to `constellation config generate`/`constellation iam create`
  - The value is now set in the config file if one wants to customize it at a later point
- Use the name flag together with the Cluster UID to create unique cluster and user name in the kubeconfig file generated by `constellation init`
  - This avoids issues when running multiple clusters and merging kubeconfig files

### Additional info
- e2e test run on GCP with the changes from this PR: https://github.com/edgelesssys/constellation/actions/runs/4075138413/jobs/7021142980

### Checklist

- [x] Link to Milestone


cli/internal/cmd/configgenerate.go
```diff
@@ -31,12 +32,14 @@ func newConfigGenerateCmd() *cobra.Command {
 		RunE:              runConfigGenerate,
 	}
 	cmd.Flags().StringP("file", "f", constants.ConfigFilename, "path to output file, or '-' for stdout")
+	cmd.Flags().String("name", "constell", "name to use for cluster creation and configuration")
```
3u13r, line 35: Do we even want to have a flag? We're already generating a config for the user to fill out. We can also provide a default value in there. Also we already do the name validation when reading the config.

3u13r (dismissed):

bootstrapper/internal/kubernetes/k8sapi/k8sutil.go
```diff
@@ -88,24 +89,24 @@ func (k *KubernetesUtil) InstallComponents(ctx context.Context, kubernetesCompon
 
 // InitCluster instruments kubeadm to initialize the K8s cluster.
```
datosh, line 90: Extend docs to mention returned kubeconf

datosh (approved):
Only one minor nit, otherwise LGTM!

datosh:
I just noticed we have the `--name` parameter in the animation of our repository README.md.

Any idea how that svg was created? How it can be modified?

bootstrapper/internal/initserver/initserver.go
```diff
@@ -161,9 +161,15 @@ func (s *Server) Init(ctx context.Context, req *initproto.InitRequest) (*initpro
 	// Check if we are running on a CVM
 	_, isCVM := s.issuer.(*snp.Issuer)
 
+	clusterName := req.ClusterName
```
3u13r, line 164: Should we add a `TODO` or similar to keep track of when we can depend on `req.ClusterName` to be set?

daniel-weisse: Since these changes only affect init, we can probably depend on the value being set as soon as the PR is merged.
We might want to do other checks or leave this check in as is if we want to ever offer the bootstrappers gRPC endpoints as a public interface.

3u13r: I thought that this was to eventually allow skew between released the CLI and Bootstrapper versions. 
But I agree that we should either decide that we only do validation in the config, or perform even more checks here. The discussion behind this is if we want to enforce that during create and init CLI version == VM image version == microservice version holds. 

I think that even the discussion if we want to expose the init endpoint as public interface is very far in the future. Right now, I think that the negative side outweighs the positive by a good margin.

malt3 (approved):

PR #1135, opened 2023-02-03
ci: e2e tests for upgrade

datosh:

### Proposed change(s)

- Add e2e test for `constellation upgrade execute`
  - Uses the latest released version of Constellation (currently v2.5.1) and upgrades to latest nightly build on main. This image should always be available in weekly e2e test run, since we now have https://github.com/edgelesssys/constellation/blob/main/.github/workflows/build-os-image-scheduled.yml
- Pipeline run: https://github.com/edgelesssys/constellation/actions/runs/4083108093

I will remove the `push:` event once the implementation is accepted.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/actions/e2e_test/action.yml
+++ b/.github/actions/e2e_test/action.yml
@@ -112,8 +112,9 @@ runs:
       run: |
         curl -fsSL https://github.com/edgelesssys/constellation/releases/download/${{ inputs.cliVersion }}/constellation-linux-amd64 > constellation
         chmod u+x constellation
-        ./constellation version
         echo "$(pwd)" >> $GITHUB_PATH
+        export PATH="$PATH:$(pwd)"
+        constellation version
         # Do not spam license server from pipeline
         sudo sh -c 'echo "127.0.0.1 license.confidential.cloud" >> /etc/hosts'
 
--- a/.github/actions/e2e_test/action.yml
+++ b/.github/actions/e2e_test/action.yml
@@ -23,6 +23,9 @@ inputs:
     description: "Is OS img a debug img?"
     default: "true"
     required: true
+  cliVersion:
+    description: "Version of a released CLI to download, e.g. 'v2.3.0', leave empty to build it."
+    required: false
   kubernetesVersion:
     description: "Kubernetes version to create the cluster from."
     required: false
@@ -96,11 +99,23 @@ runs:
         echo "hostArch=$(go env GOARCH)" >> $GITHUB_OUTPUT
 
     - name: Build CLI
+      if: inputs.cliVersion == ''
       uses: ./.github/actions/build_cli
       with:
         targetOS: ${{ steps.determine-build-target.outputs.hostOS }}
         targetArch: ${{ steps.determine-build-target.outputs.hostArch }}
         enterpriseCLI: ${{ inputs.keepMeasurements }}
+
+    - name: Fetch CLI
+      if: inputs.cliVersion != ''
+      shell: bash
+      run: |
+        curl -L https://github.com/edgelesssys/constellation/releases/download/${{ inputs.cliVersion }}/constellation-linux-amd64 > constellation
+        chmod u+x constellation
+        ./constellation version
+        echo "$(pwd)" >> $GITHUB_PATH
+        # Do not spam license server from pipeline
+        sudo sh -c 'echo "127.0.0.1 license.confidential.cloud" >> /etc/hosts'
 
     # macOS runners don't have Docker preinstalled, so they cannot build the bootstrapper.
     # But we can use a Linux runner to build it and store/retrieve it from the action cache.
--- a/.github/actions/e2e_test/action.yml
+++ b/.github/actions/e2e_test/action.yml
@@ -110,7 +110,7 @@ runs:
       if: inputs.cliVersion != ''
       shell: bash
       run: |
-        curl -L https://github.com/edgelesssys/constellation/releases/download/${{ inputs.cliVersion }}/constellation-linux-amd64 > constellation
+        curl -fsSL https://github.com/edgelesssys/constellation/releases/download/${{ inputs.cliVersion }}/constellation-linux-amd64 > constellation
         chmod u+x constellation
         ./constellation version
         echo "$(pwd)" >> $GITHUB_PATH
--- a/.github/workflows/e2e-upgrade.yml
+++ b/.github/workflows/e2e-upgrade.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Setup Go environment
         uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
         with:
-          go-version: "1.19.4"
+          go-version: "1.19.5"
 
       - name: Login to Azure
         if: matrix.provider == 'azure'
@@ -73,7 +73,7 @@ jobs:
 
       - name: Run upgrade test
         run: |
-          go test ./e2e/internal/upgrade -v --timeout=0 --tags=e2eupgrade
+          go test ./e2e/internal/upgrade -v --timeout=5h30m --tags=e2eupgrade
         env:
           KUBECONFIG: ${{ steps.e2e_test.outputs.kubeconfig }}
 
@@ -104,7 +104,7 @@ jobs:
           kubeconfig: ${{ steps.e2e_test.outputs.kubeconfig }}
 
       - name: Notify teams channel
-        if: ${{ failure() && github.ref == 'refs/heads/main' }}
+        if: failure() && github.ref == 'refs/heads/main'
         continue-on-error: true
         shell: bash
         working-directory: .github/actions/e2e_test
--- a/.github/workflows/e2e-upgrade.yml
+++ b/.github/workflows/e2e-upgrade.yml
@@ -1,10 +1,34 @@
 name: e2e test upgrade
 
 on:
-  # TODO: remove push
-  push:
   workflow_dispatch:
+    inputs:
+      fromVersion:
+        description: CLI version to create a new cluster with. This has to be a released version, e.g., 'v2.1.3'.
+        type: string
+        required: true
+      toVersion:
+        description: CLI version to execute upgrade with, e.g., 'v2.1.3', or empty to build HEAD.
+        type: string
+        required: false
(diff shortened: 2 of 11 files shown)


.github/actions/e2e_test/action.yml
```diff
@@ -96,12 +99,24 @@ runs:
...
       with:
         targetOS: ${{ steps.determine-build-target.outputs.hostOS }}
         targetArch: ${{ steps.determine-build-target.outputs.hostArch }}
         enterpriseCLI: ${{ inputs.keepMeasurements }}
 
+    - name: Fetch CLI
+      if: inputs.cliVersion != ''
+      shell: bash
+      run: |
+        curl -L https://github.com/edgelesssys/constellation/releases/download/${{ inputs.cliVersion }}/constellation-linux-amd64 > constellation
```
katexochen, line 113:
```suggestion
        curl -fsSL https://github.com/edgelesssys/constellation/releases/download/${{ inputs.cliVersion }}/constellation-linux-amd64 > constellation
```

.github/workflows/e2e-upgrade.yml
```diff
@@ -0,0 +1,130 @@
+name: e2e test upgrade
+
+on:
+  # TODO: remove push
+  push:
```
katexochen, line 5: TODO

.github/workflows/e2e-upgrade.yml
```diff
@@ -0,0 +1,130 @@
...
+      - name: Check out repository
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # v3.2.0
+        with:
+          fetch-depth: 0
+          ref: ${{ !github.event.pull_request.head.repo.fork && github.head_ref || '' }}
+
+      - name: Setup Go environment
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
+        with:
+          go-version: "1.19.4"
```
katexochen, line 35:
```suggestion
          go-version: "1.19.5"
```

.github/workflows/e2e-upgrade.yml
```diff
@@ -0,0 +1,130 @@
...
+          azureUserAssignedIdentity: ${{ secrets.AZURE_E2E_USER_ASSIGNED_IDENTITY }}
+          azureResourceGroup: ${{ steps.az_resource_group_gen.outputs.res_group_name }}
+          gcpProject: ${{ secrets.GCP_E2E_PROJECT }}
+          gcp_service_account: "constellation-e2e@constellation-331613.iam.gserviceaccount.com"
+          gcpClusterServiceAccountKey: ${{ secrets.GCP_CLUSTER_SERVICE_ACCOUNT }}
+          test: "nop"
+
+      - name: Run upgrade test
+        run: |
+          go test ./e2e/internal/upgrade -v --timeout=0 --tags=e2eupgrade
```
katexochen, line 76:
```suggestion
          go test ./e2e/internal/upgrade -v --timeout=5h30m --tags=e2eupgrade
```
Wouldn't it be better to timeout so we can clean up instead of getting killed by GitHub?

.github/workflows/e2e-upgrade.yml
```diff
@@ -0,0 +1,130 @@
...
+
+      - name: Always terminate cluster
+        if: always()
+        continue-on-error: true
+        uses: ./.github/actions/constellation_destroy
+        with:
+          kubeconfig: ${{ steps.e2e_test.outputs.kubeconfig }}
+
+      - name: Notify teams channel
+        if: ${{ failure() && github.ref == 'refs/heads/main' }}
```
katexochen, line 107:
```suggestion
        if: failure() && github.ref == 'refs/heads/main'
```
Shouldn't be needed

e2e/internal/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	os.Exit(m.Run())
+}
+
+func mustDecodeHex(s string) [32]byte {
+	val, _ := hex.DecodeString(s)
+	return *(*[32]byte)(val)
+}
+
```
katexochen, line 63: nit: Helper functions should be at the bottom.

e2e/internal/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,341 @@
...
+func testNodesEventuallyHaveImage(t *testing.T, k *kubernetes.Clientset, wantImage string) {
+	require.Eventually(t, func() bool {
+		nodes, err := k.CoreV1().Nodes().List(context.Background(), metaV1.ListOptions{})
+		if err != nil {
+			fmt.Println(err)
+			return false
+		}
+
+		allUpdated := true
+		fmt.Printf("Node status (%v):\n", time.Now())
```
katexochen, line 115:
```suggestion
		t.Log(fmt.Sprintf("Node status (%v):", time.Now()))
```
Might be better to use t.Log ?

e2e/internal/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,341 @@
...
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	coreV1 "k8s.io/api/core/v1"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	constellationCLIVersion = "2.5.1"
```
katexochen, line 43: If we do a release, does this value need to be changed? Or are we always starting at v2.5.1?

katexochen (changes requested):
Some with a bit more understanding of the upgrade process should review this, too.

e2e/internal/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,341 @@
...
+		info.wantImage = imageInfo.GCP["sev-es"]
+	case cloudprovider.Azure:
+		info.wantImage = imageInfo.Azure["cvm"]
+	case cloudprovider.AWS:
+		info.wantImage = imageInfo.AWS["eu-central-1"]
+	default:
+		return upgradeInfo{}, errors.New("finding wanted image")
+	}
+
+	return info, nil
```
derpsteb, line 340: I think this could be replaced with a call to [getCompatibleImageMeasurements](https://github.com/edgelesssys/constellation/blob/main/cli/internal/cmd/upgradeplan.go#L200)? The `images` slice is expected to be a slice of short paths.

derpsteb:
I think it would make sense to have at least these three arguments for the workflow: `from`, `to` and `toImage`. 
It should be possible to set `from` and `to` to release versions. In that case the pipeline would download the release CLIs, create the cluster with the `from` CLI and run the upgrade with the `to` CLI. The target image is always `toImage`. Alternatively `to` could be left empty. In that case the CLI will be built from HEAD. 

This would allow us to continuously see that released version can still be upgraded to current main. It also makes it easy to see if two released versions are still compatible. E.g. after a change to the versionsapi. 

Thoughts?

e2e/internal/upgrade/upgrade_test.go
datosh: Made a pipeline input.

e2e/internal/upgrade/upgrade_test.go
datosh: Good suggestion to reuse existing code! 

I did look into it, but it seems the suggested version uses [measurements.json](https://cdn.confidential.cloud/constellation/v1/ref/-/stream/stable/v2.5.0/image/csp/gcp/measurements.json) which is only available in a released version. I keep the code based on [measurements.image.json](https://cdn.confidential.cloud/constellation/v1/ref/-/stream/stable/v2.5.0/image/csp/gcp/measurements.image.json) since it is available for all branches. 

I will have another look why we have two files in a seperate PR.

.github/workflows/e2e-upgrade.yml
```diff
@@ -0,0 +1,183 @@
...
+          sudo sh -c 'echo "127.0.0.1 license.confidential.cloud" >> /etc/hosts'
+
+      - name: Find latest nightly image
+        id: find-image
+        if: inputs.toImage == ''
+        uses: ./.github/actions/versionsapi
+        with:
+          command: latest
+          ref: main
+          stream: nightly
```
derpsteb, line 121: Could you add a test after this step that checks that the located image is not older than x hrs? Maybe 2hrs or sth. I already see myself wondering why the upgrade failed only to realize that I didn't build an image with an up2date bootstrapper. 

Alternatively we could trigger the image build here. Or use debug images. Not sure what the best approach would be.

e2e/internal/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,344 @@
...
+	k, err := kubectl.New()
+	require.NoError(err)
+	assert.NotNil(k)
+
+	testNodesEventuallyAvailable(t, k, *wantControl, *wantWorker)
+	testPodsEventuallyReady(t, k, "kube-system")
+
+	testCLIHasVersion(t, *toVersion)
+	wantImage := writeUpgradeConfig(t, *toImage)
+	cmd := exec.CommandContext(context.Background(), "constellation", "upgrade", "execute")
```
derpsteb, line 73:
```suggestion
	cmd := exec.CommandContext(context.Background(), "constellation", "upgrade", "apply")
```

e2e/internal/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,344 @@
...
+	msg, err := cmd.CombinedOutput()
+	require.NoErrorf(err, "%s", string(msg))
+
+	testNodesEventuallyHaveImage(t, k, wantImage)
+}
+
+func writeUpgradeConfig(t *testing.T, toImage string) string {
```
derpsteb, line 80: This would need some slight changes. We don't have the upgrade conf anymore since the `upgrade check`/`upgrade apply` refactoring that I merged Wednesday. Maybe we can do the required changes in a pair programming session later today?

derpsteb:
I will have to read up on the existing code and see if there are any more relevant changes. Since I will pick this topic up from Fabian I am closing this PR and open up a new PR once ready.

PR #1136, opened 2023-02-03
cli: add option to automatically merge kubeconfig file on init

daniel-weisse:
### Proposed change(s)

Add `--merge-kubeconfig` flag to `constellation init`.
If set, automatically merges the config generated by `init` into the default configuration in `$HOME/.kube/config` and sets it as the current context.
Users are then able to connect to the cluster without having to do any extra setup.

If the `KUBECONFIG` env variable is set, the new config is still merged into the default config, but we cant reliably tell if `kubectl` will load this config, so a warning is printed in this case telling the user they might need to unset the `KUBECONFIG` env var

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/init.go
+++ b/cli/internal/cmd/init.go
@@ -62,7 +62,8 @@ func NewInitCmd() *cobra.Command {
 }
 
 type initCmd struct {
-	log debugLog
+	log    debugLog
+	merger configMerger
 }
 
 // runInitialize runs the initialize command.
@@ -86,7 +87,7 @@ func runInitialize(cmd *cobra.Command, args []string) error {
 	ctx, cancel := context.WithTimeout(cmd.Context(), time.Hour)
 	defer cancel()
 	cmd.SetContext(ctx)
-	i := &initCmd{log: log}
+	i := &initCmd{log: log, merger: &kubeconfigMerger{log: log}}
 	return i.initialize(cmd, newDialer, fileHandler, license.NewClient(), spinner)
 }
 
@@ -261,7 +262,7 @@ func (i *initCmd) writeOutput(
 	i.log.Debugf("Wrote kubeconfig to file: %s", constants.AdminConfFilename)
 
 	if mergeConfig {
-		if err := i.mergeKubeonfig(fileHandler); err != nil {
+		if err := i.merger.mergeConfigs(constants.AdminConfFilename, fileHandler); err != nil {
 			return fmt.Errorf("merging kubeconfig: %w", err)
 		}
 		writeRow(tw, "Kubernetes configuration merged with default config", "")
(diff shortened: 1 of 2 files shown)


cli/internal/cmd/init.go
```diff
@@ -249,17 +259,72 @@ func (i *initCmd) writeOutput(idFile clusterid.File, resp *initproto.InitRespons
 	if err := fileHandler.Write(constants.AdminConfFilename, resp.Kubeconfig, file.OptNone); err != nil {
 		return fmt.Errorf("writing kubeconfig: %w", err)
 	}
-	i.log.Debugf("Wrote out kubeconfig")
+	i.log.Debugf("Wrote kubeconfig to file")
```
datosh, line 262: Since it is `debug` level anyway we could be extra verbose and add the file name it was written to as well.

cli/internal/cmd/init.go
```diff
@@ -249,17 +259,72 @@ func (i *initCmd) writeOutput(idFile clusterid.File, resp *initproto.InitRespons
...
+		fmt.Fprintln(wr, "Constellation kubeconfig merged with default config.")
 
-	fmt.Fprintln(wr, "You can now connect to your cluster by executing:")
-	fmt.Fprintf(wr, "\texport KUBECONFIG=\"$PWD/%s\"\n", constants.AdminConfFilename)
+		if os.Getenv(clientcmd.RecommendedConfigPathEnvVar) != "" {
+			fmt.Fprintln(wr, "Warning: KUBECONFIG environment variable is set.")
+			fmt.Fprintln(wr, "You may need to unset it to use the default config and connect to your cluster.")
+		} else {
+			fmt.Fprintln(wr, "You can now connect to your cluster.")
+		}
```
datosh, line 289: great, very thorough checking 🙂 

Should we add some tests to make sure all branches produce the expected result?

cli/internal/cmd/init.go
```diff
@@ -249,17 +259,72 @@ func (i *initCmd) writeOutput(idFile clusterid.File, resp *initproto.InitRespons
...
+	cfg, err := loadingRules.Load()
+	if err != nil {
+		return fmt.Errorf("loading merged kubeconfig: %w", err)
+	}
+	if err := clientcmdapi.FlattenConfig(cfg); err != nil {
+		return fmt.Errorf("flattening merged kubeconfig: %w", err)
+	}
+
+	// Set the current context to the cluster we just created
+	cfg.CurrentContext = fmt.Sprintf("%s-admin@%s", clusterName, clusterName)
```
datosh, line 312: This feels error prone to me, if we change context name in future we need to remember to change it here as well. Could we easily read out the context name instead of regenerating it?

datosh (approved):

PR #1137, opened 2023-02-03
cli: move Talos YAML dependencies into its own smaller package

Nirusu:
### Proposed change(s)
- Move Talos' YAML encoder & docgen to: https://github.com/edgelesssys/annotated-yaml

### DIFF
--- a/.github/actions/install_docgen/action.yml
+++ b/.github/actions/install_docgen/action.yml
@@ -4,23 +4,17 @@ description: |
 runs:
   using: "composite"
   steps:
-    - name: Checkout talos
+    - name: Checkout annotated-yaml
       uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
       with:
         fetch-depth: 0
-        repository: "siderolabs/talos"
-        ref: "v1.3.1"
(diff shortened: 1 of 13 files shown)


datosh:
I'm a bit suprised there are basically no changes in our `go.sum`, since it sounded like we are pulling in a lot of things via this `talos/pkg/machinery` dependency. Not sure if there is any benefit now, if we need to maintain our own "fork" instead of letting Renovate bump the version every now and then.

Nirusu:
There's not that many dependencies:
https://github.com/siderolabs/talos/blob/main/pkg/machinery/go.mod

Nirusu:
Test is failing because I initially tagged v1.0.0 but switched to v0.1.0. I will also change and replace the v0.1.0 a few times, so I hope tests will pass in later when the Go proxy empties its cache.

Nirusu:
But maybe we should decide first whether to go through with this or not, if it doesn't serve any benefit 😄

Nirusu:
From what I can tell Go really only imports the encoder package, not the rest of the machinery package.

In this case I think this can definitely be closed for now. In case Talos breaks something / adds more dependencies, we can revisit this again.

PR #1148, opened 2023-02-06
cli: fix iam rollback

msanft:
### Proposed change(s)

- Check whether Terraform state directory already exists on the `iam create` command and if it exists, do not continue with execution
- Rename `DestroyCluster` to `DestroyResources` as it can also destroy IAM resources

### Additional info
- Previously, if the `iam create` command is executed twice in a row, the second execution would error because the Terraform state dir already exists. On this error, it would also rollback the existing directory which deletes the Terraform state of a previous execution of the command

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cloudcmd/clients.go
+++ b/cli/internal/cloudcmd/clients.go
@@ -19,7 +19,7 @@ type terraformClient interface {
 	PrepareWorkspace(path string, input terraform.Variables) error
 	CreateCluster(ctx context.Context) (terraform.CreateOutput, error)
 	CreateIAMConfig(ctx context.Context, provider cloudprovider.Provider) (terraform.IAMOutput, error)
-	DestroyResources(ctx context.Context) error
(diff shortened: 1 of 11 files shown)


cli/internal/cloudcmd/clients.go
```diff
@@ -19,7 +19,7 @@ type terraformClient interface {
 	PrepareWorkspace(path string, input terraform.Variables) error
 	CreateCluster(ctx context.Context) (terraform.CreateOutput, error)
 	CreateIAMConfig(ctx context.Context, provider cloudprovider.Provider) (terraform.IAMOutput, error)
-	DestroyCluster(ctx context.Context) error
+	DestroyResources(ctx context.Context) error
```
katexochen, line 22: Maybe just call it destroy, as @miampf did on #946 .

daniel-weisse (approved):

PR #1149, opened 2023-02-06
cli: add missing gcp values to config

msanft:
### Proposed change(s)

- The `iam create` command was not adding the GCP zone, region and project id values to the configuration file or the terminal output. To create a better UX and to align with the behavior for AWS and Azure, write these values to the configuration file and print them if no config file is being generated
- Print a reminder to fill the values in the configuration file for all provide

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/cli/internal/cmd/iam.go
+++ b/cli/internal/cmd/iam.go
@@ -353,6 +353,7 @@ func (c *awsIAMCreator) printOutputValues(cmd *cobra.Command, flags iamFlags, ia
 	cmd.Printf("zone:\t\t\t%s\n", flags.aws.zone)
 	cmd.Printf("iamProfileControlPlane:\t%s\n", iamFile.AWSOutput.ControlPlaneInstanceProfile)
 	cmd.Printf("iamProfileWorkerNodes:\t%s\n\n", iamFile.AWSOutput.WorkerNodeInstanceProfile)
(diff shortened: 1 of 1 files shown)


cli/internal/cmd/iam.go
```diff
@@ -353,6 +353,7 @@ func (c *awsIAMCreator) printOutputValues(cmd *cobra.Command, flags iamFlags, ia
 	cmd.Printf("zone:\t\t\t%s\n", flags.aws.zone)
 	cmd.Printf("iamProfileControlPlane:\t%s\n", iamFile.AWSOutput.ControlPlaneInstanceProfile)
 	cmd.Printf("iamProfileWorkerNodes:\t%s\n\n", iamFile.AWSOutput.WorkerNodeInstanceProfile)
+	cmd.Println("Your IAM configuration was created successfully. Please fill the above values into your configuration file.")
```
daniel-weisse, line 356:
```suggestion
```
This message is already printed by the [main `create` function](https://github.com/edgelesssys/constellation/blob/18a6007356dbde4789a51f016f6dd68b88cc0e98/cli/internal/cmd/iam.go#L229).
Instead of also adding the output to the AWS and GCP IAM creator, it should be removed from the Azure creator

daniel-weisse (changes requested):

daniel-weisse (approved):

PR #1151, opened 2023-02-07
bootstrapper: retry helm chart installation

derpsteb:
### Proposed change(s)

Sometimes cluster initilization fails on Azure with the error: `Error: init call: rpc error: code = Internal desc = initializing cluster: installing cert-manager: helm install: failed post-install: timed out waiting for the condition`.

This PR introduces a retry loop that aborts after 5 mins for all helm installations.
An alternative solution would be to wait after konnectivity installation, however this approach seems more robust to me and might mitigate other unknown race conditions.

Judging from the logs and the frequency that we see this bug 9/10 installation should not run into the retry case.
For those that do, installation should work on the second try (in the logs the api check becomes healthy seconds after the first try is aborted).

### Background Information
The timeout errors are most likely connected to konnectivty agents not being ready when starting 
the installation. This may become a problem during installation when test-requests are sent to the webhook that cert-manager installs. If the sender is on a different node than the webhook pod and the sender's konnectivity-agent is not ready, the request will fail.

During installation we can see the following error messages on control plane node 2:
- cert-manager-startupapicheck: `failed calling webhook "webhook.cert-manager.io": failed to call webhook: Post "[https://cert-manager-webhook.kube-system.svc:443/mutate?timeout=10s":](https://cert-manager-webhook.kube-system.svc/mutate?timeout=10s%22:) No agent available`
- konnectivity-agent: `cannot connect once" err="rpc error: code = Unavailable desc = connection error: desc = \"transport: authentication handshake failed: x509: certificate signed by unknown authority`

The `cert-manager-webhook` pod is scheduled on node 0.

The problem resolves itself after <1min and `cert-manager-startupapicheck` reports: `The cert-manager API is ready`.

[Opensearch query](https://search-e2e-logs-y46renozy42lcojbvrt3qq7csm.eu-central-1.es.amazonaws.com/_dashboards/goto/c700eeee7321e6f2671a5d3cc1c66a10?security_tenant=global).

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/bootstrapper/internal/helm/helm.go
+++ b/bootstrapper/internal/helm/helm.go
@@ -206,7 +206,7 @@ func (h *Client) install(ctx context.Context, chartRaw []byte, values map[string
 	// Other errors will not be retried.
 	newCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
 	defer cancel()
-	if err :=  retrier.Do(newCtx); err != nil {
+	if err := retrier.Do(newCtx); err != nil {
 		return fmt.Errorf("helm install: %w", err)
 	}
 	return nil
--- a/bootstrapper/internal/helm/helm.go
+++ b/bootstrapper/internal/helm/helm.go
@@ -21,10 +21,13 @@ import (
 	"github.com/edgelesssys/constellation/v2/internal/constants"
 	"github.com/edgelesssys/constellation/v2/internal/deploy/helm"
 	"github.com/edgelesssys/constellation/v2/internal/logger"
+	"github.com/edgelesssys/constellation/v2/internal/retry"
 	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chart/loader"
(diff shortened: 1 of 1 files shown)


bootstrapper/internal/helm/helm.go
```diff
@@ -178,16 +183,54 @@ func (h *Client) installCiliumGCP(ctx context.Context, kubectl k8sapi.Client, re
...
+	}
+	retrier := retry.NewIntervalRetrier(doer, 30*time.Second, isTimeoutErr)
+
+	// Since we have no precise retry condition we want to stop retrying after 5 minutes.
+	// The helm library only reports a timeout error in the error cases we currently know.
+	// Other errors will not be retried.
+	newCtx, cancel := context.WithTimeout(ctx, 5*time.Minute)
+	defer cancel()
+	err = retrier.Do(newCtx)
 	if err != nil {
```
katexochen, line 210:
```suggestion
	if err :=  retrier.Do(newCtx); err != nil {
```

malt3 (approved):

PR #1152, opened 2023-02-07
ci: refactor image measurement generation

datosh:
### Proposed change(s)

- Merge `measurements.image.json` and `measurements.json` into one file
  - Delete `measurements.image.json`
  - Measurements that were previously read dynamically from a cluster in `generate-measurements.yml` are now pinned. This improves security since we have to manually change (authorize) measurements if they change in a CSP environment.
  - Remove measurements 5 & 10 from GCP and 5 from Azure. These change frequently and we are not able to precompute them. 
  - Script used to generate summarize measurements from previous relases:
  ```
  for csp in gcp azure aws qemu; do
      for version in v2.3.0 v2.4.0 v2.5.0 v2.5.1; do
          curl -sL https://cdn.confidential.cloud/constellation/v1/ref/-/stream/stable/${version}/image/csp/${csp}/measurements.json > measurements.${csp}.${version}.json
      done
  done
  
  for csp in gcp azure aws; do
      echo "========================="
      echo "=====CSP: ${csp}========="
      echo "========================="
      # Compare
      for idx in {0..15}; do
          echo "=====IDX: ${idx}========="
          yq e ".measurements.${idx}.expected" measurements.${csp}.v2.3.0.json
          yq e ".measurements.${idx}.expected" measurements.${csp}.v2.4.0.json
          yq e ".measurements.${idx}.expected" measurements.${csp}.v2.5.0.json
          yq e ".measurements.${idx}.expected" measurements.${csp}.v2.5.1.json
      done
  done
  ```
- `generate-measurements.yml` was removed altogether. We should learn about changing measurements since `e2e-verify` does the checking as well as during release pipeline

Pipeline: https://github.com/edgelesssys/constellation/actions/runs/4115538800

New `measurements.json`:
```
curl -sL https://cdn.confidential.cloud/constellation/v1/ref/ref-image-measurement-generation/stream/nightly/v2.6.0-pre.0.20230207151923-124e74419528/image/csp/gcp/measurements.json
curl -sL https://cdn.confidential.cloud/constellation/v1/ref/ref-image-measurement-generation/stream/nightly/v2.6.0-pre.0.20230207151923-124e74419528/image/csp/azure/measurements.json
curl -sL https://cdn.confidential.cloud/constellation/v1/ref/ref-image-measurement-generation/stream/nightly/v2.6.0-pre.0.20230207151923-124e74419528/image/csp/aws/measurements.json
curl -sL https://cdn.confidential.cloud/constellation/v1/ref/ref-image-measurement-generation/stream/nightly/v2.6.0-pre.0.20230207151923-124e74419528/image/csp/qemu/measurements.json
```

### Checklist

- [x] Link to Milestone


.github/workflows/build-os-image.yml
```diff
@@ -581,18 +581,163 @@ jobs:
...
+                .measurements.2.expected = "3d458cfe55cc03ea1f443f1562beec8df51c75e14a9fcf9a7234a13f198e7969" |
+                .measurements.3.warnOnly = true |
+                .measurements.3.expected = "3d458cfe55cc03ea1f443f1562beec8df51c75e14a9fcf9a7234a13f198e7969" |
+                .measurements.4.warnOnly = false |
+                .measurements.5.warnOnly = true |
+                .measurements.5.expected = "9d5f8c2e0464c69552e593be32a4b089f7e77f4ed12d60cdcd94120b6ccaccf8" |
+                .measurements.6.warnOnly = true |
+                .measurements.6.expected = "3d458cfe55cc03ea1f443f1562beec8df51c75e14a9fcf9a7234a13f198e7969" |
+                .measurements.7.warnOnly = true |
+                .measurements.7.expected = "b1e9b305325c51b93da58cbf7f92512d8eebfa01143e4d8844e40e062e9b6cd5" |
```
Nirusu, line 657: PCR 7 is secure boot. I know that's configurable in our setup, but eh, not sure to just set this to `warnOnly`.

.github/workflows/build-os-image.yml
```diff
@@ -581,18 +581,163 @@ jobs:
...
+                .measurements.8.warnOnly = false |
+                .measurements.9.warnOnly = false |
+                .measurements.10.warnOnly = true |
+                .measurements.10.expected = "7f96fbc55e1d2a0de46e5d44658c06ef102d1198703efa69f2ea6b5aa1c9a176" |
+                .measurements.11.warnOnly = false |
+                .measurements.12.warnOnly = false |
+                .measurements.13.warnOnly = false |
+                .measurements.14.warnOnly = true |
+                .measurements.14.expected = "d7c4cc7ff7933022f013e03bdee875b91720b5b86cf1753cad830f95e791926f" |
```
Nirusu, line 666: PCR 14 is coming from our bootchain IIRC.

daniel-weisse: No, we use [4,8,9,11,12,13,15].
Not 100% sure where 14 comes from, but we don't enforce it currently.
Maybe @malt3 knows if this is some part of systemd?

daniel-weisse:
>Remove measurements 5 & 10 from GCP and 5 from Azure.

Did something change on GCP for this to happen?
As far as I know these PCRs were stable since we first checked them

.github/workflows/build-os-image.yml
Nirusu: PCR 14 comes from shim:
https://github.com/rhboot/shim/blob/main/README.tpm

So part of our bootchain / secure boot.

.github/workflows/build-os-image.yml
```diff
@@ -581,18 +581,163 @@ jobs:
           jq -sSc '.[0] * .[1] * .[2] * .[3]' ${{ github.workspace }}/pcr-* > ${{ github.workspace }}/pcrs-${{ matrix.csp }}.json
           echo "::endgroup::"
 
+      - name: Add static PCRs
```
derpsteb, line 584: I am wondering if there could be a different place than a workflow file where we store the PCRs. Maybe the image folder? 

A file in that folder could hold all measurements in a json. This step then takes just the selected CSP and redirects that section into the `pcrs-$CSP.json` for upload.

.github/workflows/build-os-image.yml
datosh: We could set this to `warnOnly = false`. This value has been stable for the last 4 releases.

datosh:
> > Remove measurements 5 & 10 from GCP and 5 from Azure.
> 
> Did something change on GCP for this to happen? As far as I know these PCRs were stable since we first checked them

My understanding is that they are "stable" as in, they won't change for a given version of Constellation. Still these values are different for each binary release and we cannot statically calculate them (is that correct @malt3 ?). The only way to generate them is stand up a Constellation cluster in a CSP environment and trust the read values, which is exactly what don't do in this PR.

.github/workflows/build-os-image.yml
datosh: What is the motivation to have these in a separate JSON file in the image folder? Are there other consumers of this file? 

In general I think it makes sense to have the information as local as possible if there is only a single consumer (currently the pipeline).

.github/workflows/build-os-image.yml
datosh: We do not rely on secure boot as part of the security guarantees. We will keep `warnOnly = true` to prevent any problems when a user decides to enable secure boot anyway.

datosh:
> > > Remove measurements 5 & 10 from GCP and 5 from Azure.
> > 
> > 
> > Did something change on GCP for this to happen? As far as I know these PCRs were stable since we first checked them
> 
> My understanding is that they are "stable" as in, they won't change for a given version of Constellation. Still these values are different for each binary release and we cannot statically calculate them (is that correct @malt3 ?). The only way to generate them is stand up a Constellation cluster in a CSP environment and trust the read values, which is exactly what don't do in this PR.

As discussed in a meeting, we will remove 5 & 10 since they do not provide any additional security guarantees right now.

malt3 (approved):

PR #1159, opened 2023-02-09
image: verbose debugging options

leongross:
### Proposed change(s)

1. Add new `systemd` unit that checks if the kernel command line option `constellation.debug` is set. If it is `true`, the following services are started with the `--debug` flag via their respective systemd unit file
- [x] `bootstrapper`
- [x] `upgrade-agent`
- [x] `disk-mapper`

2. Add `lsblk` like log that shows the information for all available block devices

### Checklist

- [x] Link to Milestone

### DIFF
--- a/disk-mapper/cmd/main.go
+++ b/disk-mapper/cmd/main.go
@@ -134,7 +134,9 @@ func main() {
 		vtpm.OpenVTPM,
 	)
 
-	setupManger.LogDevices()
+	if err := setupManger.LogDevices(); err != nil {
+		log.With(zap.Error(err)).Fatalf("Failed to log devices")
+	}
 
 	// prepare the state disk
 	if mapper.IsLUKSDevice() {
--- a/disk-mapper/cmd/main.go
+++ b/disk-mapper/cmd/main.go
@@ -134,6 +134,8 @@ func main() {
 		vtpm.OpenVTPM,
 	)
 
+	setupManger.LogDevices()
+
 	// prepare the state disk
 	if mapper.IsLUKSDevice() {
 		// set up rejoin client
--- a/disk-mapper/internal/setup/setup.go
+++ b/disk-mapper/internal/setup/setup.go
@@ -192,6 +192,7 @@ func (s *Manager) LogDevices() error {
 		dev := "/dev/" + device.Name()
 		if err := syscall.Statfs(dev, &stat); err != nil {
 			s.log.Errorf("failed to statfs %s: %v", dev, err)
+			continue
 		}
 
 		// get the raw size, in bytes
--- a/disk-mapper/internal/setup/setup.go
+++ b/disk-mapper/internal/setup/setup.go
@@ -16,6 +16,7 @@ import (
 	"crypto/rand"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"os"
 	"path/filepath"
@@ -75,8 +76,8 @@ func New(log *logger.Logger, csp string, diskPath string, fs afero.Afero,
 // Once the disk is mapped, the function taints the node as initialized by updating it's PCRs.
 func (s *Manager) PrepareExistingDisk(recover RecoveryDoer) error {
 	s.log.Infof("Preparing existing state disk")
-	uuid := s.mapper.DiskUUID()
 
+	uuid := s.mapper.DiskUUID()
 	endpoint := net.JoinHostPort("0.0.0.0", strconv.Itoa(constants.RecoveryPort))
 
 	passphrase, measurementSecret, err := recover.Do(uuid, endpoint)
@@ -91,6 +92,7 @@ func (s *Manager) PrepareExistingDisk(recover RecoveryDoer) error {
 	if err := s.mounter.MkdirAll(stateDiskMountPath, os.ModePerm); err != nil {
 		return err
 	}
+
 	// we do not care about cleaning up the mount point on error, since any errors returned here should cause a boot failure
 	if err := s.mounter.Mount(filepath.Join("/dev/mapper/", stateDiskMappedName), stateDiskMountPath, "ext4", syscall.MS_RDONLY, ""); err != nil {
 		return err
@@ -100,6 +102,7 @@ func (s *Manager) PrepareExistingDisk(recover RecoveryDoer) error {
 	if err != nil {
 		return err
 	}
+
 	clusterID, err := attestation.DeriveClusterID(measurementSecret, measurementSalt)
 	if err != nil {
 		return err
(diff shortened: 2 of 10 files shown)


disk-mapper/internal/setup/setup.go
```diff
@@ -74,9 +75,8 @@ func New(log *logger.Logger, csp string, diskPath string, fs afero.Afero,
 // PrepareExistingDisk requests and waits for a decryption key to remap the encrypted state disk.
 // Once the disk is mapped, the function taints the node as initialized by updating it's PCRs.
 func (s *Manager) PrepareExistingDisk(recover RecoveryDoer) error {
-	s.log.Infof("Preparing existing state disk")
 	uuid := s.mapper.DiskUUID()
-
+	s.log.Infof("Preparing existing state disk (UUID: %s)", uuid)
```
daniel-weisse, line 79:
```suggestion
	s.log.With(zap.String("uuid", uuid)).Infof("Preparing existing state disk")
```
Use structured logging to add context to messages

disk-mapper/internal/setup/setup.go
```diff
@@ -119,7 +121,7 @@ func (s *Manager) PrepareExistingDisk(recover RecoveryDoer) error {
 
 // PrepareNewDisk prepares an instances state disk by formatting the disk as a LUKS device using a random passphrase.
 func (s *Manager) PrepareNewDisk() error {
-	s.log.Infof("Preparing new state disk")
+	s.log.Infof("Preparing new state disk (UUID: %s)", s.mapper.DiskUUID())
```
daniel-weisse, line 124:
```suggestion
	s.log.With(zap.String("uuid", s.mapper.DiskUUID())).Infof("Preparing new state disk")
```

disk-mapper/internal/setup/setup.go
```diff
@@ -165,6 +167,49 @@ func (s *Manager) saveConfiguration(passphrase []byte) error {
...
+		}
+		fileInfo, err := file.Info()
+		if err != nil {
+			panic(err)
+		}
+		devices = append(devices, fileInfo)
+	}
+	if err != nil {
+		panic(err)
+	}
```
daniel-weisse, line 186:
```suggestion
	dirs, err := os.ReadDir("/sys/class/block")
	if err != nil {
		return err
	}
	for _, file := range dirs {
		if file.IsDir() {
			continue
		}
		fileInfo, err := file.Info()
		if err != nil {
			return err
		}
		devices = append(devices, fileInfo)
	}
```
Error should be handled before checking the return value.
Also, this function should not panic, but return errors to the caller

disk-mapper/internal/setup/setup.go
```diff
@@ -165,6 +167,49 @@ func (s *Manager) saveConfiguration(passphrase []byte) error {
...
+
+	s.log.Infof("List of all available block devices and partitions:")
+	for _, device := range devices {
+		var stat syscall.Statfs_t
+		dev := "/dev/" + device.Name()
+		if err := syscall.Statfs(dev, &stat); err != nil {
+			s.log.Errorf("failed to statfs %s: %v", dev, err)
+		}
```
daniel-weisse, line 194:
```suggestion
		if err := syscall.Statfs(dev, &stat); err != nil {
			s.log.With(zap.Error(err)).Errorf("failed to statfs %s", dev )
		}
```
Is it OK to continue if statfs failed? Or should we continue with the next device?

image/Makefile
```diff
@@ -11,6 +11,7 @@ IMAGE_VERSION                    ?= v0.0.0
 DEBUG                            ?= false
 AUTOLOGIN                        ?= false
 AUTOLOGIN_ARGS                   := $(if $(filter true,$(AUTOLOGIN)),--autologin) # set "--autologin" if AUTOLOGIN is true
+KERNEL_DEBUG_CMDLNE				 := constellation.debug
 export INSTALL_DEBUGD			 ?= $(DEBUG)
```
daniel-weisse, line 15:
```suggestion
AUTOLOGIN                        ?= false
AUTOLOGIN_ARGS                   := $(if $(filter true,$(AUTOLOGIN)),--autologin) # set "--autologin" if AUTOLOGIN is true
KERNEL_DEBUG_CMDLNE              := constellation.debug
export INSTALL_DEBUGD            ?= $(DEBUG)
```
Whitespaces are off for these two. Tabs instead of spaces?

daniel-weisse (changes requested):

disk-mapper/internal/setup/setup.go
leongross: Since there are several reasons why this could fail (invalid permissions, signal interrupt, I/O error, ...) I think we should not fail but instead continue with the next device.

image/Makefile
leongross: Yes, this was not intentional.

disk-mapper/cmd/main.go
```diff
@@ -134,7 +134,10 @@ func main() {
 		vtpm.OpenVTPM,
 	)
 
-	setupManger.LogDevices()
+	err = setupManger.LogDevices()
+	if err != nil {
```
daniel-weisse, line 138:
```suggestion
	if err := setupManger.LogDevices(); err != nil {
```

daniel-weisse (approved):

image/Makefile
```diff
@@ -44,6 +45,7 @@ mkosi.output.%/fedora~37/image.raw: mkosi.files/mkosi.%.conf inject-bins inject-
 		$(AUTOLOGIN_ARGS) \
 		--environment=INSTALL_DEBUGD \
 		--environment=CONSOLE_MOTD \
+		--kernel-command-line="$(KERNEL_DEBUG_CMDLNE)=$(DEBUG)" \
```
malt3, line 48: If i understand correctly, this always sets `constellation.debug`, even for release images?

image/mkosi.skeleton/usr/lib/dracut/modules.d/39constellation-mount/prepare-state-disk.sh
```diff
@@ -6,9 +6,26 @@
 set -euo pipefail
 shopt -s inherit_errexit
 
+# hacky parsing of the command line arguments. check if argv[1] is --debug
```
malt3, line 9:
```suggestion
# parsing of the command line arguments. check if argv[1] is --debug
```

:wink:

image/mkosi.skeleton/usr/lib/dracut/modules.d/39constellation-mount/prepare-state-disk.sh
```diff
@@ -6,9 +6,26 @@
...
+verbosity=0
+if [[ $# -gt 0 ]]; then
+  if [[ $1 == "--debug" ]]; then
+    verbosity=-1
+    echo "[Constellation] Debug mode enabled"Q
```
malt3, line 14:
```suggestion
    echo "[Constellation] Debug mode enabled"
```

image/Makefile
leongross: atm this is the case, `KERNEL_DEBUG_CMDLNE` is either `true` or `false`. I did this because I thought that would make the parsing easier. If it is false, no debug output will be enabled.

Do you think we should do this differently?

malt3: I think it would pollute the cmdline less if we could leave it out by default and only set it if debugging should be enabled.
I think we have examples in this makefile on how to implement optional args:
```
AUTOLOGIN_ARGS := $(if $(filter true,$(AUTOLOGIN)),--autologin) # set "--autologin" if AUTOLOGIN is true
```

malt3 (approved):
LGTM 🚀

PR #1160, opened 2023-02-09
cli: refactor `upgrade execute` to `upgrade apply`

derpsteb:
### Proposed change(s)

- Replace the existing `upgrade execute` command with an extended command called `upgrade apply`. The new command complements the `upgrade check` command by applying the configured upgrades from the config. It extends the existing functionality by enabling k8s and helm upgrades. The upgrades are controlled via the config values `image`, `microserviceVersion` and `kubernetesVersion`.
- `upgrade apply` checks that the configured version C is a valid upgrade to the installed version I. That means skip cases where `C==I`, `C<I` or `C.Minor>I.Minor+1`.
-  `upgrade apply` also checks that there is currently no k8s or image upgrade in progress. If an upgrade is in progress the upgrade is skipped and a message is printed.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/upgradeapply.go
+++ b/cli/internal/cmd/upgradeapply.go
@@ -22,13 +22,13 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func newUpgradeExecuteCmd() *cobra.Command {
+func newUpgradeApplyCmd() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "execute",
-		Short: "Execute an upgrade of a Constellation cluster",
-		Long:  "Execute an upgrade of a Constellation cluster by applying the chosen configuration.",
+		Short: "Apply an upgrade to a Constellation cluster",
+		Long:  "Apply an upgrade to a Constellation cluster by applying the chosen configuration.",
 		Args:  cobra.NoArgs,
-		RunE:  runUpgradeExecute,
+		RunE:  runUpgradeApply,
 	}
 
 	cmd.Flags().Bool("helm", false, "execute helm upgrade\n"+
@@ -47,7 +47,7 @@ func newUpgradeExecuteCmd() *cobra.Command {
 	return cmd
 }
 
-func runUpgradeExecute(cmd *cobra.Command, args []string) error {
+func runUpgradeApply(cmd *cobra.Command, args []string) error {
 	log, err := newCLILogger(cmd)
 	if err != nil {
 		return fmt.Errorf("creating logger: %w", err)
@@ -61,11 +61,11 @@ func runUpgradeExecute(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	return upgradeExecute(cmd, imageFetcher, upgrader, fileHandler)
+	return upgradeApply(cmd, imageFetcher, upgrader, fileHandler)
 }
 
-func upgradeExecute(cmd *cobra.Command, imageFetcher imageFetcher, upgrader cloudUpgrader, fileHandler file.Handler) error {
-	flags, err := parseUpgradeExecuteFlags(cmd)
+func upgradeApply(cmd *cobra.Command, imageFetcher imageFetcher, upgrader cloudUpgrader, fileHandler file.Handler) error {
+	flags, err := parseUpgradeApplyFlags(cmd)
 	if err != nil {
 		return fmt.Errorf("parsing flags: %w", err)
 	}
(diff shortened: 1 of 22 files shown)


derpsteb:
This currently depends on having access to the [NodeVersion](https://github.com/edgelesssys/constellation/blob/main/operators/constellation-node-operator/api/v1alpha1/nodeversion_types.go#L59) type from the operators inside the CLI to to `unstructured.Unstructured <-> NodeVersion` conversions. These are desirable as it prevents us from parsing the `map[string]any` from `unstructured.Unstructured`. 

However, we can't depend on the types without adding a replace statement to the go.mod file and messing with the build Dockerfile. Since I feel like it makes sense to have the API in `internal/` and referencing it from there inside the operator I will try to move the code from `operators/constellation-node-operator/api/v1alpha1` to `constellation/internal/api`. I will have to figure out how not to break the kubebuilder code generation. But feel free to take a look at the existing code as it should not be affected by this change.

daniel-weisse (commented):
I'm unsure about using the NodeVersion struct from the operator directly since it causes a lot of issues with out go.mod file.
The data type is not that complex, the only issue I could see is parsing the status and conditions fields properly. Do you know if we can access them even in the unstructured object?

derpsteb:
They are accessible and can be parsed. I originally built it that way. But that parsing logic will break once we modify the Operator API. With the conversion we will at least get compilation errors. I would invest a little bit more time to see if there is a good way to do this. If not I will go back to the Unstructured based implementation.

cli/internal/cmd/upgradeapply.go
```diff
@@ -0,0 +1,199 @@
...
+	imageFetcher := image.New()
+	upgrader, err := cloudcmd.NewUpgrader(cmd.OutOrStdout(), log)
+	if err != nil {
+		return err
+	}
+
+	return upgradeApply(cmd, imageFetcher, upgrader, fileHandler)
+}
+
+func upgradeApply(cmd *cobra.Command, imageFetcher imageFetcher, upgrader cloudUpgrader, fileHandler file.Handler) error {
```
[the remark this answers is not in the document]
derpsteb: Done. Helm logging is already very verbose and includes all the information you are asking for.

cli/internal/cmd/upgradeapply.go
```diff
@@ -0,0 +1,205 @@
...
+	case err != nil:
+		return fmt.Errorf("upgrading Kubernetes components: %w", err)
+	}
+
+	err = u.handleImageUpgrade(cmd.Context(), conf, imageFetcher)
+	switch {
+	case errors.Is(err, cloudcmd.ErrInProgress):
+		cmd.PrintErrln("Skipping image upgrades. Another image upgrade is in progress")
+	case errors.As(err, &invalidUpgradeErr):
+		cmd.PrintErrf("Skipping image upgrades. %s", err)
```
daniel-weisse, line 101:
```suggestion
		cmd.PrintErrf("Skipping image upgrades: %s"\n, err)
```

cli/internal/cmd/upgradeapply.go
```diff
@@ -0,0 +1,205 @@
...
+		return fmt.Errorf("service upgrade: %w", err)
+	}
+
+	invalidUpgradeErr := &cloudcmd.InvalidUpgradeError{}
+	err = u.handleK8sUpgrade(cmd.Context(), conf)
+	switch {
+	case errors.Is(err, cloudcmd.ErrInProgress):
+		cmd.PrintErrln("Skipping Kubernetes components upgrades. Another Kubernetes components upgrade is in progress")
+	case errors.As(err, &invalidUpgradeErr):
+		cmd.PrintErrf("Skipping Kubernetes components upgrades. %s", err)
```
daniel-weisse, line 91:
```suggestion
		cmd.PrintErrf("Skipping Kubernetes components upgrades: %s\n", err)
```

daniel-weisse (approved):
Was able to upgrade my cluster using `constellation upgrade apply` from `v2.5.0` to `ref/main/stream/nightly/v2.6.0-pre.0.20230210173318-77bd537fb412`, while also upgrading Kubernetes version from 1.24 to 1.25.

I think it was introduced in a different PR, but I feel like the current UX of upgrading Kubernetes versions is a bit unwieldy.
Constellation does not allow a user to choose what patch version of Kubernetes to install, we always ship the latest for a given release. Yet a user needs to specify it when upgrading from one minor release to another (or choosing a different version than default when creating a cluster).
This is especially annoying since the only way of checking what patch versions are available for a given CLI is the `upgrade check` command (which doesn't list versions older than the current).

If we choose to stay with the current way of specifying Kubernetes versions in the config, we should at least provide a way to easily check what patch versions are supported by a given CLI

daniel-weisse:
As this PR is quite large, I would like at least one other reviewer to take a look over the changes before merging

derpsteb:
Agree with the point about the patch version. katexochen also raised it. The discussion on how to solve this is ongoing. Current ideas that are in line with what you suggested are:
- an interactive component during `config generate` that lists the versions
- a subcommand for `constellation config` to list versions

malt3 (approved):
I don't have any additional code review comments. Tested this locally and it works as expected.

PR #1161, opened 2023-02-09
ci: fix yq command syntax

datosh:

Apparently quoting breaks something, I was not able to quickly figure this out so I expanded the variables instead.

failed: https://github.com/edgelesssys/constellation/actions/runs/4135811513/jobs/7149764114#step:7:610
reproduced: https://github.com/edgelesssys/constellation/actions/runs/4136536714
fixed: https://github.com/edgelesssys/constellation/actions/runs/4136561357

### DIFF
--- a/.github/workflows/build-os-image.yml
+++ b/.github/workflows/build-os-image.yml
@@ -583,9 +583,6 @@ jobs:
 
       - name: Add static PCRs
         run: |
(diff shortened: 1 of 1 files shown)


leongross (approved):
lgtm, tested it and it seems to work :+1:

malt3 (approved):

PR #1163, opened 2023-02-10
bootstrapper: fix loglevel in helm install retry

derpsteb:
### Proposed change(s)

- One needs to set a different log verbosity to see logs logged with Debugf as the bootstrapper runs with default loglevel, which is info (?).

### Checklist

- [x] Link to Milestone


bootstrapper/internal/helm/helm.go
```diff
@@ -222,7 +222,7 @@ type installDoer struct {
 
 // Do logs which chart is installed and tries to install it.
 func (i installDoer) Do(ctx context.Context) error {
-	i.log.Debugf("trying helm install for chart %s", i.chart.Name())
+	i.log.Infof("trying helm install for chart %s", i.chart.Name())
```
daniel-weisse, line 225:
```suggestion
	i.log.With(zap.String("chart", i.chart.Name())).Infof("Trying to install helm chart")
```

daniel-weisse (approved):

PR #1170, opened 2023-02-12
cli: adapt "upgrade check" reference to conventions

thomasten:
Adapt the command reference to our conventions and also simplify/clarify.

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -47,7 +47,7 @@ func NewRootCmd() *cobra.Command {
 	must(rootCmd.MarkPersistentFlagFilename("config", "yaml"))
 
 	rootCmd.PersistentFlags().Bool("debug", false, "enable debug logging")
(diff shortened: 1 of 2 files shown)


cli/cmd/root.go
```diff
@@ -47,7 +47,7 @@ func NewRootCmd() *cobra.Command {
 	must(rootCmd.MarkPersistentFlagFilename("config", "yaml"))
 
 	rootCmd.PersistentFlags().Bool("debug", false, "enable debug logging")
-	rootCmd.PersistentFlags().Bool("force", false, "disables version validation errors - might result in corrupted clusters")
+	rootCmd.PersistentFlags().Bool("force", false, "disable version compatibility checks - might result in corrupted clusters")
```
thomasten, line 50: The global --force flag might get a broader meaning in the future. Would this be okay? Otherwise, we should rename it now.

derpsteb (approved):
Thanks for the changes. Wasn't aware of the convention. Looked it up now.

PR #1171, opened 2023-02-12
ci: update vale action

thomasten:
Old version of the action doesn't work anymore. Use new version of the action with older version of vale. (Newer versions of vale introduce false positives.)


katexochen:
Successful run: https://github.com/edgelesssys/constellation/actions/runs/4164955842

katexochen (approved):

PR #1172, opened 2023-02-13
deps: vendor node-maintenance-operator api

malt3:
### Proposed change(s)

- Vendor node-maintenance-operator api folder

### Additional info
- Removes some unneeded dependencies and unblocks #766

### Checklist


malt3:
@thomasten can you look at the vendored code license? Also: Ideas on how we can handle the license check? (Maybe we can ignore the 3rdparty folder?)
 https://github.com/edgelesssys/constellation/commit/80785fce9ab8788f06cbfa3373d443d29aeac067

thomasten:
I believe doing it like this is sufficient to comply with the license. At least if you haven't modified any of the copied files.
Ignoring the 3rdparty folder in the license header check seems the right solution to me.

3rdparty/node-maintenance-operator/api/v1beta1/download.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -exuo pipefail
+shopt -s inherit_errexit
+
+# this script is used to download the CRD API definition from the node-maintenance-operator.
+
+ARCHIVE_URL=$1
+PREFIX=$2
+EXPECTED_SHA256=$3
```
katexochen, line 10: nit: use lower case variables for unexported values.

3rdparty/node-maintenance-operator/go.mod
```diff
@@ -0,0 +1,25 @@
+module github.com/edgelesssys/constellation/v2/operators/constellation-node-operator/v2/3rdparty/node-maintenance-operator
```
katexochen, line 1:
```suggestion
module github.com/edgelesssys/constellation/v2/3rdparty/node-maintenance-operator
```
I would use the real path for the module name, everything else is confusing.

katexochen (changes requested):

katexochen (approved):
lgtm :hatched_chick:

PR #1175, opened 2023-02-13
rfc: specify cli version api

msanft:
### Proposed change(s)

- Add specification for the CLI version compatibility API
  - The API should be exposed via the [version API](https://github.com/edgelesssys/constellation/tree/main/internal/versionsapi) and provide information on compatibility between a specific CLI version of a Constellation version and Kubernetes versions.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/rfc/cli-api.md
+++ b/rfc/cli-api.md
@@ -4,17 +4,17 @@ The CLI API provides information about the compatibility of the Constellation CL
 
 ## CLI API Endpoints
 
-The build pipeline produces lookup tables for compatibility information that are uploaded to S3 and can be accessed via HTTP.
-The lookup tables are organized in a directory structure that allows to look up the compatibility for a given Constellation version.
+The build pipeline produces artifacts for compatibility information that are uploaded to S3 and can be accessed via HTTP.
+The artifacts are organized in a directory structure that allows to look up the compatibility for a given Constellation version.
 
 The following HTTP endpoints are available:
 
 - `GET /constellation/v1/ref/<REF>/stream/<STREAM>/<VERSION>/cli/`
-  - [`info.json` returns the CLI compatibility lookup table for the given Constellation version.](#cli-lookup-table)
+  - [`info.json` returns the CLI compatibility information artifact for the given Constellation version.](#cli-lookup-table)
 
-## CLI lookup table
+## CLI information artifact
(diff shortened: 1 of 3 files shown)


rfc/cli-api.md
```diff
@@ -0,0 +1,33 @@
...
+
+The build pipeline produces lookup tables for compatibility information that are uploaded to S3 and can be accessed via HTTP.
+The lookup tables are organized in a directory structure that allows to look up the compatibility for a given Constellation version.
+
+The following HTTP endpoints are available:
+
+- `GET /constellation/v1/ref/<REF>/stream/<STREAM>/<VERSION>/cli/`
+  - [`info.json` returns the CLI compatibility lookup table for the given Constellation version.](#cli-lookup-table)
+
+## CLI lookup table
```
katexochen, line 15: The term "lookup table" is somewhat ambiguous in our API specs and daily usage (I know the image-api rfc still uses it). I would prefer something like "info artifact" to differentiate this from the lookup table design we use in the (tighter) versions API.

rfc/cli-api.md
```diff
@@ -0,0 +1,33 @@
...
+```
+/constellation/v1/ref/<REF>/stream/<STREAM>/<VERSION>/cli/info.json
+```
+
+```json
+{
+  "version": "<VERSION>",
+  "ref": "<REF>",
+  "stream": "<STREAM>",
+  "cli": "v1.2.3",
```
katexochen, line 28: how does the "cli" field differ from the "version" field?

msanft: Do we always build a corresponding CLI version to a Constellation release?

katexochen: This object only exists if a CLI exists. You know that there is a CLI with this versions as you used the versions API to request a CLI version (cli.json) and got the path that consists of ref/stream/version. So you know that this CLI exists. The existing information about existence is stored in the versions API.

rfc/cli-api.md
derpsteb: another idea: "cli info".

derpsteb:
Seconding katexochen's comments, lgtm apart from that.

katexochen (approved):
lgtm

PR #1186, opened 2023-02-14
bazel: add build files for go

malt3:
### Proposed change(s)

-
-

### Checklist

- [ ] Link to Milestone


.bazelrc
```diff
@@ -1,16 +1,39 @@
-build --incompatible_enable_cc_toolchain_resolution
-build --action_env BAZEL_DO_NOT_DETECT_CPP_TOOLCHAIN=1
+# share bazel cache between checkouts of the same project
+# and keep old build caches around for longer
+common --disk_cache=~/.cache/shared_bazel_action_cache
+common --repository_cache=~/.cache/shared_bazel_repository_cache
+
+# better caching / reproducibility
+common --incompatible_strict_action_env=true
```
Nirusu, line 7: Not sure if you know (it's a draft after all), but this line seems to break running pretty much anything except build? Or at least `bazel query` or really just `bazel version` is broken.

malt3: Thanks for the report. I'll have a look tomorrow.
Wondering if this is a macOS thing but it may be an oversight from my side.

malt3: You are absolutely right! Fixing this.

datosh:
Tested this in my local environment ( Ubuntu 22.04.1 LTS ) with Bazelisk v1.16.0.

Full build runs in ~12min and wrote ~35GB to disk. We should probably update the `build-test-run.md` to reflect the disk requirements. 

Full test run was super quick and ran without any problems.

👍

daniel-weisse (commented):
Build worked on my machine.

daniel-weisse (approved):

.bazelrc
```diff
@@ -1,16 +1,39 @@
...
+build --stamp --workspace_status_command=tools/workspace_status
+
+# compile for linux_amd64 by default (this is the target for any binaries that go into the cluster)
+build --platforms @zig_sdk//libc_aware/platform:linux_amd64_gnu.2.34
+
+# enable tpm simulator for tests
+test --//bazel/settings:tpm_simulator
+
+# disable test caching (rerun all test cases even if they passed before)
+test --cache_test_results=no
```
Nirusu, line 29: Why disable caching for tests globally?

cli/internal/cmd/version.go
```diff
@@ -58,18 +60,22 @@ func parseBuildInfo(info *debug.BuildInfo) (commit, state, date, goVersion, comp
 			os = value
 		case "vcs.time":
 			date = value
-		case "vcs.modified":
-			if value == "true" {
-				state = "dirty"
-			} else {
-				state = "clean"
-			}
```
Nirusu, line 66: Why remove the modified flag? If we have custom stamping tools for Bazel, can we not reimplement this part easily for the Bazel stamping?

bazel/go/go_test.bzl
```diff
@@ -0,0 +1,105 @@
...
+This module contains rules and macros for building and testing Go code.
+"""
+
+load("@io_bazel_rules_go//go:def.bzl", _go_test = "go_test")
+
+def go_test(ld = None, **kwargs):
+    """go_test is a wrapper for go_test that uses default settings for Constellation.
+
+    It adds the following:
+    - Sets test count to 10.
```
Nirusu, line 11:
```suggestion
    - Sets test count to 3.
```

bazel/go/go_test.bzl
```diff
@@ -0,0 +1,105 @@
...
+    - Sets race detector to on by default.
+    - Optionally sets the interpreter path to ld.
+
+    Args:
+          ld: path to interpreter to that will be written into the elf header.
+          **kwargs: all other arguments are passed to go_test.
+    """
+
+    # Sets test count to 10.
```
Nirusu, line 20:
```suggestion
    # Sets test count to 3.
```

dev-docs/workflows/build-test-run.md
```diff
@@ -76,6 +99,43 @@ Therefore, if you want to verify a cluster deployed with a release image you wil
 
 After collecting the measurements you can put them into your `constellation-conf.yaml` under the `measurements` key in order to enforce them.
 
+# Dependency management
+
+## Go
+
+Go dependencies are managed with [Go modules](https://blog.golang.org/using-go-modules) that are all linked from the main [go.work](/go.work) file.
+[Follow the go documentation](https://go.dev/doc/modules/managing-dependencies) on how to use go modules.
```
Nirusu, line 107:
```suggestion
[Follow the go documentation](https://go.dev/doc/modules/managing-dependencies) on how to use Go modules.
```

dev-docs/workflows/build-test-run.md
```diff
@@ -76,6 +99,43 @@ Therefore, if you want to verify a cluster deployed with a release image you wil
...
+After updating a dependency, you will have to run `bazel run //:gazelle-update-repos` to update the Bazel workspace with changes from `go.mod` and `go.sum` files.
+
+# Bazel
+
+Bazel is the primary build system for this project. It is used to build all go code and will be used to build all artifacts in the future.
```
Nirusu, line 112:
```suggestion
Bazel is the primary build system for this project. It is used to build all Go code and will be used to build all artifacts in the future.
```

dev-docs/workflows/build-test-run.md
```diff
@@ -76,6 +99,43 @@ Therefore, if you want to verify a cluster deployed with a release image you wil
...
+Still, we aim to keep the codebase compatible with `go build` and `go test` as well.
+Whenever go code is changed, you will have to run `bazel run //:gazelle` to regenerate the Bazel build files for Go code.
```
Nirusu, line 114:
```suggestion
Whenever Go code is changed, you will have to run `bazel run //:gazelle` to regenerate the Bazel build files for Go code.
```

rpm/rpms.bzl
```diff
@@ -19,31 +19,53 @@ def rpms():
...
         sha256 = "23cda2a639c358757ee35ce6270ba3d0c6cd779309ff528e7c08c0239737dffb",
         urls = [
-            "https://mirror.netzwerge.de/fedora/linux/development/37/Everything/x86_64/os/Packages/a/audit-libs-3.0.9-1.fc37.x86_64.rpm",
-            "https://ftp.fau.de/fedora/linux/development/37/Everything/x86_64/os/Packages/a/audit-libs-3.0.9-1.fc37.x86_64.rpm",
-            "https://mirror.23m.com/fedora/linux/development/37/Everything/x86_64/os/Packages/a/audit-libs-3.0.9-1.fc37.x86_64.rpm",
-            "https://ftp.halifax.rwth-aachen.de/fedora/linux/development/37/Everything/x86_64/os/Packages/a/audit-libs-3.0.9-1.fc37.x86_64.rpm",
+            "https://ftp.plusline.net/fedora/linux/releases/37/Everything/x86_64/os/Packages/a/audit-libs-3.0.9-1.fc37.x86_64.rpm",
+            "https://mirror.23m.com/fedora/linux/releases/37/Everything/x86_64/os/Packages/a/audit-libs-3.0.9-1.fc37.x86_64.rpm",
+            "https://ftp.fau.de/fedora/linux/releases/37/Everything/x86_64/os/Packages/a/audit-libs-3.0.9-1.fc37.x86_64.rpm",
+            "https://mirror.netzwerge.de/fedora/linux/releases/37/Everything/x86_64/os/Packages/a/audit-libs-3.0.9-1.fc37.x86_64.rpm",
```
Nirusu, line 25: Guess this becomes obsolete when we host the RPMs and become the sole mirror.
But in case we are not the sole mirror, is it possible to stabilize the ordering so that we don't end up with change of orderings being registered as changes? Makes it hard to review actual changes otherwise in the future.

cli/internal/helm/BUILD.bazel
```diff
@@ -0,0 +1,373 @@
...
+        "charts/cert-manager/templates/webhook-psp.yaml",
+        "charts/cert-manager/templates/webhook-rbac.yaml",
+        "charts/cert-manager/templates/webhook-service.yaml",
+        "charts/cert-manager/templates/webhook-serviceaccount.yaml",
+        "charts/cert-manager/templates/webhook-validating-webhook.yaml",
+        "charts/cert-manager/values.yaml",
+        "charts/cilium/.helmignore",
+        "charts/cilium/Chart.yaml",
+        "charts/cilium/LICENSE",
+        "charts/cilium/README.md",
```
Nirusu, line 53: This very likely was already the case before, but we embed some redundant files which are not user-visible or used by our code.

3rdparty/bazel/org_openssl/openssl_test.cc
```diff
@@ -0,0 +1,53 @@
...
+{
+    unsigned char md[SHA256_DIGEST_LENGTH] = {};
+
+    assertm(simpleSHA256(static_cast<const void*>(MESSAGE.data()), MESSAGE.size(), md), "Failed to generate hash");
+    std::string hash = fromByteArray(md, SHA256_DIGEST_LENGTH);
+
+    assertm(hash == MESSAGE_HASH, "Unexpected message hash");
+
+    return 0;
+}
```
Nirusu, line 53: Do we actually use this test somewhere or is this a copy-paste artifact?

Nirusu (commented):
Cannot fully build all targets right now on macOS since some RPM packages are down. So here's a quick review just looking at files.

Some other thoughts:
1. Is it possible to get attributions for the stuff in "3rdparty/"? This looks like copy-paste and has LICENSE files, but it would be cool to know where this was taken from in the first place (especially if this requires maintenance later with updates).
2. Is it possible for Gazelle and other auto-generated files to be labeled as such, e.g. with a "DO NOT EDIT" header?

cli/internal/helm/BUILD.bazel
malt3: Yeah this was already the case and should not be solved here

3rdparty/bazel/org_openssl/openssl_test.cc
malt3: [This test is executed with the unit tests and ensures the openssl build actually works](https://github.com/edgelesssys/constellation/blob/feat/build/bazel-go/3rdparty/bazel/org_openssl/BUILD.bazel#L11)

rpm/rpms.bzl
malt3: I think only using our archive or enforcing a stricter ordering are both good ideas. Will not be implemented here.

.bazelrc
malt3: This is the current behaviour for our go unit tests and it only reruns the tests (it doesn't force recompilation, only re-execution). Without this flag, running `bazel test ...` twice will only run the tests the first time. Not sure what is better developer UX in this case.

cli/internal/cmd/version.go
malt3: I felt like it only caused issues in our CI when non-ignored files landed in the workdir during a build which marked the release CLI as dirty. I can add it back if you feel like it provides value.

malt3:
> Is it possible to get attributions for the stuff in "3rdparty/"? This looks like copy-paste and has LICENSE files, but it would be cool to know where this was taken from in the first place (especially if this requires maintenance later with updates).

The openssl build files are taken from here: https://github.com/bazelbuild/rules_foreign_cc/tree/main/examples/third_party/openssl

I'll add a URL to the upstream version. I also asked for the correct way to do the attribution [here](https://github.com/bazelbuild/rules_foreign_cc/issues/1007). Correct format for attribution is still TBD.

Other build files in 3rdparty are written by me.

malt3:
> Is it possible for Gazelle and other auto-generated files to be labeled as such, e.g. with a "DO NOT EDIT" header?

It does not appear to be possible right now. It's also questionable how useful this is in practice:

1. The files are auto generated but some have manual overrides (marked with `keep`) so adding a `DO NOT EDIT` at the top is rather confusing.
2. Apart from external contributors, the workflow for BUILD files should be clear to everyone
3.CI will still throw an error if BUILD files are not up to date (or are manually edited without a `keep` directive)

malt3:
> Go does not fully support reproducibility when cross-compiling, but did we test reproducibility across different Linux distributions on the same platform? Especially to rule out unexpected host dependencies?

I performed manual testing for this. Automatic reproducibility testing should probably be added to our CI (can be done inside docker).

cli/internal/cmd/version.go
Nirusu: >when non-ignored files landed in the workdir during a build 

Which shouldn't happen, given that this already resulted in funny bugs (e.g. wrong SBOMs). 
Thus yes, I'd say it provides value :)

.bazelrc
Nirusu: Mh, actually it seems to cache the build steps anyway (cannot test fully since tests seem to be broken on macOS). If it does and doesn't build everything from scratch then I guess that's fine.

Nirusu:
>Apart from external contributors, the workflow for BUILD files should be clear to everyone

Well, that's mostly the point. It currently is intransparent which files are auto-generated and which are manually written just by looking at the files. The added docs are also only very brief about this:

>Whenever Go code is changed, you will have to run `bazel run //:gazelle` to regenerate the Bazel build files for Go code.

That might be easy to miss. Having a header would make this more clear, especially as an external contributor, that most of these files are auto generated.

Of course if `#keep` is set by manually modifying the file, I see the point that "DO NOT EDIT" doesn't make sense.

malt3:
> > Apart from external contributors, the workflow for BUILD files should be clear to everyone
> 
> Well, that's mostly the point. It currently is intransparent which files are auto-generated and which are manually written just by looking at the files. The added docs are also only very brief about this:
> 
> > Whenever Go code is changed, you will have to run `bazel run //:gazelle` to regenerate the Bazel build files for Go code.
> 
> That might be easy to miss. Having a header would make this more clear, especially as an external contributor, that most of these files are auto generated.
> 
> Of course if `#keep` is set by manually modifying the file, I see the point that "DO NOT EDIT" doesn't make sense.

I'll enhance the docs to better explain this

PR #1187, opened 2023-02-14
cli: load helm values during upgrade

derpsteb:
### Proposed change(s)

- This fixes multiple bugs where upgrades did not change resources in the cluster as expected. These bugs stem from the fact that the existing upgrade code did not load values locally and therefore did not render the templates correctly in some cases.
- Solve by letting go code load helm values also during upgrade. While we still want to reuse values from the cluster we need to set others. Like the images. This change therefore loads all values during an upgrade that do not depend on user input. If a value depends on user input it will not be set. Because the value is not set, the existing value from the cluster is reused.
- [Azure, 1.25, 1+2](https://github.com/edgelesssys/constellation/actions/runs/4181302080). Check that init still works as expected.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/helm/loader.go
+++ b/cli/internal/helm/loader.go
@@ -52,6 +52,7 @@ const (
 
 // ChartLoader loads embedded helm charts.
 type ChartLoader struct {
+	csp                          cloudprovider.Provider
 	joinServiceImage             string
 	keyServiceImage              string
 	ccmImage                     string
@@ -78,6 +79,7 @@ func NewLoader(csp cloudprovider.Provider, k8sVersion versions.ValidK8sVersion) 
 	}
 
 	return &ChartLoader{
+		csp:                          csp,
 		joinServiceImage:             versions.JoinImage,
 		keyServiceImage:              versions.KeyServiceImage,
 		ccmImage:                     ccmImage,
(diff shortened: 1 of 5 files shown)


cli/internal/helm/loader.go
```diff
@@ -151,9 +159,9 @@ func (i *ChartLoader) loadCilium(csp cloudprovider.Provider, conformanceMode boo
 
 // loadCiliumHelper is used to separate the marshalling step from the loading step.
```
daniel-weisse, line 160:
```suggestion
// loadCiliumValues is used to separate the marshalling step from the loading step.
```

cli/internal/helm/loader.go
```diff
@@ -480,6 +472,47 @@ func (i *ChartLoader) loadConstellationServicesValues(config *config.Config, mas
 	return values, nil
 }
 
+// extendConstellationServicesValues extends the given values map by some values depending on user input.
+// This extra step of separating the application of user input is necessary since service upgrades should
+// reuse user input from the init step. However, we can't rely on reuse-values, because
+// during upgrades we all values need to be set locally as they might have changed.
```
daniel-weisse, line 478:
```suggestion
// during upgrades all values need to be set locally as they might have changed.
```

daniel-weisse (approved):

PR #1207, opened 2023-02-16
cli: option to disable spinner via environment variable

datosh:
### Proposed change(s)

- Option to disable spinner via `CONSTELL_NO_SPINNER` environment variable. Main motivation is to automatically skip waiting times in asciinema recordings 
- Fix `constellation terminate` spinner to honor `debug` flag and new `CONSTELL_NO_SPINNER` environment variable

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/terminate.go
+++ b/cli/internal/cmd/terminate.go
@@ -37,7 +37,10 @@ func NewTerminateCmd() *cobra.Command {
 // runTerminate runs the terminate command.
 func runTerminate(cmd *cobra.Command, args []string) error {
 	fileHandler := file.NewHandler(afero.NewOsFs())
-	spinner := newSpinner(cmd.ErrOrStderr())
+	spinner, err := newSpinnerOrStderr(cmd)
+	if err != nil {
+		return fmt.Errorf("creating spinner: %w", err)
+	}
 	defer spinner.Stop()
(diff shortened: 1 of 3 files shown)


cli/internal/cmd/terminate.go
```diff
@@ -37,7 +37,10 @@ func NewTerminateCmd() *cobra.Command {
 // runTerminate runs the terminate command.
 func runTerminate(cmd *cobra.Command, args []string) error {
 	fileHandler := file.NewHandler(afero.NewOsFs())
-	spinner := newSpinner(cmd.ErrOrStderr())
+	spinner, err := newSpinnerOrStderr(cmd)
```
katexochen, line 40: Should we remove the newSpinner funciton so this cannot happen again? The code could just be inlined into the newSpinnerOrStderr function.

daniel-weisse (approved):

cli/internal/cmd/terminate.go
datosh: I checked this. `newSpinner` is used quite a few times in `spinner_test` as well. Inlining this everywhere is probably not the right call. 

The cleaner solution IMO would be to have spinner in its own package and export only required functions.

datosh: I will refactor this in another PR.

PR #1154, opened 2023-02-08
docs: embedd asciinema casts

datosh:
Signed-off-by: Fabian Kammel <fk@edgeless.systems>

### Proposed change(s)

- Move Dockerfile and scripts to generate casts from https://github.com/edgelesssys/screenrecordings to monorepo so we can generate everything from start to finish
- Use asciinema-player to directly embedd asciinema casts into our docs (without rendering them as mp4).
  - Use `termtosvg` to convert a `.cast` file to `svg` to display in GitHub README.md, since it does not support JavaScript.
- Updated workflows for SBOM & CLI verification
- Added new workflows for create config, delete IAM, create cluster, destroy cluster.
- Documented process and how to style elements of screen casts.

### Open Questions
- ~~With this solution, are mp4 renderings of the screencast still necessary? For example to upload them to YouTube?~~
  - No, we will record sessions with OBS if we intend to record webinar style sessions.
- ~~Should the user run `generate-screencasts.sh` or should the pipeline do this? I lean to the former, since the output needs to be inspected and docs preview is currently broken.~~
  - It is fine to make the user run this. 

### Additional info
- Please try to generate screencasts from your machine and let me know how it goes. Everything is in containers / shell scripts, but it would be great to confirm it works on another system as well!
- Currently we still have the spinner in our screencasts. Since the screencasts simulate what a user would do (download an actual released Constellation CLI) we need to re-run screencast generation once v2.6.0 is released.

### Checklist

- [x] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)
- [x] Add labels (e.g., for changelog category)
- [x] Link to Milestone

PR #1212, opened 2023-02-17
docs: fix terminate docs filename

msanft:
### Proposed change(s)

- Change the required files for cluster termination from `terraform.tfstate` to the Terraform state directory defined in `constants.go`

### Related issue
- As mentioned in #1154 

### Checklist

- [x] Link to Milestone

### DIFF
--- a/docs/docs/workflows/terminate.md
+++ b/docs/docs/workflows/terminate.md
@@ -1,6 +1,6 @@
 # Terminate your cluster
 
-You can terminate your cluster using the CLI. For this, you need the Terraform state file named `terraform.tfstate` in the current directory.
+You can terminate your cluster using the CLI. For this, you need the Terraform state directory named `constellation-terraform` in the current directory.
 
 :::danger
 


datosh (commented):
I reckon the change is correct. 

I was mostly confused about why we explicitly mention this here. All CLI commands need to be executed in the Constellation workspace. Maybe this is better to be merged with the Terraform documentation we have? We can mention that Constellation CLI can be used to terminate from workdir, but in order to execute Terraform one needs to `cd` into one of the two Terraform workspaces..? 

Maybe @3u13r can provide some insight.

msanft:
> I reckon the change is correct.
> 
> I was mostly confused about why we explicitly mention this here. All CLI commands need to be executed in the Constellation workspace. Maybe this is better to be merged with the Terraform documentation we have? We can mention that Constellation CLI can be used to terminate from workdir, but in order to execute Terraform one needs to `cd` into one of the two Terraform workspaces..?
> 
> Maybe @3u13r can provide some insight.

Agreed. I think this is rather a question on which (maybe obvious) details we explain how granular. I will bring this up in the stand-up

thomasten:
merging this as it is already an improvement

thomasten (approved):

PR #1217, opened 2023-02-17
docs: extend docs on Terraform iam resource destruction

msanft:
### Proposed change(s)

- Extend the docs on Terraform IAM resource destruction; Before, they didn't mention that removing the state directory is necessary to be able to create a new cluster in the same wd

### Checklist

- [x] Link to Milestone


datosh (approved):
That was quick. LGTM!

PR #1219, opened 2023-02-19
Fix links and clean lycheeignore

thomasten:
* Cilium moved pages in new docs version
* sigstore blog post no longer exists
* Some link check exceptions aren't needed anymore
* Link to our gitlab is correctly reported as broken. Someone should fix this ;)


Nirusu (approved):
> Link to our gitlab is correctly reported as broken. Someone should fix this ;)

On it, but it’s no easy operation ;)

PR #1222, opened 2023-02-20
config: allow k8s version MAJOR.MINOR  for v2.6

derpsteb:
### Proposed change(s)

- To adhere to our compatibility goal of not breaking old configs within one version, the kubernetes patch version is automatically extended for configs in the transistional version v2.6.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -70,7 +70,7 @@ type Config struct {
 	StateDiskSizeGB int `yaml:"stateDiskSizeGB" validate:"min=0"`
 	// description: |
 	//   Kubernetes version to be installed into the cluster.
-	KubernetesVersion string `yaml:"kubernetesVersion" validate:"supported_k8s_version"`
+	KubernetesVersion string `yaml:"kubernetesVersion" validate:"required,supported_k8s_version"`
 	// description: |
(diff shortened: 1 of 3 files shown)


daniel-weisse (approved):
Is there any hard reason why we wouldn't also want the option to only specify major.minor version in future releases?
I think this option is quite nice for users who just want to say "Hey give me the latest release of Kubernetes 1.25" without having to worry about setting the correct patch version.

datosh (approved):

derpsteb:
We could do that if we find a consistent UX for `upgrade apply`. The main problem I see: how to handle situations where a CLI offers patch and minor version upgrades? Right now the user sees both and can chose. But how do you (a) deny a patch version update and (b) prioritize a minor version upgrade over a patch version update?

We could introduce a confirmation dialogue. If a patch update is available, `upgrade apply` will tell the user about the new patch version and ask for confirmation to apply the update. Minor version upgrades will only be controllable via the config.

If a full version is specified in the config, the same interaction would happen, but `upgrade apply` automatically writes the new patch version to the config. 

To be honest - I like this approach better than specifying the patch version. However, if we were to go for this I would remove the need to specify the patch version completely.

PR #1224, opened 2023-02-20
cli: add `config kubernetes-versions` subcommand

derpsteb:
### Proposed change(s)

- Allows users to learn which k8s versions are supported by the current CLI. This can be helpful when creating new cluster, applying upgrades or understanding the output of `upgrade check`.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/config.go
+++ b/cli/internal/cmd/config.go
@@ -23,6 +23,7 @@ func NewConfigCmd() *cobra.Command {
 	cmd.AddCommand(newConfigGenerateCmd())
 	cmd.AddCommand(newConfigFetchMeasurementsCmd())
 	cmd.AddCommand(newConfigInstanceTypesCmd())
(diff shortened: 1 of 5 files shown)


datosh (commented):
Code change LGTM.

Two things came to mind:

1. Where do you think this would come up in the user's journey? Should we add a hint for this command in the respective section in the docs?
2. This seems in line with the location of `config instance-types`. To make it consistent with other commands: `config generate`, `iam create`, the consistent way would probably be `config list instance-types|kubernetes-versions`. What do you think?

malt3 (approved):

derpsteb:
Re 1: added docs comment
Re 2: I agree that your suggestion seems more consistent. However, I would like to wait a bit longer with refactoring this until we see the need for one or two more list commands. Or multiple people complaining about it being very counter intuitive.

PR #1232, opened 2023-02-21
bazel: update Perl for macOS & Linux arm64

Nirusu:
(This PR merges into the Bazel branch, not main)
### Proposed change(s)
- Update Perl for macOS & Linux arm64

Compared to yesterday's use of release version, I bumped it to the latest main so it also has support for Linux arm64 building (which I haven't tried yet at all - I trust them on this).

Still the remaining issues seem to be OpenSSL complaining about wrong ARM architecture and the Go race detector complaining about missing tsan definitions since we cross-compile tests. We can also handle this as a "add macOS / Linux arm64" support PR if you like a larger PR.

### DIFF
--- a/3rdparty/bazel/org_openssl/openssl_repositories.bzl
+++ b/3rdparty/bazel/org_openssl/openssl_repositories.bzl
@@ -31,9 +31,9 @@ def openssl_repositories():
     maybe(
         http_archive,
         name = "rules_perl",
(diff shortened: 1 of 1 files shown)


malt3:
Thanks for looking into this. I would keep this open while we are investigating and merge once we feel ready for bazel or solved the remaining issues.

Nirusu:
@malt3 Maybe after all it makes sense to merge this earlier. Otherwise I'll constantly have to rebase this when you rebase and the change is very small.

malt3 (approved):
Yes - lets merge this and add additional fixes as needed

PR #1245, opened 2023-02-21
helm: retry on connection refused

3u13r:
### Proposed change(s)

- Let the helm installer retry on "connection refused" errors

Problem: 
`constellation init` fails with:
```
Initializing cluster   
Cluster initialization failed. This error is not recoverable.
Terminate your cluster and try again.
Error: init call: rpc error: code = Internal desc = initializing cluster: installing pod network: helm install: Kubernetes cluster unreachable: Get "https://X.X.X.X:6443/version": dial tcp X.X.X.X:6443: connect: connection refused
```

Detailed description:

What is interesting that after kubeadm init we have an explicit kubewaiter, which waits until it can list all K8s namespaces using the kubeclient and further kubeclient requests such as annotating the current node. But the execution of the first helm command fails. 

The kubeclient automatically retries on almost all error cases. The helm client just on timeout errors (which the error we see is not, it is a "connection refused" error). 
During all those requests mentioned above all targets (i.e. nodes) are unhealthy. This means that the AWS LB "fails open", routing to all targets. 
This becomes most apparent on AWS since the switch from unhealthy to healthy takes at least 20 seconds. On GCP and Azure it should take 2-5. 

### Checklist

- [x] Link to Milestone

### DIFF
--- a/bootstrapper/internal/helm/helm.go
+++ b/bootstrapper/internal/helm/helm.go
@@ -15,6 +15,7 @@ import (
 	"net"
 	"os/exec"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/edgelesssys/constellation/v2/bootstrapper/internal/kubernetes/k8sapi"
@@ -188,6 +189,11 @@ func (h *Client) installCiliumGCP(ctx context.Context, kubectl k8sapi.Client, re
 // The function will wait 30 seconds before retrying a failed installation attempt.
 // After 5 minutes the retrier will be canceld and the function returns with an error.
 func (h *Client) install(ctx context.Context, chartRaw []byte, values map[string]any) error {
+	retriable := func(err error) bool {
+		return errors.Is(err, wait.ErrWaitTimeout) ||
+			strings.Contains(err.Error(), "connection refused")
+	}
+
 	reader := bytes.NewReader(chartRaw)
 	chart, err := loader.LoadArchive(reader)
 	if err != nil {
@@ -200,7 +206,7 @@ func (h *Client) install(ctx context.Context, chartRaw []byte, values map[string
 		values,
 		h.log,
 	}
-	retrier := retry.NewIntervalRetrier(doer, 30*time.Second, isTimeoutErr)
+	retrier := retry.NewIntervalRetrier(doer, 30*time.Second, retriable)
 
 	// Since we have no precise retry condition we want to stop retrying after 5 minutes.
 	// The helm library only reports a timeout error in the error cases we currently know.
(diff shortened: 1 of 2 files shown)


bootstrapper/internal/helm/helm.go
```diff
@@ -188,6 +189,16 @@ func (h *Client) installCiliumGCP(ctx context.Context, kubectl k8sapi.Client, re
...
 // After 5 minutes the retrier will be canceld and the function returns with an error.
 func (h *Client) install(ctx context.Context, chartRaw []byte, values map[string]any) error {
+	retriable := func(err error) bool {
+		if errors.Is(err, wait.ErrWaitTimeout) {
+			return true
+		} else if strings.Contains(err.Error(), "connection refused") {
+			return true
+		} else {
+			return false
+		}
```
derpsteb, line 199:
```suggestion
		switch {
		case errors.Is(err, wait.ErrWaitTimeout):
			return true
		case strings.Contains(err.Error(), "connection refused"):
			return true
		default:
			return false
		}
```
I like the switch syntax a lot :D. But do as you like.

derpsteb (approved):

katexochen:
> :green_circle: [AWS, 1.25, 2+3](https://github.com/edgelesssys/constellation/actions/runs/4234338625)

Doesn't seem to be related to this Pr, but https://github.com/edgelesssys/constellation/actions/runs/4234338625/jobs/7356490507#step:9:3468

katexochen (approved):
lgtm

bootstrapper/internal/kubernetes/kubewaiter/kubewaiter.go
```diff
@@ -33,7 +34,7 @@ func (w *CloudKubeAPIWaiter) Wait(ctx context.Context, kubernetesClient Kubernet
 	doer := &kubeDoer{kubeClient: kubernetesClient}
 	retrier := retry.NewIntervalRetrier(doer, 5*time.Second, funcAlwaysRetriable)
 	if err := retrier.Do(ctx); err != nil {
-		return doer.Do(context.Background())
+		return fmt.Errorf("failed to wait for Kubernetes API: %w", err)
```
thomasten, line 37:
```suggestion
		return fmt.Errorf("waiting for Kubernetes API: %w", err)
```

thomasten (approved):

bootstrapper/internal/helm/helm.go
thomasten: I'd even shorten this to
```go
	return errors.Is(err, wait.ErrWaitTimeout) ||
		strings.Contains(err.Error(), "connection refused")
```
but opinions on readability may differ

PR #1247, opened 2023-02-21
iam: assign uami role to base resource group

3u13r:
### Proposed change(s)

- Assign role at the base resource group and not the iam resource group

This bug was introduced in https://github.com/edgelesssys/constellation/commit/b8648261e3a89c69c46ec313fda6bfd16db1e415

### Checklist

- [x] Link to Milestone


thomasten (approved):
works

3u13r:
now the autoscaler also works

katexochen (approved):

PR #1255, opened 2023-02-22
Revert "deps: update Terraform azurerm to v3.44.1"

derpsteb:
Revert edgelesssys/constellation#1197

There seems to be an issue with 3.44.1: https://github.com/hashicorp/terraform-provider-azurerm/issues/20546
Fix is already on their main branch, but no release ready yet. 

Found because the mini constellation e2e pipeline failed: https://github.com/edgelesssys/constellation/actions/runs/4240682413/jobs/7370885385

### DIFF
--- a/.github/actions/e2e_mini/.terraform.lock.hcl
+++ b/.github/actions/e2e_mini/.terraform.lock.hcl
@@ -2,32 +2,32 @@
 # Manual edits may be lost in future updates.
 
 provider "registry.terraform.io/hashicorp/azurerm" {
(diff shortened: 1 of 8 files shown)


katexochen:
Why does this only happen in the minicon test?

Nirusu:
Other tests set ARM_CLIENT_ID etc. in their environment for authentication. The minicon e2e test doesn't seem to do that, but instead uses Azure OIDC.

katexochen (approved):

PR #1256, opened 2023-02-22
internal: add custom version type

msanft:
### Proposed change(s)

- Add a custom version type as proposed in [AB#2907](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/2907/)
- This should be a superset of the functionality in `internal/compatibility`, featuring a better API to other components and a more concise syntax. It should also make the use of version strings obsolete
- TODO: Implement the new versioning in versionsapi / CLI after feedback for the coarse architecture has been gathered

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -4,7 +4,10 @@ Copyright (c) Edgeless Systems GmbH
 SPDX-License-Identifier: AGPL-3.0-only
 */
 
-package versions
+/*
+Package version provides functionality to parse and process semantic versions, as they are used in multiple components of Constellation.
+*/
+package version
 
 import (
 	"encoding/json"
--- a/internal/version/version_test.go
+++ b/internal/version/version_test.go
@@ -4,7 +4,7 @@ Copyright (c) Edgeless Systems GmbH
 SPDX-License-Identifier: AGPL-3.0-only
 */
 
-package versions
+package version
 
 import (
 	"testing"
--- a/internal/versions/version.go
+++ b/internal/versions/version.go
@@ -1,0 +1,57 @@
+/*
+Copyright (c) Edgeless Systems GmbH
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+package versions
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"golang.org/x/mod/semver"
+)
+
+// Version represents a semantic version.
+type Version string
+
+// NewVersion returns a Version from a string.
+func NewVersion(version string) (Version, error) {
+	if !semver.IsValid(version) {
+		return Version(""), fmt.Errorf("invalid semver: %s", version)
+	}
+
+	return Version(version), nil
+}
+
+// String returns the string representation of the Version.
+func (v Version) String() string {
+	return string(v)
+}
+
+// Compare compares two versions. It relies on the semver.Compare function internally.
+func (v Version) Compare(other Version) int {
+	return semver.Compare(v.String(), other.String())
+}
+
+// MarshalJSON implements the json.Marshaler interface.
+func (v Version) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf(`"%s"`, v.String())), nil
+}
+
+// UnmarshalJSON implements the json.Unmarshaler interface.
+func (v *Version) UnmarshalJSON(data []byte) error {
+	var s string
+	if err := json.Unmarshal(data, &s); err != nil {
+		return err
+	}
+
+	version, err := NewVersion(s)
+	if err != nil {
+		return err
+	}
+
+	*v = version
+	return nil
+}
--- a/internal/versions/version.go
+++ b/internal/versions/version.go
@@ -58,26 +58,31 @@ func (v Version) Compare(other Version) int {
 	return semver.Compare(v.String(), other.String())
 }
 
-// CanUpgradeTo returns if a version can be upgraded to another version.
-// It checks if the version of other is greater than the current version and allows a drift of at most one minor version.
-func (v Version) CanUpgradeTo(other Version) bool {
-	return v.Compare(other) < 0 && other.Minor-v.Minor <= 1
+// IsUpgradeTo returns if a version is an upgrade to another version.
+// It checks if the version of v is greater than the version of other and allows a drift of at most one minor version.
+func (v Version) IsUpgradeTo(other Version) bool {
+	return v.Compare(other) > 0 && v.Major == other.Major && v.Minor-other.Minor <= 1
 }
 
-// CanUpgradeToBinary returns if a version can upgrade to the version of the current built binary.
-// It checks if the version of the binary is greater than the current version and allows a drift of at most one minor version.
-func (v Version) CanUpgradeToBinary() bool {
+// CompatibleWithBinary returns if a version is compatible version of the current built binary.
+// It checks if the version of the binary is equal or greater than the current version and allows a drift of at most one minor version.
+func (v Version) CompatibleWithBinary() bool {
 	binaryVersion, err := NewVersion(constants.VersionInfo)
 	if err != nil {
 		return false
 	}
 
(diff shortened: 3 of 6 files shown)


internal/versions/version.go
```diff
@@ -0,0 +1,57 @@
+/*
+Copyright (c) Edgeless Systems GmbH
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+package versions
```
katexochen, line 7: Please add a package comment and add yourself as code owner.

internal/versions/version_test.go
```diff
@@ -0,0 +1,247 @@
...
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+			require := require.New(t)
+
+			v1, err := NewVersion(tc.version1)
+			require.NoError(err)
+
+			v2, err := NewVersion(tc.version2)
+			if tc.wantErr {
```
derpsteb, line 207: I think it would be easier to read if there were a separate `TestNewVersion` and this test only checked for valid upgrades.

internal/versions/version_test.go
```diff
@@ -0,0 +1,247 @@
...
+			version1: "v1.18.0",
+			version2: "v1.18.1",
+			want:     true,
+		},
+		"minor less then": {
+			version1: "v1.18.0",
+			version2: "v1.19.0",
+			want:     true,
+		},
+		"minor too big drift": {
```
derpsteb, line 176: The description does not match the test case.

internal/versions/version_test.go
```diff
@@ -0,0 +1,247 @@
...
+			version1: "v1.18.0",
+			version2: "v1.19.1",
+			want:     true,
+		},
+		"major too big drift": {
+			version1: "v1.18.0",
+			version2: "v2.18.1",
+			want:     true,
```
derpsteb, line 184: These versions should not upgrade to each other.

internal/versions/version_test.go
```diff
@@ -0,0 +1,247 @@
...
+		},
+		"greater than": {
+			version1: "v1.18.1",
+			version2: "v1.18.0",
+			want:     false,
```
derpsteb, line 189: This would be a downgrade and should not be allowed.

internal/versions/version.go
```diff
@@ -0,0 +1,102 @@
...
+
+// Compare compares two versions. It relies on the semver.Compare function internally.
+func (v Version) Compare(other Version) int {
+	return semver.Compare(v.String(), other.String())
+}
+
+// CanUpgradeTo returns if a version can be upgraded to another version.
+// It checks if the version of other is greater than the current version and allows a drift of at most one minor version.
+func (v Version) CanUpgradeTo(other Version) bool {
+	return v.Compare(other) < 0 && other.Minor-v.Minor <= 1
```
derpsteb, line 64:
```suggestion
	return v.Compare(other) < 0 && other.Minor-v.Minor <= 1 && other.Major == v.Major
```

internal/versions/version.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Compare compares two versions. It relies on the semver.Compare function internally.
+func (v Version) Compare(other Version) int {
+	return semver.Compare(v.String(), other.String())
+}
+
+// CanUpgradeTo returns if a version can be upgraded to another version.
+// It checks if the version of other is greater than the current version and allows a drift of at most one minor version.
+func (v Version) CanUpgradeTo(other Version) bool {
```
derpsteb, line 63: nit: I would prefer `IsUpgradeTo` as name, since version a either is an upgrade to version b or it is not. The act of upgrading is not done by the version object. What do you think?

derpsteb, line 63: nit: I would prefer `IsUpgradeTo` as a name. This way you could call `CanUpgradeTo` and `CanUpgradeToBinary` on the same version object. Obviously this is a purely subjective preference. Feel free to ignore.

Would need to switch the direction in which the function works if you decide for changing it.

internal/versions/version.go
```diff
@@ -0,0 +1,102 @@
...
+
+// CanUpgradeToBinary returns if a version can upgrade to the version of the current built binary.
+// It checks if the version of the binary is greater than the current version and allows a drift of at most one minor version.
+func (v Version) CanUpgradeToBinary() bool {
+	binaryVersion, err := NewVersion(constants.VersionInfo)
+	if err != nil {
+		return false
+	}
+
+	return v.CanUpgradeTo(binaryVersion)
```
derpsteb, line 75: nice.

internal/versions/version.go
```diff
@@ -0,0 +1,102 @@
...
+
+// CanUpgradeTo returns if a version can be upgraded to another version.
+// It checks if the version of other is greater than the current version and allows a drift of at most one minor version.
+func (v Version) CanUpgradeTo(other Version) bool {
+	return v.Compare(other) < 0 && other.Minor-v.Minor <= 1
+}
+
+// CanUpgradeToBinary returns if a version can upgrade to the version of the current built binary.
+// It checks if the version of the binary is greater than the current version and allows a drift of at most one minor version.
+func (v Version) CanUpgradeToBinary() bool {
```
derpsteb, line 69: Maybe something like `CompatibleWithBinary` makes the intention of the function clearer?

internal/versions/version_test.go
```diff
@@ -0,0 +1,247 @@
...
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+			require := require.New(t)
+
+			v1, err := NewVersion(tc.version1)
+			require.NoError(err)
+
+			v2, err := NewVersion(tc.version2)
+			if tc.wantErr {
```
derpsteb, line 144: Same as for `TestCanUpgrade`

derpsteb:
Looks really good to me :+1:

internal/version/version_test.go
```diff
@@ -0,0 +1,243 @@
...
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+
+			assert.Equal(tc.want, tc.version2.IsUpgradeTo(tc.version1))
+		})
+	}
+}
+
+func TestNewVersion(t *testing.T) {
```
derpsteb, line 190: Duplicate to `TestVersionValidation`?

internal/version/version.go
```diff
@@ -0,0 +1,110 @@
...
+	return v.Equals(binaryVersion) || binaryVersion.IsUpgradeTo(v)
+}
+
+// NextMinor returns the next minor version in the format "vMAJOR.MINOR".
+func (v Version) NextMinor() string {
+	return fmt.Sprintf("v%d.%d", v.Major, v.Minor+1)
+}
+
+// Equals returns if two versions are equal.
+func (v Version) Equals(other Version) bool {
```
derpsteb, line 87: Duplicate to `Compare`? Or is there another reason for implementing this that I am not seeing?

Nirusu:
> Should this really live in `internal/versions`? Seems rather unrelated to the rest that happens there. Maybe move it `internal/version` ?

Uh, then we have three packages in internal:
`version`
`versions`
`versionsapi`

Seems super confusing imo, is it possible to give these packages more distinct naming?

internal/version/version.go
```diff
@@ -0,0 +1,110 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/edgelesssys/constellation/v2/internal/constants"
+	"golang.org/x/mod/semver"
+)
+
+// Version represents a semantic version.
+type Version struct {
```
derpsteb, line 22: Rename this to `Semver`? Package could also be renamed to `semver`. Would remove the naming conflict with the type from `internal/versionsapi` and would partially address @Nirusu comment.

derpsteb (approved):

PR #1257, opened 2023-02-22
attestation: MAA fallback

daniel-weisse:
### Proposed change(s)

- Change the `EnforceIDKeyDigest` option in the Constellation config from boolean to an enum.  The following values are available (naming not final, looking for feedback/suggestions):
  - `StrictChecking`: requires the IDKeyDigest of the node to be in the list of trusted IDKeyDigests, error otherwise
  - `MAAFallback`: if a node's IDKeyDigest is not in the list of trusted digests, MAA is used to verify the SNP properties of the VM.
  - `WarnOnly`: emit a warning, but accept IDKeyDigests not in the list of trusted digests
- Enable MAA fallback in SNP attestation Validator
  - This requires some changes to the internal Validator API, mainly the `GetTPMTrustedAttestationPublicKey` and `GetInstanceInfo` functions now also accept `extraData` (`userData` hashed with `nonce`)

### Additional info
- blocked until MAA issues are resolved on our node images

### Checklist

- [ ] Link to Milestone


malt3:

internal/attestation/idkeydigest/idkeydigest.go
```diff
@@ -4,17 +4,107 @@ Copyright (c) Edgeless Systems GmbH
...
 
 	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
 )
 
+//go:generate stringer -type=EnforceIDKeyDigest
+
+// EnforceIDKeyDigest defines the behavior of the validator when the ID key digest is not found in the expected list.
+type EnforceIDKeyDigest uint32
+
+// TODO: Decide on final value naming.
```
thomasten, line 26: we should also think about renaming the key, but not sure what this means for backwards compat

internal/attestation/azure/snp/maa.go
```diff
@@ -0,0 +1,64 @@
...
+
+func newMAAClient(url string) *maaClient {
+	return &maaClient{
+		url:    url,
+		client: http.DefaultClient,
+	}
+}
+
+func (m *maaClient) createToken(ctx context.Context, data []byte) (string, error) {
+	return maa.Attest(ctx, data, m.url, m.client)
```
thomasten, line 31: We could use maa.NewParameters etc. here if we wanted to remove some code from the snp package that does the same.

internal/attestation/azure/snp/issuer.go
```diff
@@ -53,60 +53,71 @@ func GetIDKeyDigest(open vtpm.TPMOpenFunc) ([]byte, error) {
...
+	hclReport = hclReport[lenHclHeader:]
+
+	runtimeData, _, _ := bytes.Cut(hclReport[lenSnpReport+lenSnpReportRuntimeDataPadding:], []byte{0})
+
+	vcekResponse, err := i.imds.getVcek(context.TODO())
+	if err != nil {
+		return nil, fmt.Errorf("retrieving VCEK certificate from IMDS API: %w", err)
+	}
+
+	maaToken, err := i.maa.createToken(context.TODO(), userData)
```
thomasten, line 103: I think we should do this only if MAA is configured to be used

thomasten, line 103: we should fix context usage in this func

internal/attestation/azure/snp/validator.go
```diff
@@ -191,19 +205,22 @@ func validateSNPReport(
...
 	if !hasExpectedIDKeyDigest {
-		if enforceIDKeyDigest {
-			return &idKeyError{report.IDKeyDigest[:], expectedIDKeyDigests}
-		}
-		if log != nil {
-			log.Warnf("configured idkeydigests %x doesn't contain reported idkeydigest %x", expectedIDKeyDigests, report.IDKeyDigest[:])
+		switch v.enforceIDKeyDigest {
+		case idkeydigest.MAAFallback:
+			v.log.Infof("configured idkeydigests %x don't contain reported idkeydigest %x, falling back to MAA validation", v.idKeyDigests, report.IDKeyDigest[:])
+			return v.maa.validateToken(context.TODO(), maaToken, extraData)
```
thomasten, line 219: and here

internal/attestation/azure/snp/issuer.go
daniel-weisse: I don't think that is reasonably possible to do.
The issuer would need to know about our attestation policy before creating its statement.
However, we only provide this information after we successfully verified the node.

thomasten: How will it get the maa url?

daniel-weisse: Hmmm, maybe we can use the metadata API to set an MAA url?
We already use the metadata api in the issuer, if we dont find a url we assume no MAA is to be used?

internal/attestation/azure/snp/maa.go
daniel-weisse: Requires some [changes in the MAA repo](https://github.com/edgelesssys/go-azguestattestation/pull/2), or we can keep the current implementation (which just copies some unexported code)

daniel-weisse:
Did a bunch of manual testing, but there is also this e2e test: https://github.com/edgelesssys/constellation/actions/runs/4417455739
It doesnt cover actually using the MAA fallback, but should make sure the new configs are set correctly.
For manual testing, one has to set the `constellation-maa-url` tag to the url of a MAA provider manually (at least until the changes from https://github.com/edgelesssys/constellation/pull/1409 are in main)

thomasten (approved):

internal/attestation/idkeydigest/idkeydigest.go
derpsteb: Would be a breaking change from my pov. Maybe we can rename it when implementing #1436?
For that one we will need migration docs/code anyways.

internal/config/config.go
```diff
@@ -192,7 +192,7 @@ type AzureConfig struct {
 	IDKeyDigest Digests `yaml:"idKeyDigest" validate:"required_if=EnforceIdKeyDigest true,omitempty"`
 	// description: |
 	//   Enforce the specified idKeyDigest value during remote attestation.
-	EnforceIDKeyDigest *bool `yaml:"enforceIdKeyDigest" validate:"required"`
+	EnforceIDKeyDigest idkeydigest.EnforceIDKeyDigest `yaml:"enforceIdKeyDigest" validate:"required"`
```
derpsteb, line 195: So we either need to implement a custom validator that maps from bool to enum or add a docs migration section. I would vote for the custom validator.

derpsteb (changes requested):

internal/config/config.go
daniel-weisse: https://github.com/edgelesssys/constellation/blob/c2b11777c864c44f971c2fced278b4e24bce93e4/internal/attestation/idkeydigest/idkeydigest.go#L69

Already implemented

derpsteb (approved):

PR #1258, opened 2023-02-22
ci: specify URL predicate because shortnames are unreliable

Nirusu:
### Proposed change(s)
- specify CycloneDX spec via URL rather than using a shortname

See: https://github.com/sigstore/cosign/issues/2596
Apparently, Syft and Cosign have different specs for CycloneDX. Cosign wants to update them in v2, but on the current release, you end up with stuff that is _slightly_ incompatible if you use attestation with these two tools.

Granted, this happened because the verification already used the "new" URL while the attestation didn't. Not sure if this is also related to the update or not... could be. Let's just use the new version.

And hopefully I won't have to deal with this again because this in-toto stuff is so early days it makes my head hurt dealing with these minor issues over multiple PRs now.

This is truly a mess, I hate it.

### DIFF
--- a/.github/actions/container_sbom/action.yml
+++ b/.github/actions/container_sbom/action.yml
@@ -34,9 +34,9 @@ runs:
       run: |
         set -ex
         syft packages ${{ inputs.containerReference }} -o cyclonedx-json > container-image-predicate.json
-        cosign attest ${{ inputs.containerReference }} --key env://COSIGN_PRIVATE_KEY --predicate container-image-predicate.json --type cyclonedx > container-image.att.json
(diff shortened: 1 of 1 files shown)


Nirusu:
constellation-node-operator: https://github.com/edgelesssys/constellation/actions/runs/4243631952
join-service: https://github.com/edgelesssys/constellation/actions/runs/4243636890
KeyService: https://github.com/edgelesssys/constellation/actions/runs/4243638162
qemu-metadata-api: https://github.com/edgelesssys/constellation/actions/runs/4243642124
verification-service: https://github.com/edgelesssys/constellation/actions/runs/4243643748

_Now_ this should hopefully work for every image.

derpsteb (approved):

PR #1259, opened 2023-02-22
cli: handle invalid upgrades for microservice consistently

derpsteb:
### Proposed change(s)

- Handle invalid upgrade errors similarly as for images and k8s.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/helm/client.go
+++ b/cli/internal/helm/client.go
@@ -20,7 +20,6 @@ import (
 	"github.com/edgelesssys/constellation/v2/internal/versions"
 	"github.com/pkg/errors"
 	"github.com/spf13/afero"
-	"golang.org/x/mod/semver"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/cli"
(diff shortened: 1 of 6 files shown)


cli/internal/helm/client.go
```diff
@@ -166,12 +165,10 @@ func (c *Client) upgradeRelease(
...
 	c.log.Debugf("New %s version: %s", releaseName, chart.Metadata.Version)
 
-	if !isUpgrade(currentVersion, chart.Metadata.Version) {
-		c.log.Debugf(
-			"Skipping upgrade of %s: new version (%s) is not an upgrade for current version (%s)",
-			releaseName, chart.Metadata.Version, currentVersion,
-		)
-		return nil
+	// This may break for cert-manager or cilium if we decide to upgrade more than one minor version at a time.
+	// Leaving it as is since it is not clear to me what kind of sanity check we could do.
```
daniel-weisse, line 169: Is upgrading Cilium/cert-manager from vX.Y to vX.Y+2 something we are likely to do?

I think we should make sure on our side to no do anything that breaks compatibility in the services we add to Constellation.
We could either enforce this using this `IsValidUpgrade` check, or remove the check entirely for non Edgeless services (instead rely on our e2e tests and QA to make sure Cilium/cert-manager upgrades are save).

In the second case we would need to rely on first checking if the Edgeless services upgrade is valid, in which case we can execute all helm upgrades.
If its not, we don't upgrade anything

derpsteb: I rechecked the cilium and cert-manager docs. It seems pretty unlikely that we would upgrade in a different pattern that we are doing it for our own services. Hence I would leave the check as is.

PR #1263, opened 2023-02-23
docs: add docs on general Terraform usage

msanft:
### Proposed change(s)

- Our docs featured multiple references to Terraform state files, using Terraform, etc. This PR should add a general documentation page on how Terraform is being used by Constellation to manage cloud resources as stated [here](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/2935/). This includes:
- Instructions on interacting with Terraform state manually 
- Instructions on how and where to keep Terraform state files

### Checklist

- [x] Link to Milestone

### DIFF
--- a/docs/docs/architecture/orchestration.md
+++ b/docs/docs/architecture/orchestration.md
@@ -23,10 +23,10 @@ Altogether, the following files are generated during the creation of a Constella
 * a configuration file
 * an ID file
 * a Base64-encoded master secret
-* Terraform artifacts such as `terraform.tfstate`
+* Terraform artifacts such as `terraform.tfstate`, stored in the `constellation-terraform` subdirectory
 * a Kubernetes `kubeconfig` file.
 
-Constellation uses Terraform for infrastructure management. No setup of Terraform is needed. The CLI automatically fetches a copy of Terraform when required.
+Constellation uses Terraform for infrastructure management. No setup of Terraform is needed. The CLI automatically fetches a copy of Terraform when required. The policy on usage of Terraform by Constellation can be found [here](../reference/terraform.md).
 
 After the creation of your cluster, the CLI will provide you with a Kubernetes `kubeconfig` file.
 This file grants you access to your Kubernetes cluster and configures the [kubectl](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/) tool.
@@ -51,7 +51,7 @@ Post-installation the CLI provides a configuration for [accessing the cluster us
 The `kubeconfig` file provides the credentials and configuration for connecting and authenticating to the API server.
 Once configured, orchestrate the Kubernetes cluster via `kubectl`.
 
-Make sure to keep the state files such as `terraform.tfstate` in the workspace directory to be able to manage your cluster later on.
+Make sure to keep the Terraform subdirectory such as `constellation-terraform` in the workspace directory to be able to manage your cluster later on.
 Without it, you won't be able to modify or terminate your cluster.
 
 After the initialization, the CLI will present you with a couple of tokens:
--- a/docs/docs/architecture/orchestration.md
+++ b/docs/docs/architecture/orchestration.md
@@ -23,10 +23,8 @@ Altogether, the following files are generated during the creation of a Constella
 * a configuration file
 * an ID file
 * a Base64-encoded master secret
-* Terraform artifacts such as `terraform.tfstate`, stored in the `constellation-terraform` subdirectory
+* [Terraform artifacts](../reference/terraform.md), stored in the `constellation-terraform` subdirectory
 * a Kubernetes `kubeconfig` file.
-
-Constellation uses Terraform for infrastructure management. No setup of Terraform is needed. The CLI automatically fetches a copy of Terraform when required. The policy on usage of Terraform by Constellation can be found [here](../reference/terraform.md).
 
 After the creation of your cluster, the CLI will provide you with a Kubernetes `kubeconfig` file.
 This file grants you access to your Kubernetes cluster and configures the [kubectl](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/) tool.
--- a/docs/docs/architecture/orchestration.md
+++ b/docs/docs/architecture/orchestration.md
@@ -23,7 +23,7 @@ Altogether, the following files are generated during the creation of a Constella
 * a configuration file
 * an ID file
 * a Base64-encoded master secret
-* [Terraform artifacts](../reference/terraform.md), stored in the `constellation-terraform` subdirectory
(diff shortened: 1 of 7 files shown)


msanft:
I'm not yet a 100% happy with how these turned out but I can't really figure out what I don't like neither. Maybe you (@thomasten & @datosh ) can have a look over these with the briskness of a new reader and see where room for improvement could be.

On another note, when running the docs locally, the new page does not show up in the sidebar. I assume this is expected and caused by the versioned sidebars?

docs/docs/reference/terraform.md
```diff
@@ -0,0 +1,22 @@
+# Terraform Usage
```
thomasten, line 1: > the new page does not show up in the sidebar

You need to add this page to sidebars.js. If you then set the version to Next, it will show up.

docs/docs/architecture/orchestration.md
```diff
@@ -23,10 +23,10 @@ Altogether, the following files are generated during the creation of a Constella
 * a configuration file
 * an ID file
 * a Base64-encoded master secret
-* Terraform artifacts such as `terraform.tfstate`
+* Terraform artifacts such as `terraform.tfstate`, stored in the `constellation-terraform` subdirectory
```
thomasten, line 26:
```suggestion
* [Terraform artifacts](../reference/terraform.md) stored in the `constellation-terraform` subdirectory
```
If we do it like this ...

docs/docs/architecture/orchestration.md
```diff
@@ -23,10 +23,10 @@ Altogether, the following files are generated during the creation of a Constella
...
 * a Kubernetes `kubeconfig` file.
 
-Constellation uses Terraform for infrastructure management. No setup of Terraform is needed. The CLI automatically fetches a copy of Terraform when required.
+Constellation uses Terraform for infrastructure management. No setup of Terraform is needed. The CLI automatically fetches a copy of Terraform when required. The policy on usage of Terraform by Constellation can be found [here](../reference/terraform.md).
```
thomasten, line 29:
```suggestion
```
... we may be able to just drop this one.

docs/docs/architecture/orchestration.md
```diff
@@ -51,7 +51,7 @@ Post-installation the CLI provides a configuration for [accessing the cluster us
 The `kubeconfig` file provides the credentials and configuration for connecting and authenticating to the API server.
 Once configured, orchestrate the Kubernetes cluster via `kubectl`.
 
-Make sure to keep the state files such as `terraform.tfstate` in the workspace directory to be able to manage your cluster later on.
+Make sure to keep the Terraform subdirectory such as `constellation-terraform` in the workspace directory to be able to manage your cluster later on.
 Without it, you won't be able to modify or terminate your cluster.
```
thomasten, line 55:
```suggestion
```
I'm not sure if this hint is necessary at all after explaining what the workspace directory is. What do you think?

docs/docs/workflows/create.md
```diff
@@ -30,7 +34,7 @@ constellation create --control-plane-nodes 1 --worker-nodes 2
 
 For details on the flags, consult the command help via `constellation create -h`.
 
-*create* stores your cluster's state into a [`terraform.tfstate`](../architecture/orchestration.md#cluster-creation-process) file in your workspace.
+*create* stores your cluster's state in a [`constellation-terraform`](../architecture/orchestration.md#cluster-creation-process) directory in your workspace.
 
 </tabItem>
 <tabItem value="terraform" label="Terraform">
```
thomasten, line 40: maybe shorten the paragraph that begins below this line ("Constellation supports managing ...") and put the new info box there?

thomasten (commented):
What I'm not yet happy with (and maybe it's the same with you) is that we do have the reference page now, but still haven't reduced the Terraform references and duplications throughout the docs (which I believe is what you actually wanted to achieve). Below are some ideas for this.

msanft:
> What I'm not yet happy with (and maybe it's the same with you) is that we do have the reference page now, but still haven't reduced the Terraform references and duplications throughout the docs (which I believe is what you actually wanted to achieve). Below are some ideas for this.

That definitely is a good point. But I was rather unsure about how the Terraform page itself is looking like. As a user (who possibly didn't work with Terraform before), do you think it's comprehensible?

docs/docs/reference/terraform.md
```diff
@@ -0,0 +1,22 @@
...
+
+[Terraform](https://www.terraform.io/) is an open-source Infrastructure as Code (IaC) framework which is being used by multiple Constellation components to manage cloud resources. This page describes our policy on using Terraform in Constellation.
+
+:::info
+This page assumes basic familiarity with Terraform. The Terraform documentation can be found [here](https://developer.hashicorp.com/terraform/docs).
```
datosh, line 6:
```suggestion
This page assumes basic familiarity with Terraform. Refer to the [Terraform documentation](https://developer.hashicorp.com/terraform/docs) for an introduction.
```

Embedding link into more natural language.

docs/docs/architecture/orchestration.md
```diff
@@ -23,11 +23,9 @@ Altogether, the following files are generated during the creation of a Constella
 * a configuration file
 * an ID file
 * a Base64-encoded master secret
-* Terraform artifacts such as `terraform.tfstate`
+* [Terraform artifacts](../reference/terraform.md), stored in the `constellation-terraform` subdirectory
```
datosh, line 26: Should we also mention the iam subfolder?

docs/docs/workflows/config.md
```diff
@@ -231,13 +231,13 @@ The following describes the configuration fields and how you obtain the required
 
 * **iamProfileControlPlane**: The name of an IAM instance profile attached to all control-plane nodes.
 
-  Use the [provided Terraform script](https://github.com/edgelesssys/constellation/tree/release/v2.2/hack/terraform/aws/iam) to generate the necessary profile. The profile name will be provided as Terraform output value: `control_plane_instance_profile`.
+  The resource can be created with [Terraform](../reference/terraform.md). For that, use the [provided Terraform script](https://github.com/edgelesssys/constellation/tree/release/v2.2/hack/terraform/aws/iam) to generate the necessary profile. The profile name will be provided as Terraform output value: `control_plane_instance_profile`.
```
datosh, line 234: I think the reference could be confusing. While the new page introduced Terraform, it does not mention the generation of IAM instance profiles. Profile generation is also not supported (hack), keeping it seperated from official features will help to set expectations.

thomasten:
> I was rather unsure about how the Terraform page itself is looking like. As a user (who possibly didn't work with Terraform before), do you think it's comprehensible?

I think users should only use Terraform with Constellation if they have worked with Terraform before, so IMO the page only needs to target such users. Maybe change "basic familiarity" to "familiarity" in the info box.

datosh (approved):

PR #1264, opened 2023-02-23
internal: add logging to attestation issuer

daniel-weisse:
### Proposed change(s)

- Enable easier debugging by adding logging to attestation issuer


derpsteb (approved):
:green_circle: https://github.com/edgelesssys/constellation/actions/runs/4280456285

:+1:

PR #1265, opened 2023-02-23
attestation: tdx issuer/validator

daniel-weisse:
### Proposed change(s)

- Add experimental TDX Issuer/Validator logic to run Constellation on TDX platforms

### Checklist

- [ ] Link to Milestone


internal/attestation/initialize/initialize.go
```diff
@@ -0,0 +1,129 @@
...
+	ownerState := "not initialized"
+	if ownerInitialized {
+		ownerState = "initialized"
+	}
+	clusterState := "not initialized"
+	if clusterInitialized {
+		clusterState = "initialized"
+	}
+	return false, fmt.Errorf("PCRs %v and %v are not consistent: PCR[%v]=%v (%v), PCR[%v]=%v (%v)", idxOwner, idxCluster, idxOwner, pcrs[idxOwner], ownerState, idxCluster, pcrs[idxCluster], clusterState)
+	*/
```
Nirusu, line 115: I know it's old code, but will this ever happen? 😄

internal/attestation/initialize/initialize.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func tdxMarkNodeAsBootstrapped(handle tdx.Device, clusterID []byte) error {
+	return tdx.ExtendRTMR(handle, clusterID, 4)
```
Nirusu, line 119:
```suggestion
	return tdx.ExtendRTMR(handle, clusterID, 2)
```
`ExtendRTMR` has an index parameter that is unused so the 4 here does effectively nothing... But in any case that should be a 2 likely - there's no RTMR[4].

internal/attestation/measurements/measurements.go
```diff
@@ -43,6 +43,10 @@ const (
 	// The value used to extend is derived from Constellation's master key.
 	// TODO: move to stable, non-debug PCR before use.
 	PCRIndexOwnerID = tpmutil.Handle(16)
+
+	// RTMRIndexClusterID is the RTMR we extend to mark the node as initialized.
+	// The value is the index of the RTMR + 1, since index 0 of the TDX measurements is reserved for MRTD.
+	RTMRIndexClusterID = 3
```
Nirusu, line 49: The RTMR + 1 one seems like it will end up confusion later on.
Maybe I would remove RTMR here to remove and call this `TDXAttestationIndexClusterID` to disassociate that this could be an RTMR index?

internal/attestation/measurements/measurements.go
```diff
@@ -259,11 +264,11 @@ func (m *Measurement) unmarshal(eM encodedMeasurement) error {
 		}
 	}
 
-	if len(expected) != 32 {
+	if !((len(expected) == 32) || len(expected) == 48) {
```
Nirusu, line 267:
```suggestion
	if len(expected) != 32 && len(expected) != 48 {
```

internal/attestation/measurements/measurements.go
```diff
@@ -180,7 +184,8 @@ func (m *M) SetEnforced(enforced []uint32) error {
 // Measurement wraps expected PCR value and whether it is enforced.
 type Measurement struct {
 	// Expected measurement value.
-	Expected [32]byte `json:"expected" yaml:"expected"`
+	// 32 bytes for vTPM attestation, 48 for TDX.
+	Expected []byte `json:"expected" yaml:"expected"`
 	// WarnOnly if set to true, a mismatching measurement will only result in a warning.
 	WarnOnly bool `json:"warnOnly" yaml:"warnOnly"`
 }
```
Nirusu, line 191: Mh, I am a bit unhappy about switching from fixed byte size arrays back to slices.

For the config I guess it makes sense for UX. For the issue/verification logic, that's a bit meh since we know need to double-check sizes everywhere.

For the demo it's fine I guess, but for production code maybe refactoring the validators to have distinct types for TPM & TDX might be better?

internal/attestation/tdx/tdx.go
```diff
@@ -0,0 +1,67 @@
...
+// GetSelectedMeasurements returns the selected measurements from the RTMRs.
+func GetSelectedMeasurements(open OpenFunc, selection []int) (measurements.M, error) {
+	if len(selection) > 5 {
+		return nil, fmt.Errorf("invalid measurement selection: max 5 measurements allowed, got %d", len(selection))
+	}
+	for _, idx := range selection {
+		if idx < 0 || idx >= 5 {
+			return nil, fmt.Errorf("invalid measurement index %d", idx)
+		}
+	}
```
Nirusu, line 36: Is the first check really useful here? The second already sets bounds for the indices?

internal/attestation/tdx/tdx.go
```diff
@@ -0,0 +1,67 @@
...
+
+// Available returns true if the TDX device is available and can be opened.
+func Available() bool {
+	handle, err := tdx.Open()
+	if err != nil {
+		return false
+	}
+	defer handle.Close()
+	return true
+}
```
Nirusu, line 67: If we have stuff like this already here, shouldn't `IsTDXDevice` from the TDX library be moved here to keep the library more generic and not Constellation-like?

bootstrapper/cmd/bootstrapper/main.go
```diff
@@ -157,16 +160,28 @@ func main() {
...
-		if err != nil {
-			log.With(zap.Error(err)).Fatalf("Failed to get selected PCRs")
-		}
+		var measurements measurements.M
+		if tdx.Available() {
+			measurements, err = tdx.GetSelectedMeasurements(tdxapi.Open, []int{0, 1, 2, 3, 4})
+			if err != nil {
+				log.With(zap.Error(err)).Fatalf("Failed to get selected RTMRs")
+			}
+			issuer = tdx.NewIssuer(log)
```
Nirusu, line 173: Shouldn't other issuers also have logs? 
If this is solely for debugging and to be removed later, maybe marking a `TODO` might be useful to spot this in later reviews?

Nirusu (commented):
Given that this touches attestation data again and we recently had a vulnerable in there, can another reviewer more knowledgable with this code (@thomasten ?) double-check the logic here?

I am not really familiar with the code here though on first side the changes look okay to me on first side.

bootstrapper/cmd/bootstrapper/main.go
daniel-weisse: :wink: 
https://github.com/edgelesssys/constellation/pull/1264

internal/attestation/measurements/measurements.go
daniel-weisse: Since we don't do any copying into/from the measurements struct, this should be fine.
We mostly just switched to fixed sized to save on some validation logic for the config, which we now have to do regardless since we need to accept 2 different sizes

internal/attestation/initialize/initialize.go
daniel-weisse: There is a feature and rfc for this
If we get a request to implement it we might assign it some priority

internal/attestation/measurements/measurements.go
Nirusu: I don't know, I'd think I would rather copy/cast into fixed sized arrays (since then the Go runtime guarantees us the size) make the vTPM issue/verification implementation only take 32 bytes and the TDX issue/verification only 48 bytes. 

That makes both paths more clear and resilient imo. But a second opinion might be good here.

However, there's an issue I see with generally having the same measurement field: How do we ship the default measurements if they are in the same field? Make `fetch-measurements` dependent on type?

Nirusu: I don't know, personally I am more a fan of having fixed size arrays depending on what verification path we take. Makes things more explicit what is expected.

However, there's an issue I see with generally having the same measurement field: How do we ship the default measurements if they are in the same field? We currently hard-code measurements for a CSP, right? So how would we solve this?

For `fetch-measurements` I could see making this dependent on whether you want TDX or vTPM, but for hard-coding... no idea.

Dockerfile.build
```diff
@@ -18,7 +19,7 @@ COPY operators/constellation-node-operator/api/go.mod ./operators/constellation-
 COPY operators/constellation-node-operator/api/go.sum ./operators/constellation-node-operator/api/go.sum
 COPY go.mod ./
 COPY go.sum ./
-RUN go mod download all
+RUN --mount=type=secret,id=netrc,dst=/root/.netrc,required=true go mod download all
```
malt3, line 22: This is okay while we keep it away from main but we should maybe add a comment here to ensure this never gets merged by accident.
It should always be possible to build the main and release versions without requiring authentication.

joinservice/Dockerfile
```diff
@@ -10,12 +10,15 @@ RUN wget -q https://go.dev/dl/go${GO_VER}.linux-amd64.tar.gz && \
...
+ENV GOPRIVATE=github.com/edgelesssys/*
 
 # Download go dependencies
 WORKDIR /constellation/
 COPY go.mod ./
 COPY go.sum ./
-RUN go mod download all
+COPY ./operators/constellation-node-operator/api/go.mod ./operators/constellation-node-operator/api/go.mod
+COPY ./operators/constellation-node-operator/api/go.mod ./operators/constellation-node-operator/api/go.sum
+RUN --mount=type=secret,id=netrc,dst=/root/.netrc,required=true go mod download all
```
malt3, line 21: See other comment

verify/Dockerfile
```diff
@@ -10,12 +10,15 @@ RUN wget -q https://go.dev/dl/go${GO_VER}.linux-amd64.tar.gz && \
...
+ENV GOPRIVATE=github.com/edgelesssys/*
 
 # Download go dependencies
 WORKDIR /constellation/
 COPY go.mod ./
 COPY go.sum ./
-RUN go mod download all
+COPY ./operators/constellation-node-operator/api/go.mod ./operators/constellation-node-operator/api/go.mod
+COPY ./operators/constellation-node-operator/api/go.mod ./operators/constellation-node-operator/api/go.sum
+RUN --mount=type=secret,id=netrc,dst=/root/.netrc,required=true go mod download all
```
malt3, line 21: See other comment

internal/attestation/measurements/measurements.go
Nirusu: I don't know, personally I am more a fan of having fixed size arrays depending on what issuing/verification path we take. Makes things more explicit what is expected. Other opinions could be good here?

Now there's a check for len == 32 or len == 48 on unmarshal, updateValidators has a check for len == 48 (on only one element) or anything else is vTPM.

The current implementation also allows you to mix 32 byte and 48 byte measurement definitions in the config and seems to happily reach them into the validator. Though if I see correctly, the checks are always done on the 48 byte since the RTMRs are defined as 48 byte and bytes.Equal expects the same length. 

That's the point I am unhappy about - it's all a bit messy. So far I don't see anything that's really wrong with the current way, and having fixed sizes wouldn't fix all potential errors... but it would reduce the need for checks and also enforce more thinking about potential sizes & inputs in the code.

**But** before that discussion, there's an issue I see with generally having the same measurement field: How do we ship the default measurements if they are in the same field? We currently hard-code measurements for a CSP, right? So how would we solve this when there's now vTPM _or_ TDX measurements in the same place?

For `fetch-measurements` I could see making this dependent on whether you want TDX or vTPM(+SNP), but for hard-coding... no idea.

Nirusu: I don't know, personally I am more a fan of having fixed size arrays depending on what issuing/verification path we take. Makes things more explicit what is expected. 

One small "issue" (not really, but not super desirable imo) right now is that there's a check for len == 32 or len == 48 on unmarshal, updateValidators has a check for len == 48 (on only one element) or anything else is vTPM. The current implementation thus allows you to mix 32 byte and 48 byte measurement definitions in the config and seems to happily reach them into the validator. Of course that doesn't pass bytes.Equal (and RTMRs are fixed 48 bytes anyway) since both values are not equal. It doesn't break security, but it seems unexpected to me to get to the point of comparing expected vTPMs measurements vs TDX measurements. 

I don't really like this. On the other hand, so far I don't see anything that's really wrong with the current way, and having fixed sizes wouldn't fix all potential errors. It might help having to consider this a bit more during development or have more clear distinct code/data paths, though.

It also doesn't help in terms of complexity that there's the config, the measurements, and now _eight_ different validator types. The integration is _complicated_ to read through in my opinion, and they mainly operate on the same type measurements.M which can now have different measurement types. Seems to me like a point to make mistakes in the long run, and that's security code. So I would like to avoid this as best as we can.

I don't know. There's potentially also the discussion about what is idiomatic Go. No idea. Opinions from other people here are _greatly_ appreciated.

**But** before that discussion, there's an issue I see with generally having the same measurement field in terms of UX: How do we ship the default measurements if they are in the same field? We currently hard-code measurements for a CSP, right? So how would we solve this when there's now vTPM _or_ TDX measurements in the same place?

For `fetch-measurements` I could see making this dependent on whether you want TDX or vTPM(+SNP), but for hard-coding... no idea.

internal/attestation/tdx/tdx.go
daniel-weisse: I would think the endgoal is merging our TDX code into the main Constellation repo, which should make it unnecessary to keep it separated

internal/attestation/measurements/measurements.go
daniel-weisse: Consistent measurement length should be enforced by my latest commit. 

>So how would we solve this when there's now vTPM or TDX measurements in the same place?

I don't fully understand your issue here.
Assuming TDX is supported on a CSP we already have support for, we would most likely adjust config generation accordingly to allow choosing what runtime encryption is supported.
This should not make any issues with different Measurement lengths.
Default measurements would have to be dynamically injected into the config anyways, regardless if they are the same length or different

Nirusu: I still think the vTPM / TDX validators themselves should expect different types / size inputs, seems a bit easier to enforce equal lengths this way rather than having the (un)marshaller deal with this.

> would have to be dynamically injected into the config anyways,

The anyways depends on how we implement this ;)

>This should not make any issues with different Measurement lengths.

Not saying it's impossible, just thinking about what the UX here would be given that we can very likely assume that this will depend on the machine type. 

But yeah, thinking about this we I guess this would be a mix of `constellation config create` and having to run `constellation config fetch-measurements` if you change the CVM type from e.g. vTPM to TDX. It's an extra step a user would have to take, but I guess that's better than having separate fields.

internal/attestation/tdx/tdx.go
Nirusu: Okay, if that's the plan then we can handle this later.

internal/attestation/measurements/measurements.go
Nirusu: >Consistent measurement length should be enforced by my latest commit.

Thanks. But I still lean into having the validators enforce different sizes/types and let the (un)marshaller return different types for vTPM or TDX measurements if there are still multiple places in the code for these size checks. 

On the other hand, I am not sure if two different types here would be idiomatic Go.

Again, maybe a second opinion here would be good? 

>I don't fully understand your issue here.

Not saying it's impossible, but UX-wise we likely would have an extra parameter here for `constellation config generate`, which is fine. And I guess if you e.g. change the machine type later, you would then have to run `constellation config fetch-measurements` manually.

I just remembered we once wanted to get away from the `constellation config fetch-measurements` workflow. So that's not 100% ideal, but I think that's better than having two different fields in the config. So it should be fine.

daniel-weisse: >Not saying it's impossible, but UX-wise we likely would have an extra parameter here

This would be required regardless of measurement length, since we expect different measurements for TDX and SEV.

>if there are still multiple places in the code for these size checks.

There should be none as of this moment.
Since measurements are only used to compare against a returned value by the Issuer using `bytes.Equal`, if someone were to hardcode incorrectly sized measurements this would only result in that comparison returning false.

IMO not having duplicate types here makes the code a lot cleaner.
But I am also open to other opinions.

Nirusu: >This would be required regardless of measurement length, since we expect different measurements for TDX and SEV.

If they use the same field. If they were, let's say, separate you could auto-fill both.  But that would introduce other UX challenges since though there's now one step less, you end up having another useless field if you don't want to use TDX. On the other side, one theoretically could use vTPM + TDX this way. 
But yeah, it's fine as it is for now to use the same field on the config. We can discuss UX later when this is closer for main.

>There should be none as of this moment.

The general "select TDX or vTPM" logic still uses size checks. 

Using a different type here would move more logic to compile time rather than runtime, which is slight plus in my opinion.

But I also agree that at least for the TDX validator mismatched sizes should not be able to pass `bytes.Equal` due to the one side being fixed size anyways. Cannot speak for vTPM/SNP since I don't know how these work exactly.

>IMO not having duplicate types here makes the code a lot cleaner.

Could be, but generally I would say we should implement what's better or more secure.
My general opinion is that mixed type definitions are not desirable for security code.
On the other hand, I don't see something that currently should break anything and refactoring this could introduce some challenges. I agree with this.

Back on the other hand though, one could argue that if refactoring this is too difficult, there's too many consumers / usages on the same data type that ends up in the validator and that might also be not really desirable since it's too convoluted security code. I personally already think the validator logic could need a slight refactor even without TDX since it's too complex and nested, but yeah, that's maybe just my opinion. 

Maybe someone else should decide this or give input on this, especially since I really cannot comment on the vTPM(+SNP) side of things here.

daniel-weisse: Issuer selection will in the future be based on https://github.com/edgelesssys/constellation/pull/1323
Validator selection will most likely be handled by a specific config field. The current setup is just to get TDX running without having to change too much existing code.

daniel-weisse: IMO implementing two different explicit types for measurements is really not worth it when the only difference is of the types is their slice length. Especially when their only usage in code is comparison with `bytes.Equal` which works on slices, not fixes size arrays.

Nirusu: I still think it's better for compile-time enforcement of what goes into the validator / comes out of the issuer rather than run time (and if you mess up, you panic and hopefully catch this earlier, otherwise you have to check lengths anyway), but yeah, I am just repeating myself at this point. Also again I cannot really speak for the vTPM part since I am not super familiar with the exact validation process there and what could possibly go wrong (or not).

Someone else can/should give their input and we can just go with that to not block this into all eternity.

malt3: My input: As long as the selection of the issuer/validator is correct, this change is not problematic.

Nirusu: Okay. 2:1. Marking this as resolved finally.

internal/attestation/tdx/tdx.go
Nirusu: Since we have the discussion to keep it a separate library, I might reopen this since the code in the library is pretty Constellation specific.

Nirusu (approved):
Let's get that finally merged 😄 Should be fine from my side now.
Other reviews can be done later before merging this to main.

malt3 (approved):

PR #1268, opened 2023-02-24
bootstrapper: stop join-client earlier

daniel-weisse:
### Proposed change(s)

This PR fixes a rare issue where the join-client of the bootstrapper of the initial control-plane node of a new cluster would try to join on itself:
1. User runs `constellation init`
2. The init-server of a node receives the call and initiates the bootstrapping of Kubernetes
3. The join-client continues to run and keeps requesting the metadata API for available join-services
4. Kubernetes starts running, the init-server waits for all services to be fully installed and available
5. The metadata API reports the control-plane node as having a join-service available
6. The join-client of this control-plane tries to contact the joins-service (which runs on this node) to join the cluster (which it is already a part of)
7. The init-server returns and stops the join-client

This is fixed by instead of deferring the call to `Clean()` in the init-server, we call the function after we acquired the node lock (same procedure as in the join-client).


daniel-weisse:
Run on GCP: https://github.com/edgelesssys/constellation/actions/runs/4261254946/jobs/7415355875

3u13r (approved):
LGTM

katexochen (approved):

PR #1281, opened 2023-02-27
cli: use pseudoversion and forward it into helm charts

derpsteb:
### Proposed change(s)

- Cmake produces a new pseudoversion whenever `cmake` is run. This version is used to set `PROJECT_VERSION`, unless `PROJECT_VERSION` is set to a version other than `0.0.0`. The release pipeline already overwrites `PROJECT_VERSION`, thus this change does not interfere with the pipeline. Another commit is created when merging changes from a release branch to main that resets `PROJECT_VERSION` to `0.0.0`.
- Use the CLI's current version to set the helm chart versions. This allows us to test upgrades to an arbitrary development build more easily. And we don't have to update chart versions manually. Since we currently don't publish the charts standalone it does not make sense to version them independently from the CLI imo.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -378,6 +378,12 @@ jobs:
           git config --global user.email "release[bot]@users.noreply.github.com"
           git fetch
           git checkout -b "${NEW_BRANCH}"
+
+      - name: Update CMakeLists.txt
+        run: |
+          sed -i "s/project(constellation LANGUAGES C VERSION [0-9]\+\.[0-9]\+\.[0-9]\+)/project(constellation LANGUAGES C VERSION 0.0.0)/" CMakeLists.txt
+          git add CMakeLists.txt
+          git commit -m "deps: set PROJECT_VERSION to prerelease"
           git push --set-upstream origin "${NEW_BRANCH}"
 
       - name: Create PR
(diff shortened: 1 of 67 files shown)


hack/pseudo-version/pseudo-version.go
```diff
@@ -81,14 +81,14 @@ func main() {
...
-		fmt.Println(headTime.Format("20060102150405"))
+		fmt.Print(headTime.Format("20060102150405"))
 	case *printBranch:
-		fmt.Println(parsedBranch)
+		fmt.Print(parsedBranch)
 	case *printReleaseVersion:
-		fmt.Println(strings.TrimPrefix(rawBranch, "release/"))
+		fmt.Print(strings.TrimPrefix(rawBranch, "release/"))
 	default:
-		fmt.Println(version)
```
katexochen, line 92: What's the reason for this change? I don't think it is a good idea to do this in general.

derpsteb: The output is used directly written into the variable and used e.g. [here](https://github.com/edgelesssys/constellation/pull/1281/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR22). With the trailing newline that would break. It feel good to change that `Println` but wasn't quite sure how to handle it differently.

katexochen (approved):

PR #1285, opened 2023-02-27
cli: allow existing config for IAM creation without --generate-config

Nirusu:
### Proposed change(s)

- Allow constellation-conf.yaml to exist when --generate-config is not set
- Fix the check itself: the stat check checks against the _constant_, but the config path is a possibly dynamic flag

Technically Go error strings are also kinda inconvenient to give actionable advice (due to capitalitation rules and the rule that they are not supposed to end with punctation), but let's not change this as part of this PR since there are also other parts in the code where we already do this.

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/cli/internal/cmd/iamcreate.go
+++ b/cli/internal/cmd/iamcreate.go
@@ -198,7 +198,7 @@ func (c *iamCreator) create(ctx context.Context) error {
 	}
 	c.log.Debugf("Using flags: %+v", flags)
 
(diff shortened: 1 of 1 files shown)


Nirusu:
Rebased this onto the latest changes.

Generally would appreciate a review on this, given that I think this should be merged as part of v2.6.0 since I personally would consider the current behavior a bug.

msanft (approved):
LGTM. Current behavior is (was) definitely a bug

PR #1308, opened 2023-02-28
docs: mention resource provider registration for Azure

Nirusu:
### Proposed change(s)
- Mention required resource providers in an Azure subscription 
- Mention `ARM_SKIP_PROVIDER_REGISTRATION` in case an error occurs

The Azure Terraform provider has the annoying behavior by default that it tries to register _all_ possible resource providers it supports (not just the ones actually used/required!). If this does not work, which can be the case for a user with limited permissions, Terraform fails and so does `constellation create`.

Fourtunely, Terraform allows us to skip this via `skip-provider-registration = true` or via the environment variable `ARM_SKIP_PROVIDER_REGISTRATION=true`. 

However, now Terraform will not try to register any resource provider, leading to errors such as `API version '2019-01-01' was not found for 'Microsoft.Foo'`  in a fairly fresh subscription. As an alternative, you can manually register the resource providers using [`azurerm_resource_provider_registration`](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/resource_provider_registration), but this brings the following downsides:

1. The current user may not have permissions to register any resource providers, causing an error.
2. In case a resource provider is already registered, you will get an error. As far as I know, you cannot let single resources fail in Terraform and just continue with other resources. And given these are common resources and subscription global, this would happen very often.
3. By manually registering the resource provider, they become part of your Terraform state. This means that `constellation terminate` would now try to deregister the providers, which are usually subscription global. This will very likely fail.

This means solving this issue in code is annoying, since we would need to integrate either extensive error handling or handle the provider registration out-of-band from Terraform.

For now, I would suggest simply adding the required providers + what to do in error case to the docs. I haven't added instructions on how to register the resource providers, since I didn't exactly know where to put them. But it's fairly trivial anyway (e.g. `az provider register -n "Microsoft.Compute"`) and we also didn't do that for .

 Here's my proposal. I am not 100% happy with the large troubleshooting block, but as far as I know we don't have any other troubleshooting section (yet?), so not sure where else to put it.

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/docs/docs/getting-started/first-steps.md
+++ b/docs/docs/getting-started/first-steps.md
@@ -95,6 +95,27 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
     Your Constellation cluster was created successfully.
     ```
 
+   :::info
+
+   On Azure, you may receive the following error when running with limited IAM permissions:
+   ```shell-session
+   Error: Error ensuring Resource Providers are registered.
+
+   Terraform automatically attempts to register the Resource Providers it supports to
+   ensure it's able to provision resources.
+
+   If you don't have permission to register Resource Providers you may wish to use the
+   "skip_provider_registration" flag in the Provider block to disable this functionality.
+   ```
+
+   To continue, please ensure that the [required resource providers](install.md#required-permissions) have been registered in your subscription by your administrator.
+   Afterward, run `constellation create` again with `ARM_SKIP_PROVIDER_REGISTRATION` set as an environment variable:
+   ```bash
+   ARM_SKIP_PROVIDER_REGISTRATION=true constellation create --control-plane-nodes 1 --worker-nodes 2 -y
+   ```
+   
+   :::
+
 3. Initialize the cluster
 
     ```bash
--- a/docs/docs/getting-started/first-steps.md
+++ b/docs/docs/getting-started/first-steps.md
@@ -95,26 +95,26 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
     Your Constellation cluster was created successfully.
     ```
 
-   :::info
+    :::info
 
-   On Azure, you may receive the following error when running with limited IAM permissions:
-   ```shell-session
-   Error: Error ensuring Resource Providers are registered.
+    On Azure, you may receive the following error when running with limited IAM permissions:
+    ```shell-session
+    Error: Error ensuring Resource Providers are registered.
 
-   Terraform automatically attempts to register the Resource Providers it supports to
-   ensure it's able to provision resources.
+    Terraform automatically attempts to register the Resource Providers it supports to
+    ensure it's able to provision resources.
 
-   If you don't have permission to register Resource Providers you may wish to use the
-   "skip_provider_registration" flag in the Provider block to disable this functionality.
-   ```
+    If you don't have permission to register Resource Providers you may wish to use the
+    "skip_provider_registration" flag in the Provider block to disable this functionality.
+    ```
 
-   To continue, please ensure that the [required resource providers](install.md#required-permissions) have been registered in your subscription by your administrator.
-   Afterward, run `constellation create` again with `ARM_SKIP_PROVIDER_REGISTRATION` set as an environment variable:
-   ```bash
-   ARM_SKIP_PROVIDER_REGISTRATION=true constellation create --control-plane-nodes 1 --worker-nodes 2 -y
-   ```
-   
-   :::
+    To continue, please ensure that the [required resource providers](install.md#required-permissions) have been registered in your subscription by your administrator.
+    Afterward, run `constellation create` again with `ARM_SKIP_PROVIDER_REGISTRATION` set as an environment variable:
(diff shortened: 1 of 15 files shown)


docs/docs/getting-started/first-steps.md
```diff
@@ -95,6 +95,27 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
...
+   "skip_provider_registration" flag in the Provider block to disable this functionality.
+   ```
+
+   To continue, please ensure that the [required resource providers](install.md#required-permissions) have been registered in your subscription by your administrator.
+   Afterward, run `constellation create` again with `ARM_SKIP_PROVIDER_REGISTRATION` set as an environment variable:
+   ```bash
+   ARM_SKIP_PROVIDER_REGISTRATION=true constellation create --control-plane-nodes 1 --worker-nodes 2 -y
+   ```
+   
+   :::
```
thomasten, line 117:
```suggestion
    :::info

    On Azure, you may receive the following error when running with limited IAM permissions:
    ```shell-session
    Error: Error ensuring Resource Providers are registered.

    Terraform automatically attempts to register the Resource Providers it supports to
    ensure it's able to provision resources.

    If you don't have permission to register Resource Providers you may wish to use the
    "skip_provider_registration" flag in the Provider block to disable this functionality.
    ```

    To continue, please ensure that the [required resource providers](install.md#required-permissions) have been registered in your subscription by your administrator.
    Afterward, run `constellation create` again with `ARM_SKIP_PROVIDER_REGISTRATION` set as an environment variable:
    ```bash
    ARM_SKIP_PROVIDER_REGISTRATION=true constellation create --control-plane-nodes 1 --worker-nodes 2 -y
    ```

    :::
```
indentation needs one more space or otherwise the other tip admonition above breaks (no idea why)

docs/docs/getting-started/install.md
```diff
@@ -111,6 +111,14 @@ If you don't have a cloud subscription, you can try [MiniConstellation](first-st
 <tabs groupId="csp">
 <tabItem value="azure" label="Azure">
 
+The following resource providers need to be [registered](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-providers-and-types#register-resource-provider) in your subscription:
```
thomasten, line 114: > I haven't added instructions on how to register the resource providers, since I didn't exactly know where to put them.

IMO this link is sufficient.

docs/docs/getting-started/first-steps.md
```diff
@@ -95,6 +95,27 @@ If you don't have a cloud subscription, check out [MiniConstellation](first-step
     Your Constellation cluster was created successfully.
     ```
 
+   :::info
+
+   On Azure, you may receive the following error when running with limited IAM permissions:
```
thomasten, line 100: > I am not 100% happy with the large troubleshooting block, but as far as I know we don't have any other troubleshooting section

We have Troubleshooting under Workflows. You may put it there if you prefer. I'm fine with both.

docs/docs/getting-started/install.md
```diff
@@ -111,6 +111,14 @@ If you don't have a cloud subscription, you can try [MiniConstellation](first-st
...
+* `Microsoft.Compute`
+* `Microsoft.ManagedIdentity`
+* `Microsoft.Network`
+* `microsoft.insights`
```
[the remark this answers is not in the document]
Nirusu: That's actually spelled lowercase by Microsoft.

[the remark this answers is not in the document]
thomasten: The docs page you linked contains a link to https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/azure-services-resource-providers where it's spelled uppercase. However, if in reality/technically it's spelled lowercase, it might be safer to keep it lowercase.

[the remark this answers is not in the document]
Nirusu: Good thing they themselves mess this up. This is what it looks like on the portal:
Screenshot 2023-03-01 at 12 32 33

The API client also returns "microsoft.insights", but at least the Azure CLI seems to accept the uppercase one?
No idea, it's a bit of the mess. Lowercase technically is safer.

thomasten (approved):
Thanks! Remember to apply the changes to the versioned docs before merge.

docs/docs/workflows/troubleshooting.md
```diff
@@ -2,6 +2,34 @@
 
 This section aids you in finding problems when working with Constellation.
 
+## Azure: Resource Providers can't be registered
+
+On Azure, you may receive the following error when running `create` or `terminate` step with limited IAM permissions:
```
thomasten, line 7: either
```suggestion
On Azure, you may receive the following error when running the `create` or `terminate` step with limited IAM permissions:
```
or
```suggestion
On Azure, you may receive the following error when running `create` or `terminate` with limited IAM permissions:
```

thomasten (approved):

PR #1071, opened 2023-01-25
internal: refactor storage credentials

daniel-weisse:
### Proposed change(s)

- Refactor KMS storage backends to use credentials in a similar way to the KMS backends.
- Add storage URI parsing functions

### Checklist

- [ ] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)
- [ ] Add labels (e.g., for changelog category)
- [ ] Link to Milestone

PR #1318, opened 2023-03-01
cli/bootstrapper: remove deprecated master secret & KMS related fields

Nirusu:
### Proposed change(s)
- Remove the deprecated, unused master secret & KMS related fields

From what I can see kms_uri does fully replace them, so no need to carry around that old luggage given that it is confusing while browsing through the code and understanding how things work.

Please double-check whether these fields are not really required anymore. Last time I asked around, they were deprecated. But the rfc/external-kms.md still mentions them, though plans obviously derived from that RFC. For example,  master secret and salt are definitely not used anymore as they are part of the KMS URI. Same for the KEK ID, I suppose.

### Checklist
- [x] Link to Milestone


bootstrapper/initproto/init.proto
```diff
@@ -10,15 +10,15 @@ service API {
...
 message InitRequest {
   // repeated string autoscaling_node_groups = 1; removed
-  bytes master_secret = 2;
+  // bytes master_secret = 2; removed
   string kms_uri = 3;
   string storage_uri = 4;
-  string key_encryption_key_id = 5;
-  bool use_existing_kek = 6;
+  // string key_encryption_key_id = 5; removed
+  // bool use_existing_kek = 6; removed
```
derpsteb, line 17: without double checking: maybe they are needed when using eKMS? Will check tomorrow.

Nirusu: I think I saw KEK ID being part of the KMS URI. Not sure about the bool, though.

daniel-weisse: Both removed.
However, I would like to hold off on this PR until https://github.com/edgelesssys/constellation/pull/1071 is merged

Nirusu: Okay, I'll mark it as blocked for now then.

Nirusu: Has been merged now. I rebased the PR. Marking it again ready for review.

daniel-weisse (approved):

PR #1322, opened 2023-03-02
oid: add alternative string representations for attestation variants

malt3:
### Proposed change(s)

- oid: add alternative string representations for attestation variants

### Additional info

This is the first part of a small refactor. aTLS Issuers/Validators should be selectable on a finer level than just the CSP. Adding this now simplifies adoption of more attestation variants in the future.
Next step is adding a kernel cmdline parameter indicating the wanted attestation variant using one of the string constants added here.

### Checklist

### DIFF
--- a/internal/oid/oid.go
+++ b/internal/oid/oid.go
@@ -25,11 +25,31 @@ package oid
 
 import (
 	"encoding/asn1"
+	"errors"
 )
 
 // Getter returns an ASN.1 Object Identifier.
 type Getter interface {
 	OID() asn1.ObjectIdentifier
+}
+
+// FromString returns the OID for the given string.
+func FromString(oid string) (Getter, error) {
+	switch oid {
+	case dummy:
+		return Dummy{}, nil
+	case awsNitroTPM:
+		return AWS{}, nil
+	case gcpSEVES:
+		return GCP{}, nil
+	case azureSEVSNP:
+		return AzureSNP{}, nil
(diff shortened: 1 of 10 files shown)


daniel-weisse (approved):
While we're at it, might as well rename the AWS and GCP Oid structs to AWSNitroTPM, and GCPES (or GCPSEVES)

internal/oid/oid.go
```diff
@@ -72,10 +112,29 @@ func (AzureTrustedLaunch) OID() asn1.ObjectIdentifier {
...
+func (QEMU) String() string {
+	return qemuVTPM
+}
+
+const (
+	dummy       = "dummy"
+	awsNitroTPM = "aws-nitrotpm"
+	gcpSEVES    = "gcp-seves"
+	azureSEVSNP = "azure-sevsnp"
+	azureTL     = "azure-trustedlaunch"
```
daniel-weisse, line 138:
```suggestion
	awsNitroTPM = "aws-nitro-tpm"
	gcpSEVES    = "gcp-se-ves"
	azureSEVSNP = "azure-sev-snp"
	azureTL     = "azure-trustedlaunch"
```
looks a bit cleaner imo

malt3: I was going for `<CSP>-<SUBVARIANT>` but this does look cleaner.

daniel-weisse (approved):

PR #1324, opened 2023-03-02
cli: log grpc connection state for init call

malt3:
This is a measure to detect cases where an aTLS handshake is performed but the long running call is interrupted, leading to a retry of the init call. Whenever the grpc connection state reaches ready, we know that the aTLS handshake has succeeded:

> READY: The channel has successfully established a connection all the way through TLS handshake (or equivalent) and protocol-level (HTTP/2, etc) handshaking, and all subsequent attempt to communicate have succeeded (or are pending without any known failure).

### Proposed change(s)

- cli: log grpc connection state for init call

### Additional info

Logs look like this:

```
[...]
2023-03-02T12:27:18+01:00       DEBUG   cmd/init.go:181 Sending initialization request
2023-03-02T12:27:18+01:00       DEBUG   cmd/init.go:217 Making initialization call, doer is &{dialer:0xc0012e09c0 endpoint:10.42.1.100:9000 req:0xc0006b3e00 resp:<nil> log:0xc00081c0f8}
2023-03-02T12:27:18+01:00       DEBUG   cmd/init.go:242 Created protoClient
2023-03-02T12:27:18+01:00       DEBUG   cmd/init.go:254 Connection state started as CONNECTING
2023-03-02T12:27:19+01:00       DEBUG   cmd/init.go:256 Connection state changed to CONNECTING
2023-03-02T12:27:19+01:00       DEBUG   cmd/init.go:259 Connection ready
2023-03-02T12:29:36+01:00       DEBUG   cmd/init.go:192 Initialization request succeeded
2023-03-02T12:29:36+01:00       DEBUG   cmd/init.go:193 Writing Constellation ID file
Your Constellation cluster was successfully initialized.
```

### Checklist

### DIFF
--- a/cli/internal/cmd/init.go
+++ b/cli/internal/cmd/init.go
@@ -40,6 +40,7 @@ import (
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/connectivity"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
@@ -236,6 +237,7 @@ func (d *initDoer) Do(ctx context.Context) error {
 		return fmt.Errorf("dialing init server: %w", err)
 	}
 	defer conn.Close()
+	d.logGRPCStateChanges(ctx, conn)
 	protoClient := initproto.NewAPIClient(conn)
 	d.log.Debugf("Created protoClient")
 	resp, err := protoClient.Init(ctx, d.req)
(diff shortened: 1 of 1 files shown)


cli/internal/cmd/init.go
```diff
@@ -246,6 +248,21 @@ func (d *initDoer) Do(ctx context.Context) error {
 	return nil
 }
 
+func (d *initDoer) logGRPCStateChanges(ctx context.Context, conn *grpc.ClientConn) {
+	go func() {
```
katexochen, line 252: We should wait for the goroutine to finishe after the connection is done.

malt3: While adding a waitgroup and waiting for the routine to end I noticed that the context we pass into `conn.WaitForStateChange` has to be canceled or the function would block since it doesn't know if `conn` may be reused for other calls.
This is also consistent with the function doc:

> // WaitForStateChange waits until the connectivity.State of ClientConn changes from sourceState or
> // ctx expires. A true value is returned in former case and false in latter.

katexochen (approved):

PR #1325, opened 2023-03-02
renovate: fix versioning for containers

katexochen:
without prefix

### Proposed change(s)

- Currently the gcp-guest-agent isn't updated correctly

### Checklist


malt3 (approved):
I hope this doesn't interfere with `ghcr.io/edgelesssys/cloud-provider-gcp` since it needs a different versioning scheme. But let's try!

PR #1326, opened 2023-03-02
cli: add `--kubernetes` flag to `iam create` (when used with `--create-config`)

Nirusu:
### Proposed change(s)

- Put K8s version parsing logic into an own function in configgenerate.go
- Add `--kubernetes` as a separate flag for IAM create using the new function (only respected when `--generate-config` is set)

This is mainly here to get both "config generate" and "iam create --generate-config" on feature parity. After v2.6.0, I hope we can refactor this a bit more to unify both config generation functions.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/configgenerate.go
+++ b/cli/internal/cmd/configgenerate.go
@@ -70,11 +70,7 @@ func (cg *configGenerateCmd) configGenerate(cmd *cobra.Command, fileHandler file
 	cg.log.Debugf("Parsed flags as %v", flags)
 	cg.log.Debugf("Using cloud provider %s", provider.String())
 	conf := createConfig(provider)
-	extendedVersion := config.K8sVersionFromMajorMinor(flags.k8sVersion)
-	if extendedVersion == "" {
-		return fmt.Errorf("kubernetes (%s) does not specify a valid Kubernetes version. Supported versions: %s", strings.TrimPrefix(flags.k8sVersion, "v"), supportedVersions())
-	}
-	conf.KubernetesVersion = extendedVersion
+	conf.KubernetesVersion = flags.k8sVersion
 	if flags.file == "-" {
 		content, err := encoder.NewEncoder(conf).Encode()
 		if err != nil {
@@ -134,14 +130,14 @@ func parseGenerateFlags(cmd *cobra.Command) (generateFlags, error) {
 	if err != nil {
 		return generateFlags{}, fmt.Errorf("parsing kuberentes flag: %w", err)
 	}
-	prefixedVersion := compatibility.EnsurePrefixV(k8sVersion)
-	if !semver.IsValid(prefixedVersion) {
-		return generateFlags{}, fmt.Errorf("kubernetes flag does not specify a valid semantic version: %s", k8sVersion)
+	resolvedVersion, err := resolveK8sVersion(k8sVersion)
+	if err != nil {
+		return generateFlags{}, fmt.Errorf("resolving kuberentes version from flag: %w", err)
 	}
 
 	return generateFlags{
 		file:       file,
-		k8sVersion: prefixedVersion,
+		k8sVersion: resolvedVersion,
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


cli/internal/cmd/iamcreate.go
```diff
@@ -56,8 +57,9 @@ func newIAMCreateCmd() *cobra.Command {
 		Args:  cobra.ExactArgs(0),
 	}
 
-	cmd.PersistentFlags().Bool("generate-config", false, "automatically generate a configuration file and fill in the required fields")
 	cmd.PersistentFlags().Bool("yes", false, "create the IAM configuration without further confirmation")
+	cmd.PersistentFlags().Bool("generate-config", false, "automatically generate a configuration file and fill in the required fields")
+	cmd.PersistentFlags().StringP("kubernetes", "k", semver.MajorMinor(config.Default().KubernetesVersion), "Kubernetes version to use in format MAJOR.MINOR, only usable in combination with --generate-config")
```
derpsteb, line 62: nit:

```suggestion
	cmd.PersistentFlags().StringP("kubernetes", "k", semver.MajorMinor(config.Default().KubernetesVersion), "Kubernetes version to use in format MAJOR.MINOR. Only usable in combination with --generate-config")
```

cli/internal/cmd/configgenerate.go
```diff
@@ -155,3 +151,19 @@ func generateCompletion(cmd *cobra.Command, args []string, toComplete string) ([
...
+// MAJOR.MINOR.PATCH release.
+func resolveK8sVersion(k8sVersion string) (string, error) {
+	prefixedVersion := compatibility.EnsurePrefixV(k8sVersion)
+	if !semver.IsValid(prefixedVersion) {
+		return "", fmt.Errorf("kubernetes flag does not specify a valid semantic version: %s", k8sVersion)
+	}
+
+	extendedVersion := config.K8sVersionFromMajorMinor(prefixedVersion)
+	if extendedVersion == "" {
+		return "", fmt.Errorf("--kubernetes (%s) does not specify a valid Kubernetes version. Supported versions: %s", strings.TrimPrefix(k8sVersion, "v"), supportedVersions())
```
derpsteb, line 165:
```suggestion
		return "", fmt.Errorf("--kubernetes (%s) does not specify a valid Kubernetes version. Supported versions: %s", k8sVersion, supportedVersions())
```

User should see their input in the error, shouldn't they?

derpsteb (approved):
cool. thanks for adding this in time :)

cli/internal/cmd/configgenerate.go
derpsteb: Makes sense. It _is_ confusing. My idea is that the error msg should show the user what the program "sees". The reason that the supported versions don't show a v prefix is because the user doesn't need to specify a v. 

Once we add the Semver type we should be able to loose all v prefixes the user sees.

cli/internal/cmd/iamcreate.go
Nirusu: Switched to `-` for now. Looks better in my opinion.

derpsteb: sgtm

cli/internal/cmd/configgenerate.go
Nirusu: >My idea is that the error msg should show the user what the program "sees"

On the other hand, at the point where the error message is thrown the program sees the version prefixed with `v` (since `prefixedVersion` is actually the input to `K8sVersionFromMajorMinor`). And indeed this can be a bi*** from a development standpoint give that if you miss the `v` in the call to `K8sVersionFromMajorMinor`, you get back an empty string even if you enter a correct version.

So by the standpoint of what the program "sees" at this point, returning `prefixedVersion` would make the most sense. But yeah, that would mean adding `v` everywhere to `supportedVersions` (or rather: remove the TrimPrefix in this function).

I don't know what eventually the plan is to display versions across our tools. I think you know this better on what everyone agreed on, so you decide :)

derpsteb: From what I have heard most want support for v-less versions. Internally we always need to add the v prefix. Lets go with the original input for now. I am happy to go back and change it if we come to a different conclusion later.

PR #1327, opened 2023-03-02
feat: journald package

miampf:
### Proposed change(s)

- add an internal package that can read systemd logs via journalctl.

### Checklist

- [x] Link to Milestone


internal/journald/journald.go
```diff
@@ -0,0 +1,48 @@
+/*
+Copyright (c) Edgeless Systems GmbH
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+/*
+Package journald provides functions to read and collect journald logs.
+*/
+package journald
```
katexochen, line 10: Better name it `journalctl`, as this is the binary we are wrapping. Also, this can live inside bootstrapper/internal, as it is currently not shared between different microservices.

internal/journald/journald.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+type collectionCommand interface {
+	executeCommand() ([]byte, error)
+}
+
+// Command represents a command that is executed by journalctl.
+type Command struct {
+	ctx     context.Context
+	service string
```
katexochen, line 24: In general, context's shouldn't be stored in structs. These two fields are also unused.

internal/journald/journald.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+	"os/exec"
+)
+
+type collectionCommand interface {
+	executeCommand() ([]byte, error)
+}
+
+// Command represents a command that is executed by journalctl.
+type Command struct {
```
katexochen, line 22: A package should expose an abstraction to its caller. We should name this after what it does, I suggest `Collector`. The caller doesn't need to know if we execute a command underneath or write all those logs by hand... :wink:

internal/journald/journald.go
```diff
@@ -0,0 +1,48 @@
...
+func NewCommand(ctx context.Context, service string) (*Command, error) {
+	cmd := exec.CommandContext(ctx, "journalctl", "-u", service)
+	if cmd.Err != nil {
+		return nil, cmd.Err
+	}
+	return &Command{ctx, service, cmd}, nil
+}
+
+// GetServiceLog gets all journald logs from a service and returns a byte array with the plain text logs.
+func GetServiceLog(jcmd collectionCommand) ([]byte, error) {
```
katexochen, line 38: Given the renamed struct above, this should be a method on the struct, not a function. Per convention, the name would be `Collect()`.

internal/journald/journald.go
```diff
@@ -0,0 +1,48 @@
...
+	out, err := jcmd.executeCommand()
+	if err != nil {
```
katexochen, line 40: We need to check the error and, if it is an exit error, return the Stderr that command printed.
```suggestion
func (c *Collector) Collect() ([]byte, error) {
	out, err := c.cmd.Output()
	var exitErr *exec.ExitError
	if errors.As(err, &exitErr) {
		return nil, fmt.Errorf("executing %q failed: %s", c.cmd, exitErr.Stderr)
	} else if err != nil {
		return nil, fmt.Errorf("executing command: %w", err)
	}
	return out, nil
```

internal/journald/journald.go
```diff
@@ -0,0 +1,48 @@
...
+
+type collectionCommand interface {
+	executeCommand() ([]byte, error)
+}
+
+// Command represents a command that is executed by journalctl.
+type Command struct {
+	ctx     context.Context
+	service string
+	command *exec.Cmd
```
katexochen, line 25: we shouldn't keep a concrete type in the struct, but an interface that exec.Cmd fulfills, like 
```go
type command interface {
	Output() ([]byte, error)
	String() string
}

```

katexochen (changes requested):

internal/journald/journald.go
miampf: Thanks for the tip :)

bootstrapper/internal/journald/journald.go
```diff
@@ -0,0 +1,47 @@
...
+	"errors"
+	"fmt"
+	"os/exec"
+)
+
+type command interface {
+	Output() ([]byte, error)
+}
+
+// Collector represents a command that is executed by journalctl.
```
katexochen, line 23:
```suggestion
// Collector collects logs from journald.
```

bootstrapper/internal/journald/journald.go
```diff
@@ -0,0 +1,47 @@
...
+type Collector struct {
+	cmd command
+}
+
+// NewCollector creates a new JournaldCommand.
```
katexochen, line 28:
```suggestion
// NewCollector creates a new Collector for journald logs.
```

bootstrapper/internal/journald/journald.go
```diff
@@ -0,0 +1,47 @@
...
+func NewCollector(ctx context.Context, service string) (*Collector, error) {
+	cmd := exec.CommandContext(ctx, "journalctl", "-u", service)
+	if cmd.Err != nil {
+		return nil, cmd.Err
+	}
+	return &Collector{cmd}, nil
+}
+
+// Collect gets all journald logs from a service and returns a byte array with the plain text logs.
```
katexochen, line 37:
```suggestion
// Collect gets all journald logs from a service and returns a byte slice with the plain text logs.
```

bootstrapper/internal/journald/journald.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// Collector represents a command that is executed by journalctl.
+type Collector struct {
+	cmd command
+}
+
+// NewCollector creates a new JournaldCommand.
+func NewCollector(ctx context.Context, service string) (*Collector, error) {
+	cmd := exec.CommandContext(ctx, "journalctl", "-u", service)
```
katexochen, line 30: Wasn't the plan to get all logs, to keep it simple?

bootstrapper/internal/journald/journald_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"errors"
+	"os/exec"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type stubJournaldCommand struct {
+	OutputReturn []byte
+	OutputError  error
```
katexochen, line 19:
```suggestion
	outputReturn []byte
	outputError  error
```

katexochen (changes requested):
only minor things left :hamster:

bootstrapper/internal/journald/journald.go
miampf: I thought that it would be better to have a more flexible package that can read logs from any service if we require logs from other services one day. In the bootstrapper, this will be only used to read logs from the bootstrapper and the kubelet service for now.

katexochen: As said before, imo we should just collect all units right away if init fails. How are you going to store the logs of the units you want to collect on the CLI side? In multiple files? Just concatenate the strings?

miampf: I thought of using multiple files in the form of `error-{SERVICE_NAME}.log`. But I believe I understand your point, if you want me to collect and stream all systemd logs then I can change the code accordingly.

katexochen: Multiple files will likely make it harder to request that data from the user.

malt3:

katexochen (approved):
lgtm. :dodo:

PR #1328, opened 2023-03-02
cli: set OpenStack service account credentials

malt3:
### Proposed change(s)

- cli: set OpenStack service account credentials

### Checklist

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -238,6 +238,9 @@ type OpenStackConfig struct {
 	//   Floating IP pool to use for the VMs. For details see: https://docs.openstack.org/ocata/user-guide/cli-manage-ip-addresses.html
 	FloatingIPPoolID string `yaml:"floatingIPPoolID" validate:"required"`
 	// description: |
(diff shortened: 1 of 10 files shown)


internal/config/config.go
```diff
@@ -360,11 +363,16 @@ func New(fileHandler file.Handler, name string, force bool) (*Config, error) {
 	}
 
 	// Read secrets from env-vars.
-	clientSecretValue := os.Getenv(constants.EnvVarAzureClientSecretValue)
-	if clientSecretValue != "" && c.Provider.Azure != nil {
+	clientSecretValue, ok := os.LookupEnv(constants.EnvVarAzureClientSecretValue)
+	if ok && c.Provider.Azure != nil {
```
katexochen, line 367: What's the reason for this change? If I understand it right, you are now allowing to explicitly set the clientSecretValue to the empty string, right?

malt3: I was looking for a more general pattern in case we want to use this for other config options (where we may want an empty string). For the environment variables we currently use for overriding we could disallow the empty string.

katexochen (approved):

PR #1337, opened 2023-03-03
deps: fix dependabot security warnings

derpsteb:
### Proposed change(s)

- These are fixed on main because of the local replace rules but dependabot doesn't recognize this. 

### Checklist

- [x] Link to Milestone

### DIFF
--- a/operators/constellation-node-operator/go.mod
+++ b/operators/constellation-node-operator/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.3.1
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.2.1
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v4 v4.1.0
-	github.com/edgelesssys/constellation/v2 v2.0.0
+	github.com/edgelesssys/constellation/v2 v2.5.2
 	github.com/edgelesssys/constellation/v2/3rdparty/node-maintenance-operator v0.0.0-00010101000000-000000000000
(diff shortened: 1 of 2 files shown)


operators/constellation-node-operator/go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.3.1
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.2.1
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v4 v4.1.0
-	github.com/edgelesssys/constellation/v2 v2.0.0
+	github.com/edgelesssys/constellation/v2 v2.5.2
 	github.com/edgelesssys/constellation/v2/3rdparty/node-maintenance-operator v0.0.0-00010101000000-000000000000
 	github.com/edgelesssys/constellation/v2/operators/constellation-node-operator/v2/api v0.0.0
```
malt3, line 18: We should think about those too

malt3: It feels like they should be upgraded in sync. Although i'm not sure if they are also on v2

derpsteb: They didn't show up in the warnings so I didn't touch them. Let me quickly update them.

derpsteb: Thinking about them again: They don't need updating. These modules are not part of the advisory. Right?

katexochen: They are versioned independent and are not v2

malt3 (approved):

PR #1341, opened 2023-03-03
ci: store additional data in bench results

m1ghtym0:
### Proposed change(s)

- Store MTU and volume size as part of the benchmark results
- Fix typo in udp fields


katexochen (approved):
lgtm

PR #1343, opened 2023-03-05
attestation: validate GCP machine state instead of PCR 0

thomasten:
PCR 0 changed on GCP. The new value can be calculated as follows:

```go
func main() {
	ev1 := &bytes.Buffer{}
	binary.Write(ev1, binary.LittleEndian, utf16.Encode([]rune("GCE Virtual Firmware v2\x00")))
	pcr := extend(make([]byte, 32), ev1.Bytes())                                      // EV_S_CRTM_VERSION
	pcr = extend(pcr, append([]byte("GCE NonHostInfo\x00\x01"), make([]byte, 15)...)) // EV_NONHOST_INFO
	pcr = extend(pcr, []byte{0, 0, 0, 0})                                             // EV_SEPARATOR
	fmt.Printf("%x\n", pcr)
}

func extend(last, data []byte) []byte {
	digest := sha256.Sum256(data)
	h := sha256.New()
	h.Write(last)
	h.Write(digest[:])
	return h.Sum(nil)
}
```

If you replace `v2` with `v1` in the first event, you get the old PCR value.

This means that PCR 0 will change if the GCE firmware version number is increased. We should stop comparing PCR 0 against a fixed value. Instead, we can replay the event log to validate the PCR and then validate the events. Fortunately, VerifyAttestation of go-tpm-tools already does most of this for us. This PR checks the result of the func and drops the fixed value for PCR 0.

I'm not sure if I catched all places that need to be changed to drop PCR 0. Please check.

For now I only tested the change using `constellation verify` against a cluster created with the released CLI v2.5.3 (with PCR 0 set to warnOnly).


daniel-weisse (commented):
PCR[10] is used by Linux IMA, which writes a hash hash over TPM registers 0-7 to it.
I'm not 100% certain if this still happens in out images, but we might want to remove this value from our config as well.

derpsteb:
The default config also needs to be updated (?).

thomasten:
> PCR[10] is used by Linux IMA, which writes a hash hash over TPM registers 0-7 to it. I'm not 100% certain if this still happens in out images, but we might want to remove this value from our config as well.

Yes, I saw a warning for PCR 10. Removed it.

thomasten:
> The default config also needs to be updated (?). I think [this](https://github.com/edgelesssys/constellation/blob/main/internal/attestation/measurements/measurements_oss.go#L38-L41) should be removed.

Done. Not sure if I have to do anything with measurements_enterprise.go.

derpsteb:
I don't think so. Seems to me like those are always updated based on the build-os-image.yaml measurements.

derpsteb (approved):

daniel-weisse (approved):

PR #1347, opened 2023-03-06
bootstrapper: make sure InitServer is only shut down after Init has returned

daniel-weisse:
### Proposed change(s)

-
-

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/bootstrapper/internal/initserver/initserver.go
+++ b/bootstrapper/internal/initserver/initserver.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/edgelesssys/constellation/v2/bootstrapper/initproto"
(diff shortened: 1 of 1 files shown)


bootstrapper/internal/initserver/initserver.go
```diff
@@ -198,7 +203,9 @@ func (s *Server) Init(ctx context.Context, req *initproto.InitRequest) (*initpro
 func (s *Server) Stop() {
 	s.log.Infof("Stopping")
 
+	s.shutdownLock.Lock()
```
derpsteb, line 206: Could we have a comment here giving some information on why we need to do the locking? This seems like a perfect candidate for a "refactoring" of the "oh, this seems duplicate"-kind :D.

derpsteb:
lgtm. AWS with 3+2 nodes succeeded. The same setup failed reliably before.

daniel-weisse:
init passed on AWS and Azure:

* https://github.com/edgelesssys/constellation/actions/runs/4344499301/jobs/7587908653
* https://github.com/edgelesssys/constellation/actions/runs/4344496507/jobs/7587902964

bootstrapper/internal/initserver/initserver.go
```diff
@@ -113,6 +115,9 @@ func (s *Server) Serve(ip, port string, cleaner cleaner) error {
 
 // Init initializes the cluster.
 func (s *Server) Init(ctx context.Context, req *initproto.InitRequest) (*initproto.InitResponse, error) {
+	s.shutdownLock.Lock()
```
3u13r, line 118: I think we should maintain the parallel nature of the server. Either by using a semaphore or a RWLock. 

The idea is to allow sending init requests to test the behavior for the checks before the nodeLock. Also this would emphasize, that the security is still based on the nodeLock and not the shutdownLock.

derpsteb (approved):
still works for me :)

malt3 (approved):

PR #1349, opened 2023-03-07
ci: use latest ver of versionsapi cli container

katexochen:
### Checklist


msanft:
I don't want to be the one to judge on your work, but wouldn't it be idiomatic to pass the tag as an input if we want to use the action elsewhere and tag the container differently?

msanft (approved):
🎯

PR #1351, opened 2023-03-07
image: add systemd-boot as explicit dependency

malt3:
### Proposed change(s)

- image: add systemd-boot as explicit dependency

### Additional info

Fedora's systemd package now does not bundle systemd-boot anymore, so we have to install it explicitly: https://bodhi.fedoraproject.org/updates/FEDORA-2023-f72fbc711a.

[Testrun](https://github.com/edgelesssys/constellation/actions/runs/4352383126)

Otherwise, `bootctl install` will fail:

```
  ‣    Installing boot loader…
  Failed to open "/usr/lib/systemd/boot/efi": No such file or directory
```

### Checklist

- [x] Link to Milestone

### DIFF
--- a/image/mkosi.conf.d/secure-boot-tpm.conf
+++ b/image/mkosi.conf.d/secure-boot-tpm.conf
@@ -5,4 +5,5 @@ Packages=
            sbsigntools,
            efitools,
            mokutil,
-           tpm2-tools
+           tpm2-tools,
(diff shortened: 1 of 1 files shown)


derpsteb (approved):

derpsteb:
Verifying run: https://github.com/edgelesssys/constellation/actions/runs/4352383126/jobs/7605296824

PR #1355, opened 2023-03-07
attestation: create issuer based on kernel cmd line

daniel-weisse:
### Proposed change(s)

- Instead of relying on just the Cloud Provider to identify what Attestation Issuer we need to set up, we now use information from the Kernel command line to do so
- Add `choose.Issuer()` and `choose.Validator()` utility functions to create Issuer or Validator from a given oid

### Additional info
- Set up of Issuer for the VerifyService will be handled in a different PR
- Dynamic creation of Validators using `choose.Validator()` will be handled in a different PR

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/internal/attestation/choose/choose.go
+++ b/internal/attestation/choose/choose.go
@@ -36,7 +36,6 @@ func Issuer(variant oid.Getter, log vtpm.AttestationLogger) (atls.Issuer, error)
 		return qemu.NewIssuer(log), nil
 	case oid.Dummy{}:
 		return atls.NewFakeIssuer(oid.Dummy{}), nil
-	// TODO(malt3): Add Openstack Issuer
 	default:
 		return nil, fmt.Errorf("unknown attestation variant: %s", variant)
 	}
@@ -61,7 +60,6 @@ func Validator(
 		return qemu.NewValidator(measurements, log), nil
 	case oid.Dummy{}:
 		return atls.NewFakeValidator(oid.Dummy{}), nil
-	// TODO(malt3): Add Openstack Validator
 	default:
 		return nil, fmt.Errorf("unknown attestation variant: %s", variant)
 	}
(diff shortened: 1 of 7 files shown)


internal/attestation/choose/choose.go
```diff
@@ -0,0 +1,68 @@
...
+		return trustedlaunch.NewIssuer(log), nil
+	case oid.AzureSEVSNP{}:
+		return snp.NewIssuer(log), nil
+	case oid.GCPSEVES{}:
+		return gcp.NewIssuer(log), nil
+	case oid.QEMUVTPM{}:
+		return qemu.NewIssuer(log), nil
+	case oid.Dummy{}:
+		return atls.NewFakeIssuer(oid.Dummy{}), nil
+	// TODO(malt3): Add Openstack Issuer
```
malt3, line 39:
```suggestion
```

I think at this point we will either use the dummy issuer (for testing) and later the qemu vtpm issuer. A custom issuer is not planned.

internal/attestation/choose/choose.go
```diff
@@ -0,0 +1,68 @@
...
+		return trustedlaunch.NewValidator(measurements, log), nil
+	case oid.AzureSEVSNP{}:
+		return snp.NewValidator(measurements, idKeyDigest, enfoceIDKeyDigest, log), nil
+	case oid.GCPSEVES{}:
+		return gcp.NewValidator(measurements, log), nil
+	case oid.QEMUVTPM{}:
+		return qemu.NewValidator(measurements, log), nil
+	case oid.Dummy{}:
+		return atls.NewFakeValidator(oid.Dummy{}), nil
+	// TODO(malt3): Add Openstack Validator
```
malt3, line 64:
```suggestion
```

I think at this point we will either use the dummy validator (for testing) and later the qemu vtpm validator. A custom validator is not planned.

malt3 (commented):
Change so far looks good. 
Do we also want to implement a chooser to select the validator for the joinservice? We would need to pass this as either an env var or a command line flag and then refactor `internal/watcher/validator.go` a bit.
What do you think? Otherwise, a todo would also suffice.

daniel-weisse:
>Do we also want to implement a chooser to select the validator for the joinservice? 

See my PR description. Related Ticket: [AB#2975](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/2975)

PR #1356, opened 2023-03-07
cli: use registry.k8s.io instead of k8s.gcr.io (GCP CSI driver)

Nirusu:
### Proposed change(s)

- cli: use registry.k8s.io instead of k8s.gcr.io
See: https://kubernetes.io/blog/2023/02/06/k8s-gcr-io-freeze-announcement/

This shouldn't break any existing CLI release as long as the K8s team really only freezes the repository (versus actually shutting it down). But since it might trigger one of these fancy scanning tools on GCP that an outdated repository is used, let's update here.

@daniel-weisse Is it possible to also update the external CSI images without the driver breaking? I don't think these are tracked by Renovate, so we might also bump them is this wouldn't break anything.

### Checklist

- [ ] Link to Milestone


Nirusu:
Ah, these are pulled from outside. Will update there.

PR #1375, opened 2023-03-08
cli: support custom attestation policies for maa

msanft:
### Proposed change(s)

- Add custom MAA to Azure cluster creation
- To do that, create the attestation provider and then update it's policy afterwards (creating an attestation provider with a specific policy is not yet supported by the Azure Terraform provider)

Outcome:
Azure Outcome

### Additional info
- The way the update of the policy works now is super hacky, but after hours of  checking out outdated SDKs and trying to reverse the API calls, it was the only way that turned out to work

### Checklist

- [x] Link to Milestone


cli/internal/terraform/terraform_test.go
```diff
@@ -330,6 +333,34 @@ func TestCreateCluster(t *testing.T) {
...
+						Outputs: map[string]*tfjson.StateOutput{},
+					},
+				},
+			},
+			fs:      afero.NewMemMapFs(),
+			wantErr: true,
+		},
+		"attestation url has wrong type": {
+			pathBase: "terraform",
+			provider: cloudprovider.QEMU,
```
daniel-weisse, line 352:
```suggestion
			provider: cloudprovider.Azure,
```

daniel-weisse (commented):
While very hacky, I think this is the only way to get MAA working currently.

Should we gate creation of the MAA policy provider behind actually having that attestation mode enabled in the config?
(A TODO comment should work for now, since validator/issuer MAA support is not yet merged in main)

cli/internal/cloudcmd/create.go
```diff
@@ -215,14 +217,42 @@ func (c *Creator) createAzure(ctx context.Context, cl terraformClient, config *c
 		return clusterid.File{}, err
 	}
 
+	// very, very hacky way to update the MAA attestation policy. This should be changed as soon as either the Terraform provider supports it
+	// or the Go SDK gets updated to a recent API version.
+	// https://github.com/hashicorp/terraform-provider-azurerm/issues/20804
+	cmd := exec.CommandContext(ctx,
+		"az", "rest",
```
katexochen, line 224: Having runtime dependencies like this is too hacky for main imo.

If we want to merge nevertheless, this is missing appropriate documentation about az as required dependency.

msanft: If I'm not mistaken the [Azure CLI should be a requirement](https://docs.edgeless.systems/constellation/getting-started/install#authentication) already?

katexochen: For authentication, yes. For running our CLI, no. 
There are other ways to authenticate against Azure, like the Azure login action, that don't require to have az available. We should explicitly state that this required to execute the CLI (which is also a breaking change, btw).

msanft: Let's talk about that tomorrow. I may not be as well informed about which parts of the setup process need to have this available and which do not, but I think for executing the Terraform file (which also happens at cluster creation time), the Azure CLI is required? Nonetheless docs are necessary though, I agree

Nirusu: If I may give my 2.5 cents to this:
>There are other ways to authenticate against Azure, like the Azure login action, that don't require to have az available.
I think the Azure login action also requires the CLI exist.
However, Terraform can directly authenticate over a Service Principal and Client Secret, what we currently use for e2e tests for example. If I understand correctly, _that_ doesn't need the Azure CLI to be present (in PATH).

However, I might have a suggestion for a slightly less hacky version:
Use `azidentity` from the Azure Go SDK to authenticate with Azure, retrieve the Bearer token from it (that is directly exportable) and just manually do a PUT request using Go's http library. Then you can use the SDK to authenticate and don't need the Azure CLI necessarily if you fall back to other authentication methods the SDK supports.

However, then you still need to keep in mind that Terraform expects the credentials to be prefixed with `ARM` while `azidentity` doesn't. But some glue code might be able to copy from one to another if they can be used equally otherwise.

msanft: > Or maybe the Azure SDK also offers a Go HTTP client that is authenticated
That's an idea I also came across, the [`autorest`](https://pkg.go.dev/github.com/Azure/go-autorest/autorest#Prepare) package does exactly that.
The only thing I'm concerned about is: How would we retrieve the login info from the user? Doesn't he also need the CLI installed for that?

thomasten: What was the outcome of Azure/azure-sdk-for-go#20349 ?

msanft: Found that the [`attestation`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/services/attestation/2020-10-01/attestation) package should provide this functionality, but it's outdated as well (doesn't support auth adapters nor AzureVM TEE type, while latter can be hacked around though)

thomasten: If they would regenerate it, could we use it instead of this hack? If yes, maybe you should reopen it?

msanft: I do not think so. 

There are multiple layers to the problem:

1. **Outdated Azure REST API** 
	The documentation of the [Azure REST API](https://learn.microsoft.com/en-us/rest/api/attestation/policy/set?tabs=HTTP) (from which I believe the SDKs get generated?) does not mention that Azure VMs ( `attestationType=AzureGuest` ) are supported. I only found the type string "by accident" when inspecting the HTTP request of the Azure web interface. Furthermore, the Azure docs do not mention the Azure VMs anywhere, as far as I found out. The API docs claim to be on API version `2022-08-01`.
2. **Outdated Go SDKs**
	Both the [`armattestation`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/attestation/armattestation) as well as the [`attestation`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/services/attestation/2020-10-01/attestation) package provide an API to interact with Azure attestation providers. Both packages have the problem of not explicitly supporting the Azure VM type. This can be hacked around though by just inserting the string `AzureGuest` since in both packages, the `TeeType` implementation boils down to being just a string internally. For the `armattestation` package, the next problem is that the package does not support policy management at all. The only thing an [update request](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/attestation/armattestation#ServicePatchParams) can contain is a map of tags. For the `attestation` package, there is another problem. First, it does not support credential auth (like any other Azure SDK), but only the old [`adal`](https://pkg.go.dev/github.com/Azure/go-autorest/autorest#NewBearerAuthorizer) OAuth, which will be EOL on March 31st. I thought this can be hacked around by creating a shim which implements the interface, but it turned out that the [`WithAuthorization`](https://github.com/Azure/azure-sdk-for-go/blob/2187aca40bbf6931ffac666977e5d29128d82585/services/attestation/2020-10-01/attestation/policy.go#L252) method is just not being called at all. Both packages claim to be on API version `2020-10-01`.
3. **Outdated Terraform provider** 
	The functionality provided by the [Azure Terraform provider](https://registry.terraform.io/providers/hashicorp/azurerm/3.46.0/docs/resources/attestation) is similar to what the `armattestation` package can do. There's just not support for policy management at all. I don't know how the Terraform provider gets generated internally, but I assume it would also use either the REST API definitions or the SDKs? However, the functionality is not sufficient as is.

I don't know how I would proceed with this. I assume a request to first-level-support asking for the API to be updated. Once that is in we could try and get the SDKs updated. Either way, this process will surely not be fast.. 

Let me know what you think @thomasten @azure-proficient-folks 😄

msanft: According to Daniel, for cluster creation the Azure CLI currently is not needed. I don't know how Terraform authenticates with Azure exactly, but maybe we could require to set credentials when executing cluster creation with MAA selected as attestation type and then authenticate with the Azure API and send the HTTP request as @Nirusu said?

daniel-weisse: I *think* the Azure CLI is not needed. Terraform should be able to load credentials from file or environment. The azure CLI would only be "needed" to generate those credentials in the first place.
The Azure SDK has a credentials loading function that handles looking for credentials in all the common places (this also does not require the Azure CLI), I would assume Terraform uses this function (or something similar).

Using those credentials to send HTTP request might work

msanft: Just verified: Constellation CLI is needed for cluster creation.

```
moritz@moritzs-qemu-test:~/const/const-setup$ constellation create --control-plane-nodes 1 --worker-nodes 2 -y
Creating
An error occurred: exit status 1

Error: building AzureRM Client: please ensure you have installed Azure CLI version 2.0.79 or newer. Error parsing json result from the Azure CLI: launching Azure CLI: exec: "az": executable file not found in $PATH.

Comment (Nirusu) [reply cli/internal/cloudcmd/create.go:224]: It is needed if you don't specify any environment variables with credentials (see what we do in the e2e-tests) or used managed identities. But yes, the _default_ authentication method is over the CLI.

**Comment by msanft** [comment]
Thoughts on this "PoC"? Better than using the CLI?
```
func main() {
	cred, _ := azidentity.NewDefaultAzureCredential(nil)
	token, _ := cred.GetToken(context.Background(), policy.TokenRequestOptions{
		Scopes: []string{"https://attest.azure.net/.default"},
	})

	client := attestation.NewPolicyClient()

	req, _ := client.SetPreparer(context.Background(),
		"https://snpattestationtester.neu.attest.azure.net", "azureGuest", encodeAttestationPolicy(constants.EncodedAzureMAAPolicy))
	req.Header.Set("Authorization", "Bearer "+token.Token)

	res, _ := client.Send(req)
}

msanft:
Implemented this as a HTTP request now, prepared by the official Azure package, then decorated with the Authorization header by ourselves. One thing to consider are the environment variables used for Azure credentials, as @Nirusu told me.

The Terraform provider uses the `ARM` prefix while the [Azure SDK](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#readme-authenticating-service-principals) uses the `AZURE` one.

For example: https://github.com/edgelesssys/constellation/blob/02694c0648baaf30f4bd7b51b26c25053b78394c/.github/workflows/e2e-test-manual.yml#L111-L115

malt3:

cli/internal/cloudcmd/create.go
```diff
@@ -54,7 +59,7 @@ func NewCreator(out io.Writer) *Creator {
 }
 
 // Create creates the handed amount of instances and all the needed resources.
-func (c *Creator) Create(ctx context.Context, provider cloudprovider.Provider, config *config.Config, insType string, controlPlaneCount, workerCount int,
+func (c *Creator) Create(ctx context.Context, provider cloudprovider.Provider, config *config.Config, azurePolicyPatcher PolicyPatcher, insType string, controlPlaneCount, workerCount int,
```
thomasten, line 62: create the PolicyPatcher in NewCreator instead of creating it in the cmd package and passing it

cli/internal/cloudcmd/create.go
```diff
@@ -215,14 +220,84 @@ func (c *Creator) createAzure(ctx context.Context, cl terraformClient, config *c
...
+
+// Patch updates the attestation policy to the base64-encoded attestation policy JWT for the given attestation URL.
+// https://learn.microsoft.com/en-us/azure/attestation/author-sign-policy#next-steps
+func (p policyPatcher) Patch(ctx context.Context, attestationURL, policy string) error {
+	// hacky way to update the MAA attestation policy. This should be changed as soon as either the Terraform provider supports it
+	// or the Go SDK gets updated to a recent API version.
+	// https://github.com/hashicorp/terraform-provider-azurerm/issues/20804
+	cred, err := azidentity.NewDefaultAzureCredential(nil)
+	if err != nil {
+		return fmt.Errorf("failed to create default Azure credentials: %w", err)
```
thomasten, line 257:
```suggestion
		return fmt.Errorf("creating default Azure credentials: %w", err)
```
This reads better when errors are chained. Also adapt the other error strings.

cli/internal/cloudcmd/create.go
```diff
@@ -215,14 +220,84 @@ func (c *Creator) createAzure(ctx context.Context, cl terraformClient, config *c
...
+	}
+	token, err := cred.GetToken(context.Background(), azpolicy.TokenRequestOptions{
```
thomasten, line 259:
```suggestion
	token, err := cred.GetToken(ctx, azpolicy.TokenRequestOptions{
```

cli/internal/cloudcmd/create.go
```diff
@@ -215,14 +220,84 @@ func (c *Creator) createAzure(ctx context.Context, cl terraformClient, config *c
...
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create default Azure credentials: %w", err)
+	}
+
+	client := attestation.NewPolicyClient()
+
+	// azureGuest is the id for the "Azure VM" attestation type. Other types are documented here:
+	// https://learn.microsoft.com/en-us/rest/api/attestation/policy/set
+	req, err := client.SetPreparer(context.Background(), attestationURL, "azureGuest", p.encodeAttestationPolicy(constants.EncodedAzureMAAPolicy))
```
thomasten, line 270:
```suggestion
	req, err := client.SetPreparer(ctx, attestationURL, "azureGuest", p.encodeAttestationPolicy(constants.EncodedAzureMAAPolicy))
```

cli/internal/cloudcmd/create.go
```diff
@@ -215,14 +220,84 @@ func (c *Creator) createAzure(ctx context.Context, cl terraformClient, config *c
...
+	}
+
+	res, err := client.Send(req)
+	if err != nil {
+		return fmt.Errorf("failed to send request: %w", err)
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
+		return fmt.Errorf("failed to update attestation policy: %s", res.Status)
```
thomasten, line 283:
```suggestion
	resp, err := client.Send(req)
	if err != nil {
		return fmt.Errorf("failed to send request: %w", err)
	}
	resp.Body.Close()

	if resp.StatusCode != http.StatusOK {
		return fmt.Errorf("failed to update attestation policy: %s", resp.Status)
```
This is usually called resp.
You can immediately close the Body if you don't need it.

cli/internal/cloudcmd/create.go
```diff
@@ -215,14 +220,84 @@ func (c *Creator) createAzure(ctx context.Context, cl terraformClient, config *c
...
+	}
+
+	return nil
+}
+
+// encodeAttestationPolicy encodes the base64-encoded attestation policy in the JWS format specified here:
+// https://learn.microsoft.com/en-us/azure/attestation/author-sign-policy#creating-the-policy-file-in-json-web-signature-format
+func (p policyPatcher) encodeAttestationPolicy(encodedPolicy string) string {
+	header := `{"alg":"none"}`
+	payload := fmt.Sprintf(`{"AttestationPolicy": "%s"}`, encodedPolicy)
```
thomasten, line 293:
```suggestion
	const header = `{"alg":"none"}`
	payload := fmt.Sprintf(`{"AttestationPolicy":"%s"}`, encodedPolicy)
```

cli/internal/cloudcmd/create.go
```diff
@@ -215,14 +220,84 @@ func (c *Creator) createAzure(ctx context.Context, cl terraformClient, config *c
...
+
+	encodedHeader := base64.RawStdEncoding.EncodeToString([]byte(header))
+	encodedPayload := base64.RawStdEncoding.EncodeToString([]byte(payload))
```
thomasten, line 296:
```suggestion
	encodedHeader := base64.RawURLEncoding.EncodeToString([]byte(header))
	encodedPayload := base64.RawURLEncoding.EncodeToString([]byte(payload))
```
This must be url encoding. This may also be the problem you faced with the policy encoding.

cli/internal/terraform/terraform.go
```diff
@@ -128,10 +128,20 @@ func (c *Client) CreateCluster(ctx context.Context) (CreateOutput, error) {
 		return CreateOutput{}, errors.New("invalid type in uid output: not a string")
 	}
 
+	attestationURLOutput, ok := tfState.Values.Outputs["attestationURL"]
+	if !ok {
+		return CreateOutput{}, errors.New("no attestationUrl output found")
```
thomasten, line 133: this will fail for other CSPs

cli/internal/terraform/terraform/azure/main.tf
```diff
@@ -41,6 +41,12 @@ resource "random_password" "initSecret" {
   override_special = "_%@"
 }
 
+resource "azurerm_attestation_provider" "attestation_provider" {
```
thomasten, line 44: > Should we gate creation of the MAA policy provider behind actually having that attestation mode enabled in the config?

Yes, we should.

thomasten (changes requested):

cli/internal/cloudcmd/create.go
msanft: Looking at this, is it correct that we use the background context and not the (passed down) command context at the rollbacker? If so, can I add this to the godoc of the `rollback` method so that nobody ends up using the wrong context by accident.
https://github.com/edgelesssys/constellation/blob/5355e1cc9c963f7749beb571a60bbd8860fc50d6/cli/internal/cloudcmd/create.go#L217

cli/internal/cloudcmd/create.go
msanft: Out of interest: Has closing the body immediately any benefits (except for potential memory being freed)? Wouldn't it be more convenient for other developers who may extend the function (i.e. by further processing of the body) to keep it deferred? Don't take this as criticism to your review please, I really just want to extend my knowledge on this :)

cli/internal/cloudcmd/create.go
msanft: Agree! The issue with the policy encoding was a whitespace issue. I wasn't able to format the policy string so that the whitespace matched up with the one needed(?) by Azure? I tested some ways using backticks but didn't get it to work that way. (500 status code)

cli/internal/terraform/terraform/azure/main.tf
msanft: @daniel-weisse should I just use pass an extra variable into the Terraform call and then [let Terraform create the resource conditionally](https://developer.hashicorp.com/terraform/language/expressions/conditionals) depending on that variable? Or do we have a best-practice way to go here?

cli/internal/cloudcmd/create.go
thomasten: IMO it also makes the intent of the code clearer. I can see at once that Body isn't processed on purpose. Anyway, this is just a nit (forgot to prefix "nit: ") and you can choose what you prefer.

cli/internal/cloudcmd/create.go
thomasten: Yes, this is correct because if the context is canceled, the rollback should still be performed. @katexochen, I wonder whether we should remove the context parameter from rollbackOnError and create the background context in the func to prevent wrong use of the func.

cli/internal/terraform/terraform/azure/main.tf
daniel-weisse: >should I just use pass an extra variable into the Terraform call

Yes.
Configuration will happen through the already existing `EnforceIDKeyDigest` key in the config.

cli/internal/cloudcmd/create.go
katexochen: Yes, good idea!

cli/internal/terraform/terraform.go
```diff
@@ -128,10 +128,23 @@ func (c *Client) CreateCluster(ctx context.Context) (CreateOutput, error) {
...
+	if provider == cloudprovider.Azure {
+		attestationURLOutput, ok := tfState.Values.Outputs["attestationURL"]
+		if !ok {
+			return CreateOutput{}, errors.New("no attestationUrl output found")
+		}
+		attestationURL, ok = attestationURLOutput.Value.(string)
+		if !ok {
+			return CreateOutput{}, errors.New("invalid type in attestationUrl output: not a string")
+		}
+	}
```
daniel-weisse, line 141:
```suggestion
	var attestationURL string
	if attestationURLOutput, ok := tfState.Values.Outputs["attestationURL"]; 
		attestationURL = attestationURLOutput
	}
```
This would allow us to once again remove the CSP variable from `CreateCluster`

msanft:
Just created 2 clusters (1 with maa, 1 without) again with the latest changes, everything works.

internal/constants/constants.go
```diff
@@ -116,6 +116,11 @@ const (
 	IDKeyDigestFilename = "idkeydigests"
 	// EnforceIDKeyDigestFilename is the name of the file configuring whether idkeydigest is enforced or not.
 	EnforceIDKeyDigestFilename = "enforceIdKeyDigest"
+	// AzureCVM is the name of the file indicating whether the cluster is expected to run on CVMs or not.
+	AzureCVM = "azureCVM"
```
thomasten, line 120:
```suggestion
```

cli/internal/cloudcmd/create.go
```diff
@@ -215,14 +222,81 @@ func (c *Creator) createAzure(ctx context.Context, cl terraformClient, config *c
...
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("updating attestation policy: unexpected status code: %s", resp.Status)
+	}
+
+	return nil
+}
+
+// encodeAttestationPolicy encodes the base64-encoded attestation policy in the JWS format specified here:
+// https://learn.microsoft.com/en-us/azure/attestation/author-sign-policy#creating-the-policy-file-in-json-web-signature-format
+func (p policyPatcher) encodeAttestationPolicy(encodedPolicy string) string {
```
thomasten, line 290:
```suggestion
func (p policyPatcher) encodeAttestationPolicy() string {
	const policy = `
                version= 1.0;
                authorizationrules
                {
                    [type=="x-ms-azurevm-default-securebootkeysvalidated", value==false] => deny();

                    [type=="x-ms-azurevm-debuggersdisabled", value==false] => deny();

                    // The line below was edited by the Constellation CLI. Do not edit manually.
                    //[type=="secureboot", value==false] => deny();

                    [type=="x-ms-azurevm-signingdisabled", value==false] => deny();

                    [type=="x-ms-azurevm-dbvalidated", value==false] => deny();

                    [type=="x-ms-azurevm-dbxvalidated", value==false] => deny();

                    => permit();
                };
                issuancerules
                {
                };`

	encodedPolicy := base64.RawURLEncoding.EncodeToString([]byte(policy))
```

msanft: > The issue with the policy encoding was a whitespace issue. I wasn't able to format the policy string so that the whitespace matched up with the one needed(?) by Azure? I tested some ways using backticks but didn't get it to work that way. (500 status code)

I think the 500 error was caused by something else, as it works with the example you suggested. It doesn't match up with the original encoded policy being set by Azure (due to the whitespace, I believe?) but I don't think this is a problem

internal/constants/constants.go
```diff
@@ -116,7 +116,6 @@ const (
 	IDKeyDigestFilename = "idkeydigests"
 	// EnforceIDKeyDigestFilename is the name of the file configuring whether idkeydigest is enforced or not.
 	EnforceIDKeyDigestFilename = "enforceIdKeyDigest"
-
```
thomasten, line 119: nit: revert unintentional change

thomasten (approved):

PR #1380, opened 2023-03-08
image: print RTMR state upon boot (TDX)

Nirusu:
### Proposed change(s)
- Create a new tool called rtmr-reader which prints the MRTD/initial RTMR state upon boot
- Only print PCRs when a TPM is present 

The format strings are weird because the output is supposed to mimic tpm2_tools (since they don't seem to use \t?).
Maybe could be simplified by just using spaces in strings.

Maybe later we can also only print either PCRs or RTMRs depending on what is set on the kernel command line.

Still misses CI integration to build the tool (due to GOPRIVATE). Draft for now?


Nirusu:
Will rebase this later to use the new measurement reader from #1381, but we need to rebase the `tdx-attest` branch first to current main.

PR #1384, opened 2023-03-09
docs: minor adjustements to the release docs

derpsteb:
### Proposed change(s)

- write down advice on some uncertainties I had during the release

### Checklist

- [x] Link to Milestone


dev-docs/workflows/release.md
```diff
@@ -23,8 +23,9 @@ Releases should be performed using [the automated release pipeline](https://gith
     gh workflow run release.yml --ref release/v$minor -F version=$ver -F kind=patch
    ```
 3. wait for the pipeline to finish
-4. look over the autogenerated draft release and publish it
-
+4. while in editing mode for the release, clear the textbox, select the last patch release for the current release branch and click "Generate release notes".
```
thomasten, line 26: nit: instead of "clear the textbox", should the pipepline not generate it in the first place?

malt3: Github should auto detect the last and current release tags and autogenerate release notes. The detection of the last and current tags does not work properly. We can make the last release tag an input for the release pipeline. This way we don't have to let github guess.

derpsteb: True, however, as Malte said, the initial changelog is from 0.0.0 to "current". Will go back and change the description once we implement the "from"-version input to the pipeline.

PR #1390, opened 2023-03-09
measurement-reader: Add support for TDX

Nirusu:
### Proposed change(s)

- Add support for TDX measurements in the measurement-reader

In the end I did decide to unify the sort logic of both TPM and TDX measurements just to save on redunant test code.
I also to make `MeasurementType` an internal type here. The alternative would be to detect on length of one measurement or on the number of PCRs, but the first one is always kinda ugly and the last one would make writing the test code very annoying. Though the condition theoretically could just be:
```golang
var attestationType AttestationType
if len(m) == 5 {
  AttestationType = TDX
} else {
  AttestationType = TPM
}
```
Though I still prefer just passing this directly to clean up the function from more if/else code.

On a side-note: `measurements.WithAllBytes` technically only returns 32 byte values, which again would make the length check also currently not working correctly for the test code. I think I will refactor this tomorrow, so this is a draft PR for now.


Nirusu:
Last two commits add consts + make adjustments to `WithAllBytes` to take a length as input and return measurements of arbitrary length (though we only expect to use 32 or 48 bytes). Opening this for review.

daniel-weisse (approved):

PR #1377, opened 2023-03-08
ci: upload cli version list

msanft:
### Proposed change(s)

- Upload a version list of CLI versions to the versionsapi as a CI action
- [Successful run](https://github.com/edgelesssys/constellation/actions/runs/4363889664/jobs/7631996913)

### Checklist

- [x] Add labels (e.g., for changelog category)
- [x] Link to Milestone

PR #1218, opened 2023-02-19
ci: add cli k8s compatibility table artifact upload to ci

msanft:
### Proposed change(s)

- Add uploading of CLI / K8s version compatibility information to the CI pipeline to satisfy the API specified in [this RFC](https://github.com/edgelesssys/constellation/blob/main/rfc/cli-api.md)
- [Successful run](https://github.com/edgelesssys/constellation/actions)

### Additional info
- A refactoring might be done at some point to not have the cli info upload in the image build pipeline for a better structure, but I considered it as not being worth it at the moment

### Checklist

- [x] Add labels (e.g., for changelog category)
- [x] Link to Milestone

PR #1394, opened 2023-03-10
cli: show available cli upgrades on upgrade check command

msanft:
### Proposed change(s)

- Add listing of valid CLI version upgrades with the `upgrade check` command
  -  Upgrade is considered valid if 1) compatible with current K8s version and 2) semver is greater and has at most 1 minor version drift
- Uploading of CLI version lists and compatibility information was done in #1218 & #1377 

### Additional info
- Possible follow-ups
  - Upgrade check command is in (urgent) need of a refactoring. Files are very large. See [this AB ticket.](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/2957/)
  - Custom version type should be used eventually. See [this AB ticket](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/2956/).

Output of `constellation upgrade check` looks like:
```
The following updates are available with this CLI:
  Kubernetes: v1.25.7 --> v1.26.2
  CLI: v2.5.0 --> v2.6.0-pre.0.20230308113257-643afd38b347 v2.6.0-pre.0.20230309101241-15faa64ef6cc
  ```

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/upgradecheck.go
+++ b/cli/internal/cmd/upgradecheck.go
@@ -81,7 +81,7 @@ func runUpgradeCheck(cmd *cobra.Command, _ []string) error {
 			flags:          flags,
 			cliVersion:     compatibility.EnsurePrefixV(constants.VersionInfo()),
 			log:            log,
-			verFetcher:     fetcher.NewFetcher(),
+			versionsapi:    fetcher.NewFetcher(),
 		},
 		log: log,
 	}
@@ -251,7 +251,7 @@ type versionCollector struct {
 	client         *http.Client
 	rekor          rekorVerifier
 	flags          upgradeCheckFlags
-	verFetcher     versionFetcher
+	versionsapi    versionFetcher
 	cliVersion     string
 	log            debugLog
 }
@@ -453,7 +453,7 @@ func (v *versionUpgrade) buildString() (string, error) {
 		return result.String(), nil
 	}
 
-	result.WriteString("No upgrades available with this CLI.\nNewer versions may be available at: https://github.com/edgelesssys/constellation/releases\n")
+	result.WriteString("No upgrades available with this CLI.\n")
 
 	return result.String(), nil
 }
@@ -567,7 +567,7 @@ func (v *versionCollector) newCompatibleCLIVersions(ctx context.Context, current
 		Base:        "v2",
 		Kind:        versionsapi.VersionKindCLI,
 	}
-	minorList, err := v.verFetcher.FetchVersionList(ctx, list)
+	minorList, err := v.versionsapi.FetchVersionList(ctx, list)
 	if err != nil {
 		return nil, fmt.Errorf("listing minor versions: %w", err)
 	}
@@ -585,7 +585,7 @@ func (v *versionCollector) newCompatibleCLIVersions(ctx context.Context, current
 			Base:        version,
 			Kind:        versionsapi.VersionKindCLI,
 		}
-		patchList, err := v.verFetcher.FetchVersionList(ctx, list)
+		patchList, err := v.versionsapi.FetchVersionList(ctx, list)
 		if err != nil {
 			return nil, fmt.Errorf("listing minor versions: %w", err)
 		}
@@ -604,7 +604,7 @@ func (v *versionCollector) newCompatibleCLIVersions(ctx context.Context, current
 			Stream:  v.flags.stream,
 			Version: version,
 		}
-		info, err := v.verFetcher.FetchCLIInfo(ctx, req)
+		info, err := v.versionsapi.FetchCLIInfo(ctx, req)
 		if err != nil {
 			return nil, fmt.Errorf("fetching CLI info: %w", err)
 		}
--- a/cli/internal/cmd/upgradecheck.go
+++ b/cli/internal/cmd/upgradecheck.go
@@ -145,7 +145,7 @@ func (u *upgradeCheckCmd) upgradeCheck(cmd *cobra.Command, fileHandler file.Hand
 		return err
 	}
 
-	supported, err := u.collect.supportedVersions(cmd.Context(), csp, current.image, current.k8s)
+	supported, err := u.collect.supportedVersions(cmd.Context(), current.image, current.k8s)
 	if err != nil {
 		return err
 	}
@@ -236,10 +236,10 @@ func filterK8sUpgrades(currentVersion string, newVersions []string) []string {
 
 type collector interface {
 	currentVersions(ctx context.Context) (currentVersionInfo, error)
-	supportedVersions(ctx context.Context, csp cloudprovider.Provider, version, currentK8sVersion string) (supportedVersionInfo, error)
-	newImages(ctx context.Context, version string, csp cloudprovider.Provider) ([]versionsapi.Version, error)
+	supportedVersions(ctx context.Context, version, currentK8sVersion string) (supportedVersionInfo, error)
+	newImages(ctx context.Context, version string) ([]versionsapi.Version, error)
 	newMeasurements(ctx context.Context, csp cloudprovider.Provider, images []versionsapi.Version) (map[string]measurements.M, error)
-	newerVersions(ctx context.Context, currentVersion string, allowedVersions []string) ([]versionsapi.Version, error)
+	newerVersions(ctx context.Context, allowedVersions []string) ([]versionsapi.Version, error)
 	newCompatibleCLIVersions(ctx context.Context, currentKubernetesVersion string) ([]string, error)
 }
 
@@ -311,13 +311,13 @@ type supportedVersionInfo struct {
 }
 
 // supportedVersions returns slices of supported versions.
-func (v *versionCollector) supportedVersions(ctx context.Context, csp cloudprovider.Provider, version, currentK8sVersion string) (supportedVersionInfo, error) {
+func (v *versionCollector) supportedVersions(ctx context.Context, version, currentK8sVersion string) (supportedVersionInfo, error) {
 	k8sVersions := versions.SupportedK8sVersions()
 	serviceVersion, err := helm.AvailableServiceVersions()
 	if err != nil {
 		return supportedVersionInfo{}, fmt.Errorf("loading service versions: %w", err)
 	}
-	imageVersions, err := v.newImages(ctx, version, csp)
+	imageVersions, err := v.newImages(ctx, version)
 	if err != nil {
 		return supportedVersionInfo{}, fmt.Errorf("loading image versions: %w", err)
 	}
--- a/cli/internal/cmd/upgradecheck.go
+++ b/cli/internal/cmd/upgradecheck.go
@@ -586,7 +586,7 @@ func (v *versionCollector) newCLIVersions(ctx context.Context, currentKubernetes
 	}
 	minorList, err := v.versionsapi.FetchVersionList(ctx, list)
 	if err != nil {
-		return nil, fmt.Errorf("listing minor versions: %w", err)
+		return nil, fmt.Errorf("listing major versions: %w", err)
 	}
 
 	var patchVersions []string
--- a/cli/internal/cmd/upgradecheck.go
+++ b/cli/internal/cmd/upgradecheck.go
@@ -139,42 +139,49 @@ func (u *upgradeCheckCmd) upgradeCheck(cmd *cobra.Command, fileHandler file.Hand
 	csp := conf.GetProvider()
 	u.log.Debugf("Using provider %s", csp.String())
 
-	currentServices, currentImage, currentK8s, err := u.collect.currentVersions(cmd.Context())
+	current, err := u.collect.currentVersions(cmd.Context())
 	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: cli/internal/cmd/upgradecheck.go [hunk 1]
```diff
@@ -519,6 +576,61 @@ func getCompatibleImageMeasurements(ctx context.Context, writer io.Writer, clien
...
+		Base:        "v2",
+		Kind:        versionsapi.VersionKindCLI,
+	}
+	minorList, err := versionFetcher.FetchVersionList(ctx, list)
+	if err != nil {
+		return nil, fmt.Errorf("listing minor versions: %w", err)
+	}
+
+	var patchVersions []string
+	for _, version := range minorList.Versions {
```
derpsteb, line 596: No need to check all minor versions, as only the current and upcoming minor versions are relevant.

cli/internal/cmd/upgradecheck.go
```diff
@@ -519,6 +576,61 @@ func getCompatibleImageMeasurements(ctx context.Context, writer io.Writer, clien
...
+		list := versionsapi.List{
+			Ref:         "feat-upload-cli-version-list",
+			Stream:      "debug",
```
derpsteb, line 599: This should respect the flags that can be provided to `upgrade check`.

msanft: Oh, of course🤦‍♂️

cli/internal/cmd/upgradecheck.go
```diff
@@ -519,6 +576,61 @@ func getCompatibleImageMeasurements(ctx context.Context, writer io.Writer, clien
...
+		}
+		patchVersions = append(patchVersions, patchList.Versions...)
+	}
+
+	// filter out invalid upgrades and versions which are not compatible with the current Kubernetes version
+	var compatibleVersions []string
+	for _, version := range patchVersions {
+		req := versionsapi.CLIInfo{
+			Ref:     "feat-upload-cli-version-list",
+			Stream:  "debug",
```
derpsteb, line 616: same as above.

cli/internal/cmd/upgradecheck.go
```diff
@@ -258,43 +281,71 @@ func (v *versionCollector) newMeasurementes(ctx context.Context, csp cloudprovid
...
-	return serviceVersion, imageVersion, k8sVersion, nil
+	return currentVersionInfo{
+		service: serviceVersion,
+		image:   imageVersion,
+		k8s:     k8sVersion,
+		cli:     v.cliVersion,
+	}, nil
+}
+
+type supportedVersionInfo struct {
```
derpsteb, line 320: good idea! :see_no_evil:

cli/internal/cmd/upgradecheck.go
```diff
@@ -227,11 +236,25 @@ func filterK8sUpgrades(currentVersion string, newVersions []string) []string {
 	return result
 }
 
+// filterCLIUpgrades filters out CLI versions that are not valid upgrades.
+func filterCLIUpgrades(currentVersion string, newVersions []string) []string {
+	result := []string{}
+	for i := range newVersions {
+		if err := compatibility.IsValidUpgrade(currentVersion, newVersions[i]); err != nil {
+			fmt.Println(err)
```
derpsteb, line 244: User should not see all those incompatible upgrades.
Maybe log them to debug output.

cli/internal/cmd/upgradecheck.go
```diff
@@ -139,42 +139,49 @@ func (u *upgradeCheckCmd) upgradeCheck(cmd *cobra.Command, fileHandler file.Hand
...
-	newServices := supportedServices
-	if err := compatibility.IsValidUpgrade(currentServices, supportedServices); err != nil {
+	newServices := supported.service
+	if err := compatibility.IsValidUpgrade(current.service, supported.service); err != nil {
 		newServices = ""
 	}
 
-	newKubernetes := filterK8sUpgrades(currentK8s, supportedK8s)
+	newKubernetes := filterK8sUpgrades(current.k8s, supported.k8s)
 	sort.Strings(newKubernetes)
```
derpsteb, line 165: This should use [semver.Sort](https://pkg.go.dev/golang.org/x/mod@v0.7.0/semver#Sort) :thinking:

cli/internal/cmd/upgradecheck.go
```diff
@@ -139,42 +139,49 @@ func (u *upgradeCheckCmd) upgradeCheck(cmd *cobra.Command, fileHandler file.Hand
...
-	supportedImages = filterImageUpgrades(currentImage, supportedImages)
-	newImages, err := u.collect.newMeasurementes(cmd.Context(), csp, supportedImages)
+	supported.image = filterImageUpgrades(current.image, supported.image)
+	newImages, err := u.collect.newMeasurements(cmd.Context(), csp, supported.image)
 	if err != nil {
 		return err
 	}
 
+	newCLI := filterCLIUpgrades(current.cli, supported.cli)
+	sort.Strings(newCLI)
```
derpsteb, line 174: Same as above.

derpsteb:
Could you also extend the unittests for `TestBuildString`, `newCompatibleCLIVersions` and possibly `TestUpgradeCheck`?

LGTM apart from that :)

msanft:
> Could you also extend the unittests for TestBuildString, newCompatibleCLIVersions and possibly TestUpgradeCheck?

Sure, I totally forgot about the tests here. How would you go about testing `newCompatibleCLIVersions`? Should I put the internal calls to versionsapi behind an interface?

derpsteb:
Yes, I think that is how I would do it :). Make `versionFetcher` a property of `versionCollector`. Then you can stub the calls to the API.

cli/internal/cmd/upgradecheck.go
```diff
@@ -81,6 +81,7 @@ func runUpgradeCheck(cmd *cobra.Command, args []string) error {
 			flags:          flags,
 			cliVersion:     compatibility.EnsurePrefixV(constants.VersionInfo()),
 			log:            log,
+			verFetcher:     fetcher.NewFetcher(),
```
derpsteb, line 84:
```suggestion
			versionsapi:     fetcher.NewFetcher(),
```

Maybe? `versionCollector.versionsapi.FetchVersionList(ctx, list)` reads a bit more clear to me.

cli/internal/cmd/upgradecheck_test.go
```diff
@@ -42,20 +42,24 @@ func TestBuildString(t *testing.T) {
...
 				newImages:         map[string]measurements.M{},
 				newKubernetes:     []string{},
+				newCLI:            []string{},
 				currentServices:   "v2.5.0",
 				currentImage:      "v2.5.0",
 				currentKubernetes: "v1.25.6",
+				currentCLI:        "v2.5.0",
 			},
 			expected: "No upgrades available with this CLI.\nNewer versions may be available at: https://github.com/edgelesssys/constellation/releases\n",
 		},
```
derpsteb, line 65: What about a testcase that shows that new CLI versions are displayed as expected?

cli/internal/cmd/upgradecheck.go
```diff
@@ -379,6 +418,10 @@ func (v *versionUpgrade) buildString() (string, error) {
 		upgradeMsg.WriteString(fmt.Sprintf("  Kubernetes: %s --> %s\n", v.currentKubernetes, strings.Join(v.newKubernetes, " ")))
 	}
 
+	if len(v.newCLI) > 0 {
+		upgradeMsg.WriteString(fmt.Sprintf("  CLI: %s --> %s\n", v.currentCLI, strings.Join(v.newCLI, " ")))
+	}
+
```
derpsteb, line 424: If we print the new CLI versions like this we need to change the "No updates" [message](https://github.com/edgelesssys/constellation/pull/1394/files#diff-06df0ccd1ff71379f21375eff055cdd48759f2d6b56366e06e18fb858940a835R456). The message should not suggest to check the release page anymore and instead just state that the user is up-to-date.

cli/internal/cmd/upgradecheck_test.go
msanft: I think this is what the test case "update everything" does? I will add test cases specific for version upgrades of CLI / K8s / .. only though

derpsteb:
Summary of our conversation on the phone:
it probably makes more sense to only point the user to newer CLI versions in case the current CLI does not offer any more upgrades. So these are the required output formats.

Current CLI offers upgrades:
```bash
The following updates are available with this CLI:
 Kubernetes: v1.24.5 --> v1.24.12 v1.25.6
 Services: v2.4.0 --> v2.5.0
 Images: v2.4.0 --> v2.5.0
 Includes these measurements:
  4:
    expected: \"1234123412341234123412341234123412341234123412341234123412341234\"
    warnOnly: false
  8:
    expected: \"0000000000000000000000000000000000000000000000000000000000000000\"
    warnOnly: false
  9:
    expected: \"1234123412341234123412341234123412341234123412341234123412341234\"
    warnOnly: false
  11:
    expected: \"0000000000000000000000000000000000000000000000000000000000000000\"
    warnOnly: false
  12:
    expected: \"1234123412341234123412341234123412341234123412341234123412341234\"
    warnOnly: false
  13:
    expected: \"0000000000000000000000000000000000000000000000000000000000000000\"
    warnOnly: false
  15:
    expected: \"0000000000000000000000000000000000000000000000000000000000000000\"
    warnOnly: false
```

Current CLI offers no upgrades (apart from k8s), but newer CLIs are available:
```
The following updates are available with this CLI:
 Kubernetes: v1.24.5 --> v1.24.12 v1.25.6
Newer CLI versions that are compatible with your cluster are: v2.5.0.
```

There are newer CLIs, but none are compatible with the current k8s version:
```
The following updates are available with this CLI:
 Kubernetes: v1.24.5 --> v1.24.12 v1.25.6
There are newer CLIs available, however, you need to upgrade your cluster's Kubernetes version first.
```

No newer versions at all:
``` 
You are up to date.
```

cli/internal/cmd/upgradecheck.go
```diff
@@ -510,6 +570,81 @@ func getCompatibleImageMeasurements(ctx context.Context, writer io.Writer, clien
...
+	FetchCLIInfo(ctx context.Context, cliInfo versionsapi.CLIInfo) (versionsapi.CLIInfo, error)
+}
+
+// newCLIVersions returns a list of versions of the CLI which are a valid upgrade.
+func (v *versionCollector) newCLIVersions(ctx context.Context, currentKubernetesVersion string) ([]string, error) {
+	list := versionsapi.List{
+		Ref:         v.flags.ref,
+		Stream:      v.flags.stream,
+		Granularity: versionsapi.GranularityMajor,
+		Base:        "v2",
```
derpsteb, line 584: Probably not a problem in the near future, but this should likely be based on the current major version.

cli/internal/cmd/upgradecheck.go
```diff
@@ -510,6 +570,81 @@ func getCompatibleImageMeasurements(ctx context.Context, writer io.Writer, clien
...
+		Kind:        versionsapi.VersionKindCLI,
+	}
+	minorList, err := v.versionsapi.FetchVersionList(ctx, list)
+	if err != nil {
+		return nil, fmt.Errorf("listing minor versions: %w", err)
```
derpsteb, line 589:
```suggestion
		return nil, fmt.Errorf("listing major versions: %w", err)
```

cli/internal/cmd/upgradecheck.go
```diff
@@ -510,6 +570,81 @@ func getCompatibleImageMeasurements(ctx context.Context, writer io.Writer, clien
...
+		list := versionsapi.List{
+			Ref:         v.flags.ref,
+			Stream:      v.flags.stream,
+			Granularity: versionsapi.GranularityMinor,
+			Base:        version,
+			Kind:        versionsapi.VersionKindCLI,
+		}
+		patchList, err := v.versionsapi.FetchVersionList(ctx, list)
+		if err != nil {
+			return nil, fmt.Errorf("listing minor versions: %w", err)
```
derpsteb, line 607:
```suggestion
			return nil, fmt.Errorf("listing minor versions for major version %s: %w", version, err)
```

derpsteb:
The output rework looks good, but this might need some further manual testing? Currently breaks for me due to missing cli.json in the bucket, when listing the major versions.

cli/internal/cmd/upgradecheck.go
derpsteb, line 584: Probably not a problem in the near future, but this should likely be based on the current major version.
```suggestion
func (v *versionCollector) newCLIVersions(ctx context.Context) ([]string, error) {
	cliVersion, err := conSemver.New(constants.VersionInfo())
	if err != nil {
		return nil, fmt.Errorf("parsing current CLI version: %w", err)
	}
	list := versionsapi.List{
		Ref:         v.flags.ref,
		Stream:      v.flags.stream,
		Granularity: versionsapi.GranularityMajor,
		Base:        fmt.Sprintf("v%d", cliVersion.Major),
```

derpsteb (approved):

PR #1403, opened 2023-03-10
bootstrapper: try to mitigate timeout issues

Nirusu:
### Proposed change(s)
- Split FixCilium into WaitForCilium and FixCilium
- Put both directly after Cilium installation so we have a distinct error when Cilium doesn't come up instead of having the same cert-manager context deadline exceeded error
- Set a timeout for 20 minutes for WaitForCilium (pulling can sometimes be slow with ghcr.io - in local testing the worst case was ~16 minutes occasionally) -> This should likely be reduced after the next Cilium update when/if we switch repositories that have more consistent performance.
- Set a timeout for 10 minutes for Helm install instead of 5 minutes (cert-manager can sometimes take longer)
- Track duration of Cilium and cert-manager install (since both can stall)
- Minor fixes

Generally, the issue is the following:
- We try to install too many things at once almost immediately the cluster is up, so when "cert-manager" fails, we have ~10 Pods that need to be created, need to pull images, scheduled etc.
- The node can still be tainted as uninitialized for a bit when the cluster recently came up
- Repositories can be slow sometimes (e.g. hit a bad PoP from a CDN - definitely happens with ghcr.io)
- Sometimes Kubernetes takes time to schedule things

Often it's not things being completely broken - they are just occasionally slow. Given that we depend on external resources, this is not too unsurprising (though certainly annoying).

I hope we can use OpenSearch to maybe capture the duration of successful installs and derive good timeouts from them. 

I choose a middle ground with 20 minutes and 10 minutes. We could also set them higher in case they still make trouble and/or we want to run more statistics on them. 20 minutes for Cilium is already graceful (only doing this because of the ghcr.io issue), 10 minutes for Helm might still be a little bit too short for worst-case scenarios.

In the later run though, however, I wish we can refactor the bootstrapper to handle the Helm installations after we return the kubeconfig to the client. Often things just take longer. Or they can be fixed manually.

Maybe we can also disable the cert-manager API health check? Not sure how that works though with the dependencies on the operator. Maybe @derpsteb can comment on that?

### DIFF
--- a/.github/actions/constellation_create/action.yml
+++ b/.github/actions/constellation_create/action.yml
@@ -194,6 +194,7 @@ runs:
         echo "KUBECONFIG=$(pwd)/constellation-admin.conf" >> $GITHUB_OUTPUT
         echo "MASTERSECRET=$(pwd)/constellation-mastersecret.json" >> $GITHUB_OUTPUT
 
+    # TODO(nirusu): Temporarily increase kubectl wait timeout here - might be related to all the Cilium / cert-manager issues?
     - name: Wait for nodes to join and become ready
       shell: bash
       env:
@@ -214,7 +215,13 @@ runs:
             exit 1
         fi
         echo "$(kubectl get nodes -o json | jq '.items | length')/"${NODES_COUNT}" nodes have joined"
-        kubectl wait --for=condition=ready --all nodes --timeout=10m
+        kubectl wait --for=condition=ready --all nodes --timeout=20m
+        kubectlErrorCode=$?
+        if [[ "${kubectlErrorCode}" -ne 0 ]]; then
+          kubectl get pods -n kube-system
+          kubectl get events -n kube-system
+          exit "${kubectlErrorCode}"
+        fi
         echo "::endgroup::"
 
     - name: Download boot logs
--- a/.github/actions/constellation_create/action.yml
+++ b/.github/actions/constellation_create/action.yml
@@ -215,12 +215,12 @@ runs:
             exit 1
         fi
         echo "$(kubectl get nodes -o json | jq '.items | length')/"${NODES_COUNT}" nodes have joined"
-        kubectl wait --for=condition=ready --all nodes --timeout=20m
-        kubectlErrorCode=$?
-        if [[ "${kubectlErrorCode}" -ne 0 ]]; then
+        if ! kubectl wait --for=condition=ready --all nodes --timeout=20m; then
           kubectl get pods -n kube-system
           kubectl get events -n kube-system
-          exit "${kubectlErrorCode}"
+          echo "::error::kubectl wait timed out before all nodes became ready"
+          echo "::endgroup::"
+          exit 1
         fi
         echo "::endgroup::"
 
--- a/bootstrapper/internal/kubernetes/kubernetes.go
+++ b/bootstrapper/internal/kubernetes/kubernetes.go
@@ -248,12 +248,10 @@ func (k *KubeWrapper) InitCluster(
 
 	// cert-manager is necessary for our operator deployments.
 	log.Infof("Installing cert-manager")
-	timeToStartWaiting = time.Now()
 	if err = k.helmClient.InstallCertManager(ctx, helmReleases.CertManager); err != nil {
 		return nil, fmt.Errorf("installing cert-manager: %w", err)
 	}
-	timeUntilFinishedWaiting = time.Since(timeToStartWaiting)
-	log.With(zap.Duration("duration", timeUntilFinishedWaiting)).Infof("cert-manager installation finished")
+
 	operatorVals, err := k.setupOperatorVals(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("setting up operator vals: %w", err)
--- a/bootstrapper/internal/kubernetes/kubernetes.go
+++ b/bootstrapper/internal/kubernetes/kubernetes.go
@@ -200,7 +200,7 @@ func (k *KubeWrapper) InitCluster(
 		return nil, fmt.Errorf("waiting for Cilium to become healthy: %w", err)
 	}
 	timeUntilFinishedWaiting := time.Since(timeToStartWaiting)
-	log.Infof("Cilium took %s to become healthy", timeUntilFinishedWaiting.Round(time.Second).String())
+	log.With(zap.Duration("duration", timeUntilFinishedWaiting)).Infof("Cilium became healthy")
 
 	log.Infof("Restart Cilium")
 	if err := k.clusterUtil.FixCilium(ctx); err != nil {
@@ -253,8 +253,7 @@ func (k *KubeWrapper) InitCluster(
 		return nil, fmt.Errorf("installing cert-manager: %w", err)
 	}
 	timeUntilFinishedWaiting = time.Since(timeToStartWaiting)
-	log.Infof("cert-manager took %s to be fully installed", timeUntilFinishedWaiting.Round(time.Second).String())
-
+	log.With(zap.Duration("duration", timeUntilFinishedWaiting)).Infof("cert-manager installation finished")
 	operatorVals, err := k.setupOperatorVals(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("setting up operator vals: %w", err)
--- a/bootstrapper/internal/kubernetes/kubernetes.go
+++ b/bootstrapper/internal/kubernetes/kubernetes.go
@@ -189,6 +189,26 @@ func (k *KubeWrapper) InitCluster(
 		return nil, fmt.Errorf("installing pod network: %w", err)
 	}
 
+	// TODO: The timeout here is high as ghcr.io can be slow sometimes. Reduce this later when we move the repository.
+	// Also remove the logging later.
+	log.Infof("Waiting for Cilium to become healthy")
+	timeToStartWaiting := time.Now()
+	// TODO(Nirusu): Reduce the timeout when we switched the package repository - this is only this high because I once
+	// saw polling times of ~16 minutes when hitting a slow PoP from Fastly (GitHub's / ghcr.io CDN).
+	waitCtx, cancel = context.WithTimeout(ctx, 20*time.Minute)
+	defer cancel()
+	if err := k.clusterUtil.WaitForCilium(waitCtx, log); err != nil {
+		return nil, fmt.Errorf("waiting for Cilium to become healthy: %w", err)
+	}
+	timeUntilFinishedWaiting := time.Since(timeToStartWaiting)
+	log.Infof("Cilium took %s to become healthy", timeUntilFinishedWaiting.Round(time.Second).String())
+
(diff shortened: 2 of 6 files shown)


bootstrapper/internal/kubernetes/kubernetes.go
```diff
@@ -218,30 +237,34 @@ func (k *KubeWrapper) InitCluster(
...
 
 	// cert-manager is necessary for our operator deployments.
+	log.Infof("Installing cert-manager")
+	timeToStartWaiting = time.Now()
 	if err = k.helmClient.InstallCertManager(ctx, helmReleases.CertManager); err != nil {
 		return nil, fmt.Errorf("installing cert-manager: %w", err)
 	}
-
+	timeUntilFinishedWaiting = time.Since(timeToStartWaiting)
+	log.With(zap.Int("duration", int(timeUntilFinishedWaiting.Seconds()))).Infof("cert-manager installation finished")
```
daniel-weisse, line 257:
```suggestion
	log.With(zap.Duration("duration", timeUntilFinishedWaiting)).Infof("cert-manager installation finished")
```

daniel-weisse (commented):
If we are already logging the install time for cert-manager, why not also log the time for all the other helm deployments?

Nirusu:
Mainly because only Cilium and cert-manager are the only blocking ones since they require a healthy state. The other installs go through quickly - they don't wait for their deployments to be fully functional.

But I could refactor the cert-manager duration into the general Helm install code. Cilium needs extra treatment nevertheless.

.github/actions/constellation_create/action.yml
```diff
@@ -214,7 +215,13 @@ runs:
             exit 1
         fi
         echo "$(kubectl get nodes -o json | jq '.items | length')/"${NODES_COUNT}" nodes have joined"
-        kubectl wait --for=condition=ready --all nodes --timeout=10m
+        kubectl wait --for=condition=ready --all nodes --timeout=20m
+        kubectlErrorCode=$?
```
katexochen, line 219: This won't work due to the shell flags set in the CI. If the above kubectl fails, the step ends. You could use something like

```bash
if ! kubectl ... ; then
```

.github/actions/constellation_create/action.yml
```diff
@@ -214,7 +215,13 @@ runs:
...
+        if [[ "${kubectlErrorCode}" -ne 0 ]]; then
+          kubectl get pods -n kube-system
+          kubectl get events -n kube-system
+          exit "${kubectlErrorCode}"
```
katexochen, line 223:
```suggestion
          echo "::error::Some error description"
          exit "${kubectlErrorCode}"
```
Please write a descriptive error message.

.github/actions/constellation_create/action.yml
```diff
@@ -194,6 +194,7 @@ runs:
         echo "KUBECONFIG=$(pwd)/constellation-admin.conf" >> $GITHUB_OUTPUT
         echo "MASTERSECRET=$(pwd)/constellation-mastersecret.json" >> $GITHUB_OUTPUT
 
+    # TODO: Temporarily increase kubectl wait timeout here - might be related to all the Cilium / cert-manager issues?
```
katexochen, line 197:
```suggestion
    # TODO(nirusu): Temporarily increase kubectl wait timeout here - might be related to all the Cilium / cert-manager issues?
```

katexochen (changes requested):

bootstrapper/internal/kubernetes/kubernetes.go
Nirusu: Changed to zap.Duration and also made it log for every Helm install (outside the retry loop so we get the total duration of a _successful_ install).

Wasn't sure if it makes sense to log duration for non-successful installations.

3u13r (approved):
LGTM

bootstrapper/internal/kubernetes/kubernetes.go
```diff
@@ -186,10 +186,29 @@ func (k *KubeWrapper) InitCluster(
...
+	if err := k.clusterUtil.WaitForCilium(waitCtx, log); err != nil {
+		return nil, fmt.Errorf("waiting for Cilium to become healthy: %w", err)
+	}
+	timeUntilFinishedWaiting := time.Since(timeToStartWaiting)
+	log.With(zap.Duration("duration", timeUntilFinishedWaiting)).Infof("Cilium became healthy")
+
+	log.Infof("Restart Cilium")
+	if err := k.clusterUtil.FixCilium(ctx); err != nil {
+		log.With(zap.Error(err), "FixCilium failed", zap.Error(err))
+		// Continue and don't throw an error here - things might be okay.
```
derpsteb, line 209: Could you explain why things might still be ok if we see an error? And why we throw an error even if we want to continue in that case?

derpsteb:
I really look forward to having this bug gone :D 
Thanks for investigating the issue further!

What is the reason for moving the `WaitForCilium` code into it's own function?

derpsteb (approved):

bootstrapper/internal/kubernetes/kubernetes.go
```diff
@@ -186,10 +186,29 @@ func (k *KubeWrapper) InitCluster(
...
+	defer cancel()
+	if err := k.clusterUtil.WaitForCilium(waitCtx, log); err != nil {
+		return nil, fmt.Errorf("waiting for Cilium to become healthy: %w", err)
+	}
+	timeUntilFinishedWaiting := time.Since(timeToStartWaiting)
+	log.With(zap.Duration("duration", timeUntilFinishedWaiting)).Infof("Cilium became healthy")
+
+	log.Infof("Restart Cilium")
+	if err := k.clusterUtil.FixCilium(ctx); err != nil {
+		log.With(zap.Error(err), "FixCilium failed", zap.Error(err))
```
daniel-weisse, line 208:
```suggestion
		log.With(zap.Error(err)).Errorf("FixCilium failed")
```

bootstrapper/internal/kubernetes/kubernetes.go
```diff
@@ -298,7 +319,16 @@ func (k *KubeWrapper) JoinCluster(ctx context.Context, args *kubeadm.BootstrapTo
...
 
-	k.clusterUtil.FixCilium(log)
+	log.Infof("Waiting for Cilium to become healthy")
+	if err := k.clusterUtil.WaitForCilium(context.Background(), log); err != nil {
+		return fmt.Errorf("waiting for Cilium to become healthy: %w", err)
+	}
+
+	log.Infof("Restart Cilium")
+	if err := k.clusterUtil.FixCilium(context.Background()); err != nil {
+		log.With(zap.Error(err), "FixCilium failed")
```
daniel-weisse, line 329:
```suggestion
		log.With(zap.Error(err).Errorf("FixCilium failed")
```

bootstrapper/internal/kubernetes/kubernetes.go
```diff
@@ -361,7 +391,17 @@ func (k *KubeWrapper) StartKubelet(log *logger.Logger) error {
...
```
daniel-weisse, line 401:
```suggestion
		log.With(zap.Error(err)).Errorf("FixCilium failed")
```

bootstrapper/internal/kubernetes/kubernetes.go
Nirusu: >Could you explain why things might still be ok if we see an error?
The `FixCilium` function only restarts Cilium the first time to work around some Pod2Pod networking bug. However, given that it seemed to mostly work fine in the old version where we only did this after we installed all other Helm charts, it will likely be fine if this fails (for whatever reason) and the rest of the init likely is still able to go through. 

Worst case if the Pod2Pod networking bug still exists but everything else was set up correctly, you just restart Cilium or the node after the init.

>And why we throw an error even if we want to continue in that case?

For the logs. The old versions passed the logger down, but why do that when we can just return the error and let the caller choose to handle this? Seems more generalized and straightforward to me.

Nirusu:
>I really look forward to having this bug gone :D

Unfortunately it won't be fully gone but it should be a bit better.

> What is the reason for moving the `WaitForCilium` code into it's own function?

No super important reason, mainly:

- to split the functionality here (since it had two logical components - waiting and restarting the Pod)
- have different contexts for waiting and killing (otherwise you *could* theoretically pass the wait but fail the rest in an unfortunate situation). Also only the first part needed the logger for warning. 
- Also helps to count the time this way.

Could also still be one function but I thought it is cleaner this way.

bootstrapper/internal/kubernetes/kubernetes.go
derpsteb: Speaking purely from a UX perspective: it seems less than ideal to me to force the user to go in and restart pods after we have "Successfully initiated" a cluster. If I understand you correctly the more acurate CLI output would be sth like: "We may have successfully initiated a cluster, but maybe it broke, so you better go back to check and fix those Cilium pods"  :D

If pod2pod networking is broken until `FixCilium` completed successfully, why not retry that function?

Nirusu: >If pod2pod networking is broken until FixCilium completed successfully, why not retry that function?

1. We don't know for sure we into the bug.
2. Because the logic itself rarely fails. Looking in OpenSearch, we have only ~40 fails and most of them are where the context deadline exceeded. All remaining issues, except one were "container is already in removing state" which should be okay I guess. The last one was "container is in starting state, can't be removed" which shouldn't have happened due to the wait, but yeah, this _very_ rare.

Let's assume we try. How often should we try? Should we block & wait? Should we do this in a separate goroutine? There's not even a guarantee we ran into the bug.

>If I understand you correctly the more acurate CLI output would be sth like: "We may have successfully initiated a cluster, but maybe it broke, so you better go back to check and fix those Cilium pods" :D

But it's not permanently broken. If we hard fail (which we also did not before, so it's not a regression but rather could introduce a regression), why would you force the user to recreate their whole infrastructure? 

Generally just printing a warning to the CLI would be fine here (and better than `terminate` and `create`, but we don't have that mechanism unfourtunaly and would require larger refactors to make the bootstrapped init process a multi-message gRPC.

daniel-weisse (approved):

Nirusu:
@katexochen Can you re-review please? It's stuck on change requested otherwise.
Will squash some commits together then and merge this and see how everything behaves in the e2e tests.

bootstrapper/internal/kubernetes/kubernetes.go
derpsteb: Oh. I didn't realize that FixCilium also didn't break init in case it failed. Nvm then :D

Alright, lets see where this goes.

Nirusu: To be honest I was surprised init even worked this well before despite `FixCilium` only being called at the end (since IIRC the cert-manager check is Pod2Pod networking) :D

katexochen (approved):

PR #1404, opened 2023-03-11
cli: split spinner into "Connecting" and "Initializing cluster" state for init (and prevent double initialization cases when an error is mistakenly retried)

Nirusu:
### Proposed change(s)
- Add "Connecting" spinner state for "constellation init"
- Show "Initializing cluster" only when our gRPC connection was successful 

This is based on @malt3 gRPC debug logging functions.
I figured we can also just use them for the normal user output :) 
Not sure if this is the best place to put it there, though since I am not sure if the state can theoretically change multiple times to READY. 
Hope I can get some feedback on that? Not super familiar with gRPC connection states and how this works in our init.
Otherwise it might make sense to make this a channel and let the caller handle this.

Second commit also adds a note that it can take a few minutes.

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/cli/internal/cmd/init.go
+++ b/cli/internal/cmd/init.go
@@ -65,8 +65,9 @@ func NewInitCmd() *cobra.Command {
 }
 
 type initCmd struct {
-	log    debugLog
-	merger configMerger
+	log     debugLog
+	merger  configMerger
+	spinner spinnerInterf
 }
 
 // runInitialize runs the initialize command.
@@ -90,13 +91,13 @@ func runInitialize(cmd *cobra.Command, args []string) error {
 	ctx, cancel := context.WithTimeout(cmd.Context(), time.Hour)
 	defer cancel()
 	cmd.SetContext(ctx)
-	i := &initCmd{log: log, merger: &kubeconfigMerger{log: log}}
-	return i.initialize(cmd, newDialer, fileHandler, license.NewClient(), spinner)
+	i := &initCmd{log: log, spinner: spinner, merger: &kubeconfigMerger{log: log}}
+	return i.initialize(cmd, newDialer, fileHandler, license.NewClient())
 }
 
 // initialize initializes a Constellation.
 func (i *initCmd) initialize(cmd *cobra.Command, newDialer func(validator *cloudcmd.Validator) *dialer.Dialer,
-	fileHandler file.Handler, quotaChecker license.QuotaChecker, spinner spinnerInterf,
+	fileHandler file.Handler, quotaChecker license.QuotaChecker,
 ) error {
 	flags, err := i.evalFlagArgs(cmd)
 	if err != nil {
@@ -160,7 +161,8 @@ func (i *initCmd) initialize(cmd *cobra.Command, newDialer func(validator *cloud
 	clusterName := conf.Name + "-" + idFile.UID
 	i.log.Debugf("Setting cluster name to %s", clusterName)
 
-	spinner.Start("Initializing cluster ", false)
+	cmd.PrintErrln("Note: If you just created the cluster, it can take a few minutes to connect.")
+	i.spinner.Start("Connecting ", false)
 	req := &initproto.InitRequest{
 		KmsUri:                 masterSecret.EncodeToURI(),
 		StorageUri:             uri.NoStoreURI,
@@ -176,7 +178,7 @@ func (i *initCmd) initialize(cmd *cobra.Command, newDialer func(validator *cloud
 	}
 	i.log.Debugf("Sending initialization request")
 	resp, err := i.initCall(cmd.Context(), newDialer(validator), idFile.IP, req)
-	spinner.Stop()
+	i.spinner.Stop()
 	if err != nil {
 		var nonRetriable *nonRetriableError
 		if errors.As(err, &nonRetriable) {
@@ -201,6 +203,7 @@ func (i *initCmd) initCall(ctx context.Context, dialer grpcDialer, ip string, re
 		endpoint: net.JoinHostPort(ip, strconv.Itoa(constants.BootstrapperPort)),
 		req:      req,
 		log:      i.log,
+		spinner:  i.spinner,
 	}
 
 	// Create a wrapper function that allows logging any returned error from the retrier before checking if it's the expected retriable one.
@@ -224,6 +227,7 @@ type initDoer struct {
 	req      *initproto.InitRequest
 	resp     *initproto.InitResponse
 	log      debugLog
+	spinner  spinnerInterf
 }
 
 func (d *initDoer) Do(ctx context.Context) error {
@@ -239,8 +243,7 @@ func (d *initDoer) Do(ctx context.Context) error {
 
 	grpcStateLogCtx, grpcStateLogCancel := context.WithCancel(ctx)
 	defer grpcStateLogCancel()
-	wg.Add(1)
-	d.logGRPCStateChanges(grpcStateLogCtx, &wg, conn)
+	d.handleGRPCStateChanges(grpcStateLogCtx, &wg, conn)
 
 	protoClient := initproto.NewAPIClient(conn)
 	d.log.Debugf("Created protoClient")
@@ -252,7 +255,8 @@ func (d *initDoer) Do(ctx context.Context) error {
 	return nil
 }
 
-func (d *initDoer) logGRPCStateChanges(ctx context.Context, wg *sync.WaitGroup, conn *grpc.ClientConn) {
+func (d *initDoer) handleGRPCStateChanges(ctx context.Context, wg *sync.WaitGroup, conn *grpc.ClientConn) {
+	wg.Add(1)
 	go func() {
 		defer wg.Done()
 		state := conn.GetState()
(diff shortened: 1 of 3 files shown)


malt3:
>  I am not sure if the state can theoretically change multiple times to READY.

Currently it certainly can change to READY multiple times. This is something we do want to prevent in the future by disallowing further init retries once we see that the connection became ready once.

cli/internal/cmd/init.go
```diff
@@ -262,6 +265,8 @@ func (d *initDoer) logGRPCStateChanges(ctx context.Context, wg *sync.WaitGroup,
 		}
 		if state == connectivity.Ready {
 			d.log.Debugf("Connection ready")
+			d.spinner.Stop()
+			d.spinner.Start("Initializing cluster ", false)
```
malt3, line 269: I think we should either prevent this spinner change from happening more than once in this func or prevent multiple init calls all together (see my other comment).

@daniel-weisse @3u13r any thoughts on preventing multiple init calls after the atls conn was established at least once?

daniel-weisse: Dropping from an established connection to connecting should result in a failure imo.
We need to refactor the whole init logic anyway. Currently, a failed attestation will result in a warning about this being a non recoverable error (which is not true).

Nirusu: Should we make a double `connectivity.Ready` a failure for now and discuss a larger refactor to init in an upcoming dev sync?

Because I would also +1 for a refactor just because of #1403 for which proper fixes also might include a refactor to init (from the bootstrapper side) to remove or better mitigate the pesky timeout logic.

daniel-weisse:
```shell
2023-03-14T08:39:56+01:00       DEBUG   cmd/init.go:246 Created protoClient
2023-03-14T08:39:56+01:00       DEBUG   cmd/init.go:259 Connection state started as CONNECTING
2023-03-14T08:39:56+01:00       DEBUG   cloudcmd/validators.go:170      Validating attestation document
2023-03-14T08:39:57+01:00       DEBUG   cloudcmd/validators.go:170      Successfully validated attestation document
2023-03-14T08:39:57+01:00       DEBUG   cmd/init.go:261 Connection state changed to CONNECTING
2023-03-14T08:39:57+01:00       DEBUG   cmd/init.go:264 Connection ready
// Init actually starts
```
Connection changes to `connecting` twice, so we should definitely not cancel here. But aborting with unretriable error after "Connection read" seems ok to me.

Nirusu: Alright, I gave it a try with channels in the new commit. Check it out,  this makes things definitely more complicated though 😄

cli/internal/cmd/init.go
```diff
@@ -209,21 +243,25 @@ func (i *initCmd) initCall(ctx context.Context, dialer grpcDialer, ip string, re
...
 		return isServiceUnavailable
 	}
-
 	i.log.Debugf("Making initialization call, doer is %+v", doer)
 	retrier := retry.NewIntervalRetrier(doer, 30*time.Second, serviceIsUnavailable)
-	if err := retrier.Do(ctx); err != nil {
+	err := retrier.Do(initCtx)
+	if err != nil {
+		if errors.Is(err, context.Canceled) && connectedTwice {
+			return nil, &nonRetriableError{fmt.Errorf("init call: temporarily lost gRPC connection but resumption is not supported")}
```
daniel-weisse, line 251:
```suggestion
			return nil, &nonRetriableError{errors.New("init call: temporarily lost gRPC connection but resumption is not supported")}
```

daniel-weisse (commented):
Overall implementation idea looks good to me, though I would like this to be extensively tested before merging.
Is it possible to write a unit test for this?

Nirusu:
Okay, now the unit test actually complain about data races with the use of the logger in the goroutine. Will investigate.

cli/internal/cmd/init.go
```diff
@@ -195,12 +196,45 @@ func (i *initCmd) initialize(cmd *cobra.Command, newDialer func(validator *cloud
...
-func (i *initCmd) initCall(ctx context.Context, dialer grpcDialer, ip string, req *initproto.InitRequest) (*initproto.InitResponse, error) {
+func (i *initCmd) initCall(ctx context.Context, dialer grpcDialer, ip string, spinner spinnerInterf, req *initproto.InitRequest) (*initproto.InitResponse, error) {
+	initCtx, initCtxCancel := context.WithCancel(ctx)
+	defer initCtxCancel()
+
+	// This goroutine tracks if we established a gRPC connection successfully.
+	// If yes, it switches the spinner to "Initializing cluster".
+	// If it happens multiple times, we cancel the context and abort.
+	// This usually means we lost the connection and opened a new session (which will likely fail).
+	var connectedTwice bool
```
malt3, line 207:
```suggestion
	var connectedTwice bool
```

I think what you should to do here is really just use a single variable (`connected`), to check if a connection was established at least once.
We don't even want to try the second time.
So `connected` starts with the value `false`, if we get a signal from `connReadyChan`, we set `connected` to true and in the error handling below we should check if `connected` is true + we received an error and then return an unretriable error.

Nirusu: I mainly used the second variable to differentiate the context was canceled from some other circumstance (e.g. pressing CTRL+C) during a "normal" connection (since I believe it is inherited from Cobra).

Nirusu: Has been superseded by just using `context.CancelWithCause`.

Nirusu:
Added a test and removed that ugly `conntectedTwice` variable by just using `context.CancelWithCause`.

Now there's still two issues:
1. The data race still exists with the logger (effectively only reachable in the test I guess though). I don't know how to fix this.
2. Channels are blocking by default so I guess in a very unfortunate event `logGRPCStateChanges` could try to send on a blocking channel if the events change rapidly. I don't think that's really likely to happen, though?

This cancelling on multiple events is way more painful than I thought it would be... Might actually be easier to move this down to `logGRPCStateChanges` rather than using channels.

Nirusu:
Bye channels and bye data races, this should be easier if everything is just in the same function 😅

cli/internal/cmd/init.go
```diff
@@ -288,8 +260,9 @@ func (d *initDoer) Do(ctx context.Context) error {
 	return nil
 }
 
-func (d *initDoer) logGRPCStateChanges(ctx context.Context, wg *sync.WaitGroup, conn *grpc.ClientConn) {
+func (d *initDoer) handleGRPCStateChanges(ctx context.Context, wg *sync.WaitGroup, conn *grpc.ClientConn) {
 	go func() {
```
daniel-weisse, line 264: Why does this function start a new Go routine instead of letting the caller handle this? @malt3

daniel-weisse (approved):

Nirusu:
Actually thanks to the linter warning I see that this doesn't work as aspected given that the goroutine is shutdown when the state changes to READY. Will need to fix this still.

cli/internal/cmd/init.go
malt3: This is almost in line with how the rest of our code handles go routines. CC @katexochen 

The general pattern we currently follow:

```go
func foo(ctx context.Context) {
  var wg sync.WaitGroup
  defer wg.Wait()
  
  usesAGoRoutine(ctx, wg)
  // do something ....
}

func usesAGoRoutine(ctx context.Context, wg sync.WaitGroup) {
  wg.Add(1)
  go func() {
    defer wg.Done()
    // do something in the background
  }
}
```

cli/internal/cmd/init.go
```diff
@@ -240,7 +251,8 @@ func (d *initDoer) Do(ctx context.Context) error {
 	grpcStateLogCtx, grpcStateLogCancel := context.WithCancel(ctx)
 	defer grpcStateLogCancel()
 	wg.Add(1)
```
malt3, line 253: nit (that i produced): wg.Add should be in `handleGRPCStateChanges`

cli/internal/cmd/init.go
```diff
@@ -252,7 +263,8 @@ func (d *initDoer) Do(ctx context.Context) error {
 	return nil
 }
 
-func (d *initDoer) logGRPCStateChanges(ctx context.Context, wg *sync.WaitGroup, conn *grpc.ClientConn) {
+func (d *initDoer) handleGRPCStateChanges(ctx context.Context, wg *sync.WaitGroup, conn *grpc.ClientConn) {
+	wg.Add(1)
 	go func() {
```
daniel-weisse, line 268: I think this function would be a lot more readable if you moved the Go routine from being called in this function to the caller.
I.e  instead of doing `wg.Add(1)` and `go func() {` in `handleGRPCStateChanges`, do so in the calling function.
This would also have the benefit of making things a bit more clear to what is happening in the calling function.

Nirusu: I personally agree with you but the others don‘t, so I don’t know ¯\\_(ツ)_/¯

However if we have this pattern already somewhere else in the code I would just go for consistency.

daniel-weisse: @katexochen could you explain why you want to keep calling the go routine in the function?

katexochen: It bundles concurrency related code in a single place, which makes it easier to reason about. There is a single place where mistakes could be made. If the caller needs to handle it, you'd need to check every caller for correct use. Functions should always aim to hide complexity like this from the caller, same goes for synchronization primitives like mutex etc.

PR #1406, opened 2023-03-13
image: update Azure and GCP to kernel 6.1.18

Nirusu:
### Proposed change(s)

- Update Azure & GCP kernel to 6.1.18

6.1.18 contains the [backport for the GCP disk fix](https://lore.kernel.org/all/20230310133720.407990040@linuxfoundation.org/) we reported, so we can finally get away from EOL 5.19.
The kernel also does not contain yet the backport of the MTRR fix revert that breaks Azure disks again. Wonder if this will ever happen with kernel 6.1, but 6.2 will be broken again I presume.


malt3 (approved):

Nirusu:
e2e run full: https://github.com/edgelesssys/constellation/actions/runs/4405181580
CSI functionality manually tested. Unmounting PV(C)s from a host seems to work as expected. Will merge.

PR #1412, opened 2023-03-13
Use version.txt to generate (pseudo-)version information

3u13r:
### Proposed change(s)

Instead of walking the git tree to the first label found, we rely on a check-in `version.txt`. This removes the need to check-out the git history and the tags to build the binaries with correct stamps/version information.

build-os-image run: https://github.com/edgelesssys/constellation/actions/runs/4407760616
miniconstellation test this change is needed for: https://github.com/edgelesssys/constellation/actions/runs/4408501059/jobs/7723591433

### Checklist

- [x] Link to Milestone


hack/pseudo-version/pseudo-version.go
```diff
@@ -91,9 +94,11 @@ func main() {
...
 		fmt.Println(parsedBranch)
-	case *printReleaseVersion:
-		fmt.Println(strings.TrimPrefix(rawBranch, "release/"))
 	default:
-		fmt.Println(version)
+		if !strings.Contains(*base, "pre") {
+			fmt.Println(*base) // v2.7.0
+		} else {
+			fmt.Println(version) // 2.7.0-pre.0.20230313121936-bab76e8a9acf
+		}
```
malt3, line 102: not sure i like the example but it does kind of help to understand what will be printed so feel free to keep it.

hack/pseudo-version/pseudo-version.go
```diff
@@ -44,19 +45,21 @@ func main() {
...
 	if *base == "" {
-		_, versionTag, err := gitc.FirstParentWithVersionTag()
+		rootPath, err := gitc.Path()
 		if err != nil {
-			log.With(zap.Error(err)).Warnf("Failed to find base version. Using default.")
-			versionTag = ""
+			log.With(zap.Error(err)).Fatalf("Failed to get git root path. Using default base version.")
+			//
+		} else {
+			versionTag, err := os.ReadFile(path.Join(rootPath, "/version.txt"))
```
malt3, line 54: nit for 🪟

```suggestion
			versionTag, err := os.ReadFile(filepath.Join(rootPath, "version.txt"))
```

malt3 (approved):
thanks for making our builds more reproducible

hack/pseudo-version/pseudo-version.go
Nirusu: [Windows does not care whether you use `/` or `\`](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats#canonicalize-separators) ;)
`C:/Windows\\system32/cmd.exe` for example is perfectly acceptable as path and works fine.

Though yes, `filepath` is better.

hack/pseudo-version/pseudo-version.go
3u13r: I think I know what you mean. Let me add a bit more verbosity here.

malt3: Thanks

PR #1413, opened 2023-03-13
config: add attestation variant

daniel-weisse:
### Proposed change(s)

- Add `AttestationVariant` to config file
  - Used to choose attestation validator in the CLI
- Dynamically create validator in `JoinService` based on configured variant from cmd line args

### Additional info
- Apart from the extra config entry, these changes should not yet be user facing
  - We may enforce setting the value with v2.8, or wait until we actually provide the option to choose different attestation variants to do so
- Azure e2e (verify) test: https://github.com/edgelesssys/constellation/actions/runs/4413629182
- GCP e2e (verify) test: https://github.com/edgelesssys/constellation/actions/runs/4414056826

### Checklist

- [ ] Link to Milestone


malt3 (commented):
Code LGTM. Still want to perform some tests :)

internal/config/validation.go
```diff
@@ -466,3 +467,66 @@ func (c *Config) validateName(fl validator.FieldLevel) bool {
...
+	variant, err := oid.FromString(c.AttestationVariant)
+	if err != nil {
+		return false
+	}
+
+	// make sure the variant is valid for the chosen CSP
+	switch variant {
+	case oid.AWSNitroTPM{}:
+		return c.Provider.AWS != nil
+	case oid.AzureSEVSNP{}, oid.AzureTrustedLaunch{}:
```
derpsteb, line 494: that syntax :exploding_head:

internal/config/validation.go
```diff
@@ -466,3 +467,66 @@ func (c *Config) validateName(fl validator.FieldLevel) bool {
...
+			printWarning()
+			c.AttestationVariant = oid.GCPSEVES{}.String()
+		}
+	case cloudprovider.QEMU:
+		if c.AttestationVariant == "" {
+			printWarning()
+			c.AttestationVariant = oid.QEMUVTPM{}.String()
+		}
+	}
+}
```
derpsteb, line 532: This may be more concise?

```suggestion
	if c.AttestationVariant != "" {
		return 
	}

	printWarning()
	switch c.GetProvider() {
	case cloudprovider.AWS:
		c.AttestationVariant = oid.AWSNitroTPM{}.String()
	case cloudprovider.Azure:
		c.AttestationVariant = oid.AzureTrustedLaunch{}.String()
	case cloudprovider.GCP:
		c.AttestationVariant = oid.GCPSEVES{}.String()
	case cloudprovider.QEMU:
		c.AttestationVariant = oid.QEMUVTPM{}.String()
	}
```

internal/config/config.go
```diff
@@ -77,6 +77,9 @@ type Config struct {
 	//   DON'T USE IN PRODUCTION: enable debug mode and use debug images. For usage, see: https://github.com/edgelesssys/constellation/blob/main/debugd/README.md
 	DebugCluster *bool `yaml:"debugCluster" validate:"required"`
 	// description: |
+	//   Attestation variant used to verify the integrity of a node.
+	AttestationVariant string `yaml:"attestationVariant" validate:"valid_attestation_variant"` // TODO: v2.8: Mark required
```
derpsteb, line 81: (irrespective of the question "should we wait for TDX support to make this required?"): Why start making it required in v2.8 and not v2.7?

cli/internal/helm/loader_test.go
```diff
@@ -59,25 +60,34 @@ func TestConstellationServices(t *testing.T) {
 		cnmImage           string
 	}{
 		"GCP": {
-			config: &config.Config{Provider: config.ProviderConfig{GCP: &config.GCPConfig{
-				DeployCSIDriver: func() *bool { b := true; return &b }(),
-			}}},
+			config: &config.Config{
+				AttestationVariant: oid.GCPSEVES{}.String(),
+				Provider: config.ProviderConfig{GCP: &config.GCPConfig{
+					DeployCSIDriver: func() *bool { b := true; return &b }(),
```
derpsteb, line 66: Could define and use [toPtr[T any](v T) *T](https://github.com/edgelesssys/constellation/blob/8679988b6c6989affd330ab07dc924c164b0202e/internal/kubernetes/configmaps.go#L60-L62).

daniel-weisse: This is not code I wrote for this PR, but yes, it can be changed

internal/config/config.go
daniel-weisse: Because it is not yet required in v2.6 and iirc our "breaking-changes" policy says we want to keep compatibility between two minor version, which we dont do if we start requiring it in v2.7

derpsteb: Right. I am a bit confused why I asked the question :D

malt3 (approved):

PR #1414, opened 2023-03-13
docs: refer to known issues

thomasten:
We want to maintain a list of known issues for the latest release. These should help users to work around a problem before we are able to bring a patch.
Please check if locations for the references make sense and if it should additionally be added elsewhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ If you're already familiar with Kubernetes, it's easy to get started with Conste
 
 ## Live demos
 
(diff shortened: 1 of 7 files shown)


m1ghtym0 (approved):
The locations make sense to me.
IMO, people without issue swill just ignore it, and for those that need it will make it a tiny bit less frustrating. So better to have one too many of these notes.

3u13r (approved):
LGTM

PR #1416, opened 2023-03-13
bazel: add tidy and check

katexochen:
### Proposed change(s)

- Improving the developer experience by introducing two new Bazel targets: `//:tidy` and `//:check`
- More formatting and checks will be added in the future

### Checklist


bazel/ci/def.bzl
```diff
@@ -0,0 +1,65 @@
...
+    attrs = {
+        "data": attr.label_list(
+            allow_files = True,
+        ),
+        "substitutions": attr.string_dict(),
+        "template": attr.label(
+            allow_single_file = True,
+        ),
+    },
+)
```
3u13r, line 37: This is defined in the global scope of this file, since it is not indented, right? Can we define this in a order which keeps those "bazel" definitions separate from the starlark functions?

BUILD.bazel
```diff
@@ -1,34 +1,20 @@
...
-    lint_warnings = ["all"],
-    mode = "fix",
-    verbose = True,
+alias(
+    name = "check",
+    actual = "//bazel/ci:check",
 )
+
+# gazelle:map_kind go_test go_test //bazel/go:go_test.bzl
+# gazelle:prefix github.com/edgelesssys/constellation/v2
```
3u13r, line 20: Is this still needed? If yes, what is this for?

3u13r (dismissed):

.github/workflows/test-tidy.yml
```diff
@@ -32,9 +32,11 @@ jobs:
...
-          bazelisk run //bazel/ci:tidy
-          bazelisk run //:gazelle-check
+        run: bazelisk run //:tidy
+
+      - name: Run Bazel check
+        shell: bash
+        run: bazelisk run //:check
 
       - name: Check if tidy made modifications
         id: tidycheck
```
malt3, line 42: Cannot comment on the unchanged code below..

nit for consistency on line 55:
```diff
- echo "::error::The repo is not tidy. Please run 'bazel run bazel/ci:tidy' and commit the changes."
+ echo "::error::The repo is not tidy. Please run 'bazel run //:tidy' and commit the changes."
```

bazel/toolchains/multirun_deps.bzl
```diff
@@ -0,0 +1,10 @@
+"""multirun_deps"""
+
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
+def multirun_deps():
+    git_repository(
+        name = "com_github_ash2k_bazel_tools",
+        commit = "4e045b9b4e3e613970ab68941b556a356239d433",
+        remote = "https://github.com/ash2k/bazel-tools.git",
+    )
```
malt3, line 10: nit: using a git_repository dep is harder for us to add to our own cache. (Implementing automatically adding `http_archive` rules to the edgeless archive is still an open todo, so this will only become an issue then.) I think using `http_archive` for everything makes this a bit easier.
Also it is less obvious how to ensure it is actually reproducible.
The current code leads to the message:
```
DEBUG: Rule 'com_github_ash2k_bazel_tools' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1672749937 +1100"
DEBUG: Repository com_github_ash2k_bazel_tools instantiated at:
  /home/malte/projects/clones/constellation/WORKSPACE.bazel:143:14: in <toplevel>
  /home/malte/projects/clones/constellation/bazel/toolchains/multirun_deps.bzl:6:19: in multirun_deps
Repository rule git_repository defined at:
  /home/malte/.cache/bazel/_bazel_malte/e58bd5cf140e0accfbab91a2f501a0a3/external/bazel_tools/tools/build_defs/repo/git.bzl:176:33: in <toplevel>
```

You can create a `.tar.gz` from any GitHub commit using this scheme:
https://github.com/ash2k/bazel-tools/archive/4e045b9b4e3e613970ab68941b556a356239d433.tar.gz

bazel/ci/def.bzl
malt3: Just to clarify: I think the rule definition always has to live in the global scope of a `.bzl` file. [See also.](https://bazel.build/extending/rules)

The order can probably we changed though.

.github/workflows/test-tidy.yml
katexochen: I've also moved the check step after this one.

malt3 (approved):

PR #1418, opened 2023-03-13
docs: simplify readme svg

m1ghtym0:
### Proposed change(s)

- Removes unnecessary steps from SVG in the readme
- Updates tooling to generate that svg

### Checklist

- [ ] Link to Milestone


Nirusu:
The new SVG seems to be a bit better but it's still super sluggish on Safari and maxes out the CPU.
On Chrome it seems to be fully smooth now but also sometimes peaks at 60-70% CPU usage on an M1.

So still not ideal unfortunately.

(For comparison: The old SVG before the new versions runs at about 20-40% CPU usage in Safari and 5-20% in Chrome and was fully smooth in both of them).

thomasten:
> The old SVG before the new versions runs at about 20-40% CPU usage in Safari and 5-20% in Chrome and was fully smooth in both of them

On my machine, the old SVG is already sluggish in Chrome on Windows. (No issues in Firefox.)

docs/screencasts/generate-readme-svg.sh
```diff
@@ -19,9 +19,10 @@ docker run -it \
 
 # Fix meta data: width and height are always zero in Docker produced cast files.
 # Header is the first lint of cast file in JSON format, we read, fix and write it.
-head recordings/github-readme.cast -n 1 | yq e -M '.width = 95 | .height = 17' - > new_header.cast
+head -n 1 recordings/github-readme.cast | yq -M '.width = 95 | .height = 17 | . + {"idle_time_limit": 0.5}' - > new_header.cast
 # Then append everything, expect first line from original cast file.
 tail -n+2 recordings/github-readme.cast >> new_header.cast
+# Check if header containers "idle_time_limit": 0.5
```
thomasten, line 25: misplaced comment?

docs/static/img/shell-windowframe.svg
```diff
@@ -715,5 +503,5 @@
...
...32"/><use xlink:href="#g307" y="118949"/><use xlink:href="#g350" y="118966"/><use xlink:href="#g359" y="118983"/><rect x="0" y="119000" width="8" height="17" class="foreground"/><use xlink:href="#g1" y="119000"/></g><g><use xlink:href="#g293" y="119034"/><use xlink:href="#g296" y="119051"/><use xlink:href="#g297" y="119068"/><use xlink:href="#g299" y="119085"/><use xlink:href="#g300" y="119102"/><use xlink:href="#g301" y="119119"/><use xlink:href="#g302" y="119136"/><use xlink:href="#g303" y="119170"/><use xlink:href="#g304" y="119187"/><use xlink:href="#g305" y="119204"/><use xlink:href="#g306" y="119238"/><use xlink:href="#g307" y="119255"/><use xlink:href="#g350" y="119272"/><use xlink:href="#g359" y="119289"/><rect x="16" y="119306" width="8" height="17" class="foreground"/><use xlink:href="#g2" y="119306"/></g><g><use xlink:href="#g293" y="119340"/><use xlink:href="#g296" y="119357"/><use xlink:href="#g297" y="119374"/><use xlink:href="#g299" y="119391"/><use xlink:href="#g300" y="119408"/><use xlink:href="#g301" y="119425"/><use xlink:href="#g302" y="119442"/><use xlink:href="#g303" y="119476"/><use xlink:href="#g304" y="119493"/><use xlink:href="#g305" y="119510"/><use xlink
```
thomasten, line 506: remember to edit out the pcr warning before merge

thomasten (approved):

docs/screencasts/generate-readme-svg.sh
m1ghtym0: Yeah, you can check the header after either step doesn't really matter.
Ideally, we wouldn't even need that comment.
But I gave up to make `yq` insert `idle_time_limit` correctly. It always omits the double quotes around the object key...

m1ghtym0:
With the latest version only Safari is bad for me.
Can you confirm @Nirusu and @flxflx?

Nirusu:
Even Safari is now somewhat okay. Sluggish, but not freezing (still 90% CPU usage though).
Chrome is okay (30% CPU usage).

I still think that's a bit high though, not sure how that would perform on lower end machines?

m1ghtym0:
> Even Safari is now somewhat okay. Sluggish, but not freezing (still 90% CPU usage though). Chrome is okay (30-40% CPU usage).
> 
> I still think that's a bit high though, not sure how that would perform on lower end machines?

Is that different with the "original version"?

Nirusu:
>(For comparison: The old SVG before the new versions runs at about 20-40% CPU usage in Safari and 5-20% in Chrome and was fully smooth in both of them).

old SVG = original Version

Though for @thomasten the old one was already sluggish 😅 But still the new new one still seems to be 2x CPU intensive.

m1ghtym0:
> On my personal laptop the new one definitely performs better. The current one is a sluggish mesh on it. Only recommendation I would have for now is to not _type_ the Step comments but display them fully instantly. I think this could still give a decent boost as this would reduce quite a bit of animation in there.
> 
> Otherwise for now I think it's okay and for later we can investigate how to handle this better maybe.
> 
> I was experimenting with animated WebPs which would definitely perform better (but look slightly worse due to chroma subsampling) but turns out extracting single frames from animated SVGs is not as trivial as I thought it was so I couldn't come up with a demo that works as expected. But yeah, a smaller SVG might be good enough to bring us back were we there.

Thanks, I can remove those `#comments`. Our tooling actually allows to output single frames instead of an animated svg. So if that's what you need that would not be a problem

m1ghtym0:
Will merge this now and we replace this with WebPs in a separate PR.

PR #1426, opened 2023-03-14
fix: use TDX device to mark node as initialized

daniel-weisse:
### Proposed change(s)

- Fix disk-mapper rejoining by using TDX device instead of trying to mark node using TPM


Nirusu (approved):
Reboots work now :)

PR #1429, opened 2023-03-14
bootstrapper: create /etc/kubernetes/manifests for all nodes

Nirusu:
### Proposed change(s)
- Create `/etc/kubernetes/manifests` on worker nodes, too.

Worker nodes are spamming their logs with the Kubelet not being able to find `/etc/kubernetes/manifests`.
Currently we only create this directory for control planes. But it turns out every Kubelet seems to search for it.

So let's create it on every node to stop the log spam.


3u13r (approved):
LGTM

malt3 (approved):

PR #1432, opened 2023-03-15
cli: `upgrade apply` now allows to upgrade measurements only

derpsteb:
### Proposed change(s)

- `upgrade apply` can now upgrade only measurements. Previously, the command exited if it detected no changes in image _and_ Kubernetes version to prevent accidental measurement modification. Now, this abort is replaced by a confirmation prompt. Alternatively, the `yes` flag can be used to skip the manual confirmation.
- This PR made me realize the following: currently the config measurements are used to verify the first bootstrapper node A. Next, the config measurements are used to select which of A's measurements should be enforced. Lastly, the cluster uses the combination of A's initial measurements with the enforce-info from the config. A config generated with `config generate` usually only specifies a subset of the measurements the bootstrapper/join-service sees. Namely the subset of values that should be verified. After using `upgrade apply`, the subset from the config becomes the whole set of cluster measurements. Do you see a problem with this? I don't see a problem with this, but we might want to stop putting A's initial measurements with "warnOnly" into the join-config? @daniel-weisse 

### Checklist

- [x] Link to Milestone


derpsteb:
Link checker is failing because linux-kvm.org's cert expired 4 days ago. Would ignore that if no other solution is preferred.

daniel-weisse:
We might want to rethink how we set the initial expected measurements of the cluster.
Since we have a pretty good grasp of what measurements we want to enforce (and which wants we can simply ignore), we could  just set the measurements in the helm chart using the CLI, and remove loading the measurements on the first bootstrapper node.

docs/docs/workflows/change-measurements.md
daniel-weisse, line 1: Just testing a new GitHub feature, feel free to ignore

docs/docs/workflows/change-measurements.md
```diff
@@ -0,0 +1,14 @@
+# Changing measurements
+
+There might be situations in which you have a running Constellation cluster, but want to change the expected measurements.
```
thomasten, line 3: I'm unsure if this should be its own workflow. Can we think of any use case besides fixing attestation failures? If yes, there should be a motivation at the beginning of the workflow. If not, there should either also be a motivation at beginning and in addition an entry on the Troubleshooting page linking to this, or the content should be completely moved to the troubleshooting page.

daniel-weisse (approved):
Code lgtm
Discuss with Thomas on how to proceed with the doc changes

docs/docs/workflows/change-measurements.md
derpsteb: Can't think of other reasons right now. Moved to troubleshooting.

docs/docs/workflows/troubleshooting.md
```diff
@@ -106,3 +106,19 @@ Debugging via a shell on a node is [directly supported by Kubernetes](https://ku
     ```sh
     kubectl delete pod node-debugger-constell-worker-xksa0-000000-bjthj
     ```
+
+# Changing measurements
+
+There might be situations in which you have a running Constellation cluster, but want to change the expected measurements.
+For example because the cloud provider updated the firmware.
+You can achieve this by using the `upgrade apply` command:
```
thomasten, line 114:
```suggestion
## Nodes fail to join with error `untrusted PCR value`

This error indicates that a node's [attestation statement](../architecture/attestation.md) contains measurements that don't match the trusted values expected by the [JoinService](../architecture/microservices.md#joinservice).
This may for example happen if the cloud provider updates the VM's firmware such that it influences the [runtime measurements](../architecture/attestation.md#runtime-measurements) in an unforeseen way.
You can change the expected measurements to resolve the failure.

:::caution

Attestation and trusted measurements are crucial for the security of your cluster.
Be extra careful when manually changing these settings.

When in doubt, check if the encountered [issue is known](https://github.com/edgelesssys/constellation/issues?q=is%3Aopen+is%3Aissue+label%3A%22known+issue%22) or [contact support](https://github.com/edgelesssys/constellation#support).

:::

You can use the `upgrade apply` command to change measurements of a running cluster:
```
Suggest to emphasize the issue instead of the action. We currently have both types of sections on this page, but to me this fits better for this item. If you agree, I'll rearrange this page a bit after this PR is merged.
I also added a bit more context and motivation. Do you agree with that?

derpsteb: I do.

PR #1436, opened 2023-03-16
rfc: attestation config options

daniel-weisse:
### Proposed change(s)

Add RFC for attestation options in Constellation's config file.

Open Discussion: comments, suggestions and feedback are highly appreciated

### DIFF
--- a/rfc/attestation-config.md
+++ b/rfc/attestation-config.md
@@ -34,7 +34,7 @@ provider:
 # See our docs for an overview on attestation: https://docs.edgeless.systems/constellation/architecture/attestation
 attestation:
   # Azure SEV-SNP attestation configuration
-  # See our docs for information on about configurable values [Link to our docs explaining SEV-SNP on Azure]
+  # See our docs for information about configurable values [Link to our docs explaining SEV-SNP on Azure]
   azureSEVSNP:
     # Expected confidential VM image measurements.
     measurements:
--- a/rfc/attestation-config.md
+++ b/rfc/attestation-config.md
@@ -83,6 +83,7 @@ since these are currently used by all attestation variants.
 
 If an attestation config specifies the minimum version of a parameter as `latest`,
 that value is substituted with the most recent version of that parameter for the given CSP from our API.
+Substituting values should use a similar signature verification logic for the config's signature as is used by the measurements `*M.FetchAndVerify(...)` flow.
 The value substitution is part of the unmarshalling logic.
 
 ## Attestation config API
@@ -96,6 +97,7 @@ The following HTTP endpoint is available:
 - `GET /constellation/v1/attestation/<ATTESTATION_VARIANT>/`
   - `list` returns a sorted list of available configurations for a given attestation variant
   - `<YEAR>-<MONTH>-<DAY>-<HOUR>-<MINUTE>.json`, e.g. `2023-01-23-14-32.json` returns an attestation config for the given date, if it exists. A list of available configs can be queried using the `list` endpoint.
+  - `<YEAR>-<MONTH>-<DAY>-<HOUR>-<MINUTE>.json.sig` returns the signature of the attestation config file.
 
 While this API should stay compatible with old release, extensive changes to our code may require breaking changes to the format of the attestation config files.
 In this case a new API version will be used to retrieve the config in the updated format, e.g. `/constellation/v2/attestation/<ATTESTATION_VARIANT>/`.
--- a/rfc/attestation-config.md
+++ b/rfc/attestation-config.md
@@ -1,0 +1,76 @@
+# Attestation configuration options
+
+To allow users more in-depth control over validating attestation statements, a separate, attestation-specific entry in the user's Constellation config file needs to be added.
+
+This option should allow for more in-depth configuration of how and what parts of an attestation statement are validated.
+For example, one should be able to set the minimum microcode version of an SEV-SNP attestation statement, that is deemed acceptable,
+or configure acceptable TCB statuses for Intel TDX attestation.
+
+## In the user's config
+
+A new `attestation` entry in the config file will be added, holding the options used for attestation validation.
+Existing attestation options, like `measurements` and `idKeyDigests` will be moved to this entry.
+
+This entry will replace the `attestationVariant` entry.
+
+Example configuration for Azure SEV-SNP:
+
+```yaml
+version: v3
+image: vX.Y.Z
+name: constell
+stateDiskSizeGB: 30
+kubernetesVersion: "v1.25.6"
+debugCluster: false
+provider:
+  azure: {} # Configuration for creating Azure cloud resources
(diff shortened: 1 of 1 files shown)


rfc/attestation-config.md
```diff
@@ -0,0 +1,76 @@
...
+  # See our docs for information on about configurable values [Link to our docs explaining SEV-SNP on Azure]
+  azureSEVSNP:
+    # Expected confidential VM image measurements.
+    measurements:
+      15:
+        expected: "0000000000000000000000000000000000000000000000000000000000000000"
+        warnOnly: false
+    # Configuration for validating the SEV-SNP ID Key Digest
+    firmwareSignerConfig:
+      # List of accepted key digests
```
derpsteb, line 40:
```suggestion
      # List of accepted values for the firmware signing key digest.
```

rfc/attestation-config.md
```diff
@@ -0,0 +1,76 @@
...
+      acceptedKeyDigests:
+        - 7486a447ec0f1958002a22a06b7673b9fd27d11e1c6527498056054c5fa92d23c50f9de44072760fe2b6fb89740cc96
+      # Key digest enforcement policy. One of {'equal', 'maaFallback', 'warnOnly'}
+      #     - 'equal'       : Error if the reported ID Key Digest does not match any of the values in 'acceptedKeyDigests'
```
derpsteb, line 44:
```suggestion
      #     - 'equal'       : Error if the reported signing key digest does not match any of the values in 'acceptedKeyDigests'
```

rfc/attestation-config.md
```diff
@@ -0,0 +1,76 @@
...
+      #     - 'maaFallback' : Use 'equal' checking for validation, but fallback to using MAA for validation if the reported digest does not match any of the values in 'acceptedKeyDigests'
+      #     - 'warnOnly'    : Same as 'equal', but only prints a warning instead of returning an error if no match is found
+      enforcementPolicy: maaFallback
```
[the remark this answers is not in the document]
Nirusu: I mean in the current example it is listed under `azureSEVSNP` so MAA would be fine in this context. Not sure how that would change though if GCP finally gets SNP support.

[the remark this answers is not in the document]
daniel-weisse: As this configuration is Azure specific, I think it is fine (or maybe even better) if we explain what service we are using as a fallback.
Maybe the following is better (linked webpage is subject to change, just took the first thing mentioning MAA)?
```
#     - 'maaFallback' : Use 'equal' checking for validation, but fallback to using Microsoft Azure Attestation (MAA) for validation if the reported digest does not match any of the values in 'acceptedKeyDigests'. See the Azure docs for more details: https://learn.microsoft.com/en-us/azure/attestation/overview#amd-sev-snp-attestation
```

rfc/attestation-config.md
```diff
@@ -0,0 +1,76 @@
...
+kubernetesVersion: "v1.25.6"
+debugCluster: false
+provider:
+  azure: {} # Configuration for creating Azure cloud resources
+# Configuration for attestation validation. This configuration provides sensible defaults for the Constellation version it was created for.
+# See our docs for an overview on attestation: https://docs.edgeless.systems/constellation/architecture/attestation
+attestation:
+  # Azure SEV-SNP attestation configuration
+  # See our docs for information on about configurable values [Link to our docs explaining SEV-SNP on Azure]
+  azureSEVSNP:
```
Nirusu, line 32: I am wondering, if we need to implement special attestation schemas for every provider (e.g. SNP on Azure and on GCP cannot be the same due to MAA), wouldn't it make sense to move this to provider?

rfc/attestation-config.md
```diff
@@ -0,0 +1,76 @@
...
+      #     - 'equal'       : Error if the reported signing key digest does not match any of the values in 'acceptedKeyDigests'
+      #     - 'maaFallback' : Use 'equal' checking for validation, but fallback to using MAA for validation if the reported digest does not match any of the values in 'acceptedKeyDigests'
+      #     - 'warnOnly'    : Same as 'equal', but only prints a warning instead of returning an error if no match is found
+      enforcementPolicy: maaFallback
+    # Lowest acceptable microcode version
+    microcodeVersion: 115
+    # Lowest acceptable SNP version
+    snpVersion: 8
+    # Lowest acceptable bootloader version
+    bootloaderVersion: 3
```
thomasten, line 53: For these, we need some form of auto update for users that don't touch attestation config. Currently, they get updates for these values by updating to a new Constellation version.

rfc/attestation-config.md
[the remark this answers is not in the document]
daniel-weisse: I think that is out of scope for this PR, especially since we dont yet know how different CSPs would implement their TDX solutions

rfc/attestation-config.md
malt3: We could implement this as part of a config migration step when using a new cli version for the first time.

daniel-weisse: Could also be implemented using the versions API couple with a CLI command

derpsteb: Another idea: omit non-required values when creating a config with `generate`. For those omitted values, use the the CLI's defaults. The user learns about the config values through the docs. Wouldn't need the migration step and users who don't care automatically use the latest version for each CLI.

But Malte's and Daniel's ideas are closer to what we are currently doing.

malt3: > Another idea: omit non-required values when creating a config with generate. For those omitted values, use the the CLI's defaults. The user learns about the config values through the docs. Wouldn't need the migration step and users who don't care automatically use the latest version for each CLI.

I like the idea 🤩. What about a sentinel value `auto` or `managed` (or similar) to indicate that the CLI will take care of upgrading the value?

daniel-weisse: Added specification for how I imagine this API to look like

rfc/attestation-config.md
```diff
@@ -0,0 +1,102 @@
...
+## Attestation config API
+
+Config values are uploaded to S3 and can be accessed via HTTP.
+
+The attestation config API uses the same CSP names as [the image API](./image-api.md#image-api-endpoints).
+
+The following HTTP endpoint is available:
+
+- `GET /constellation/v1/attestation/<ATTESTATION_VARIANT>/`
+  - `latest` returns the latest configuration available for a given attestation variant, as well as a reference to all past configurations
```
katexochen, line 97: Imo it shouldn't give you a list if it is called latest. This will lead to much confusion with the other parts of our api. Why not just call it 'list' and return all versions? As long as they sort (what they should), the caller can find the latest version herself.

rfc/attestation-config.md
Nirusu: Okay, I don't have strong opinions on this so I am fine either way.

malt3 (approved):

Nirusu (approved):

PR #1441, opened 2023-03-17
cli: minor restructuring for loading helm charts

derpsteb:
### Proposed change(s)

- Use one `loadRelease` function instead of one function for each release.

### Checklist

- [x] Link to Milestone


cli/internal/helm/loader_test.go
```diff
@@ -50,6 +50,17 @@ func TestLoad(t *testing.T) {
 	assert.NotNil(chart.Dependencies())
 }
 
+func BenchmarkLoad(b *testing.B) {
+	require := require.New(b)
```
daniel-weisse, line 54: Not sure how helpful this test is as the helm loading is not a performance critical step, but we can keep it if you want

daniel-weisse (approved):

PR #1443, opened 2023-03-17
openstack: implement constellation init

malt3:
### Proposed change(s)

- use dummy attestation for OpenStack (temporary change while building OpenStack support)
- use OpenStack metadata api clients in debugd and bootstrapper
- enable constellation init for OpenStack
- add CCM image for OpenStack
- add helm charts for OpenStack
- add helm chart tests for AWS and OpenStack

### Additional info
- [OS image build](https://github.com/edgelesssys/constellation/actions/runs/4446416128)

### Checklist

- [ ] Link to Milestone


malt3:
I'm seeing two issues when trying `constellation init` on OpenStack:

- bootstrapper always tries to open the vtpm to check if node was already bootstrapped
- cert-manager does not get ready in time

You can use this patch for testing (it is valid for OpenStack tests but cannot go into main):

```diff
diff --git a/bootstrapper/cmd/bootstrapper/run.go b/bootstrapper/cmd/bootstrapper/run.go
index 0a6639b2..b74b91a8 100644
--- a/bootstrapper/cmd/bootstrapper/run.go
+++ b/bootstrapper/cmd/bootstrapper/run.go
@@ -46,7 +46,8 @@ func run(issuer atls.Issuer, tpm vtpm.TPMOpenFunc, fileHandler file.Handler,
 
        nodeBootstrapped, err := vtpm.IsNodeBootstrapped(tpm)
        if err != nil {
-               log.With(zap.Error(err)).Fatalf("Failed to check if node was previously bootstrapped")
+               // TODO(malt3): convert back to error
+               log.With(zap.Error(err)).Warnf("Failed to check if node was previously bootstrapped")
        }
 
        if nodeBootstrapped {
diff --git a/bootstrapper/internal/helm/helm.go b/bootstrapper/internal/helm/helm.go
index a1491c2e..3abd5039 100644
--- a/bootstrapper/internal/helm/helm.go
+++ b/bootstrapper/internal/helm/helm.go
@@ -85,7 +85,8 @@ func (h *Client) InstallCertManager(ctx context.Context, release helm.Release) e
        h.Timeout = 10 * time.Minute
 
        if err := h.install(ctx, release.Chart, release.Values); err != nil {
-               return err
+               // TODO(malt3): Remove this once cert-manager is stable.
+               h.log.With(zap.Error(err)).Warnf("failed to install cert-manager")
        }
 
        return nil
```

internal/cloud/openstackshared/accountkey.go
katexochen, line 1: We should just do this in the openstack package, the separation of of x and xshared doesn't make any sense since x packages were moved from /bootstrapper to /internal.

internal/cloud/openstackshared/accountkey.go
```diff
@@ -0,0 +1,142 @@
...
+// String returns the string representation of the CloudINI.
+func (i CloudINI) String() string {
+	authURL := regexp.MustCompile(`[\r\n]+`).ReplaceAllString(i.AuthURL, "")
+	username := regexp.MustCompile(`[\r\n]+`).ReplaceAllString(i.Username, "")
+	password := regexp.MustCompile(`[\r\n]+`).ReplaceAllString(i.Password, "")
+	tenantID := regexp.MustCompile(`[\r\n]+`).ReplaceAllString(i.TenantID, "")
+	tenantName := regexp.MustCompile(`[\r\n]+`).ReplaceAllString(i.TenantName, "")
+	userDomainName := regexp.MustCompile(`[\r\n]+`).ReplaceAllString(i.UserDomainName, "")
+	tenantDomainName := regexp.MustCompile(`[\r\n]+`).ReplaceAllString(i.TenantDomainName, "")
+	region := regexp.MustCompile(`[\r\n]+`).ReplaceAllString(i.Region, "")
```
katexochen, line 130: regexp.MustCompile must be used in global vars only!

katexochen (changes requested):
reviewed up to and including d6fdd29e54088cc1e36bb1501d26c3f1fee28b0f

internal/config/config.go
```diff
@@ -241,9 +241,24 @@ type OpenStackConfig struct {
...
 	// description: |
+	// AuthURL is the OpenStack Identity endpoint to use inside the cluster.
+	AuthURL string `yaml:"authURL" validate:"required"`
+	// description: |
+	//   ProjectID is the ID of the project where a user resides.
+	ProjectID string `yaml:"projectID" validate:"required"`
+	// description: |
+	//   ProjectName is the name of the project where a user resides.
+	ProjectName string `yaml:"projectName" validate:"required"`
+	// description: |
```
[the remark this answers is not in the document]
malt3: In general, we have three sources of csp credentials:

- implicit credentials on the devops laptop used by terraform
- permissions attached to VMs (this is faked using imds on openstack but will be improved later)
- In-cluster credentials that are passed during constellation init

For the in-cluster credentials, we use a value that is retrieved from the constellation CLI and send during constellation init. This value is then stored as a kubernetes secret.

The code used to create the kubernetes secret can use the creds passed via the init call but cannot use the cloud metadata api client. For all other cloud providers this would not make sense since any credentials that can be retrieved from that API could also be retrieved from within the cloud controller manager (or similar deployments) using the cloudprovider sdk directly.
This form of authentication is only used while we are waiting for the better authentication flow so it will change soon.

3u13r:
Tested with the additional patch from the comment :heavy_check_mark:. Code LGTM after Paul's changes.

katexochen (approved):
lgtm. We should work on adding openstack e2e tests next.

PR #1446, opened 2023-03-17
e2e: Remove Ubuntu builder for debug artifacts required for macOS runner

Nirusu:
### Proposed change(s)
- Remove Ubuntu + cache build steps for macOS e2e test

Shouldn't be required anymore as we use Bazel and we can now cross-compile correctly.


Nirusu:
e2e run: https://github.com/edgelesssys/constellation/actions/runs/4448782421/jobs/7812063894

PR #1453, opened 2023-03-17
e2e: fix junit token warnings for sonobuoy

Nirusu:
### Proposed change(s)

- Currently any e2e test involving sonobuoy currently has the following errors in its annotations:
"Failed to create checks using the provided token. (HttpError: Resource not accessible by integration)"
"⚠️ This usually indicates insufficient permissions. More details: https://github.com/mikepenz/action-junit-report/issues/23"

See: https://github.com/edgelesssys/constellation/actions/runs/4414262630

This PR fixes this issue by adding the required permission + only enabling JUnit upload whenever we do "sonobuoy full", given that only then sonobuoy actually outputs a junit file.


katexochen (approved):
Nice, thanks!

Nirusu:
The if logic here seems to work but it still does not correctly upload the JUnit files:
https://github.com/edgelesssys/constellation/actions/runs/4450126596/jobs/7815193951

I guess this was broken before? Might have to investigate. I'll keep the PR open right now, hopefully this should be a relatively quick fix.

Nirusu:
Now fully working: https://github.com/edgelesssys/constellation/actions/runs/4468238341

katexochen (approved):

PR #1454, opened 2023-03-19
cli: remove ctx parameter from rollbackOnError to prevent wrong use

thomasten:
rollbackOnError must not be called with the ctx of the to-be-rollbacked action, so we shouldn't allow to pass a (wrong) context in.


katexochen (approved):
lgtm, thanks!

PR #1455, opened 2023-03-19
Use maa.NewParameters

thomasten:
Proposal to remove the following code from the SNP issuer that was also needed by the MAA library:
* get VCEK via IMDS
* get and parse HCL report


daniel-weisse (approved):
Yeah, this makes a lot more sense

PR #1466, opened 2023-03-20
ci: add unittests for the benchmark actions

m1ghtym0:
### Proposed change(s)

- Fix the current problem with the compare method
- Refactor and add unit tests
- It's not the cleanest python code and unittests. Not sure how much engineering we want to put into this since our goal is to parse some numbers and put them into json files.

### Checklist

### DIFF
--- a/.github/actions/e2e_benchmark/evaluate/compare.py
+++ b/.github/actions/e2e_benchmark/evaluate/compare.py
@@ -169,6 +169,7 @@ class BenchmarkComparer:
 
         return f'| {subtest} | {metric} ({unit}) | {val_curr} | {val_prev} | {ratio} |'
 
+
 def set_failed() -> None:
     os.environ['COMPARISON_SUCCESS'] = str(False)
 
--- a/.github/actions/e2e_benchmark/evaluate/compare.py
+++ b/.github/actions/e2e_benchmark/evaluate/compare.py
@@ -157,11 +157,11 @@ class BenchmarkComparer:
             if is_bigger_better(bench_suite=metric):
                 ratio_num = val_curr / val_prev
                 if ratio_num < ALLOWED_RATIO_DELTA.get(metric, 1):
-                    self.set_failed()
+                    set_failed()
             else:
                 ratio_num = val_prev / val_curr
                 if ratio_num > ALLOWED_RATIO_DELTA.get(metric, 1):
-                    self.set_failed()
+                    set_failed()
 
             ratio_num = round(ratio_num, 3)
             emoji = PROGRESS[int(ratio_num >= 1)]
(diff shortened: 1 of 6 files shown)


.github/actions/e2e_benchmark/evaluate/compare.py
```diff
@@ -55,69 +70,78 @@ def get_paths() -> Tuple[str, str]:
...
+                ch=bench_prev['metadata']['github.sha']),
+            '',
+            '| Benchmark suite | Metric | Current | Previous | Ratio |',
+            '|-|-|-|-|-|',
+        ]
+
+        # compare FIO results
+        for subtest in FIO_TESTS:
+            if subtest not in bench_prev['fio']:
+                raise ValueError(f'Previous benchmnarks do not include the "{subtest}" test.')
```
Nirusu, line 130:
```suggestion
                raise ValueError(f'Previous benchmarks do not include the "{subtest}" test.')
```

.github/actions/e2e_benchmark/evaluate/compare.py
```diff
@@ -55,69 +70,78 @@ def get_paths() -> Tuple[str, str]:
...
+            for metric in bench_prev['fio'][subtest].keys():
+                md_lines.append(self.compare_test('fio', subtest, metric, bench_prev, bench_curr))
+
+        # compare knb results
+        for subtest in KNB_TESTS:
+            if subtest not in bench_prev['knb']:
+                raise ValueError(f'Previous benchmnarks do not include the "{subtest}" test.')
```
Nirusu, line 137:
```suggestion
                raise ValueError(f'Previous benchmarks do not include the "{subtest}" test.')
```

.github/actions/e2e_benchmark/evaluate/compare.py
```diff
@@ -133,20 +157,28 @@ def compare_test(test, subtest, metric, bench_prev, bench_curr) -> str:
...
             ratio_num = round(ratio_num, 3)
             emoji = PROGRESS[int(ratio_num >= 1)]
             ratio = f'{ratio_num} {emoji}'
 
         return f'| {subtest} | {metric} ({unit}) | {val_curr} | {val_prev} | {ratio} |'
 
-def set_failed() -> None:
-    os.environ['COMPARISON_SUCCESS'] = str(False)
+    def set_failed(self) -> None:
+        os.environ['COMPARISON_SUCCESS'] = str(False)
```
Nirusu, line 173: This function can be moved outside the class. It does not use `self` in any form and has global state due to the OS environment.

Nirusu (changes requested):
I added a commit that auto-formatted the files again to match PEP 8 (we don't have any Python style checks in the CI).

Otherwise some typos + another linter issue.

Theoretically the test could also test error cases but generally I don't think we really care about error handling - it should just fail and not produce any garbage output. 

If it works in the CI and does everything as expected LGTM apart from the nits.

Nirusu:
CI fails due to lack of rebase (Bazel is fragile), should be fine if you merge into main I believe. Otherwise you can rebase once if you want to go sure.

PR #1468, opened 2023-03-20
cli: fix `upgrade apply` for image-only upgrades

derpsteb:
### Proposed change(s)

- This fixes a bug where `upgrade apply` fails if only the image version is upgraded. This is due to mishandling of an empty configmap. This is a bug that is only located within the CLI and does not affect the cluster's ability to upgrade.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cloudcmd/upgrade.go
+++ b/cli/internal/cloudcmd/upgrade.go
@@ -124,15 +124,15 @@ func (u *Upgrader) UpgradeNodeVersion(ctx context.Context, conf *config.Config) 
 
 	components, err := u.updateK8s(&nodeVersion, versionConfig.ClusterVersion, versionConfig.KubernetesComponents)
 	switch {
-	case errors.As(err, &upgradeErr):
-		upgradeErrs = append(upgradeErrs, fmt.Errorf("skipping Kubernetes upgrades: %w", err))
-	case err != nil:
-		return fmt.Errorf("updating Kubernetes version: %w", err)
 	case err == nil:
 		err := u.applyComponentsCM(ctx, components)
 		if err != nil {
 			return fmt.Errorf("applying k8s components ConfigMap: %w", err)
 		}
+	case errors.As(err, &upgradeErr):
+		upgradeErrs = append(upgradeErrs, fmt.Errorf("skipping Kubernetes upgrades: %w", err))
+	default:
+		return fmt.Errorf("updating Kubernetes version: %w", err)
 	}
 
 	if len(upgradeErrs) == 2 {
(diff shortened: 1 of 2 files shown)


cli/internal/cloudcmd/upgrade.go
```diff
@@ -131,32 +142,41 @@ func (u *Upgrader) UpgradeNodeVersion(ctx context.Context, conf *config.Config)
...
-	updatedNodeVersion, err := u.applyUpgrade(ctx, &components, nodeVersion)
+
+	updatedNodeVersion, err := u.applyNodeVersion(ctx, nodeVersion)
 	if err != nil {
 		return fmt.Errorf("applying upgrade: %w", err)
 	}
 	if updatedNodeVersion.Spec.ImageReference != imageReference ||
 		updatedNodeVersion.Spec.ImageVersion != imageVersion.Version ||
-		updatedNodeVersion.Spec.KubernetesComponentsReference != components.ObjectMeta.Name ||
+		updatedNodeVersion.Spec.KubernetesComponentsReference != nodeVersion.Spec.KubernetesComponentsReference ||
```
3u13r, line 152: I'd like to have those checks as separate if cases, which say which of those checks is false, and also both values. 

Use case: I forgot to set the correct config for `constellation upgrade apply` and it selected the wrong one by default which had a mismatch on the image. I know this should not happen for an end user, since they should have dedicated folders for each Constellation/config, but nevertheless, it is more verbose in such cases.

cli/internal/cloudcmd/upgrade.go
```diff
@@ -131,32 +142,41 @@ func (u *Upgrader) UpgradeNodeVersion(ctx context.Context, conf *config.Config)
...
 
 	return errors.Join(upgradeErrs...)
 }
 
-func (u *Upgrader) applyUpgrade(ctx context.Context, components *corev1.ConfigMap, nodeVersion updatev1alpha1.NodeVersion) (updatev1alpha1.NodeVersion, error) {
+// applyComponentsCM applies the k8s components ConfigMap to the cluster.
+func (u *Upgrader) applyComponentsCM(ctx context.Context, components *corev1.ConfigMap) error {
+	if components == nil {
+		return errors.New("components ConfigMap is nil")
+	}
```
3u13r, line 164: Having dedicated input checks which simply error, is uncommon in Go at least from what I've seen. If the components is nil, the function who actually uses this value should return an error, which we handle i.e. `createConfigMap(ctx, components)`. Furthermore, the `components` passed to this function needed to be created in a function before, which should return either a valid `components` or an error we handle.

cli/internal/cloudcmd/upgrade.go
```diff
@@ -115,13 +115,24 @@ func (u *Upgrader) UpgradeNodeVersion(ctx context.Context, conf *config.Config)
...
+	switch {
+	case errors.As(err, &upgradeErr):
 		upgradeErrs = append(upgradeErrs, fmt.Errorf("skipping Kubernetes upgrades: %w", err))
+	case err != nil:
+		return fmt.Errorf("updating Kubernetes version: %w", err)
+	case err == nil:
+		err := u.applyComponentsCM(ctx, components)
+		if err != nil {
+			return fmt.Errorf("applying k8s components ConfigMap: %w", err)
+		}
```
3u13r, line 135: What do you think about:
```suggestion
	case err == nil:
		err := u.applyComponentsCM(ctx, components)
		if err != nil {
			return fmt.Errorf("applying k8s components ConfigMap: %w", err)
		}
	case errors.As(err, &upgradeErr):
		upgradeErrs = append(upgradeErrs, fmt.Errorf("skipping Kubernetes upgrades: %w", err))
	default:
		return fmt.Errorf("updating Kubernetes version: %w", err)
```

I think having the `err==nil` case is a bit worse, but I prefer having a `default` which might makes it clearer that this is the "catch everything unexpected" case which should never be deleted. But I have no strong opinion about this.

3u13r (changes requested):

cli/internal/cloudcmd/upgrade.go
derpsteb: Had this for more descriptive error messages. Removed. All other aspects you mentioned are already true.

3u13r (approved):
LGTM

PR #1469, opened 2023-03-20
ci: add e2e-upgrade test

derpsteb:
### Proposed change(s)

- Add e2e test for upgrade. The test is implemented as a go test. It can be executed as a bazel target. The general workflow is to setup a cluster, point the test to the workspace in which to find the kubeconfig and the constellation config and specify a target image, k8s and service version. The test will succeed if it detects all target versions in the cluster within the configured timeout. The CI automates the above steps. A separate workflow is introduced as there are multiple input fields to the test. Adding all of these inputs to the manual e2e test seemed to overload the inputs interface.
- Labeled as hold since both dependencies need to be merged first. 
- **EDIT** we currently don't support upgrades on AWS. Will add a check to `upgrade apply` to fail early in a separate PR.

### Additional info
- [Pipeline run](https://github.com/edgelesssys/constellation/actions/runs/4470733744)
- Needs: https://github.com/edgelesssys/constellation/pull/1468, https://github.com/edgelesssys/constellation/pull/1467 

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/helm/BUILD.bazel
+++ b/cli/internal/helm/BUILD.bazel
@@ -319,6 +319,7 @@ go_library(
     importpath = "github.com/edgelesssys/constellation/v2/cli/internal/helm",
     visibility = ["//cli:__subpackages__"],
     deps = [
+        "//cli/internal/clusterid",
         "//internal/attestation/idkeydigest",
         "//internal/cloud/cloudprovider",
         "//internal/compatibility",
--- a/cli/internal/helm/client.go
+++ b/cli/internal/helm/client.go
@@ -8,11 +8,14 @@ package helm
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
 	"time"
 
+	"github.com/edgelesssys/constellation/v2/cli/internal/clusterid"
+	"github.com/edgelesssys/constellation/v2/internal/attestation/idkeydigest"
 	"github.com/edgelesssys/constellation/v2/internal/compatibility"
 	"github.com/edgelesssys/constellation/v2/internal/config"
 	"github.com/edgelesssys/constellation/v2/internal/constants"
@@ -128,8 +131,10 @@ func (c *Client) Upgrade(ctx context.Context, config *config.Config, timeout tim
 		return fmt.Errorf("creating CR backup: %w", err)
 	}
 
+	// TODO: v2.8: remove fileHanlder.
+	fileHandler := file.NewHandler(afero.NewOsFs())
 	for _, chart := range upgradeReleases {
-		err = c.upgradeRelease(ctx, timeout, config, chart, allowDestructive)
+		err = c.upgradeRelease(ctx, timeout, config, chart, allowDestructive, fileHandler)
 		if err != nil {
 			return fmt.Errorf("upgrading %s: %w", chart.Metadata.Name, err)
 		}
@@ -172,26 +177,9 @@ func (c *Client) currentVersion(release string) (string, error) {
 // ErrConfirmationMissing signals that an action requires user confirmation.
 var ErrConfirmationMissing = errors.New("action requires user confirmation")
 
-// TODO: v2.8: remove. This function is only temporarily needed as a migration from 2.6 to 2.7.
-// setAttestationVariant sets the attesationVariant value on verification-service and join-service value maps.
-func setAttestationVariant(values map[string]any, variant string) error {
-	joinServiceVals, ok := values["join-service"].(map[string]any)
-	if !ok {
-		return errors.New("invalid join-service values")
-	}
-	joinServiceVals["attestationVariant"] = variant
-
-	verifyServiceVals, ok := values["verification-service"].(map[string]any)
-	if !ok {
-		return errors.New("invalid verification-service values")
-	}
-	verifyServiceVals["attestationVariant"] = variant
-
-	return nil
-}
-
+// TODO: v2.8: remove fileHandler argument.
 func (c *Client) upgradeRelease(
-	ctx context.Context, timeout time.Duration, conf *config.Config, chart *chart.Chart, allowDestructive bool,
+	ctx context.Context, timeout time.Duration, conf *config.Config, chart *chart.Chart, allowDestructive bool, fileHandler file.Handler,
 ) error {
 	// We need to load all values that can be statically loaded before merging them with the cluster
 	// values. Otherwise the templates are not rendered correctly.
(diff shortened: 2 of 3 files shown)


e2e/internal/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,353 @@
...
+	targetImage         = flag.String("target-image", "", "Image (shortversion) to upgrade to.")
+	targetKubernetes    = flag.String("target-kubernetes", "", "Kubernetes version (MAJOR.MINOR.PATCH) to upgrade to. Defaults to default version of target CLI.")
+	targetMicroservices = flag.String("target-microservices", "", "Microservice version (MAJOR.MINOR.PATCH) to upgrade to. Defaults to default version of target CLI.")
+	// When executing the test as a bazel target the workspace path is supplied through an env variable that bazel sets.
+	workspace = flag.String("workspace", "", "Constellation workspace in which to run the tests.")
+	// When executing the test as a bazel target the CLI path is supplied through an env variable that bazel sets.
+	// When executing via `go test` extra care should be taken that the supplied CLI is built on the same commit as this test.
+	cliPath     = flag.String("cli", "", "Constellation CLI to run the tests.")
+	wantWorker  = flag.Int("want-worker", 0, "Number of wanted worker nodes.")
+	wantControl = flag.Int("want-control", 0, "Number of wanted control nodes.")
```
katexochen, line 45: This can be done in a function, no need for global state.

e2e/internal/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,353 @@
...
+
+// TestUpgrade checks that the workspace's kubeconfig points to a healthy cluster,
+// we can write an upgrade config, we can trigger an upgrade
+// and the cluster eventually upgrades to the target version.
+func TestUpgrade(t *testing.T) {
+	assert := assert.New(t)
+	require := require.New(t)
+	k, err := kubectl.New()
+	require.NoError(err)
+	assert.NotNil(k)
```
katexochen, line 84:
```suggestion
	require.NotNil(k)
```
We can't continue without kubectl, right?

internal/versionsapi/version.go
```diff
@@ -142,9 +144,32 @@ func (v Version) ListPath(gran Granularity) string {
...
+	if err != nil {
+		return &url.URL{}, &url.URL{}, fmt.Errorf("parsing path for measurement: %w", err)
+	}
+
+	signatureURL, err := url.Parse(signaturePath)
+	if err != nil {
+		return &url.URL{}, &url.URL{}, fmt.Errorf("parsing path for signature: %w", err)
+	}
+	return measurementURL, signatureURL, nil
+}
```
katexochen, line 168: This can't be implemented as method of the version struct. The measurement artifact is special to a image version, but the version struct represents versions of multiple kind. There is no such thing as a measuremntURL for a version of versionKindCLI.

bazel/ci/BUILD.bazel
```diff
@@ -3,10 +3,16 @@ load("@com_github_ash2k_bazel_tools//multirun:def.bzl", "multirun")
 load("@com_github_bazelbuild_buildtools//buildifier:def.bzl", "buildifier", "buildifier_test")
 load("//bazel/sh:def.bzl", "noop_warn", "repo_command", "sh_template")
 
-gazelle(name = "gazelle_generate")
+required_tags = ["e2e"]
+
+gazelle(
+    name = "gazelle_generate",
+    build_tags = required_tags,
```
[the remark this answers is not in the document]
derpsteb: If `build_tags` is not set, gazelle does not generate build files for files that require a build tag. In this case, [e2e/internal/upgrade/BUILD.bazel](https://github.com/edgelesssys/constellation/pull/1469/files#diff-cc17dc16068258edef10efdad177053aa9c3298a6ba58c79f4d72c32b132b408) would not include the required `upgrade_test` target, since `upgrade_test.go` requires the build tag `e2e`.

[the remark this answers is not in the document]
katexochen: That makes sense, thanks. :)

.github/actions/e2e_test/action.yml
```diff
@@ -114,13 +117,26 @@ runs:
...
         targetOS: ${{ steps.determine-build-target.outputs.hostOS }}
         targetArch: ${{ steps.determine-build-target.outputs.hostArch }}
         enterpriseCLI: ${{ inputs.keepMeasurements }}
         outputPath: "build/constellation"
 
+    - name: Fetch CLI
+      if: inputs.cliVersion != ''
+      shell: bash
+      run: |
+        curl -fsSL https://github.com/edgelesssys/constellation/releases/download/${{ inputs.cliVersion }}/constellation-linux-amd64 > constellation
```
katexochen, line 132:
```suggestion
        curl -fsSL -o constellation https://github.com/edgelesssys/constellation/releases/download/${{ inputs.cliVersion }}/constellation-linux-amd64
```

internal/versionsapi/version.go
derpsteb: Made this a function that receives a function as argument. Error out for versions that are of wrong kind.

e2e/internal/upgrade/image.go
```diff
@@ -0,0 +1,132 @@
...
+		shortPath:    toImage,
+	}
+	versionsClient := fetcher.NewFetcher()
+
+	ver, err := versionsapi.NewVersionFromShortPath(toImage, versionsapi.VersionKindImage)
+	if err != nil {
+		return upgradeInfo{}, err
+	}
+
+	measurementsURL, _, err := ver.MeasurementURL(csp)
```
daniel-weisse, line 60:
```suggestion
	measurementsURL, _, err := versionsapi.MeasurementURL(ver, csp)
```
`versionsapi.Version` does not have a method `MeasurementURL`

e2e/internal/upgrade/upgrade_test.go
daniel-weisse, line 1: Consider using `log.PrintX` instead of `fmt.PrintX` to add timestamps to messages during the test.
We could also use `logger.NewTest(t)`

e2e/internal/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,378 @@
...
+			return false
+		}
+
+		if version != wantMicroserviceVersion {
+			fmt.Printf("Microservices still at version: %v\n", version)
+			return false
+		}
+
+		return true
+	}, time.Hour*3, time.Minute)
```
daniel-weisse, line 226: Make this timeout configurable through cmd line flags?

e2e/internal/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,378 @@
...
+						fmt.Printf("Status %s for node %s is %s\n", coreV1.NodePIDPressure, node.Name, condition.Status)
+						return false
+					}
+				case coreV1.NodeNetworkUnavailable:
+					if condition.Status != coreV1.ConditionFalse {
+						fmt.Printf("Status %s for node %s is %s\n", coreV1.NodeNetworkUnavailable, node.Name, condition.Status)
+						return false
+					}
+				}
+			}
```
daniel-weisse, line 358:
```suggestion
				switch condition.Type {
				case coreV1.NodeReady, coreV1.NodeMemoryPressure, coreV1.NodeDiskPressure, coreV1.NodePIDPressure, coreV1.NodeNetworkUnavailable:
					if condition.Status != coreV1.ConditionTrue {
						fmt.Printf("Status %s for node %s is %s\n", condition.Type, node.Name, condition.Status)
						return false
					}
				}
			}
```
Not 100% sure I understand this code segment correctly, but this seems a lot simpler?

daniel-weisse (changes requested):

e2e/internal/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,378 @@
...
+		wantWorker:          *wantWorker,
+		wantControl:         *wantControl,
+	}
+}
+
+// TestMain checks that the prerequisites for the test are met:
+// - a workspace is set
+// - a CLI path is set
+// - the constellation-upgrade folder does not exist
+func setup() (testFlags, error) {
```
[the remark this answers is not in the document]
derpsteb: Yes this was previously a TestMain method. I switched to this style of defining and parsing the CLI flags based on katexochen's comment about removing global state. To address your concern about panics when calling this twice I added a check using flag's [Parsed](https://pkg.go.dev/flag#Parsed) and a comment that this should only be called once.

In general I would not want to add other tests to this file since it is already quite big and add separate pkgs/files for new e2e tests.

Hope that addresses your concern.

derpsteb:
Still waiting for a successful pipeline run since upgrade was broken 3x now. Currently waiting for the smoketest to complete: https://github.com/edgelesssys/constellation/pull/1491. Merge that, build new images, rebase, retrigger, :crossed_fingers: ..

derpsteb:
[GCP run](https://github.com/edgelesssys/constellation/actions/runs/4499931797) :green_circle: 
[Azure run](https://github.com/edgelesssys/constellation/actions/runs/4499046373) :green_circle: 

We currently don't support upgrades on AWS. Will add a check so `upgrade apply` fails early in a separate PR.

katexochen (approved):

PR #1465, opened 2023-03-20
terraform: valid Azure attestation provider name

katexochen:
### Proposed change(s)

- Fix [bug in attestation provider naming](https://github.com/edgelesssys/constellation/actions/runs/4469849456/jobs/7852552415#step:9:1089), introduced in #1375
- e2e test run, Kubernetes 1.25, nop: https://github.com/edgelesssys/constellation/actions/runs/4470212067

### Checklist

- [x] Add labels (e.g., for changelog category)

PR #1472, opened 2023-03-21
terraform: unique Azure attestation provider name

katexochen:
### Proposed change(s)

- While #1465 made the name valid, is also has to be unique. The name should use local.name instead of var.name, as first contains the uid.
- [`Attestation Provider Name: "e2etestap"): attestationproviders.AttestationProvidersClient#Create: Failure sending request: StatusCode=409 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="DuplicateTenant" Message="The given service URI 'e2etestap' is already in use. Please try again with a different name."`](https://github.com/edgelesssys/constellation/actions/runs/4471281390/jobs/7855959048#step:9:929)

### Checklist

### DIFF
--- a/cli/internal/terraform/terraform/azure/main.tf
+++ b/cli/internal/terraform/terraform/azure/main.tf
@@ -44,7 +44,7 @@ resource "random_password" "initSecret" {
 resource "azurerm_attestation_provider" "attestation_provider" {
   count = var.create_maa ? 1 : 0
   # name must be between 3 and 24 characters in length and use numbers and lower-case letters only.
-  name                = format("%sap", replace(var.name, "/[^a-z0-9]/", ""))
+  name                = format("constell%s", local.uid)
   resource_group_name = var.resource_group
   location            = var.location
 }


cli/internal/terraform/terraform/azure/main.tf
```diff
@@ -44,7 +44,7 @@ resource "random_password" "initSecret" {
 resource "azurerm_attestation_provider" "attestation_provider" {
   count = var.create_maa ? 1 : 0
   # name must be between 3 and 24 characters in length and use numbers and lower-case letters only.
-  name                = format("%sap", replace(var.name, "/[^a-z0-9]/", ""))
+  name                = format("%sap", replace(local.name, "/[^a-z0-9]/", ""))
```
daniel-weisse, line 47: Constellation limits names on Azure (without uid suffix) to 37 characters.
However, MAA names are limited to only 24 characters. Since we also add `sap` and a uid suffix to the name, we can easily run into configurations that are accepted by our config validation, but will fail during creation with Terraform.

I would suggest using only the uid of a constellation to generate the MAA name.
Example: `constellXXXX`, where `XXXX` is the Constellation's UID

malt3 (approved):

daniel-weisse (approved):

PR #1476, opened 2023-03-21
bazel: use pure go implementation of netdns for all builds 

malt3:
All builds (with or without cgo support) should use netdns. This prevents weird issues we occasionally see when building linux binaries on macOS.

### Proposed change(s)

- bazel: use pure go implementation of netdns for all builds 

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -20,6 +20,9 @@ build --strip=always
 # set build mode to opt by default (better reproducibility and performance)
 build --compilation_mode=opt
 
+# use pure go implementation of netdns
+build --define=gotags=netgo
+
 # enable tpm simulator for tests
 test --//bazel/settings:tpm_simulator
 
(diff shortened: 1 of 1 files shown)


Nirusu (commented):
From what I can see this seems to fix the early panics - though personally I am not really happy about not knowing what exactly the cause here is 😅 Some odd behavior also happened when Linux as build host so not sure if cross-compiling alone is the culprit here.

But also not sure how to debug this.

malt3:
From @katexochen: we should also keep [os/user](https://pkg.go.dev/os/user) in mind.
The `osusergo` flag can be used to disable cgo here.

katexochen (approved):

Nirusu (approved):
Not 100% happy not knowing the cause here, but we can give it a shot in the CI and see if things blow up elsewhere. There's still bit of time before the next release.

PR #1483, opened 2023-03-21
debugd: use nanosecond precision for logs

Nirusu:
### Proposed change(s)
- Use nanosecond precision in Filebeat
- Remove re-timestamping the bootstrapper events (this likely leads to a loss of precision)

journald logs are still slightly too unpricise to have distinct timestamps for every events compared to the Pod container logs which have a full timestamp, but Kibana now seems to (mostly?) correctly display everything in order. As a last resort we can always sort by `_id` or, in the future, tell log stash to rewrite `_id` to an integer.

But now at least things the logs are readable :)

PS: Don't trust ChatGPT. It was a liar and told me I didn't need to change the Filebeat configuration and told me to just use a date filter with `ISO8601_WITH_NANOS`. That's a lie - it doesn't exist. The AI uprising is not happening yet.

Screenshot 2023-03-21 at 17 37 19


malt3 (approved):
I saw this working on @Nirusu machine so I approve it. The options look correct to me as well.

Nirusu:
Small update (in case someone looks this up in the future):
journald only has microsecond precision. 
But the Pod logs are logged with nanosecond precision.

Not sure if there's any huge performance or feature downside to nanoseconds. In case this makes any issues maybe we can consider downgrading to microseconds. But for now it lets go with nanoseconds and it should be fine, I guess. The target data type in OpenSearch can either only be `date` (millisecond) or `date_nano` (nanosecond) anyways, so at least storage wise it doesn't seem to matter - only processing wise it could make a different I assume.

PR #1493, opened 2023-03-22
add version.txt step to release pipeline

3u13r:
### Proposed change(s)

- bumps the version.txt which bazel uses to version stamp all artifacts such as the CLI (one can verify the correct stamp via `constellation version`
- Also removed the CMake version replacement, as it is not used anymore

CI run: https://github.com/edgelesssys/constellation/actions/runs/4491528068

### Checklist

- [x] Link to Milestone


.github/workflows/release.yml
```diff
@@ -117,6 +117,15 @@ jobs:
           git checkout "${BRANCH}" || git checkout -B "${BRANCH}"
           git push origin "${BRANCH}"
 
+      - name: Write version to version.txt
+        run: |
+          echo "${{ inputs.version }}" > version.txt
+          git config --global user.name "release[bot]"
+          git config --global user.email "release[bot]@users.noreply.github.com"
+          git add version.txt
+          git commit -m "deps: update stamping version to ${{ inputs.version }}"
```
katexochen, line 126: I don't think "deps:" in the commit message fits, as we update our own version.

katexochen (approved):

PR #1494, opened 2023-03-22
bazel: add golangci-lint to //:check target

katexochen:
### Checklist


malt3 (approved):
works for me :) thanks for moving everything to Bazel

PR #1495, opened 2023-03-22
cli: add verify support for TDX (and various refactors + fixes regarding attesation variants for TDX)

Nirusu:
### Proposed change(s)
- Plenty of bug fixes regarding QEMU TDX when selecting the attestation variant
- Refactor the cloudcmd validator to remove most mentions of PCRs unless we actually talk about PCRs (to generalize for TPM & TDX measurements)
- Add `constellation verify` support for TDX

Plus one cherry-pick commits to get Bazel to work since the Zig toolchain couldn't be downloaded anymore from the URL and I don't want to rebase this to the current main branch yet. Gotta love this toolchain :/

### DIFF
--- a/cli/internal/cloudcmd/validators.go
+++ b/cli/internal/cloudcmd/validators.go
@@ -22,10 +22,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// Validator validates Platform Configuration Registers (PCRs).
+// Validator validates Measurements (Platform Control Registers = PCRs).
 type Validator struct {
 	attestationVariant oid.Getter
-	pcrs               measurements.M
+	measurements       measurements.M
 	idkeydigests       idkeydigest.IDKeyDigests
 	enforceIDKeyDigest bool
 	validator          atls.Validator
@@ -41,7 +41,7 @@ func NewValidator(conf *config.Config, log debugLog) (*Validator, error) {
 	}
 	v.attestationVariant = variant // valid variant
 
-	if err := v.setPCRs(conf); err != nil {
+	if err := v.setMeasurements(conf); err != nil {
 		return nil, err
 	}
 
@@ -53,22 +53,22 @@ func NewValidator(conf *config.Config, log debugLog) (*Validator, error) {
 	return &v, nil
 }
 
-// UpdateInitPCRs sets the owner and cluster PCR values.
-func (v *Validator) UpdateInitPCRs(ownerID, clusterID string) error {
-	if err := v.updatePCR(uint32(measurements.PCRIndexOwnerID), ownerID); err != nil {
+// UpdateInitMeasurements sets the owner and cluster measurement values.
+func (v *Validator) UpdateInitMeasurements(ownerID, clusterID string) error {
+	if err := v.updateMeasurement(uint32(measurements.PCRIndexOwnerID), ownerID); err != nil {
 		return err
 	}
-	return v.updatePCR(uint32(measurements.PCRIndexClusterID), clusterID)
+	return v.updateMeasurement(uint32(measurements.PCRIndexClusterID), clusterID)
 }
 
-// updatePCR adds a new entry to the measurements of v, or removes the key if the input is an empty string.
+// updateMeasurement adds a new entry to the measurements of v, or removes the key if the input is an empty string.
 //
 // When adding, the input is first decoded from hex or base64.
-// We then calculate the expected PCR by hashing the input using SHA256,
-// appending expected PCR for initialization, and then hashing once more.
-func (v *Validator) updatePCR(pcrIndex uint32, encoded string) error {
+// We then calculate the expected measurement by hashing the input using SHA256,
+// appending expected measurement for initialization, and then hashing once more.
+func (v *Validator) updateMeasurement(measurementIndex uint32, encoded string) error {
 	if encoded == "" {
-		delete(v.pcrs, pcrIndex)
+		delete(v.measurements, measurementIndex)
 		return nil
 	}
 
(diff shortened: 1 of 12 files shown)


Nirusu:
golangci-lint fixes will require a rebase though.

cli/internal/cloudcmd/validators.go
```diff
@@ -53,22 +56,38 @@ func NewValidator(conf *config.Config, log debugLog) (*Validator, error) {
...
-	if err := v.updatePCR(uint32(measurements.PCRIndexOwnerID), ownerID); err != nil {
+// UpdateInitMeasurements sets the owner and cluster measurement values.
+func (v *Validator) UpdateInitMeasurements(ownerID, clusterID string) error {
+	switch v.attestationVariant {
+	case oid.AWSNitroTPM{}, oid.AzureTrustedLaunch{}, oid.AzureSEVSNP{}, oid.GCPSEVES{}, oid.QEMUVTPM{}:
+		return v.updateInitMeasurementsTPM(ownerID, clusterID)
+	case oid.QEMUTDX{}:
+		return v.updateInitMeasurementsTDX(clusterID)
+	default:
+		return fmt.Errorf("selecting attestation variant: unknown attestation variant")
```
daniel-weisse, line 67:
```suggestion
		return fmt.Errorf("selecting attestation variant: unknown attestation variant: %s", v.attestationVariant.String())
```

cli/internal/cloudcmd/validators.go
```diff
@@ -53,22 +56,38 @@ func NewValidator(conf *config.Config, log debugLog) (*Validator, error) {
...
+	return v.updateMeasurement(uint32(measurements.TDXIndexClusterID), clusterID)
+}
+
+// updateMeasurement adds a new entry to the measurements of v, or removes the key if the input is an empty string.
 //
 // When adding, the input is first decoded from hex or base64.
-// We then calculate the expected PCR by hashing the input using SHA256,
-// appending expected PCR for initialization, and then hashing once more.
-func (v *Validator) updatePCR(pcrIndex uint32, encoded string) error {
+// We then calculate the expected measurement by hashing the input using SHA256,
```
daniel-weisse, line 86:
```suggestion
// We then calculate the expected measurement by hashing the input,
```

cli/internal/cloudcmd/validators.go
```diff
@@ -81,44 +100,58 @@ func (v *Validator) updatePCR(pcrIndex uint32, encoded string) error {
...
+		oldExpected := v.measurements[measurementIndex].Expected
+		expectedMeasurementSum := sha256.Sum256(append(oldExpected[:], hashedInput[:]...))
+		expectedMeasurement = expectedMeasurementSum[:]
+	case oid.QEMUTDX{}:
+		hashedInput := sha512.Sum384(decoded)
+		oldExpected := v.measurements[measurementIndex].Expected
+		expectedMeasurementSum := sha512.Sum384(append(oldExpected[:], hashedInput[:]...))
+		expectedMeasurement = expectedMeasurementSum[:]
+	default:
+		return fmt.Errorf("updating measurement: unknown attestation variant")
```
daniel-weisse, line 118:
```suggestion
		return fmt.Errorf("updating measurement [%d]: unknown attestation variant: %s", measurementIndex, v.attestationVariant.String())
```

cli/internal/cloudcmd/validators.go
```diff
@@ -81,44 +100,58 @@ func (v *Validator) updatePCR(pcrIndex uint32, encoded string) error {
...
+		v.measurements = gcpMeasurements
+	case oid.QEMUVTPM{}, oid.QEMUTDX{}:
+		qemuMeasurements := config.Provider.QEMU.Measurements
+		if len(qemuMeasurements) == 0 {
+			return errors.New("no expected QEMU measurements provided")
 		}
-		v.pcrs = qemuPCRs
+		v.measurements = qemuMeasurements
+	default:
+		return errors.New("unknown attestation variant")
```
daniel-weisse, line 154:
```suggestion
		return fmt.Errorf("selecting measurements: unknown attestation variant: %s", v.attestationVariant.String())
```
Consistent error messages for the same error

cli/internal/cloudcmd/validators_test.go
```diff
@@ -28,6 +30,7 @@ import (
 )
 
 func TestNewValidator(t *testing.T) {
+	// TODO: Add test case for TDX default RTMRs when we know how to handle default values in the config
```
daniel-weisse, line 33: This test doesn't use any default measurements though?
Why not just add a test case for qemu TDX and define a list of measurements with TDX measurement length for this test?

cli/internal/cloudcmd/validators.go
```diff
@@ -129,16 +162,16 @@ func (v *Validator) V(cmd *cobra.Command) atls.Validator {
 	return v.validator
 }
 
-// PCRS returns the validator's PCR map.
-func (v *Validator) PCRS() measurements.M {
-	return v.pcrs
+// GetMeasurements returns the validator's measurements map.
+func (v *Validator) GetMeasurements() measurements.M {
```
daniel-weisse, line 166:
```suggestion
// Measurements returns the validator's measurements map.
func (v *Validator) Measurements() measurements.M {
```
nit

cli/internal/cloudcmd/validators.go
daniel-weisse: Yeah, you are right. Validator uses `oid.Getter` interface which doesn't implement `String()`.
This will be reworked in the future.
Should be fine to just return a simple error message without naming the variant (which shouldn't happen in the first place if the `NewValidator` constructor is used)

cli/internal/cloudcmd/validators.go
Nirusu: see other comment

cli/internal/cloudcmd/validators.go
Nirusu: see other comment

daniel-weisse (approved):

PR #1518, opened 2023-03-24
cli: warn about missing support for upgrades on AWS, OpenStack, QEMU

derpsteb:
### Proposed change(s)

- The constellation-operator currently doesn't support the necessary operations for node upgrades on AWS, OpenStack and QEMU. Previously `upgrade apply` just executed on the aforementioned CSPs, no error reported. However, the nodes never successfully upgraded. Now the CLI prints a warning and does not apply any upgrades.
- Mention support in the docs.
- Ran [e2e-upgrade test on azure](https://github.com/edgelesssys/constellation/actions/runs/4509186713/jobs/7939036115) to see that the upgrade is still correctly triggered. Manually initiated an upgrade for an AWS cluster and go this output:
```
$ ./build/constellation upgrade apply --debug
WARNING: the config key `attestationVariant` is not set. This key will be required in the next version.
WARNING: Skipping service & image upgrades. Only supported for Azure and GCP.
You are about to change your cluster's measurements. Are you sure you want to continue? [y/n]: y
2023-03-24T09:29:56+01:00       DEBUG   cloudcmd/upgrade.go:207 Triggering measurements config map update now
Successfully updated the cluster's expected measurements
```

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/upgradeapply.go
+++ b/cli/internal/cmd/upgradeapply.go
@@ -15,6 +15,7 @@ import (
 	"github.com/edgelesssys/constellation/v2/cli/internal/cloudcmd"
 	"github.com/edgelesssys/constellation/v2/cli/internal/helm"
 	"github.com/edgelesssys/constellation/v2/internal/attestation/measurements"
+	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
 	"github.com/edgelesssys/constellation/v2/internal/compatibility"
 	"github.com/edgelesssys/constellation/v2/internal/config"
 	"github.com/edgelesssys/constellation/v2/internal/file"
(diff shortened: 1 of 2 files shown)


cli/internal/cmd/upgradeapply.go
```diff
@@ -80,23 +81,27 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, fileHandler file.Hand
...
+		switch {
+		case errors.Is(err, cloudcmd.ErrInProgress):
+			cmd.PrintErrln("Skipping image & Kubernetes upgrades. Another upgrade is in progress")
+		case errors.As(err, &upgradeErr):
+			cmd.PrintErrln(err)
+		case err != nil:
+			return fmt.Errorf("upgrading NodeVersion: %w", err)
+		}
+	} else {
+		cmd.PrintErrln("WARNING: Skipping service & image upgrades. Only supported for Azure and GCP.")
```
thomasten, line 104:
```suggestion
			cmd.PrintErrln("Skipping image and Kubernetes upgrades. Another upgrade is in progress.")
		case errors.As(err, &upgradeErr):
			cmd.PrintErrln(err)
		case err != nil:
			return fmt.Errorf("upgrading NodeVersion: %w", err)
		}
	} else {
		cmd.PrintErrln("WARNING: Skipping service and image upgrades, which are currently only supported for Azure and GCP.")
```

docs/docs/workflows/upgrade.md
```diff
@@ -6,6 +6,9 @@ You configure the desired versions in your local Constellation configuration and
 To learn about available versions you use the `upgrade check` command.
 Which versions are available depends on the CLI version you are using.
 
+Currently, Azure and GCP are supported for full upgrades.
+Please open an [issue](https://github.com/edgelesssys/constellation/issues/new?assignees=&labels=&template=feature_request.md&title=) or [contact us](https://github.com/edgelesssys/constellation#support) directly if you are looking for AWS or OpenStack support.
```
thomasten, line 10:
```suggestion
:::caution

Upgrades arent't yet implemented for AWS. If you require this feature, [let us know](https://github.com/edgelesssys/constellation/issues/new?assignees=&labels=&template=feature_request.md)!

:::
```
For consistency, I'd keep this in line with the existing hints on missing features.
We shouldn't mention OpenStack until we have real support for it.
Is it okay to just speak of "upgrades" instead of "full upgrades" or is a subset of it working on AWS?

derpsteb: Measurements can be changed. That's why I included "full". But I wasn't really happy with that either. I think just "upgrades" is fine since this page does not talk about changing measurements anyways.

thomasten (approved):

3u13r (approved):
LGTM

PR #1538, opened 2023-03-28
internal: refactor oid package to variant package

daniel-weisse:
### Proposed change(s)

- Rename `oid` package to `variant`
- Add new `Variant` interface:
  - embeds the `Getter` interface
  - has a `String() string` method to return the string representation of the variant
  - has an `Equal(other Getter) bool` method to check if the OID of `other` is equal to the one of the interface


3u13r (approved):
LGTM

malt3 (approved):

PR #1539, opened 2023-03-28
config: deprecate confidentialVM option for Azure clusters in favor of using attestationVariant option 

daniel-weisse:
### Proposed change(s)

- Remove usage of `confidentialVM` option in Azure provider config in favor of using `attestationVariant`
  - `confidentialVM` has been marked as deprecated, setting it has no effect but will print a warning about future removal

### Additional info
- Should be merged to main after https://github.com/edgelesssys/constellation/pull/1538


3u13r (approved):
LGTM

thomasten (approved):

PR #1540, opened 2023-03-28
cli: set cluster's measurements from user's config using Helm instead of loading them from node on boot

daniel-weisse:
### Proposed change(s)

A cluster's expected measurements are set in a user's config.
These measurements are set in the cluster only during an upgrade using `constellation upgrade apply`.
The initial measurements of a cluster a loaded by the bootstrapper from the initial control-plane node, we only apply the list of "enforced measurements" using values received from the init request.

This PR aligns the behavior of `constellation init` with the behavior of `constellation upgrade apply` by setting the initial measurements of a cluster using the expected values from a user's config.
This happens by setting them in the `JoinService` helm chart.

### Additional info
- Azure e2e test: https://github.com/edgelesssys/constellation/actions/runs/4544562555

### Checklist

- [ ] Link to Milestone


3u13r (approved):
LGTM

PR #1541, opened 2023-03-29
config: remove deprecated upgradeConfig and require name and microserviceVersion fields

daniel-weisse:
### Proposed change(s)

- Remove `upgradeConfig`  from config
  - This field was only partially functional in v2.5.0 and has been deprecated in v2.6.0
  - Setting it in v2.7.0 will return an error
- Require `name` field
  - Was optional in v2.6.0 with a warning for it to be required in the next version
- Required `microserviceVersion` field
  - This field was optional in v2.6.0
  - There was no hint about it being required in future versions, we might want to consider keeping it optional in v2.7.0 with a warning
- Removed a warning about unset `attestationVariant` field
  - This field was not yet present in v2.6.0
  - ~This field will be superseded in v2.8.0 by the `attestation` struct, as outlined in https://github.com/edgelesssys/constellation/pull/1436 and therefore won't throw a warning for not being set~
  - The value is no longer written to a users config on `constellation config generate` since it will be superseded in v2.8.0 by the `attestation` struct, as outlined in https://github.com/edgelesssys/constellation/pull/1436. We only use the default values (or the value specified by a users if they manually edited the config)

### DIFF
--- a/cli/internal/cmd/BUILD.bazel
+++ b/cli/internal/cmd/BUILD.bazel
@@ -63,7 +63,6 @@ go_library(
         "//internal/logger",
         "//internal/retry",
         "//internal/sigstore",
-        "//internal/variant",
         "//internal/versions",
         "//internal/versionsapi",
         "//internal/versionsapi/fetcher",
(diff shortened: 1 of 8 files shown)


derpsteb:
Smoketest: [Azure, 1+2, quick](https://github.com/edgelesssys/constellation/actions/runs/4572597252)

derpsteb (approved):
Changes lgtm :+1:

PR #1544, opened 2023-03-29
dev-docs: etcd disaster recovery

Nirusu:
### Proposed change(s)
- Add dev-docs documentation on how to do etcd disaster recovery

I am still not 100% sure about what a good way would be to approach the last step. The restored control plane still holds the manually uploaded VHD, while all other newly created VMs will have no data disk if you want to scale up again. The problem is that while the one VM holds a disk in LUN 0 (which is hard-coded in the disk-mapper), we _cannot_ add a VMSS global disk definition on LUN 0.

 To get back to the original state (which honestly might not be super desirable given that these disks disappear), we would somehow:

1. Create a new disk in the whole VMSS in LUN 1 or higher
2. Upgrade the control plane node to the newest VMSS state
3. Clone the disk onto the VMSS state (difficulty: doing this while the cluster is online?)
4. Turn off the control plane
5. Remove the attached disk
6. Change the LUN from the new VMSS disk we cloned the VHD onto to LUN 0 (difficulty: is this possible without losing data and not ending up with a VM specific disk again?)
7. Start the control plane again
8. Perform Constellation recovery once

I haven't played this though (and honestly don't want to), but I think this is what it *roughly* would look like. This isn't really needed if you just need one control plane, though. But if you have multiple ones, this likely would need to be done in some way without modifying the disk-mapper.


derpsteb:
Held by other tasks. Will test the documentation asap.

dev-docs/howto/etcd-disaster-recovery.md
```diff
@@ -0,0 +1,139 @@
...
+```
+*(replace name & IP with the name and the private IP of the remaining control plane VM you are to perform the restore process on - this information can be found in the Azure portal)*
+
+10. Copy the contents of the newly created recovery directory to the mounted state disk and remove any remaining old files. 
+**Make sure the permissions are the same as before!**
+
+11. Unmount the partition:
+```bash 
+sudo umount /your/mount/path
+sudo luksClose constellation-state
```
msanft:
```suggestion
sudo cryptsetup luksClose constellation-state
```

PR #1442, opened 2023-03-17
docs: list minimal permissions for Constellation setup

msanft:
### Proposed change(s)

- List the minimal permissions to be able to execute the `constellation iam` commands (`iam create` & `iam destroy`) for each CSP.

### Checklist

- [x] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)
- [x] Add labels (e.g., for changelog category)
- [x] Link to Milestone

PR #1555, opened 2023-03-30
ci: separate e2e permissions

msanft:
### Proposed change(s)

- Use the minimal e2e permissions (documented in #1442) in the e2e tests to be aligned with documentation and notice changes in permission requirements at test time.
  - e2e permissions are now managed in [edgeless-iam](https://github.com/edgelesssys/edgeless-iam/tree/feat/seperate-iam-accounts)
- Use [ad-hoc IAM creation](https://docs.edgeless.systems/constellation/workflows/config#creating-an-iam-configuration) (the way we recommend to users in the docs) for all CSPs, including Azure. With that, manual resource group creation / deletion is obsolote.

### Successful test runs
- [Manual e2e test: AWS](https://github.com/edgelesssys/constellation/actions/runs/4561618154/jobs/8047791821)
- [Manual e2e test: Azure](https://github.com/edgelesssys/constellation/actions/runs/4561611430/jobs/8047775473)
- [Manual e2e test: GCP](https://github.com/edgelesssys/constellation/actions/runs/4561617079/jobs/8047788746)
- [Daily e2e test](https://github.com/edgelesssys/constellation/actions/runs/4561612340/jobs/8050253615)
- [Weekly e2e test](https://github.com/edgelesssys/constellation/actions/runs/4561846874)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/actions/e2e_test/action.yml
+++ b/.github/actions/e2e_test/action.yml
@@ -35,10 +35,13 @@ inputs:
   gcpProject:
     description: "The GCP project to deploy Constellation in."
     required: false
-  gcp_service_account:
-    description: "Service account with permissions to create Constellation on GCP."
+  gcpIAMServiceAccount:
+    description: "Service account with permissions to create IAM configuration on GCP."
     required: false
-  gcpClusterServiceAccountKey:
+  gcpClusterServiceAccount:
+    description: "Service account with permissions to create a Constellation cluster on GCP."
+    required: false
+  gcpInClusterServiceAccountKey:
     description: "Service account to use inside the created Constellation cluster on GCP."
     required: false
   awsOpenSearchDomain:
@@ -56,20 +59,26 @@ inputs:
   azureTenant:
     description: "The Azure tenant ID to deploy Constellation in."
     required: false
-  azureClientID:
-    description: "The client ID of the application registration created for Constellation in Azure."
+  azureIAMClientID:
+    description: "The client ID of the application registration created for Constellation in Azure. (IAM creation)"
     required: false
-  azureClientSecret:
-    description: "The client secret value of the used secret"
+  azureIAMClientSecret:
+    description: "The client secret value of the used secret. (IAM creation)"
     required: false
-  azureUserAssignedIdentity:
-    description: "The Azure user assigned identity to use for Constellation."
+  azureClusterClientID:
+    description: "The client ID of the application registration created for Constellation in Azure. (cluster creation)"
     required: false
-  azureResourceGroup:
-    description: "The resource group to use"
+  azureClusterClientSecret:
+    description: "The client secret value of the used secret. (cluster creation)"
     required: false
+  azureClusterCredentials:
+    description: "Azure credentials authorized to create a Constellation cluster."
+    required: true
+  azureIAMCredentials:
+    description: "Azure credentials authorized to create an IAM configuration."
+    required: true
   test:
-    description: "The test to run. Can currently be one of [sonobuoy full, sonobuoy quick, autoscaling, lb, perf-bench, verify, recover, nop, iamcreate]."
+    description: "The test to run. Can currently be one of [sonobuoy full, sonobuoy quick, autoscaling, lb, perf-bench, verify, recover, nop]."
     required: true
   sonobuoyTestSuiteCmd:
     description: "The sonobuoy test suite to run."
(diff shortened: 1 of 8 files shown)


.github/workflows/e2e-test-weekly.yml
katexochen, line 1: Line 102: We should pick another test to run on the latest release, sonobuoy full would be a good choice. Currently this wouldn't run any tests...

katexochen (changes requested):

.github/actions/e2e_test/action.yml
```diff
@@ -50,26 +53,14 @@ inputs:
...
-  azureClientSecret:
-    description: "The client secret value of the used secret"
-    required: false
-  azureUserAssignedIdentity:
-    description: "The Azure user assigned identity to use for Constellation."
-    required: false
-  azureResourceGroup:
-    description: "The resource group to use"
-    required: false
+  azureClusterCredentials:
```
3u13r, line 56: Same as above

.github/workflows/e2e-test-manual.yml
```diff
@@ -269,16 +244,10 @@ jobs:
           kubeconfig: ${{ steps.e2e_test.outputs.kubeconfig }}
 
       - name: Always delete IAM configuration
-        if: always() && inputs.test == 'iamcreate' && inputs.cloudProvider != 'azure' # skip for Azure, as the SP / MI does not have the required permissions
+        if: always()
+        continue-on-error: true
```
3u13r, line 248: I think this is needed for `name: Always terminate cluster` above but not strictly for this, since it is most likely the last step, right?

.github/actions/e2e_test/action.yml
```diff
@@ -50,26 +53,14 @@ inputs:
...
+    description: "Azure credentials authorized to create a Constellation cluster."
+    required: true
+  azureIAMCredentials:
```
3u13r, line 59: Maybe `azureCICredentials` works better as all credentials are "IAM" credentials, so it is not totally obvious that we mean IAM creation.

3u13r (changes requested):

.github/actions/e2e_test/action.yml
```diff
@@ -35,10 +35,13 @@ inputs:
   gcpProject:
     description: "The GCP project to deploy Constellation in."
     required: false
-  gcp_service_account:
-    description: "Service account with permissions to create Constellation on GCP."
+  gcpIAMServiceAccount:
+    description: "Service account with permissions to create IAM configuration on GCP."
     required: false
-  gcpClusterServiceAccountKey:
+  gcpClusterServiceAccount:
```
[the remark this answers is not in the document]
msanft: I see this point as well - hence the renaming of the in-cluster service account to `gcpInClusterServiceAccount`. The "same permissions as a user"-point certainly is true, but both the cluster-creation as well as the IAM-creation service accounts have permissions that would normally apply to a user. Therefore I would prefer other naming. Maybe something like `...clusterCreate...` and `...IAMCreate...` works better? (same goes for the other CSPs)

.github/actions/e2e_test/action.yml
msanft: s.a.

.github/workflows/e2e-test-manual.yml
3u13r: Yes, but my main point was that it is missing for the `name: Always terminate cluster` step above, right?

.github/actions/e2e_test/action.yml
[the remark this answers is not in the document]
3u13r: Sounds good to me.

.github/workflows/e2e-test-manual.yml
msanft: I think this is due to the fact that a failure on cluster termination needs to be resolved manually?

3u13r: After a discussion with @katexochen, I think that we are using the `continue-on-error` in too many cases. Since we want to know if this deletion fails, I'd go with removing is here.

.github/workflows/e2e-test-weekly.yml
3u13r: After discussion with @katexochen we used the `iamcreate` test as the only test to verify that the last release still works, everything else is excluded by the list after line 102.  If we remove the iamcreate test, we should execute the e.g. `verify` test to mainly check cluster creation and initialization.

3u13r (approved):
LGTM

PR #1561, opened 2023-03-30
internal: use config to create attestation validators

daniel-weisse:
### Proposed change(s)

- add config structs for creating attestation Validators
-

### Checklist

- [ ] Link to Milestone


internal/config/attestation.go
```diff
@@ -0,0 +1,81 @@
+/*
+Copyright (c) Edgeless Systems GmbH
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+// Package config defines types and interfaces to use for configuring attestation Issuers and Validators.
```
thomasten, line 7:
```suggestion
```

internal/config/attestation.go
```diff
@@ -0,0 +1,81 @@
...
+// MarshalJSON marshals the certificate to PEM.
+func (c Certificate) MarshalJSON() ([]byte, error) {
+	pem := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: c.Raw})
+	return json.Marshal(string(pem))
+}
+
+// MarshalYAML marshals the certificate to PEM.
+func (c Certificate) MarshalYAML() (any, error) {
+	pemData := &bytes.Buffer{}
+	if err := pem.Encode(pemData, &pem.Block{Type: "CERTIFICATE", Bytes: c.Raw}); err != nil {
```
thomasten, line 41: Why do we use different funcs for PEM encoding here?

internal/config/attestation.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	*c = Certificate(*cert)
+	return nil
+}
+
+func mustParsePEM(data string) Certificate {
+	jsonData := fmt.Sprintf("\"%s\"", data)
+	var cert Certificate
+	err := json.Unmarshal([]byte(jsonData), &cert)
+	if err != nil {
```
thomasten, line 77:
```suggestion
	if err := json.Unmarshal([]byte(jsonData), &cert); err != nil {
```

internal/config/config.go
```diff
@@ -638,6 +639,148 @@ func (c *Config) Validate(force bool) error {
...
+	bootloaderVersion = 2
+	teeVersion        = 0
+	snpVersion        = 6
+	microcodeVersion  = 93
```
thomasten, line 648: These consts feel a bit lost here. You could move them into DefaultForAzureSEVSNP. If you want them here, you should prefix them to make clear that they belong to (Azure) SNP.

internal/config/config.go
```diff
@@ -638,6 +639,148 @@ func (c *Config) Validate(force bool) error {
...
+}
+
+// SNPFirmwareSignerConfig is the configuration for validating the firmware signer.
+type SNPFirmwareSignerConfig struct {
+	// description: |
+	//   List of accepted values for the firmware signing key digest.\nValues are enforced according to the 'enforcementPolicy'\n     - 'equal'       : Error if the reported signing key digest does not match any of the values in 'acceptedKeyDigests'\n     - 'maaFallback' : Use 'equal' checking for validation, but fallback to using Microsoft Azure Attestation (MAA) for validation if the reported digest does not match any of the values in 'acceptedKeyDigests'. See the Azure docs for more details: https://learn.microsoft.com/en-us/azure/attestation/overview#amd-sev-snp-attestation\n     - 'warnOnly'    : Same as 'equal', but only prints a warning instead of returning an error if no match is found
+	AcceptedKeyDigests idkeydigest.List `json:"acceptedKeyDigests" yaml:"acceptedKeyDigests"`
+	// description: |
+	//   Key digest enforcement policy. One of {'equal', 'maaFallback', 'warnOnly'}
+	EnforcementPolicy idkeydigest.Enforcement `json:"enforcementPolicy" yaml:"enforcementPolicy" validate:"required"`
```
thomasten, line 727: should the enforcementPolicy explanations be moved to the description of EnforcementPolicy?

internal/config/config.go
```diff
@@ -638,6 +639,148 @@ func (c *Config) Validate(force bool) error {
...
+	// description: |
+	//   URL of the Microsoft Azure Attestation (MAA) instance to use for fallback validation. Only used if 'enforcementPolicy' is set to 'maaFallback'.
+	MAAURL string `json:"maaURL,omitempty" yaml:"maaURL,omitempty"`
```
thomasten, line 730: the value is ignored if the user sets it in their config file, right?

daniel-weisse: Yes, we only use the value from the ID file.
I could add a check tho to explicitly not allow a set value here

internal/config/config.go
daniel-weisse: With how the docgen tool works, setting the explanations on `EnforcementPolicy` results in the doc comment being behind, and under the entry in the config.
Unless there is an option to explicitly move the doc comment above the `EnforcementPolicy` key, I would keep it as is.

How I would like it to look:
```yaml
# ...
someKey: someValue # Some doc comment explaining the value
# First line of the enforcementPolicy doc comment
# Second line of the enforcementPolicy doc comment
# Third line of the enforcementPolicy doc comment
enforcementPolicy: MaaFallback
anotherKey: "anotherValue" # Another doc comment explaining a value
# ...
```

How it actually looks:
```yaml
# ...
someKey: someValue # Some doc comment explaining the value
enforcementPolicy: MaaFallback # First line of the enforcementPolicy doc comment
# Second line of the enforcementPolicy doc comment
# Third line of the enforcementPolicy doc comment
anotherKey: "anotherValue" # Another doc comment explaining a value
# ...
```

internal/config/attestation.go
daniel-weisse: You're right, this doesn't make a lot of sense. Will rewrite to just use `pem.EncodeToMemory` instead

internal/config/config.go
thomasten: I agree. I expected it would result in what you would like it to look. (It actually looks like this for acceptedKeyDigests, maybe because of its type. But I guess you know that already.)

internal/config/config.go
thomasten: You mean something like `validate:"len=0"`?
And regarding the description: Is there any case where a user would see it? For devs, the description is good.

daniel-weisse: The description will only show up if we set a default value (which we dont do)
`validate:"len=0"` sounds like a good idea

derpsteb:
- [Smoketest Azure](https://github.com/edgelesssys/constellation/actions/runs/4625928199)
- [Smoketest GCP](https://github.com/edgelesssys/constellation/actions/runs/4625931891)

thomasten (approved):

PR #1564, opened 2023-03-31
internal: semver support for pseudoversions

msanft:
### Proposed change(s)

- Add support for `v.1.19.2-pre`-type pseudoversions in the semver package as specified in the [semver standard](https://semver.org/).
- Prerelease-info is being put to respect when comparing semvers.

### Related issue
- [AB#3017](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3017/)

### Additional info
- Implementation using `strings.Contains` is probably not perfect, but I didn't found a way to nicely scan for "optional" characters with just `fmt.Sscanf`. Maybe Regex would be the next best option if the strings check is not appropriate.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/semver/semver.go
+++ b/internal/semver/semver.go
@@ -60,10 +60,11 @@ func New(version string) (Semver, error) {
 
 // String returns the string representation of the version.
 func (v Semver) String() string {
+	version := fmt.Sprintf("v%d.%d.%d", v.Major, v.Minor, v.Patch)
 	if v.Prerelease != "" {
-		return fmt.Sprintf("v%d.%d.%d-%s", v.Major, v.Minor, v.Patch, v.Prerelease)
+		return fmt.Sprintf("%s-%s", version, v.Prerelease)
 	}
-	return fmt.Sprintf("v%d.%d.%d", v.Major, v.Minor, v.Patch)
+	return version
 }
 
 // Compare compares two versions. It relies on the semver.Compare function internally.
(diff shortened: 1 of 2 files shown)


internal/semver/semver.go
```diff
@@ -39,20 +40,29 @@ func New(version string) (Semver, error) {
...
+		Prerelease: pre,
 	}, nil
 }
 
 // String returns the string representation of the version.
 func (v Semver) String() string {
+	if v.Prerelease != "" {
+		return fmt.Sprintf("v%d.%d.%d-%s", v.Major, v.Minor, v.Patch, v.Prerelease)
+	}
 	return fmt.Sprintf("v%d.%d.%d", v.Major, v.Minor, v.Patch)
```
daniel-weisse, line 66:
```suggestion
	version := fmt.Sprintf("v%d.%d.%d", v.Major, v.Minor, v.Patch)
	if v.Prerelease != "" {
		return fmt.Sprintf("%s-%s", version, v.Prerelease)
	}
	return version
```
nit: less error prone if we change things around

internal/semver/semver.go
```diff
@@ -39,20 +40,29 @@ func New(version string) (Semver, error) {
 	version = semver.Canonical(version)
 
 	var major, minor, patch int
+	var pre string
+	if strings.Contains(version, "-") {
+		parts := strings.Split(version, "-")
+		pre = parts[1]
+	}
```
daniel-weisse, line 47:
```suggestion
	_, pre, _ := strings.Cut(version, "-")
```

daniel-weisse (approved):

PR #1593, opened 2023-04-03
ci: reduce number of steps with continue-on-error

katexochen:
### Checklist

### DIFF
--- a/.github/actions/constellation_create/action.yml
+++ b/.github/actions/constellation_create/action.yml
@@ -268,6 +268,7 @@ runs:
 
     - name: Upload boot logs
       if: always() && !env.ACT
+      continue-on-error: true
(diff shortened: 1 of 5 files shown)


3u13r:
I think we should also remote it from `actions/constellation_destroy/actions.yml:14`, since otherwise we leak might leak cloud resources without knowing about it. The successful deletion of those resources can also be considered part of the functionality which should be tested.

Same with `workflows/e2e-test-release.yml:176`

3u13r (changes requested):
I think we should also remote it from `actions/constellation_destroy/actions.yml:14`, since otherwise we leak might leak cloud resources without knowing about it. The successful deletion of those resources can also be considered part of the functionality which should be tested.

Same with `workflows/e2e-test-release.yml:176`

PR #1609, opened 2023-04-04
versionsapi: fix list command by allowing empty set of patch versions for a given minor version

malt3:
### Proposed change(s)

- versionsapi: fix list command by allowing empty set of patch versions for a given minor version

### Checklist

### DIFF
--- a/internal/versionsapi/cli/list.go
+++ b/internal/versionsapi/cli/list.go
@@ -12,12 +12,13 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/edgelesssys/constellation/v2/internal/logger"
-	"github.com/edgelesssys/constellation/v2/internal/versionsapi"
-	verclient "github.com/edgelesssys/constellation/v2/internal/versionsapi/client"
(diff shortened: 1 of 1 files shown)


internal/versionsapi/cli/list.go
```diff
@@ -137,10 +138,6 @@ func listPatchVersions(ctx context.Context, client *verclient.Client, ref string
 			return nil, fmt.Errorf("listing patch versions: %w", err)
 		}
 
-		if len(list.Versions) == 0 {
-			return nil, fmt.Errorf("no versions found")
-		}
-
```
katexochen, line 143: This function is also called within the remove command, did you consider that? Look good nevertheless, but would we great if you could check this, too, if you didn't already. :smile_cat:

katexochen (approved):

internal/versionsapi/cli/list.go
malt3: I'm quite sure this still works correctly but I'll run a quick test to be sure!

PR #1617, opened 2023-04-05
bazel-deps-mirror: upgrade command

malt3:
This command can be used to upgrade a dependency.
Users are supposed to replace any upstream URLs and run the upgrade command. It replaces the expected hash and uploads the new dep to the mirror.

### Proposed change(s)

- bazel-deps-mirror: upgrade command
- bazel-deps-mirror: improve fix command to add hash if missing

### DIFF
--- a/hack/bazel-deps-mirror/upgrade.go
+++ b/hack/bazel-deps-mirror/upgrade.go
@@ -135,11 +135,10 @@ func upgradeRule(ctx context.Context, mirrorUpload mirrorUploader, rule *build.R
 	log.Debugf("Upgrading rule: %s", rule.Name())
 
 	upstreamURLs, err := rules.UpstreamURLs(rule)
-	if err != nil {
-		if errors.Is(err, rules.ErrNoUpstreamURL) {
-			log.Debugf("Rule has no upstream URL. Skipping.")
-			return
-		}
+	if errors.Is(err, rules.ErrNoUpstreamURL) {
+		log.Debugf("Rule has no upstream URL. Skipping.")
+		return
+	} else if err != nil {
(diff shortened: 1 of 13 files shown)


hack/bazel-deps-mirror/upgrade.go
```diff
@@ -0,0 +1,222 @@
...
+
+	upstreamURLs, err := rules.UpstreamURLs(rule)
+	if err != nil {
+		if errors.Is(err, rules.ErrNoUpstreamURL) {
+			log.Debugf("Rule has no upstream URL. Skipping.")
+			return
+		}
+		iss = append(iss, err)
+		return
+	}
```
katexochen, line 145:
```suggestion
	if errors.Is(err, rules.ErrNoUpstreamURL) {
		log.Debugf("Rule has no upstream URL. Skipping.")
		return
	} else if err != nil {
		iss = append(iss, err)
		return
	}
```

hack/bazel-deps-mirror/upgrade.go
```diff
@@ -0,0 +1,222 @@
...
+	log.Debugf("Upgrading rule: %s", rule.Name())
+
+	upstreamURLs, err := rules.UpstreamURLs(rule)
+	if err != nil {
+		if errors.Is(err, rules.ErrNoUpstreamURL) {
+			log.Debugf("Rule has no upstream URL. Skipping.")
+			return
+		}
+		iss = append(iss, err)
+		return
```
katexochen, line 144: Please don't use naked returns.

katexochen:
Works great! :rock:

katexochen (approved):

PR #1618, opened 2023-04-05
bootstrapper: collect journald logs on failure

miampf:
### Proposed change(s)

- add a gRPC endpoint into the bootstrapper to query encrypted logs from failed machines
- this happens with server side streaming, the server sends individual batches of a bit over 1KiB of data

### Checklist

- [x] Link to Milestone


bootstrapper/internal/initserver/initserver.go
```diff
@@ -187,12 +205,83 @@ func (s *Server) Init(ctx context.Context, req *initproto.InitRequest) (*initpro
...
+		return status.Errorf(codes.Internal, "invalid init secret: %s", err)
+	}
+
+	logPipe, err := s.journaldCollector.Start()
+	if err != nil {
+		return status.Errorf(codes.Internal, "failed starting the log collector: %s", err)
+	}
+
+	// decode the master secret
+	cfg, err := uri.DecodeMasterSecretFromURI(s.kmsURI)
```
3u13r, line 232: I don't think we should assume that the `kmsURI` contains a master secret, since it could also contain various different kms endpoints. We either need to re-think the crypto or we need to setup the KMS as done in `Init()` in line 146.

3u13r (changes requested):

bootstrapper/internal/initserver/initserver.go
miampf: I don't see a way in init which derives a key that the client could generate as well. We could use the init secret for deriving a key instead of the master secret, or use asymetric cryptography for a key exchange, at least those are the only two options that came to my mind. What do you think would work best/do you have a different idea which could work?

miampf:
Since there were concerns about reaching the node through a node balancer, this has to be rewritten

3u13r (changes requested):

3u13r (approved):
LGTM
You can either decide to merge it into main, or leave this as approved and direct the client-side PR to this branch and then merge the whole feature at the end. 

The second option is a little bit cleaner, but this also means more rebasing work. I'm fine with both options, so I leave it up to you.

PR #1620, opened 2023-04-06
cli: add Terraform log support

msanft:
### Proposed change(s)

- Add support for [Terraform logging](https://developer.hashicorp.com/terraform/internals/debugging) by passing the `--tf-log` flag from the CLI to the Terraform client.
  - The flag can be set to `NONE` (equals to not setting the flag), `ERROR`, `WARN`, `INFO`, `DEBUG`, `TRACE`, or `JSON` (JSON-Formatted logs at `TRACE` level)
 
### Additional info
- The [terraform-exec module](https://pkg.go.dev/github.com/hashicorp/terraform-exec@v0.18.1/tfexec#Terraform.SetLog) only allows logging to files as logs are very verbose and could leak secret values. It also does not provide streaming logs but only writes all logs to the file at once after the full apply / init / ... has finished

### Checklist

- [x] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs) #1627 
- [x] Link to Milestone

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -48,6 +48,7 @@ func NewRootCmd() *cobra.Command {
 
 	rootCmd.PersistentFlags().Bool("debug", false, "enable debug logging")
 	rootCmd.PersistentFlags().Bool("force", false, "disable version compatibility checks - might result in corrupted clusters")
+	rootCmd.PersistentFlags().String("tf-log", "NONE", "sets the Terraform log level (default \"NONE\" - no logs)")
 
 	rootCmd.AddCommand(cmd.NewConfigCmd())
 	rootCmd.AddCommand(cmd.NewCreateCmd())
--- a/cli/internal/cloudcmd/clients.go
+++ b/cli/internal/cloudcmd/clients.go
@@ -23,9 +23,9 @@ type imageFetcher interface {
 
 type terraformClient interface {
 	PrepareWorkspace(path string, input terraform.Variables) error
-	CreateCluster(ctx context.Context) (terraform.CreateOutput, error)
-	CreateIAMConfig(ctx context.Context, provider cloudprovider.Provider) (terraform.IAMOutput, error)
-	Destroy(ctx context.Context) error
+	CreateCluster(ctx context.Context, logLevel terraform.LogLevel) (terraform.CreateOutput, error)
+	CreateIAMConfig(ctx context.Context, provider cloudprovider.Provider, logLevel terraform.LogLevel) (terraform.IAMOutput, error)
+	Destroy(ctx context.Context, logLevel terraform.LogLevel) error
 	CleanUpWorkspace() error
 	RemoveInstaller()
 	Show(ctx context.Context) (*tfjson.State, error)
--- a/cli/internal/cmd/iamdestroy.go
+++ b/cli/internal/cmd/iamdestroy.go
@@ -11,6 +11,7 @@ import (
 	"os"
 
 	"github.com/edgelesssys/constellation/v2/cli/internal/cloudcmd"
+	"github.com/edgelesssys/constellation/v2/cli/internal/terraform"
 	"github.com/edgelesssys/constellation/v2/internal/cloud/gcpshared"
 	"github.com/edgelesssys/constellation/v2/internal/constants"
 	"github.com/edgelesssys/constellation/v2/internal/file"
@@ -56,9 +57,14 @@ type destroyCmd struct {
 }
 
 func (c *destroyCmd) iamDestroy(cmd *cobra.Command, spinner spinnerInterf, destroyer iamDestroyer, fsHandler file.Handler) error {
+	flags, err := c.parseDestroyFlags(cmd)
+	if err != nil {
+		return fmt.Errorf("parsing flags: %w", err)
+	}
+
 	// check if there is a possibility that the cluster is still running by looking out for specific files
 	c.log.Debugf("Checking if %q exists", constants.AdminConfFilename)
-	_, err := fsHandler.Stat(constants.AdminConfFilename)
+	_, err = fsHandler.Stat(constants.AdminConfFilename)
 	if !errors.Is(err, os.ErrNotExist) {
 		return fmt.Errorf("file %q still exists, please make sure to terminate your cluster before destroying your IAM configuration", constants.AdminConfFilename)
 	}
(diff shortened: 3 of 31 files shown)


cli/cmd/root.go
```diff
@@ -48,6 +48,7 @@ func NewRootCmd() *cobra.Command {
 
 	rootCmd.PersistentFlags().Bool("debug", false, "enable debug logging")
 	rootCmd.PersistentFlags().Bool("force", false, "disable version compatibility checks - might result in corrupted clusters")
+	rootCmd.PersistentFlags().String("tf-log", "NONE", "sets the Terraform log level (default \"NONE\" - no logs)")
```
daniel-weisse, line 51: Since only (iam) create and terminate, and mini-Constellation actually use terraform, do we actually want this as a global flag?

cli/internal/terraform/terraform.go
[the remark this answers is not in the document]
msanft: If we reuse a client for separate actions in 1 CLI invocation, it would stick to the same log level and we would still have to pass the log level to the creation of the terraform client, which in our current command architecture usually happens before flag parsing

If theres a big benefit with setting it directly, we can surely change that but I didn't find any (also thought about doing it like you said while implementing)

cli/cmd/root.go
msanft: We can do that, but as 4 commands were using it i was unsure where to put it

cli/internal/terraform/terraform.go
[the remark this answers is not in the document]
daniel-weisse: Ah I see the problem with setting it in the constructor, maybe a separate `SetLogLevel()` method would work?

>If we reuse a client for separate actions in 1 CLI invocation, it would stick to the same log level 

IMO, this should be the preferred case, no? Or do you see any possibilities where one would dynamically adjust the log level during one cli call?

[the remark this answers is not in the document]
msanft: Only for possible future uses of Terraform - which we dont have right now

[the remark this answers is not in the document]
msanft: Conclusion of our in-person talk:

Both passing the log level to the Terraform Client constructor as well as defining a method to update the client in-place add too much refactoring overhead to be considered in this PR. For now, pass down the levels as it is in this PRs current implementation and refactor later on in a separate PR

cli/internal/cloudcmd/clients.go
```diff
@@ -23,9 +23,9 @@ type imageFetcher interface {
 
 type terraformClient interface {
 	PrepareWorkspace(path string, input terraform.Variables) error
-	CreateCluster(ctx context.Context) (terraform.CreateOutput, error)
-	CreateIAMConfig(ctx context.Context, provider cloudprovider.Provider) (terraform.IAMOutput, error)
-	Destroy(ctx context.Context) error
+	CreateCluster(ctx context.Context, logLevel terraform.LogLevel) (terraform.CreateOutput, error)
+	CreateIAMConfig(ctx context.Context, provider cloudprovider.Provider, logLevel terraform.LogLevel) (terraform.IAMOutput, error)
+	Destroy(ctx context.Context, logLevel terraform.LogLevel) error
```
katexochen, line 28: This shouldn't change the interface of those calls. Instead, pass the log level to the `terraform.New` constructor and store it in the struct

cli/internal/terraform/terraform.go
```diff
@@ -33,10 +36,70 @@ import (
...
+		return "DEBUG"
+	case LogLevelTrace:
+		return "TRACE"
+	case LogLevelJSON:
+		return "JSON"
+	default:
+		return ""
+	}
+}
+
```
katexochen, line 102: Maybe this could be done in a new file so the more important code of the client is easier to find.

katexochen: Also you might want to use stringer for this.

cli/internal/cmd/iamdestroy.go
```diff
@@ -155,3 +155,32 @@ func (c *destroyCmd) deleteGCPServiceAccountKeyFile(cmd *cobra.Command, destroye
...
+type destroyFlags struct {
+	yes        bool
+	tfLogLevel terraform.LogLevel
+}
+
+// parseDestroyFlags parses the flags of the create command.
+func (c *destroyCmd) parseDestroyFlags(cmd *cobra.Command) (destroyFlags, error) {
+	yes, err := cmd.Flags().GetBool("yes")
+	if err != nil {
+		return destroyFlags{}, fmt.Errorf("%w; Set '-yes' without a value to automatically confirm", err)
```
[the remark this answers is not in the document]
msanft: I just moved the exact thing here from above, but I agree to reword it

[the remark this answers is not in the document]
msanft: I just moved the exact thing here from above, but I agree to reword it

cli/internal/terraform/terraform.go
msanft: [This one here](https://pkg.go.dev/golang.org/x/tools/cmd/stringer)?

msanft: This would mean we'd have to rename the constants to their string value though which I think isn't very intuitive when calling them from other packages (e.g. `terraform.NONE`; What indicates that this is a log level?) 

And we can't change the string values to `"LogLevelNone"` etc. as they come from Terraform

What do you think?

cli/internal/terraform/terraform.go
```diff
@@ -83,18 +85,27 @@ func (c *Client) PrepareWorkspace(path string, vars Variables) error {
...
-func (c *Client) CreateCluster(ctx context.Context) (CreateOutput, error) {
+func (c *Client) CreateCluster(ctx context.Context, logLevel LogLevel) (CreateOutput, error) {
+	if logLevel.String() != "" {
+		if err := c.tf.SetLog(logLevel.String()); err != nil {
+			return CreateOutput{}, fmt.Errorf("set log level %s: %w", logLevel.String(), err)
+		}
+		if err := c.tf.SetLogPath(filepath.Join("..", constants.TerraformLogFile)); err != nil {
+			return CreateOutput{}, fmt.Errorf("set log path: %w", err)
+		}
+	}
```
daniel-weisse, line 96: Make this into its own function?

```Go
func (c *Client) setLogLevel(logLevel LogLevel) error {
	if logLevel.String() != "" {
		if err := c.tf.SetLog(logLevel.String()); err != nil {
			return fmt.Errorf("set log level %s: %w", logLevel.String(), err)
		}
		if err := c.tf.SetLogPath(filepath.Join("..", constants.TerraformLogFile)); err != nil {
			return fmt.Errorf("set log path: %w", err)
		}
	}
	return nil
}
```

Maybe add additional error handling in case `logLevel.String() == ""`

daniel-weisse (approved):

PR #1630, opened 2023-04-12
cli: fix misleading error while applying kubernetes-only upgrade

derpsteb:
### Proposed change(s)

- Currently `upgrade apply` will show a misleading error message when the config specifies an invalid image upgrade and a valid Kubernetes upgrade. In that case the upgrade is applied as expected. However, the CLI prints an error like this: `Error: upgrading NodeVersion: expected NodeVersion to contain <new image reference>, got <current image reference>`. This patch fixes the check.
- Add unittest to check the errors are actually triggered if apply responds with an unexpected message
- Do not verify measurements during e2e test. If we are upgrading to release images we can't verify the release signatures with the dev pub key in the CLI.

### Additional info
- Test run: [GCP, 1+1, 2.6 --> main](https://github.com/edgelesssys/constellation/actions/runs/4687145738/jobs/8306066558) :green_circle:. Executes only kubernetes upgrade 1.25.7 -> 1.26.3.
- A backport would be nice, but is not urgent imo. As far as I see it, `apply` triggers any upgrades as expected. 

### Checklist
- [x] Link to Milestone

### DIFF
--- a/cli/internal/kubernetes/upgrade.go
+++ b/cli/internal/kubernetes/upgrade.go
@@ -54,6 +54,17 @@ func GetConstellationVersion(ctx context.Context, client DynamicInterface) (upda
 	return nodeVersion, nil
 }
 
+// InvalidUpgradeError present an invalid upgrade. It wraps the source and destination version for improved debuggability.
+type applyError struct {
+	expected string
+	actual   string
+}
+
+// Error returns the String representation of this error.
+func (e *applyError) Error() string {
+	return fmt.Sprintf("expected NodeVersion to contain %s, got %s", e.expected, e.actual)
+}
+
 // Upgrader handles upgrading the cluster's components using the CLI.
 type Upgrader struct {
 	stableInterface  stableInterface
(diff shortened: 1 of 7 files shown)


cli/internal/kubernetes/upgrade.go
```diff
@@ -162,14 +162,14 @@ func (u *Upgrader) UpgradeNodeVersion(ctx context.Context, conf *config.Config)
 		return fmt.Errorf("applying upgrade: %w", err)
 	}
 	switch {
-	case updatedNodeVersion.Spec.ImageReference != imageReference:
-		return fmt.Errorf("expected NodeVersion to contain %s, got %s", imageReference, updatedNodeVersion.Spec.ImageReference)
-	case updatedNodeVersion.Spec.ImageVersion != imageVersion.Version:
-		return fmt.Errorf("expected NodeVersion to contain %s, got %s", imageVersion.Version, updatedNodeVersion.Spec.ImageVersion)
+	case updatedNodeVersion.Spec.ImageReference != nodeVersion.Spec.ImageReference:
+		return fmt.Errorf("expected NodeVersion to contain %s, got %s", nodeVersion.Spec.ImageReference, updatedNodeVersion.Spec.ImageReference)
```
3u13r, line 166: should we add a test which runs in one of those errors?

derpsteb: I mean there is unittest that triggered the error that is now fixed :D.
But I can also add another one that triggers the path with the new code.

e2e/internal/upgrade/image.go
```diff
@@ -38,17 +37,15 @@ func fetchUpgradeInfo(ctx context.Context, csp cloudprovider.Provider, toImage s
...
 
-	measurementsURL, signatureURL, err := versionsapi.MeasurementURL(ver, csp)
+	measurementsURL, _, err := versionsapi.MeasurementURL(ver, csp)
 	if err != nil {
 		return upgradeInfo{}, err
 	}
 
 	var fetchedMeasurements measurements.M
-	_, err = fetchedMeasurements.FetchAndVerify(
+	err = fetchedMeasurements.Fetch(
```
3u13r, line 46: I don't think that we should always skip the verification even for release enterprise CLIs.

3u13r (changes requested):

e2e/internal/upgrade/image.go
3u13r: My point is that from a end-user perspective this is strictly worse security simply dropping the verification on upgrades. Especially since we expect the user now to be able to have long lasting cluster through upgrading it. If we drop the verification on those upgrades, I don't see a point why we wouldn't also drop it from `constellation create`.

3u13r: As discussed, I missed that this is test code only.

e2e/internal/upgrade/upgrade_test.go
```diff
@@ -107,6 +107,14 @@ func TestUpgrade(t *testing.T) {
 	require.NoError(containsUnexepectedMsg(string(msg)))
 	log.Println(string(msg))
 
+	// Show versions set in cluster.
+	// The string after "Cluster status:" in the output might not be updated yet.
+	// This is only updated after the operator finishes one reconcile loop.
+	cmd = exec.CommandContext(context.Background(), cli, "status")
+	msg, err = cmd.CombinedOutput()
+	require.NoErrorf(err, "%s", string(msg))
```
daniel-weisse, line 115:
```suggestion
	require.NoError(err, string(msg))
```
This should also work

internal/attestation/measurements/measurements.go
```diff
@@ -112,6 +112,35 @@ func (m *M) FetchAndVerify(
 	return hex.EncodeToString(shaHash[:]), nil
 }
 
+// Fetch fetches measurements via provided URLs, using client for download.
+func (m *M) Fetch(ctx context.Context, client *http.Client, measurementsURL *url.URL, metadata WithMetadata) error {
```
daniel-weisse, line 116: Not a fan of having this function exported since it is used in tests only, and shares all of its code with `FetchAndVerify`.
Is this function really necessary, or could  it be removed from the `measurements` package and replaced with a simple download of the measurements and unmarshal call in the e2e test?

derpsteb: Done.

3u13r (approved):
LGTM

PR #1649, opened 2023-04-13
upgrade: fix 2.6 -> 2.7 migration for 2.7.1 patch

derpsteb:
### Proposed change(s)

- We added migration for 2.6.X -> 2.7.0 upgrades. But we need to skip those for 2.7.0 -> 2.7.X upgrades.
- We need to think about how to handle changes to our terraform setup during upgrades for the 2.6->2.7->2.8 upgrade chain. But thats a separate issue not for this PR.
- [Azure, 1+1, 2.7.0 --> k8s-only](https://github.com/edgelesssys/constellation/actions/runs/4785701347)
- [Azure, 1+1, 2.7.0 --> head](https://github.com/edgelesssys/constellation/actions/runs/4785969862)
- [Azure, 1+1, 2.6.0 --> head](https://github.com/edgelesssys/constellation/actions/runs/4785973994)

### Checklist


malt3 (approved):
Check looks correct.

derpsteb:
[Azure, 1+1, 2.7 -> 2.7.1, k8s-only](https://github.com/edgelesssys/constellation/actions/runs/4806168907)
[Azure, 1+1, 2.7 -> 2.8-pre, k8s-only](https://github.com/edgelesssys/constellation/actions/runs/4806699128)

cli/internal/helm/client.go
```diff
@@ -109,9 +116,11 @@ func (c *Client) Upgrade(ctx context.Context, config *config.Config, timeout tim
 		if info == constellationOperatorsInfo || info == constellationServicesInfo {
```
3u13r, line 116: I don't think it makes sense to iterate over 4 values but have explicit checks for half of them inside the loop. I think we could refactor this for loop to its own function and call the function in 2 different for loops one passing the `config.MicroserviceVersion` and one `chart.Metadata.Version` to the function.

cli/internal/helm/client.go
```diff
@@ -269,26 +278,38 @@ func (c *Client) upgradeRelease(
...
 		values, err = loader.loadConstellationServicesValues()
 
-		// TODO: v2.8: remove this call.
-		// Manually setting attestationVariant is required here since upgrade normally isn't allowed to change this value.
-		// However, to introduce the value into a 2.6 cluster for the first time we have to set it nevertheless.
-		if err := setAttestationVariant(values, conf.AttestationVariant); err != nil {
-			return fmt.Errorf("setting attestationVariant: %w", err)
-		}
-
 		// TODO: v2.8: remove from here...
```
3u13r, line 281:
```suggestion
		// TODO(derpsteb): v2.8: remove from here...
```

3u13r, line 281: Since there are multiple different migration taking place here, I'd move each migration to its on function and guard it at the beginning of the function by actually checking if we have an upgrade where we need to migrate, so that even if we forget deleting the code, the migration is never applies twice. 

For example for the new migration added in this PR we can actually check if we upgrade from `v2.6.A` to `v2.7.B`. Then upgrading to `v2.7.C` or `v2.8.D` will not trigger the migration again.

3u13r (changes requested):

cli/internal/helm/client.go
derpsteb: I agree that this isn't the prettiest solution. But I don't yet see the improvement from your suggestion.

I assume you want to put the content of the for loop into a new function `new` and then have two for-loops, one with `ciliumInfo, certManagerInfo` and one with `constellationOperatorsInfo, constellationServicesInfo`. Both for-loops call `new`.

Two problems I see with that:
- The error handling still needs to happen in the caller. Therefore the [switch statement](https://github.com/edgelesssys/constellation/pull/1649/files/aec6a499733dec4266f7a3e3c0cc505d59987925#diff-9a96aebbcd10227f19a06b17202f088565d3cbd804f3b1c9ac4afd3b3ce08ab9R124-R131) needs to be duplicated, once for each loop
- Calling [updateVersions](https://github.com/edgelesssys/constellation/pull/1649/files/aec6a499733dec4266f7a3e3c0cc505d59987925#diff-9a96aebbcd10227f19a06b17202f088565d3cbd804f3b1c9ac4afd3b3ce08ab9R118) on all charts _should_ work. But I feel saver calling it just on our charts. Therefore there would be the same check inside the new function.

cli/internal/helm/client.go
derpsteb: Changed code so that migrations can be left in the codebase:
- Moved migrations into a separate `applyMigrations` function. 
- Added `migrateFrom2_6` function that is called based on the "from" version of the upgrade. This way the migration is only called where required.

cli/internal/helm/client.go
derpsteb: We discussed the problem in a call and came up with this: https://github.com/edgelesssys/constellation/pull/1649/commits/0948dc32ab66798ca1d259ab128c74c1eb6f678a

3u13r (approved):
LGTM

derpsteb:
[Azure, 1+1, 2.7.0 -> 2.7.1, microservice only](https://github.com/edgelesssys/constellation/actions/runs/4829385412) 
[Azure, 1+1, 2.7.0 -> HEAD, microservice only](https://github.com/edgelesssys/constellation/actions/runs/4829391439)

PR #1665, opened 2023-04-17
ci: add Terraform logging

msanft:
### Proposed change(s)

- Upload a Terraform log artifact at `DEBUG` level for manual inspection of Terraform failures.
- [AWS Run](https://github.com/edgelesssys/constellation/actions/runs/4718634933)
- [Azure Run](https://github.com/edgelesssys/constellation/actions/runs/4718638266)
- [GCP Run](https://github.com/edgelesssys/constellation/actions/runs/4718637100)

### Additional info
- Change to `JSON` level later on and upload the logs to OpenSearch.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/actions/upload_terraform_logs/action.yml
+++ b/.github/actions/upload_terraform_logs/action.yml
@@ -1,11 +1,16 @@
 name: Upload Terraform logs
 description: "Upload the Terraform log file as an artifact."
 
+inputs:
+  artifactNameSuffix:
+    description: "Name suffix for the created artifact."
+    required: true
+
 runs:
   using: "composite"
   steps:
     - name: Upload artifact
       uses: actions/upload-artifact@v3
       with:
-        name: terraform-logs-e2e
+        name: terraform-logs-${{ inputs.artifactNameSuffix }}
         path: terraform.log
--- a/.github/actions/upload_terraform_logs/action.yml
+++ b/.github/actions/upload_terraform_logs/action.yml
@@ -1,0 +1,11 @@
+name: Upload Terraform logs
+description: "Upload the Terraform log file as an artifact."
+
+runs:
+  using: "composite"
+  steps:
+    - name: Upload artifact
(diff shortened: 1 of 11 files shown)


.github/actions/upload_terraform_logs/action.yml
```diff
@@ -0,0 +1,11 @@
...
+description: "Upload the Terraform log file as an artifact."
+
+runs:
+  using: "composite"
+  steps:
+    - name: Upload artifact
+      uses: actions/upload-artifact@v3
+      with:
+        name: terraform-logs
+        path: terraform.log
```
katexochen, line 11: We should add the prefix from the e2e_test as suffix to the filename.

katexochen: Thats what I meant: https://github.com/edgelesssys/constellation/blob/main/.github/actions/e2e_test/action.yml#L170-L177

.github/workflows/e2e-test-weekly.yml
```diff
@@ -241,6 +241,12 @@ jobs:
           kubernetesVersion: ${{ matrix.kubernetes-version }}
           provider: ${{ matrix.provider }}
 
+      - name: Always upload Terraform logs
+        if: always()
+        uses: ./.github/actions/upload_terraform_logs
+        with:
+          artifactNameSuffix: ${{ steps.e2e_test.outputs.namePrefix }}
```
katexochen, line 248: Can't we do this step on a lower layer, like in constellation_destroy? Otherwise lgtm.

katexochen (approved):

.github/workflows/e2e-test-weekly.yml
msanft: I think it'd be a bad practice to put it in either constellation_destroy or constellation_iam_destroy. If you put it in 1 of the 2 you would implicitly require that action to be always executed last (to not lose log output)

PR #1662, opened 2023-04-15
ci: fix naming issues

katexochen:
### Proposed change(s)

- Fixing stuff that caused failure due to bad naming in https://github.com/edgelesssys/constellation/actions/runs/4705664276

### Additional info
- AWS: https://github.com/edgelesssys/constellation/actions/runs/4706876674
- GCP: https://github.com/edgelesssys/constellation/actions/runs/4706875477
- Azure:https://github.com/edgelesssys/constellation/actions/runs/4706874289

### Checklist

- [x] Add labels (e.g., for changelog category)

PR #1674, opened 2023-04-19
ci: limit prefix length of AWS IAM resources

malt3:
### Proposed change(s)

- ci: limit prefix length of AWS IAM resources

### Additional info

IAM create fails in the CI on AWS:

```
Error: expected length of name to be in the range (1 - 64), got e2e-4705664276-1-387ada92-9a16-40c1-8f35-adafacf04581_control_plane_role
```

So we have to shorten the prefix used for e2e tests.

### DIFF
--- a/.github/actions/e2e_test/action.yml
+++ b/.github/actions/e2e_test/action.yml
@@ -189,7 +189,7 @@ runs:
       id: create-prefix
       shell: bash
       run: |
-        uuid=$(uuidgen | head -c 23)
+        uuid=$(uuidgen)
         uuid=${uuid%%-*}
         echo "uuid=${uuid}" | tee -a $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


.github/actions/e2e_test/action.yml
```diff
@@ -189,7 +189,7 @@ runs:
       id: create-prefix
       shell: bash
       run: |
-        uuid=$(uuidgen)
+        uuid=$(uuidgen | head -c 23)
```
katexochen, line 192: That shouldn't be needed, as the next line should cut only the part after the last dash?

malt3: I was triaging a run and saw the old error output that was already fixed by https://github.com/edgelesssys/constellation/pull/1662

msanft (approved):

PR #1675, opened 2023-04-19
bootstrapper: pull node logs on init failure

miampf:
### Proposed change(s)

- add a gRPC streaming call to the constellation CLI that queries logs from a server that failed initialization
- extend the file package to support appending to files on write

### Checklist

- [x] Link to Milestone


cli/internal/cmd/init.go
```diff
@@ -259,7 +265,47 @@ func (d *initDoer) Do(ctx context.Context) error {
 	if err != nil {
 		return &nonRetriableError{fmt.Errorf("init call: %w", err)}
 	}
-	d.resp = resp
+
+	res, err := resp.Recv() // get first response, either success or failure
```
3u13r, line 269: Just as a heads up: I think it is cleaner to first check the error and then switch over the message type of `res`. 
For an example you can see the last code snippet here: https://protobuf.dev/reference/go/go-generated/#oneof

miampf:
This was rebased into #1618 and then merged so it should be closed, sorry for having it open

PR #1676, opened 2023-04-21
ci: login to container registry before pushing containers

malt3:
### Proposed change(s)

- ci: login to container registry before pushing containers

### Additional info
- [e2e test run](https://github.com/edgelesssys/constellation/actions/runs/4763155001)


msanft (approved):
lgtm

PR #1678, opened 2023-04-24
cli: add config migrate command

daniel-weisse:
### Proposed change(s)

- Add `constellation config migrate` command to automatically migrate configs from `v2` to `v3`
  - Makes it easier for users to adapt to config format changes
  - Allows us to easily migrate configs for our e2e tests
- Fix upgrades from v2.7 to v2.8 for GCP by deleting the `encrypted-rwo` storage-class before applying the upgrade
  - This is necessary since v2.8 introduces changes that cannot be patches on a live storage class
  - Relevant code can be removed after v2.8 has been released
- Fix e2e upgrade test by making pushing container images from current commit hash before running the test

### Additional info
The code used here keeps a backup of the previous config format in `/internal/config/migration`.
We can decide to either keep this migration logic until there is a new format (e.g. `v4`), or remove it again once it is no longer needed.

### Upgrade e2e Tests
* Azure: https://github.com/edgelesssys/constellation/actions/runs/4795538914
  * Only the IAM deletion failed, this is not part of the actual test
* GCP: https://github.com/edgelesssys/constellation/actions/runs/4795931603

### Checklist

- [ ] Link to Milestone


cli/internal/cmd/configmigrate.go
```diff
@@ -0,0 +1,64 @@
...
+
+func newConfigMigrateCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "migrate",
+		Short: "Migrate a configuration file to a new version",
+		Long:  "Migrate a configuration file to a new version.",
+		Args:  cobra.NoArgs,
+		RunE:  runConfigMigrate,
+	}
+	cmd.Flags().StringP("file", "f", constants.ConfigFilename, "path to output file, or '-' for stdout")
```
derpsteb, line 28: not used?

cli/internal/helm/client.go
```diff
@@ -278,6 +279,16 @@ func (c *Client) upgradeRelease(
 		}
 		joinServiceVals["attestationConfig"] = string(attestationConfigJSON)
 
+		// v2.8 updates the disk type of GCP default storage class
+		// This value is not updatable in Kubernetes, but we can use a workaround to update it:
+		// First, we delete the storage class, then we upgrade the chart,
+		// which will recreate the storage class with the new disk type.
+		if conf.GetProvider() == cloudprovider.GCP {
+			if err := c.kubectl.DeleteStorageClass(ctx, "encrypted-rwo"); err != nil {
```
derpsteb, line 287: Wondering if PVCs created with the old storage class will still trigger volume deletion, once the new storage class with another disk type is deployed.

Daniel already said he would test this again.

daniel-weisse: >will still trigger volume deletion, once the new storage class with another disk type is deployed.

This works without problem, and I already tested this yesterday.
Only thing I am unsure of is if volumes can be properly deleted if there is no storageclass at all

daniel-weisse: Deletion works even if the storage class doesnt exist anymore

derpsteb (approved):

PR #1685, opened 2023-04-26
cli: Terraform migrations on upgrade

msanft:
### Proposed change(s)

- Implement migration of cloud resources on Constellation version upgrades. (e.g. Constellation 2.6.0 doesn't create an MAA, while 2.7.0 does optionally and 2.8.0 always creates one)
  - In that case, a user with a cluster without an MAA (2.6.0/2.7.0) downloads the 2.8.0 CLI and runs the `upgrade apply` command to have a MAA added to his cloud resources.
- How the migration works:
  - The **old** Terraform directory `constellation-terraform` is backed up into `constellation-upgrade/constellation-upgrade-terraform-backup`. If the user wants to rollback to the pre-upgrade state, he can do it there
  - Constellation copies the **old** Terraform state from the `constellation-terraform` dir into the `constellation-upgrade/constellation-upgrade-terraform` dir.
  - Constellation populates the `constellation-upgrade/constellation-upgrade-terraform` dir with the **new** Terraform script files (embedded in the CLI)
  - Constellation performs a `terraform plan` on selected targets (as of now, only on the MAA on Azure)
  - The resulting diff is shown to the user who can confirm it or abort the upgrade.
  - If the user confirms the diff (resources to be crated / changed / etc.), the **new** script is applied on the selected target (in simpler terms; the output confirmed by the user gets created)
  - The old Terraform directory `constellation-terraform` is replaced with the new `constellation-upgrade/constellation-upgrade-terraform`. The output of the migrated resources is written into `terraform-migration-output.json` (e.g. the attestation URL of the MAA provider)

### Additional info
- TODO: Add e2e test run

### Checklist

- [x] Link to Milestone


e2e/internal/upgrade/upgrade_test.go
```diff
@@ -101,7 +101,7 @@ func TestUpgrade(t *testing.T) {
 	log.Println(string(data))
 
 	log.Println("Triggering upgrade.")
-	cmd := exec.CommandContext(context.Background(), cli, "upgrade", "apply", "--force", "--debug")
+	cmd := exec.CommandContext(context.Background(), cli, "upgrade", "apply", "--force", "--debug", "--yes", "--tf-log=DEBUG")
```
derpsteb, line 104: We need to check if the binary supports `--tf-log` before adding it to the command.

internal/constants/constants.go
```diff
@@ -211,7 +217,7 @@ func State() string {
 
 var (
 	// versionInfo is the version of a binary. Left as a separate variable to allow override during build.
-	versionInfo = "0.0.0"
+	versionInfo = "2.8.0"
```
derpsteb, line 220: This should not be changed. To force a specific version in your CLI you can change the `version.txt` file in the repo's root folder.

internal/constants/constants.go
```diff
@@ -154,6 +154,12 @@ const (
 	MiniConstellationUID = "mini"
 	// TerraformLogFile is the file name of the Terraform log file.
 	TerraformLogFile = "terraform.log"
+	// TerraformUpgradePlanFile is the file name of the zipfile created by Terraform plan for Constellation upgrades.
+	TerraformUpgradePlanFile = "plan.zip"
+	// TerraformUpgradeWorkingDir is the directory name for the Terraform workspace being used in an update.
+	TerraformUpgradeWorkingDir = "constellation-upgrade-terraform"
```
derpsteb, line 160: maybe make this a subfolder of `constellation-upgrade`? 
During service upgrades we are also creating a local folder structure. See [here](https://github.com/edgelesssys/constellation/blob/main/cli/internal/helm/backup.go#LL30C32-L30C32).

cli/internal/kubernetes/upgrade.go
```diff
@@ -73,10 +79,19 @@ type Upgrader struct {
 	imageFetcher     imageFetcher
 	outWriter        io.Writer
 	log              debugLog
+	tf               terraformUpgrader
+}
+
+// a terraformUpgrader performs the Terraform interactions in an upgrade.
+type terraformUpgrader interface {
```
derpsteb, line 86: nit: should be moved further down in the file. [ref](https://github.com/edgelesssys/wiki/blob/master/company/gostyle.md#order-of-elements-in-a-source-file).

internal/constants/constants.go
msanft: Oops! That was accidental

cli/internal/cloudcmd/create.go
```diff
@@ -218,7 +217,6 @@ func (c *Creator) createAzure(ctx context.Context, cl terraformClient, opts Crea
 		StateDiskType:        opts.Config.Provider.Azure.StateDiskType,
 		ImageID:              opts.image,
 		SecureBoot:           *opts.Config.Provider.Azure.SecureBoot,
-		CreateMAA:            opts.Config.Provider.Azure.EnforceIDKeyDigest == idkeydigest.MAAFallback,
```
derpsteb, line 221: Needs a rebase?
The option is still there on main because we don't create the MAA for TrustedLaunch. Is the removal on purpose here?

cli/internal/cmd/upgradeapply.go
```diff
@@ -180,19 +316,37 @@ func parseUpgradeApplyFlags(cmd *cobra.Command) (upgradeApplyFlags, error) {
...
+	terraformLogLevel terraform.LogLevel
 }
 
 type cloudUpgrader interface {
 	UpgradeNodeVersion(ctx context.Context, conf *config.Config) error
 	UpgradeHelmServices(ctx context.Context, config *config.Config, timeout time.Duration, allowDestructive bool) error
 	UpdateMeasurements(ctx context.Context, newMeasurements measurements.M) error
 	GetClusterMeasurements(ctx context.Context) (measurements.M, *corev1.ConfigMap, error)
+	PlanTerraformMigrations(ctx context.Context, opts kubernetes.TerraformUpgradeOptions) (bool, error)
+	ApplyTerraformMigrations(ctx context.Context, file file.Handler, opts kubernetes.TerraformUpgradeOptions) error
```
derpsteb, line 351: Is there something that speaks against putting these two into a separate type?
I think it would make sense to have the implementation of those two functions inside `cli/internal/terraform` and not in `cli/internal/kubernetes`. The functions do not interact with the cluster, but with the platform the cluster is running on.

cli/internal/kubernetes/upgrade.go
derpsteb, line 1: Following [this](https://github.com/edgelesssys/constellation/pull/1685/files#r1188174225) comment: the changes from this file would go into e.g. `cli/internal/terraform/upgrade.go` or `cli/internal/terraform/migrations.go`

cli/internal/terraform/terraform/azure/main.tf
derpsteb, line 1: same as above.

derpsteb:
re the open design descision: I would expect `constellation terminate` to delete all my resources. I don't see a reason why we would not replace the old state. Which reasons are there?

e2e/internal/upgrade/upgrade_test.go
msanft: How would you do that? Run it with the `--help` flag and see if the output contains `tf-log`?

cli/internal/cmd/upgradeapply.go
msanft: I think they should not be placed in `cli/internal/terraform` as their scope goes beyond the interaction with Terraform only. E.g., `ApplyTerraformMigrations` formats the Terraform output and writes it to the `clusterid.File`, while the `terraform` package only wraps singular Terraform operations. (See [cluster creation](https://github.com/edgelesssys/constellation/blob/fd83f3439e847ff14a51525f7bd9f55133588671/cli/internal/cloudcmd/create.go#L169) as a reference)

msanft:
re "selected targets"

[Planning](https://github.com/edgelesssys/constellation/blob/02a45d195c74edc6a86dac12db0281dd5064da81/cli/internal/terraform/terraform.go#L327) and [applying](https://github.com/edgelesssys/constellation/blob/02a45d195c74edc6a86dac12db0281dd5064da81/cli/internal/terraform/terraform.go#L109) the configuration uses the [`targets`](https://developer.hashicorp.com/terraform/tutorials/state/resource-targeting) flag which specifies resources to plan / apply the configuration for. 
The specific targets get set [here](https://github.com/edgelesssys/constellation/blob/02a45d195c74edc6a86dac12db0281dd5064da81/cli/internal/cmd/upgradeapply.go#L207).

msanft:
re "design decision"

The reason I would see is, that if an upgrade goes wrong, the user should be able to rollback to the old resource state manually. Therefore we should at least keep the old state somewhere in the upgrade folder. I will prepare something to put the current state into a backup folder and move the upgraded state into the `constellation-terraform` folder

e2e/internal/upgrade/upgrade_test.go
derpsteb: Yes. One example can be found here: https://github.com/edgelesssys/constellation/blob/main/.github/actions/constellation_create/action.yml#L211-L215

Dunno if there are better ways, but this seems like a pretty accurate check to me.

cli/internal/cmd/upgradeapply.go
derpsteb: Okay. Then lets put them into a new pkg: `cli/internal/upgrade`. Looks like the implementations don't depend on other parts of `Upgrader`. Sorry about initially pointing you towards adding the methods to that interface. But looking at it for a second time I think it makes sense to separate the terraform migrations from the kubernetes operations..

internal/constants/constants.go
```diff
@@ -148,6 +148,16 @@ const (
 	MiniConstellationUID = "mini"
 	// TerraformLogFile is the file name of the Terraform log file.
 	TerraformLogFile = "terraform.log"
+	// TerraformUpgradePlanFile is the file name of the zipfile created by Terraform plan for Constellation upgrades.
+	TerraformUpgradePlanFile = "plan.zip"
+	// TerraformUpgradeWorkingDir is the directory name for the Terraform workspace being used in an upgrade.
+	TerraformUpgradeWorkingDir = "constellation-upgrade-terraform"
+	// TerraformUpgradeBackupDir is the directory name being used to backup the pre-upgrade state in an upgrade.
+	TerraformUpgradeBackupDir = "constellation-upgrade-terraform-backup"
```
derpsteb, line 156:
```suggestion
	TerraformUpgradeWorkingDir = "terraform"
	// TerraformUpgradeBackupDir is the directory name being used to backup the pre-upgrade state in an upgrade.
	TerraformUpgradeBackupDir = "terraform-backup"
```

cli/internal/cmd/upgradeapply.go
```diff
@@ -120,6 +130,132 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, fileHandler file.Hand
...
+
+	if hasDiff {
+		// If there are any Terraform migrations to apply, ask for confirmation
+		if !flags.yes {
+			ok, err := askToConfirm(cmd, "Do you want to apply the Terraform migrations?")
+			if err != nil {
+				return fmt.Errorf("asking for confirmation: %w", err)
+			}
+			if !ok {
+				cmd.Println("Aborting upgrade.")
```
derpsteb, line 167: We should clean up the create backup files in this case, since they are not relevant in that case. 
To distinguish between runs we could add a suffix to `TerraformUpgradeBackupDir` on each invocation of `apply`. Sth like a 6-digit cut from a uuid. 

That folder name should be printed on successful completion. "You can find a backup of your terraform workspace at ...". Or similar.

derpsteb:
Not part of the PR but would be great if you could add a `-` [here](https://github.com/edgelesssys/constellation/blob/main/cli/internal/terraform/terraform/azure/main.tf#L51) to make the name uniform with the other resources.

cli/internal/kubernetes/upgrade.go
```diff
@@ -103,16 +110,97 @@ func NewUpgrader(outWriter io.Writer, log debugLog) (*Upgrader, error) {
...
+		IP:             tfOutput.IP,
+		UID:            tfOutput.UID,
+		AttestationURL: tfOutput.AttestationURL,
+	}
+
+	if err := file.RemoveAll(constants.TerraformWorkingDir); err != nil {
+		return fmt.Errorf("removing old terraform directory: %w", err)
+	}
+
+	if err := file.CopyDir(filepath.Join(constants.UpgradeDir, constants.TerraformUpgradeWorkingDir), constants.TerraformWorkingDir); err != nil {
```
derpsteb, line 193: `src` of the copy operation should be deleted once it succeeded.

The old state is in `constellation-upgrade/terraform-upgrade`. The new state is in`constellation-terraform`. But also in `constellation-upgrade/terraform`.

cli/internal/upgrade/terraform.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	for _, f := range filesToCheck {
+		if err := checkFileExists(fileHandler, &existingFiles, f); err != nil {
+			return fmt.Errorf("checking terraform migrations: %w", err)
+		}
+	}
+
+	if len(existingFiles) > 0 {
+		return fmt.Errorf("terraform migrations not possible: files / directories %s already exist", strings.Join(existingFiles, ", "))
```
derpsteb, line 68:
```suggestion
		return fmt.Errorf("file(s) %s already exist", strings.Join(existingFiles, ", "))
```

cli/internal/cmd/upgradeapply.go
```diff
@@ -120,6 +132,141 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, fileHandler file.Hand
 	return nil
 }
 
+// migrateTerraform checks if the Constellation version the cluster is being upgraded to requires a migration
+// of cloud resources with Terraform. If so, the migration is performed.
+func (u *upgradeApplyCmd) migrateTerraform(cmd *cobra.Command, file file.Handler, fetcher imageFetcher, conf *config.Config, flags upgradeApplyFlags) error {
+	u.log.Debugf("Planning Terraform migrations")
+
+	if err := u.upgrader.CheckTerraformMigrations(file); err != nil {
+		return fmt.Errorf("checking terraform migrations: %w", err)
```
derpsteb, line 141:
```suggestion
		return fmt.Errorf("checking workspace: %w", err)
```

derpsteb (approved):
Just tested it again on AWS and Azure. Looks great now :)

PR #1686, opened 2023-04-26
cli: configurable state disk type on OpenStack

malt3:
### Proposed change(s)

- cli: configurable state disk type on OpenStack


internal/config/config_test.go
```diff
@@ -188,7 +188,7 @@ func TestNewWithDefaultOptions(t *testing.T) {
 }
 
 func TestValidate(t *testing.T) {
-	const defaultErrCount = 32 // expect this number of error messages by default because user-specific values are not set and multiple providers are defined by default
+	const defaultErrCount = 33 // expect this number of error messages by default because user-specific values are not set and multiple providers are defined by default
```
daniel-weisse, line 191: Can we not set a default for the OpenStack state disk type?
Or is this something that will differ for multiple OpenStack deployments?

malt3: The default type differs between OpenStack deployments.
All we could do is make this optional and explicitly not set this value in terraform if the string is empty.
Feels a bit weird.
We have another feature planned (AB#3112) to improve the quality of prefilled values for well-known OpenStack deployments.

daniel-weisse (approved):

PR #1693, opened 2023-04-26
bazel: allow custom container_prefix

3u13r:
### Proposed change(s)

- add `container_prefix` flag which where the container images of this build should be hosted.
- Automatically push the container images on `//:devbuild`

The container references in the CLI also use this value.

I'd also like to thank @malt3 for the support on this feature.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/dev-docs/workflows/build-test-run.md
+++ b/dev-docs/workflows/build-test-run.md
@@ -28,12 +28,19 @@ mkdir build
 cd build
 # build required binaries for a dev build
 # and symlink them into the current directory
-bazel run //:devbuild
+# also push the built container images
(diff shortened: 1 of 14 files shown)


dev-docs/workflows/build-test-run.md
```diff
@@ -28,14 +28,21 @@ mkdir build
 cd build
 # build required binaries for a dev build
 # and symlink them into the current directory
-bazel run //:devbuild
+# also push the built container images
+bazel run //:devbuild --container_prefix=ghcr.io/USERNAME
```
malt3, line 32: Maybe we should add a warning here that users have to set the images to public on the first push. This may not be obvious.

malt3 (approved):

PR #1719, opened 2023-05-03
cli: remove ambiguity in path for CR backups

derpsteb:
### Proposed change(s)

- During upgrade all custom resources are backed up to files on the local file system. Since old versions are also backed up, we need to reflect the version in the name. This is a fixup for #1702 .
- Tested the change locally by running against a bigger cluster.

### Checklist


3u13r (approved):
LGTM

PR #1722, opened 2023-05-03
ci: set toImage argument in e2e-test-release

derpsteb:
### Proposed change(s)

- The default is to use the latest main-debug image. During release we want to use the release image. 

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/e2e-test-release.yml
+++ b/.github/workflows/e2e-test-release.yml
@@ -222,7 +222,18 @@ jobs:
         with:
           artifactNameSuffix: ${{ steps.e2e_test.outputs.namePrefix }}
 
+  determine-version:
+    runs-on: ubuntu-22.04
+    outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-test-release.yml
```diff
@@ -223,22 +223,30 @@ jobs:
...
-      controlNodesCount: 3
+
+    steps:
+      - name: Determine version
+        id: determine_version
+        run: |
+          echo "targetVersion=$(cat version.txt) | tee -a "$GITHUB_OUTPUT"
+
+      - name: Test upgrade
+        uses: ./.github/workflows/e2e-upgrade.yml
```
katexochen, line 246: Afaik you cannot do a workflow call in a step.

katexochen (approved):
lgtm

PR #1741, opened 2023-05-05
ci: fix measurement generation on scheduled build

katexochen:
### Proposed change(s)

- This was broken after the first replacement, as the regexp expected the version to start with a "v".
- Measurements and versions are updates with the values that should have been updated in this workflow: https://github.com/edgelesssys/constellation/actions/runs/4887467663

### Checklist


malt3 (approved):
Was working on a similar fix for the release pipeline. LGTM.

PR #1769, opened 2023-05-15
cli: new flag to set the attestation type for `config generate`

elchead:
### Proposed change(s)

- new --attestation flag for `constellation config generate` to specify which type to use for the given cloud provider

AB2976

### Checklist

- [ ] Link to Milestone


cli/internal/cmd/configgenerate.go
```diff
@@ -135,13 +157,40 @@ func parseGenerateFlags(cmd *cobra.Command) (generateFlags, error) {
...
 
+	attestationString, err := cmd.Flags().GetString("attestation")
+	if err != nil {
+		return generateFlags{}, fmt.Errorf("parsing attestation flag: %w", err)
+	}
+
+	var attestationType config.AttestationType
+	// if no attestation type is specified, use the default for the cloud provider
+	if attestationString == "" {
+		attestationType = config.AttestationType("")
```
katexochen, line 168: Calling attestation type with empty string as arg is not that easy to read on the caller side, maybe add a 

```go
const AttestationTypeDefault AttestationType = "default"
```

to the constant?

It might also be a good idea to define 

```go
const AttestationTypeUnknown AttestationType = ""
```

to have an explicit nil/empty value and not implicitly assuming default where we might have forgotten to set it.

cli/internal/cmd/configgenerate.go
```diff
@@ -105,7 +110,24 @@ func createConfig(provider cloudprovider.Provider) *config.Config {
...
+	if provider == cloudprovider.Unknown {
+		return conf, nil // TODO tests use Unknown provider... the CLI doesn't allow it.. why do we do that?
+	}
+	if attestationType == "" {
+		attestationType = config.GetDefaultAttestationType(provider)
+	} else {
+		if !attestationType.ValidProvider(provider) {
+			return nil, fmt.Errorf("provider %s does not support attestation type %s", provider, attestationType)
+		}
+	}
```
katexochen, line 122:
```suggestion
	} else if !attestationType.ValidProvider(provider) {
			return nil, fmt.Errorf("provider %s does not support attestation type %s", provider, attestationType)
	}
```

cli/internal/cmd/configgenerate_test.go
```diff
@@ -139,3 +140,97 @@ func TestConfigGenerateStdOut(t *testing.T) {
...
+		expected    config.AttestationConfig
+	}{
+		{cloudprovider.Azure, config.AttestationTypeAzureTrustedLaunch, config.AttestationConfig{AzureTrustedLaunch: defaultAttestation.AzureTrustedLaunch}},
+		{cloudprovider.Azure, config.AttestationTypeAzureSEVSNP, config.AttestationConfig{AzureSEVSNP: defaultAttestation.AzureSEVSNP}},
+
+		{cloudprovider.AWS, config.AttestationTypeAWSNitroTPM, config.AttestationConfig{AWSNitroTPM: defaultAttestation.AWSNitroTPM}},
+		{cloudprovider.GCP, config.AttestationTypeGCPSEVES, config.AttestationConfig{GCPSEVES: defaultAttestation.GCPSEVES}},
+
+		{cloudprovider.QEMU, config.AttestationTypeQEMUVTPM, config.AttestationConfig{QEMUVTPM: defaultAttestation.QEMUVTPM}},
+		{cloudprovider.OpenStack, config.AttestationTypeQEMUVTPM, config.AttestationConfig{QEMUVTPM: defaultAttestation.QEMUVTPM}},
```
katexochen, line 177: Please break long lines and nested structs.

katexochen (commented):
Just some quick thoughts on Go style, looks quite good. :+1:

internal/config/attestation.go
```diff
@@ -13,9 +13,50 @@ import (
...
 )
 
+type AttestationType (string)
+
+const (
+	AttestationTypeAWSNitroTPM        AttestationType = "aws-nitro-tpm"
+	AttestationTypeAzureSEVSNP        AttestationType = "azure-sev-snp"
+	AttestationTypeAzureTrustedLaunch AttestationType = "azure-trusted-launch"
+	AttestationTypeGCPSEVES           AttestationType = "gcp-sev-es"
+	AttestationTypeQEMUVTPM           AttestationType = "qemu-vtpm"
```
derpsteb, line 27:
```suggestion
	AWSNitroTPM        AttestationType = "aws-nitro-tpm"
	AzureSEVSNP        AttestationType = "azure-sev-snp"
	AzureTrustedLaunch AttestationType = "azure-trusted-launch"
	GCPSEVES           AttestationType = "gcp-sev-es"
	QEMUVTPM           AttestationType = "qemu-vtpm"
```
Id say the information is already encoded in the type name.

katexochen: But that doesn't help when reading code on the caller side. As the package name is `config`, it would read `config.QEMUVTPM` , which isn't enough information Imo. Would totally agree if this had its own package `attestation` .

daniel-weisse: We already have [`variant`](https://github.com/edgelesssys/constellation/blob/main/internal/variant/variant.go), could just export the string constants, if working with strings is required.

katexochen: `variant` is a really undescriptive package name. :laughing:

derpsteb: I looked at the places where the variables are used ([1](https://github.com/edgelesssys/constellation/pull/1769/files#diff-54c7c1af5fa8d5db4dc49f0e8e80e93ba2b1183ba4d5c9e2e5729e6deae6a3cdR468-R478), [2](https://github.com/edgelesssys/constellation/pull/1769/files#diff-438814676158a6febbb3ef8e77769ba76f6d70d7df4bac31c3101b1e0b64c3b4R185), [3](https://github.com/edgelesssys/constellation/pull/1769/files#diff-62a8c60cf08995632a7054539677e3f3b04a0bfb956018e12ac6804ec4a9aebbR147-R153), ...) and felt like the type information was always close to the use of the const and thus it would be pretty readable. But feel free to skip this.

elchead: I skipped the renaming due to the conflict with another equally named struct. They could be merged but I don't see the benefit at the moment to invest more refactoring.

elchead:
I added some short comments for Mac and PR in the `build.md`

internal/config/attestation.go
```diff
@@ -11,11 +11,74 @@ import (
...
 )
 
+const (
+	AttestationTypeAWSNitroTPM        AttestationType = "aws-nitro-tpm"        // AttestationTypeAWSNitroTPM
+	AttestationTypeAzureSEVSNP        AttestationType = "azure-sev-snp"        // AttestationTypeAzureSEVSNP
+	AttestationTypeAzureTrustedLaunch AttestationType = "azure-trusted-launch" // AttestationTypeAzureTrustedLaunch
+	AttestationTypeGCPSEVES           AttestationType = "gcp-sev-es"           // AttestationTypeGCPSEVES
+	AttestationTypeQEMUVTPM           AttestationType = "qemu-vtpm"            // AttestationTypeQEMUVTPM
+	AttestationTypeDefault            AttestationType = "default"              // AttestationTypeDefault is the default attestation type for the cloud provider
+)
```
daniel-weisse, line 28: We already have the [`variant`](https://github.com/edgelesssys/constellation/blob/main/internal/variant/variant.go) package, which basically does the same things `AttestationType` does.
imo, we should remove `AttestationType` in favor of using the types from `variant` to avoid duplication and having to change code in multiple places if we want to change a single thing

elchead: I agree it's better and made the change. But should we rename the package to `attestation_variant`? Variant is indeed unclear naming as @katexochen mentioned

cli/internal/cmd/configgenerate.go
```diff
@@ -135,13 +171,39 @@ func parseGenerateFlags(cmd *cobra.Command) (generateFlags, error) {
...
+	if err != nil {
+		return generateFlags{}, fmt.Errorf("parsing attestation flag: %w", err)
+	}
+
+	var attestationType variant.Variant
+	// if no attestation type is specified, use the default for the cloud provider
+	if attestationString == "" {
+		attestationType = variant.Default{}
+	} else {
+		resType := validateAttestationType(attestationString)
```
daniel-weisse, line 184: you can use `variant.FromString(attestationString)` to return a variant (and error) from a string

internal/variant/variant.go
```diff
@@ -34,17 +34,68 @@ package variant
...
 	dummy              = "dummy"
 	awsNitroTPM        = "aws-nitro-tpm"
 	gcpSEVES           = "gcp-sev-es"
 	azureSEVSNP        = "azure-sev-snp"
-	azureTrustedLaunch = "azure-trustedlaunch"
+	azureTrustedLaunch = "azure-trusted-launch"
 	qemuVTPM           = "qemu-vtpm"
 )
 
+var providerAttestationMapping map[cloudprovider.Provider][]Variant = map[cloudprovider.Provider][]Variant{
```
daniel-weisse, line 51:
```suggestion
var providerAttestationMapping = map[cloudprovider.Provider][]Variant{
```

internal/variant/variant.go
```diff
@@ -34,17 +34,68 @@ package variant
...
+	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
 )
 
 const (
 	dummy              = "dummy"
 	awsNitroTPM        = "aws-nitro-tpm"
 	gcpSEVES           = "gcp-sev-es"
 	azureSEVSNP        = "azure-sev-snp"
-	azureTrustedLaunch = "azure-trustedlaunch"
+	azureTrustedLaunch = "azure-trusted-launch"
```
daniel-weisse, line 47: I dont think we want to change the string representation of this value, since this could break compatibility with older images.
Maybe @malt3 has an opinion on this?

internal/variant/variant.go
```diff
@@ -76,7 +127,36 @@ func FromString(oid string) (Variant, error) {
...
+	return false
+}
+
+// Default is the default variant for the cloud provider.
+type Default struct{}
+
+// OID returns the struct's object identifier.
+func (Default) OID() asn1.ObjectIdentifier {
+	return asn1.ObjectIdentifier{1, 3, 9900, 0, 0}
+}
```
daniel-weisse, line 149: This variant is basically just a copy of `Dummy`.
I think it can be removed and replaced with `Dummy` in the places where it is currently used.

daniel-weisse (commented):
Dont have a strong opinion on how/if we should rename the variant package.
Feel free to discuss with @katexochen

cli/internal/cmd/configgenerate.go
```diff
@@ -36,13 +37,27 @@ func newConfigGenerateCmd() *cobra.Command {
 	}
 	cmd.Flags().StringP("file", "f", constants.ConfigFilename, "path to output file, or '-' for stdout")
 	cmd.Flags().StringP("kubernetes", "k", semver.MajorMinor(config.Default().KubernetesVersion), "Kubernetes version to use in format MAJOR.MINOR")
+	cmd.Flags().StringP("attestation", "a", "", fmt.Sprintf("Attestation variant to use %s. If not specified, the default for the cloud provider is used.", printFormattedSlice(variant.GetAvailableAttestationTypes())))
```
derpsteb, line 40:
```suggestion
	cmd.Flags().StringP("attestation", "a", "", fmt.Sprintf("attestation variant to use %s. If not specified, the default for the cloud provider is used", printFormattedSlice(variant.GetAvailableAttestationTypes())))
```

Suggestion is based on a section in [conventions.md](https://github.com/edgelesssys/constellation/blob/main/dev-docs/conventions.md#cli-reference).
```suggestion
	cmd.Flags().StringP("attestation", "a", "", fmt.Sprintf("Attestation variant to use %s. If not specified, the default for the cloud provider is used.", printFormattedSlice(variant.GetAvailableAttestationTypes())))
```

cli/internal/cmd/configgenerate.go
```diff
@@ -36,13 +37,27 @@ func newConfigGenerateCmd() *cobra.Command {
...
+	return fmt.Sprintf("{%s}", strings.Join(toString(input), "|"))
+}
+
+func toString[T any](t []T) []string {
+	var res []string
+	for _, v := range t {
+		res = append(res, fmt.Sprintf("%v", v))
+	}
+	return res
+}
```
derpsteb, line 55: These should be moved down a bit based on: https://github.com/edgelesssys/wiki/blob/master/company/gostyle.md#order-of-elements-in-a-source-file

I would prefer `GetAvailableAttestationTypes` to be a `PrintAvailableAttestationTypes`. This would remove the need for these generic functions.

internal/variant/variant.go
```diff
@@ -45,6 +48,54 @@ const (
...
+// GetDefaultAttestation returns the default attestation type for the given provider. If not found, it returns the default variant.
+func GetDefaultAttestation(provider cloudprovider.Provider) Variant {
+	res, ok := providerAttestationMapping[provider]
+	if ok {
+		return res[0]
+	}
+	return Default{}
+}
+
+func removeDuplicate(sliceList []Variant) []Variant {
```
derpsteb, line 68: Should be moved further down. https://github.com/edgelesssys/wiki/blob/master/company/gostyle.md#order-of-elements-in-a-source-file

cli/internal/cmd/configgenerate.go
```diff
@@ -69,7 +84,10 @@ func (cg *configGenerateCmd) configGenerate(cmd *cobra.Command, fileHandler file
 
 	cg.log.Debugf("Parsed flags as %v", flags)
 	cg.log.Debugf("Using cloud provider %s", provider.String())
-	conf := createConfig(provider)
+	conf, err := createConfigWithAttestationType(provider, flags.attestationVariant)
+	if err != nil {
+		return err
```
derpsteb, line 89:
```suggestion
		return fmt.Errorf("creating config: %w", err)
```

cli/internal/cmd/configgenerate.go
```diff
@@ -105,7 +123,25 @@ func createConfig(provider cloudprovider.Provider) *config.Config {
 		conf.StateDiskSizeGB = 10
 	}
 
-	return conf
+	if provider == cloudprovider.Unknown {
+		return conf, nil // TODO tests use Unknown provider... the CLI doesn't allow it.. why do we do that?
```
derpsteb, line 127: please remove todo before merging.

internal/config/config.go
```diff
@@ -434,36 +434,49 @@ func (c *Config) UpdateMeasurements(newMeasurements measurements.M) {
...
 
+// RemoveProviderAndAttestationExcept calls RemoveProviderExcept and sets the default attestations for the provider (only used for convenience in tests).
+func (c *Config) RemoveProviderAndAttestationExcept(provider cloudprovider.Provider) {
+	c.RemoveProviderExcept(provider)
+	c.SetAttestation(variant.GetDefaultAttestation(provider))
+}
+
 // RemoveProviderExcept removes all provider specific configurations, i.e.,
 // sets them to nil, except the one specified.
-// If an unknown provider is passed, the same configuration is returned.
```
derpsteb, line 439: why remove the comment?

derpsteb:
Some last minor comments. Good to merge after that that :+1:

cli/internal/cmd/configgenerate.go
elchead: I know YAGNI, but I thought its likely we will want to access all attestation types in the future. Whereas we would definitely not need the print function elsewhere

PR #1758, opened 2023-05-10
Add autoscaling and cluster upgrade support for AWS

3u13r:
### Proposed change(s)

- Implement the AWS cloudprovider for our Constellation node operator
 This allows for autoscaling, K8s upgrade, and VM image upgrades

### Checklist

- [x] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)
- [x] Add labels (e.g., for changelog category)
- [x] Link to Milestone

PR #1772, opened 2023-05-16
ci: update the e2e-upgrade fromVersion

derpsteb:
### Proposed change(s)

- ~During weekly tests we always want to test upgrades from the latests release version to current main. The `fromVersion` has fallen behind and is updated. Also add a release step to have this updated automatically during release. The idea is that the bigger version of the two versions (existing fromVersion & release version) is kept in fromVersion.~
- automation is futile. will solve during refactoring.

### DIFF
--- a/.github/workflows/e2e-test-release.yml
+++ b/.github/workflows/e2e-test-release.yml
@@ -238,7 +238,7 @@ jobs:
       fail-fast: false
       max-parallel: 1
       matrix:
-        fromVersion: ["v2.6.0"]
+        fromVersion: ["v2.7.1"]
         cloudProvider: ["gcp", "azure"]
     name: Run upgrade tests
     secrets: inherit
--- a/.github/workflows/e2e-test-weekly.yml
+++ b/.github/workflows/e2e-test-weekly.yml
@@ -253,7 +253,7 @@ jobs:
       fail-fast: false
       max-parallel: 1
       matrix:
-        fromVersion: ["v2.6.0"]
+        fromVersion: ["v2.7.1"]
         cloudProvider: ["gcp", "azure"]
     name: Run upgrade tests
     secrets: inherit


.github/workflows/release.yml
```diff
@@ -126,6 +126,15 @@ jobs:
           git diff --staged --quiet || git commit -m "chore: update version.txt to ${{ inputs.version }}"
           git push origin "${BRANCH}"
 
+      - name: Update e2e-upgrade version
+        run: |
+          fromVersion=$(grep -oP 'fromVersion: \["\Kv[0-9]+.[0-9]+.[0-9]+' .github/workflows/e2e-test-release.yml)
+          newVersion=$(printf '%s\n${{ inputs.version }}' "$fromVersion" | sort -V | tail -n 1)
```
katexochen, line 132: Add comment why

.github/workflows/release.yml
```diff
@@ -126,6 +126,15 @@ jobs:
...
+          sed -i "s/fromVersion: \[\".*\"\]/fromVersion: \[\"$newVersion\"\]/g" .github/workflows/e2e-test-release.yml
```
katexochen, line 133: Write first sed regex to var and grep frist.

katexochen (changes requested):

3u13r (changes requested):
We should probably also update `.github/workflows/e2e-test-weekly.yml:256`  and then attach a test run in the PR.

derpsteb:
[Azure 1+1, full upgrade](https://github.com/edgelesssys/constellation/actions/runs/5001813500)

3u13r (approved):
LGTM

derpsteb:
Superseeded by: https://github.com/edgelesssys/constellation/pull/1758

PR #1781, opened 2023-05-16
ci: run `upgrade check` during e2e-upgrade

derpsteb:
### Proposed change(s)

- We did not run a e2e test on `upgrade check` so far. This add a basic smoke test. A detailed verification of the output seems like too much effort for little gain to me. This patch only checks for existence of some strings.
- Fix a bug where the supported microservice versions are not reported correctly in `upgrade check`
- [Azure, 1+1, no upgrades](https://github.com/edgelesssys/constellation/actions/runs/4992674280) :green_circle: 

### Checklist

- [x] Link to Milestone

### DIFF
--- a/e2e/internal/upgrade/upgrade_test.go
+++ b/e2e/internal/upgrade/upgrade_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/edgelesssys/constellation/v2/internal/constants"
 	"github.com/edgelesssys/constellation/v2/internal/file"
 	"github.com/edgelesssys/constellation/v2/internal/semver"
+	"github.com/edgelesssys/constellation/v2/internal/versions"
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/require"
 	coreV1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


e2e/internal/upgrade/upgrade_test.go
```diff
@@ -125,6 +124,20 @@ func TestUpgrade(t *testing.T) {
...
 }
 
+func runUpgradeCheck(require *require.Assertions, cli string) {
+	cmd := exec.CommandContext(context.Background(), cli, "upgrade", "check", "--debug")
+	msg, err := cmd.CombinedOutput()
+	require.NoErrorf(err, "%s", string(msg))
+
+	// We can't check for the exact output as it depends on multiple factors.
+	// But each CI run should at least include these lines.
+	require.True(strings.Contains(string(msg), "The following updates are available with this CLI:"))
```
3u13r, line 134:
```suggestion
	require.Contains(string(msg), "The following updates are available with this CLI:")
	require.Contains(string(msg), "Kubernetes:")
	require.Contains(string(msg), "Services:")
```

Couldn't we use the test variables from above (lines 38ff) to also check for the versions the CLI wants to upgrade from/to?

e2e/internal/upgrade/upgrade_test.go
```diff
@@ -125,6 +124,20 @@ func TestUpgrade(t *testing.T) {
 	testNodesEventuallyHaveVersion(t, k, targetVersions, *wantControl+*wantWorker, *timeout)
 }
 
+func runUpgradeCheck(require *require.Assertions, cli string) {
+	cmd := exec.CommandContext(context.Background(), cli, "upgrade", "check", "--debug")
+	msg, err := cmd.CombinedOutput()
+	require.NoErrorf(err, "%s", string(msg))
```
3u13r, line 130: Why did you choose `require.NoErrorf` over `require.NoError` ?

3u13r (changes requested):

e2e/internal/upgrade/upgrade_test.go
derpsteb: can't remember. seems unnecessary. changed.

e2e/internal/upgrade/upgrade_test.go
derpsteb: We can't check images because targetImages may be unpublished. Added checks for the other two cases.

3u13r (approved):
LGTM

PR #1786, opened 2023-05-16
config: support latest as version value

elchead:
### Proposed change(s)

- support to set "latest" as value for the respective version of Azure SEVSNP. The value gets replaced with the current version when reading the config file.

For now, the values are harcoded. Version checking against a JSON in a S3 bucket will be added/

see AB3042

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -27,6 +27,7 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/edgelesssys/constellation/v2/internal/config/version"
 	"github.com/go-playground/locales/en"
 	ut "github.com/go-playground/universal-translator"
 	"github.com/go-playground/validator/v10"
@@ -704,16 +705,16 @@ type AzureSEVSNP struct {
 	Measurements measurements.M `json:"measurements" yaml:"measurements" validate:"required,no_placeholders"`
 	// description: |
 	//   Lowest acceptable bootloader version.
-	BootloaderVersion uint8 `json:"bootloaderVersion" yaml:"bootloaderVersion"`
+	BootloaderVersion version.Version `json:"bootloaderVersion" yaml:"bootloaderVersion"`
 	// description: |
 	//   Lowest acceptable TEE version.
-	TEEVersion uint8 `json:"teeVersion" yaml:"teeVersion"`
+	TEEVersion version.Version `json:"teeVersion" yaml:"teeVersion"`
 	// description: |
 	//   Lowest acceptable SEV-SNP version.
-	SNPVersion uint8 `json:"snpVersion" yaml:"snpVersion"`
+	SNPVersion version.Version `json:"snpVersion" yaml:"snpVersion"`
 	// description: |
 	//   Lowest acceptable microcode version.
-	MicrocodeVersion uint8 `json:"microcodeVersion" yaml:"microcodeVersion"`
+	MicrocodeVersion version.Version `json:"microcodeVersion" yaml:"microcodeVersion"`
 	// description: |
 	//   Configuration for validating the firmware signature.
 	FirmwareSignerConfig SNPFirmwareSignerConfig `json:"firmwareSignerConfig" yaml:"firmwareSignerConfig"`
@@ -724,14 +725,13 @@ type AzureSEVSNP struct {
 
 // DefaultForAzureSEVSNP returns the default configuration for Azure SEV-SNP attestation.
 // Version numbers are hard coded and should be updated with each new release.
-// TODO(AB#3042): replace with dynamic lookup for configurable values.
 func DefaultForAzureSEVSNP() *AzureSEVSNP {
 	return &AzureSEVSNP{
 		Measurements:      measurements.DefaultsFor(cloudprovider.Azure),
-		BootloaderVersion: 2,
-		TEEVersion:        0,
-		SNPVersion:        6,
-		MicrocodeVersion:  93,
+		BootloaderVersion: version.GetVersion(version.Bootloader),
+		TEEVersion:        version.GetVersion(version.TEE),
+		SNPVersion:        version.GetVersion(version.SNP),
+		MicrocodeVersion:  version.GetVersion(version.Microcode),
 		FirmwareSignerConfig: SNPFirmwareSignerConfig{
 			AcceptedKeyDigests: idkeydigest.DefaultList(),
 			EnforcementPolicy:  idkeydigest.MAAFallback,
@@ -772,6 +772,36 @@ func (c AzureSEVSNP) EqualTo(old AttestationCfg) (bool, error) {
 	rootKeyEqual := bytes.Equal(c.AMDRootKey.Raw, otherCfg.AMDRootKey.Raw)
 
 	return firmwareSignerCfgEqual && measurementsEqual && bootloaderEqual && teeEqual && snpEqual && microcodeEqual && rootKeyEqual, nil
+}
+
+// UnmarshalYAML implements a custom unmarshaler to resolve the latest version value.
+func (c *AzureSEVSNP) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type rawAzureSEVSNP AzureSEVSNP
+	if err := unmarshal((*rawAzureSEVSNP)(c)); err != nil {
+		return err
+	}
+	c.updateVersion(version.Bootloader)
+	c.updateVersion(version.TEE)
+	c.updateVersion(version.SNP)
+	c.updateVersion(version.Microcode)
+	return nil
+}
+
+func (c *AzureSEVSNP) updateVersion(versionType version.Type) {
+	var v *version.Version
+	switch versionType {
+	case version.Bootloader:
+		v = &c.BootloaderVersion
+	case version.TEE:
+		v = &c.TEEVersion
+	case version.SNP:
+		v = &c.SNPVersion
+	case version.Microcode:
+		v = &c.MicrocodeVersion
+	}
+	if *v == "latest" {
+		*v = version.GetVersion(versionType)
+	}
 }
 
 // SNPFirmwareSignerConfig is the configuration for validating the firmware signer.
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -87,6 +87,28 @@ func TestFromFile(t *testing.T) {
 				return conf
 			}(),
 		},
+		"latest as version value gets replaced with integer value": {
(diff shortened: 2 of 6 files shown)


internal/config/config.go
daniel-weisse, line 1: Might be a good idea to move the attestation config structs to their own file in the package

internal/config/config.go
```diff
@@ -802,6 +803,46 @@ func (c AzureSEVSNP) EqualTo(old AttestationCfg) (bool, error) {
...
+		if !convertLatestToNumber(a, versionType, aux) {
+			if !convertStringToUint(a, versionType, aux) {
+				return fmt.Errorf("failed to convert %s version to number", versionType)
+			}
+		}
+	}
+	return nil
+}
+
+func getUintAndStringPtrToVersion(versionType version.Type, c *AzureSEVSNP, aux *fusedAzureSEVSNP) (*uint8, *string) {
```
daniel-weisse, line 826:
```suggestion
func getUintAndStringPtrToVersion(c *AzureSEVSNP, versionType version.Type, aux *fusedAzureSEVSNP) (*uint8, *string) {
```
nit:  `convertStringToUint` and `convertLatestToNumber` use a different order of function arguments.

internal/config/config.go
```diff
@@ -802,6 +803,46 @@ func (c AzureSEVSNP) EqualTo(old AttestationCfg) (bool, error) {
...
+	var v *uint8
+	var stringV *string
```
daniel-weisse, line 828:
```suggestion
	var version *uint8
	var stringVersion *string
```
Prefer using more descriptive variables names, and keep single letter variable names restricted to loop variables or receiver names.
Some parts of these functions are quite hard to read otherwise

internal/config/config.go
```diff
@@ -802,6 +803,46 @@ func (c AzureSEVSNP) EqualTo(old AttestationCfg) (bool, error) {
...
+	}
+	a = (*AzureSEVSNP)(aux.auxAzureSEVSNP)
+
+	for _, versionType := range []version.Type{version.Bootloader, version.TEE, version.SNP, version.Microcode} {
+		if !convertLatestToNumber(a, versionType, aux) {
+			if !convertStringToUint(a, versionType, aux) {
+				return fmt.Errorf("failed to convert %s version to number", versionType)
+			}
+		}
+	}
```
daniel-weisse, line 822: If `convertLatestToNumber` would instead be a `convertLatestToString`, you could call `convertStringToUint` unconditionally, which improve readability a bit.
Not required though, keep what you prefer.

internal/config/config.go
```diff
@@ -947,3 +988,66 @@ func (c QEMUTDX) EqualTo(other AttestationCfg) (bool, error) {
...
+func convertStringToUint(c *AzureSEVSNP, versionType version.Type, aux *fusedAzureSEVSNP) bool {
+	v, stringV := getUintAndStringPtrToVersion(versionType, c, aux)
+
+	bvInt, err := strconv.ParseInt(*stringV, 10, 8)
+	if err != nil {
+		return false
+	}
+	*v = uint8(bvInt)
+	return true
+}
```
daniel-weisse, line 1044:
```suggestion
func convertStringToUint(c *AzureSEVSNP, versionType version.Type, aux *fusedAzureSEVSNP) error {
	v, stringV := getUintAndStringPtrToVersion(versionType, c, aux)

	bvInt, err := strconv.ParseInt(*stringV, 10, 8)
	if err != nil {
		return err
	}
	*v = uint8(bvInt)
	return nil
}
```
No need to lose information by not returning the error

internal/config/config.go
```diff
@@ -947,3 +988,66 @@ func (c QEMUTDX) EqualTo(other AttestationCfg) (bool, error) {
...
+
+func convertLatestToNumber(c *AzureSEVSNP, versionType version.Type, aux *fusedAzureSEVSNP) bool {
+	v, stringV := getUintAndStringPtrToVersion(versionType, c, aux)
+	if *stringV == "latest" {
```
daniel-weisse, line 1048:
```suggestion
	if strings.ToLower(*stringV) == "latest" {
```
nit

internal/config/config.go
elchead: I agree but it would imply 2 conversions since changing GetVersion to return a string doesn't make sense. So i rather keep it as is

elchead:
@daniel-weisse went through your feedback :)

internal/config/azure.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	if err := unmarshal(aux); err != nil {
+		return fmt.Errorf("failed to unmarshal AzureSEVSNP: %w", err)
+	}
+	c = (*AzureSEVSNP)(aux.auxAzureSEVSNP)
+
+	for _, versionType := range []version.Type{version.Bootloader, version.TEE, version.SNP, version.Microcode} {
+		if !convertLatestToNumber(c, versionType, aux) {
+			if err := convertStringToUint(c, versionType, aux); err != nil {
+				return fmt.Errorf("failed to convert %s version to number: %w", versionType, err)
```
thomasten, line 111:
```suggestion
		return fmt.Errorf("unmarshal AzureSEVSNP: %w", err)
	}
	c = (*AzureSEVSNP)(aux.auxAzureSEVSNP)

	for _, versionType := range []version.Type{version.Bootloader, version.TEE, version.SNP, version.Microcode} {
		if !convertLatestToNumber(c, versionType, aux) {
			if err := convertStringToUint(c, versionType, aux); err != nil {
				return fmt.Errorf("convert %s version to number: %w", versionType, err)
```
avoid phrases like "failed to", "error when", etc. in error strings because it gets too verbose when they get chained

internal/config/config_test.go
```diff
@@ -36,6 +37,78 @@ func TestDefaultConfig(t *testing.T) {
...
+				assert.Error(err)
+			} else {
+				require.NoError(err)
+				assert.Equal(tc.wantResult, result)
+			}
+		})
+	}
+}
+
+// getConfigAsMap returns a map of the attestation section of the config.
```
thomasten, line 98:
```suggestion
// getConfigAsMap returns a map of the config.
```

internal/config/config_test.go
```diff
@@ -36,6 +37,78 @@ func TestDefaultConfig(t *testing.T) {
...
+func getConfigAsMap(conf *Config, t *testing.T) map[string]interface{} {
+	d, err := yaml.Marshal(&conf)
+	if err != nil {
+		t.Fatal(err)
+	}
+	var m map[string]interface{}
+	err = yaml.Unmarshal(d, &m)
+	if err != nil {
```
thomasten, line 106:
```suggestion
	if if err := yaml.Unmarshal(d, &m); err != nil {
```
also, a bit more descriptive variable names would improve readability here IMO

internal/config/version/version.go
```diff
@@ -0,0 +1,33 @@
...
+
+const (
+	Bootloader Type = "bootloader" // Bootloader is the version of the Azure SEVSNP bootloader.
+	TEE        Type = "tee"        // TEE is the version of the Azure SEVSNP TEE.
+	SNP        Type = "snp"        // SNP is the version of the Azure SEVSNP SNP.
+	Microcode  Type = "microcode"  // Microcode is the version of the Azure SEVSNP microcode.
+)
+
+// Type is the type of the version to be requested.
+type Type (string)
```
thomasten, line 17:
```suggestion
type Type string
```

internal/config/version/version.go
```diff
@@ -0,0 +1,33 @@
...
+	case Bootloader:
+		return 2
+	case TEE:
+		return 0
+	case SNP:
+		return 6
+	case Microcode:
+		return 93
+	default:
+		return 1
```
thomasten, line 31: Not sure what's a good default here (if any). May also depend on how the caller uses the value. For the current use case, MaxUint8 may be the secure choice so that the attestation doesn't accidentally accept an outdated version.
One could also argue that this is a "can't happen"/logic error and panic.
Or return an error, which may be needed anyway when getting the value from the API.
(maybe @daniel-weisse also has an opinion on this)

thomasten (commented):
just some nits and one minor point to discuss

internal/config/version/version.go
daniel-weisse: I final version of the code which loads these values from our API will probably look a bit different.
IMO throwing a panic here works to make sure we dont have any logic errors in the config code

elchead: I agree, panicking makes even sense. I think failing early is a good principle. And also, the function will be replaced with the S3 logic in the next PR :) 
Would appreciate some design feedback here too https://github.com/edgelesssys/constellation/pull/1808

elchead:
went over the feedback :)

internal/config/version/version.go
derpsteb, line 1: I'd prefer this to have a more specific name, for example `snpversion`. There is another package `internal/versions`. And right now the package only holds SNP specific values. We can rename the pkg in case we want to include different version types.

internal/config/version/version.go
```diff
@@ -0,0 +1,33 @@
...
+	TEE        Type = "tee"        // TEE is the version of the Azure SEVSNP TEE.
+	SNP        Type = "snp"        // SNP is the version of the Azure SEVSNP SNP.
+	Microcode  Type = "microcode"  // Microcode is the version of the Azure SEVSNP microcode.
+)
+
+// Type is the type of the version to be requested.
+type Type string
+
+// GetVersion returns the version of the given type.
+func GetVersion(t Type) uint8 {
```
derpsteb, line 20:
```suggestion
// GetLatest returns the version of the given type.
func GetLatest(t Type) uint8 {
```

internal/config/version/version.go
```diff
@@ -0,0 +1,33 @@
...
+	case Bootloader:
+		return 2
+	case TEE:
+		return 0
+	case SNP:
+		return 6
+	case Microcode:
+		return 93
+	default:
+		panic("invalid version type")
```
derpsteb, line 31: Should be an error? I guess there will be more errors that should be handled gracefully once this function starts talking to the configapi.

internal/config/version/version.go
elchead: the package is disappearing in the next PR, where it moves to configapi pkg. Is it ok?

elchead: applied the change

derpsteb (approved):

PR #1808, opened 2023-05-19
config: dynamic attestation configuration through S3 backed API

elchead:
The PR is still missing a signature for the version upload to the S3 bucket. I suggest to do that in a new PR, because it's a big change by itself and to avoid making this PR even bigger

 

### Proposed change(s)

- introduce new API endpoints `/constellation/v1/attestation/<ATTESTATION_VARIANT>/` as specified in RFC to fetch version values https://github.com/edgelesssys/constellation/blob/eed533932e1f5e98b9b626ffe10ccc8ca780df66/rfc/attestation-config.md
- new tool `bazel run //hack/configapi:upload` to upload versions (e.g. microcode version) of Azure SEV SNP (other variants don't have versions at the moment) to an S3 bucket
- extend #1786 to fetch the version values from the new API upon loading the config file through `config.New` when the Azure SEV-SNP attestation variant is used. Furthermore, it improves the prior PR by persisting the version values as "latest" in the yaml file. Previously, the latest version values were fetched upon reading the config from file and consequently these version numbers were stored on subsequent saves to file.

see [AB#3045](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3045)

### Checklist

- [ ] Link to Milestone


elchead:
Should we add a fallback to hardcoded values when the API call fails or should we error in that case?

derpsteb:
If `latest` was specified in the config, init/verify should fail if the value for `latest` can't be determined, imo. 

I think @katexochen should have a look at this once it's ready.

elchead:
@katexochen it's ready, could you have a look please?

internal/config/config.go
```diff
@@ -358,12 +361,21 @@ func Default() *Config {
...
+			AzureSEVSNP:        azureSEVSNP,
 			AzureTrustedLaunch: &AzureTrustedLaunch{Measurements: measurements.DefaultsFor(cloudprovider.Azure, variant.AzureTrustedLaunch{})},
 			GCPSEVES:           &GCPSEVES{Measurements: measurements.DefaultsFor(cloudprovider.GCP, variant.GCPSEVES{})},
 			QEMUVTPM:           &QEMUVTPM{Measurements: measurements.DefaultsFor(cloudprovider.QEMU, variant.QEMUVTPM{})},
 		},
+	}, nil
+}
+
+// DefaultWithPanic returns a struct with the default config.
+func DefaultWithPanic() *Config {
```
katexochen, line 373: [Don't panic](https://go-proverbs.github.io/)

We try to avoid using panics. There are rare occasions where it is necessary (regexp.MustCompile), but in these cases the code must panic immediately at startup, and thus be global state. I don't see a need to panic when creating a default config.

katexochen (dismissed):
@elchead Could you please clean up your commits and provide a meaningful commit history?

internal/api/fetcher/configapi.go
derpsteb, line 1: We also need update functionality, cf. https://github.com/edgelesssys/constellation/blob/feat/config/s3-api/internal/api/versionsapi/client/client.go#L150.

cli/internal/cmd/upgradeapply_test.go
```diff
@@ -30,6 +30,8 @@ import (
 
 func TestUpgradeApply(t *testing.T) {
 	someErr := errors.New("some error")
+	azureSEVSNP, err := config.DefaultForAzureSEVSNP()
+	require.NoError(t, err)
```
daniel-weisse, line 34: Since this function makes real API calls now, we should not use it in unit tests, but rather have a function that returns placeholder values or use an empty struct

internal/config/config.go
katexochen: We don't differentiate code quality of production and test code.

Maybe something that fills default values into a struct shouldn't return an error in the first place.

internal/attestation/azure/snp/validator_test.go
```diff
@@ -348,7 +348,10 @@ func TestNewSNPReportFromBytes(t *testing.T) {
 			wantErr: true,
 		},
 	}
-
+	cfg, err := config.DefaultForAzureSEVSNP()
+	if err != nil {
+		t.Fatal(err)
+	}
```
daniel-weisse, line 354:
```suggestion
	require.NoError(t, err)
```

internal/config/config.go
daniel-weisse: If this function is only used in unit tests, it shouldnt make real API calls (as it is done when calling `DefaultForAzureSEVSNP()`)

I think a better solution for this would be rewriting `DefaultForAzureSEVSNP()` to return some placeholder values for the variables that can be dynamically fetched from our API.
These placeholders should marshal to `"latest"` when we write the config to disk.

daniel-weisse (changes requested):

daniel-weisse:
Hmm, the first option could work.
We don't expect the attestation configs to have any values set to `latest` in any other place but the user's config.
So we could go with a similar approach as for the measurements with [`FetchAndVerify`](https://github.com/edgelesssys/constellation/blob/4a8617ea18a7ffef32135d0a4888ec03e4e34626/internal/attestation/measurements/measurements.go#L75).
Keep the unmarshaling logic free of API calls, but add a method to parse `latest` to actual version.

elchead:
@daniel-weisse , @katexochen I know the commit history is a huge mess and I'm sorry about it. I will start rebasing earlier in future PRs. I tried rebasing but it's a big undertaking that is likely to introduce errors along the way. Is it ok? I'm more than happy to jump on a call to discuss and explain design details.

internal/config/azure.go
```diff
@@ -45,14 +45,14 @@ type AzureSEVSNP struct {
 }
 
 // DefaultForAzureSEVSNP returns the default configuration for Azure SEV-SNP attestation.
-// Version numbers are hard coded and should be updated with each new release.
+// Version numbers have dummy values and the true values are fetched from the configapi when calling config.readFile().
```
daniel-weisse, line 48:
```suggestion
// Version numbers have placeholder values and the latest available values can be fetched using [AzureSEVSNP.FetchAndSetLatestVersionNumbers].
```

internal/api/configapi/attestation.go
```diff
@@ -0,0 +1,95 @@
...
+
+// ValidateRequest validates the request.
+func (i AzureSEVSNPVersionGet) ValidateRequest() error {
+	return nil
+}
+
+// Validate validates the request.
+func (i AzureSEVSNPVersionGet) Validate() error {
+	return nil
+}
```
daniel-weisse, line 67: Are these functions expected to be implemented at some point?
Or will they always remain a no-op?
If so, please add a comment

daniel-weisse (commented):
Rebasing is fine for me

malt3:
nit: if you use this format for the ticket number it will cross reference the PR and ticket: AB#3045

internal/api/fetcher/configapi.go
```diff
@@ -0,0 +1,57 @@
...
+func (f *ConfigAPIFetcher) FetchAzureSEVSNPVersionList(ctx context.Context, attestation configapi.AzureSEVSNPVersionList) (configapi.AzureSEVSNPVersionList, error) {
+	return fetch(ctx, f.httpc, attestation)
+}
+
+// FetchAzureSEVSNPVersion fetches the version information from the config API.
+func (f *ConfigAPIFetcher) FetchAzureSEVSNPVersion(ctx context.Context, attestation configapi.AzureSEVSNPVersionGet) (configapi.AzureSEVSNPVersionGet, error) {
+	// TODO check sig
+	//if err := sigstore.VerifySignature(measurements, signature, publicKey); err != nil {
+	//	return "", err
+	//}
```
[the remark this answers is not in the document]
katexochen: This should be `TODO(elchead):`if you want to create a follow up ticket.

internal/api/fetcher/configapi.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// FetchAzureSEVSNPVersionList fetches the version list information from the config API.
+func (f *ConfigAPIFetcher) FetchAzureSEVSNPVersionList(ctx context.Context, attestation configapi.AzureSEVSNPVersionList) (configapi.AzureSEVSNPVersionList, error) {
+	return fetch(ctx, f.httpc, attestation)
+}
+
+// FetchAzureSEVSNPVersion fetches the version information from the config API.
+func (f *ConfigAPIFetcher) FetchAzureSEVSNPVersion(ctx context.Context, attestation configapi.AzureSEVSNPVersionGet) (configapi.AzureSEVSNPVersionGet, error) {
+	// TODO check signature (sigstore.VerifySignature)
```
malt3, line 34:
```suggestion
	// TODO(elchead): check signature (sigstore.VerifySignature)
```
nit. Have a point of contact  if someone else reads the code and wants to know what needs to be done.

hack/configapi/main.go
```diff
@@ -0,0 +1,67 @@
...
+	snpVersion        = pflag.Uint8P("snp-version", "s", 0, "SNP version number")
+	microcodeVersion  = pflag.Uint8P("microcode-version", "m", 0, "Microcode version number")
+)
+
+func main() {
+	pflag.Parse()
+	if *awsAccessKey == "" || *awsAccessKeyID == "" {
+		pflag.Usage()
+		fmt.Println("Required flags not set: --aws-access-key, --aws-access-key-id")
+		os.Exit(0)
```
daniel-weisse, line 42:
```suggestion
		os.Exit(1)
```
exit 1 to indicate failure

internal/api/configapi/attestation.go
```diff
@@ -0,0 +1,102 @@
...
+	"net/url"
+	"path"
+	"strings"
+
+	"github.com/edgelesssys/constellation/v2/internal/constants"
+	"github.com/edgelesssys/constellation/v2/internal/variant"
+)
+
+// baseURL is the base URL of the config api.
+var baseURL = constants.CDNRepositoryURL
```
daniel-weisse, line 19: Since we dont manipulate this for test anymore, I think the package can use `constants.CDNRepositoryURL` directly

internal/api/configapi/attestationversion.go
```diff
@@ -0,0 +1,80 @@
...
+
+const dummyVersionValue = 0
+
+// NewLatestDummyVersion returns the latest version with a dummy version value.
+func NewLatestDummyVersion() AttestationVersion {
+	return AttestationVersion{
+		Value:    dummyVersionValue,
+		IsLatest: true,
+	}
+}
```
daniel-weisse, line 23:
```suggestion
const placeholderVersionValue = 0

// NewPlaceholderVersion returns a placeholder for the latest version.
func NewPlaceholderVersion() AttestationVersion {
	return AttestationVersion{
		Value:    placeholderVersionValue,
		IsLatest: true,
	}
}
```

To quote the Google developer guide: Don't use dummy to refer to placeholders. Instead, use placeholder.

daniel-weisse (commented):
Just some small things that were missed in the last commit

internal/api/fetcher/configapi_test.go
```diff
@@ -0,0 +1,25 @@
+/*
+Copyright (c) Edgeless Systems GmbH
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+package fetcher_test
```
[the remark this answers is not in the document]
katexochen: That might be the case for a library or if you have many pure functions. Most of out packages are quite complex. We usually need the introspection into structs etc to inject stubs and assert. We also don't work with a stable API between packages and don't maintain a public API, so there isn't much to gain from testing from an "outside" perspective. You can find some more infos @ https://github.com/edgelesssys/testing.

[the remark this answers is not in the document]
katexochen: But I'm happy to discuss in case you think we should change this. :smile:

internal/api/fetcher/fetcher.go
```diff
@@ -90,6 +99,7 @@ func (e *NotFoundError) Unwrap() error {
 	return e.err
 }
 
-type httpc interface {
+// HttpClienter is an interface for http clients.
+type HttpClienter interface {
```
daniel-weisse, line 103:
```suggestion
// HttpClient is an interface for http clients.
type HttpClient interface {
```

internal/config/config_test.go
```diff
@@ -878,3 +880,45 @@ func getConfigAsMap(conf *Config, t *testing.T) (res configMap) {
 	}
 	return
 }
+
+type dummyConfigAPIHandler struct{}
```
daniel-weisse, line 884:
```suggestion
type fakeConfigAPIHandler struct{}
```
nit: naming

malt3:
Sorry for the merge conflicts. Those should be simple to resolve since it is just a move of the files in your PR.

internal/config/config_test.go
elchead: you really dislike dummies ;)  In the testing community there is a distinction between mocks, fakes, dummies, spies.
A fake implies some kind of logic whereas a dummy just returns canned responses.  
https://blog.cleancoder.com/uncle-bob/2014/05/14/TheLittleMocker.html

Why do you want to use fake?

daniel-weisse: As you already said yourself, a fake implies some kind of logic, which is the case for this handler, no?

The distinction between dummies, fakes, mocks, or dummies is, imo, most of the time rather unclear and a lot of those terms get used interchangeably by people (which also gets touched on in the article you linked).
Going strictly by the terminology laid out by the article the api handler would either be a stub, or a fake, since we *do* care about how it is used, and there is a slight bit of logic in the handler

daniel-weisse (approved):
Looks good from my side

malt3 (approved):

elchead:
I had to set the placeholder back to 0 because attestation tests were failing with MaxInt. I can't run them locally, but I will look into that on Tuesday

PR #1819, opened 2023-05-23
ci: add pipeline to precalc launchmeasurements

derpsteb:
### Proposed change(s)

- Adds a pipeline that precalculates launch measurements for AWS SNP. This measurement is distinct from the image measurement. The measurement is calculated by the CPU, while starting the CVM. It is a measurement of the initial firmware that is supplied by AWS.
- In a subsequent PR we will take the measurements, sign them and put them into the attestation API.
- [Test run](https://github.com/edgelesssys/constellation/actions/runs/5059204989/jobs/9080405102)
### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/aws-snp-launchmeasurement.yml
+++ b/.github/workflows/aws-snp-launchmeasurement.yml
@@ -1,0 +1,68 @@
+name: AWS SNP Launch Measurement
+
+on:
+  schedule:
+    # Run daily at 22:00.
+    - cron: '0 22 * * *'
+  workflow_dispatch:
+
+jobs:
+  run:
+    runs-on: ubuntu-22.04
+
+    steps:
+    - name: Install necessary tools
+      run: |
+        sudo apt-get update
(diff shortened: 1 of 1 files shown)


.github/workflows/aws-snp-launchmeasurement.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    steps:
+    - name: Install necessary tools
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y python3 python3-pip
+        sudo python3 -m pip install sev-snp-measure==0.0.5
+
+    - name: Install Nix
+      uses: cachix/install-nix-action@v20
```
malt3, line 20: All actions should be pinned to a commit hash. Version can be specified in a comment after that so renovate can understand it.

malt3 (approved):
Just some nits. Otherwise LGTM

PR #1820, opened 2023-05-23
cli: use uami for in-cluter authentication

3u13r:
### Proposed change(s)

- Don't create App Registration anymore
- `ApplicationID` and `ApplicationClientSecretValue` are optional now, but are used for authentication if present. Also we print a deprecation warning linking to the migration guide.
- If they are not present, we assume that the config and IAM credentials were recently generated 

### Checklist

- [x] Link to Milestone

### DIFF
--- a/docs/docs/reference/config-migration.md
+++ b/docs/docs/reference/config-migration.md
@@ -1,7 +1,18 @@
-# Configuration migrations
+# Migrations
 
-This document describes breaking changes in the configuration file format between Constellation releases.
+This document describes breaking changes and migrations between Constellation releases.
 Use [`constellation config migrate`](./cli.md#constellation-config-migrate) to automatically update an old config file to a new format.
+
+## Migrating from Azure's service principal authentication to managed identity authentication
+
+- The `provider.azure.appClientID` and `provider.azure.appClientSecret` fields are no longer required and should be removed.
+- To keep using an existing UAMI add the `Owner` permission with the scope of your `resourceGroup`.
+- Otherwise, simply [create new Constellation IAM credentials](../workflows/config.md#creating-iam-credentials) and use the created UAMI.
+- To migrate the authentication for an existing Constellation on Azure to an UAMI with the necessary permissions:
+  1. Remove the `aadClientId` and `aadClientSecret` from the azureconfig secret.
+  2. Set `useManagedIdentityExtension` to `true`  and use the `userAssignedIdentity` from the Constellation config for the value of `userAssignedIdentityID`.
+  3. Restart the CSI driver, cloud controller manager, cluster autoscaler, and Constellation operator pods.
+
 
 ## Migrating from CLI versions before 2.8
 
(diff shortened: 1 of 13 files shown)


docs/docs/reference/config-migration.md
```diff
@@ -1,8 +1,14 @@
...
-This document describes breaking changes in the configuration file format between Constellation releases.
+This document describes breaking changes and migrations between Constellation releases.
 Use [`constellation config migrate`](./cli.md#constellation-config-migrate) to automatically update an old config file to a new format.
 
+## Migrating from Azure's service principal authentication to managed identity authentication
+
+- The `provider.azure.appClientID` and `provider.azure.appClientSecret` fields are no longer required and should be removed.
+- To keep using an existing UAMI add the `Owner` permission with the scope of your `resourceGroup`.
+- Otherwise, simply [create new Constellation IAM credentials](../workflows/config.md#creating-iam-credentials) and use the created UAMI.
+
```
malt3, line 11: I am not super convinced of my suggestion below so feel free to not follow it:

We could also explain how to patch the `azureconfig` k8s secret to use the uami (and no longer depend on the app id / app secret).

3u13r: Ah yes, I actually just forgot about it. I think we should at least point to the secret name and the docs of the `cloud-provider-azure`. But since it can be done with a single command, we can also just add that.

internal/config/config.go
daniel-weisse, line 1: `AppClientID` and `ClientSecretValue` should print a deprecation warning if set

internal/config/config.go
```diff
@@ -407,6 +407,14 @@ func New(fileHandler file.Handler, name string, force bool) (*Config, error) {
 		c.MicroserviceVersion = Default().MicroserviceVersion
 	}
 
+	// TODO(3u13r): Remove this deprecation warning and enforce assigned managed identity after the v2.8.0 but before the v2.9.0 release.
+	if c.Provider.Azure != nil &&
+		(c.Provider.Azure.AppClientID != "" || c.Provider.Azure.ClientSecretValue != "") {
```
daniel-weisse, line 412: We already have the validation function [`warnDeprecated`](https://github.com/edgelesssys/constellation/blob/757fccf8260b34afb5008c015600e32a6c16458a/internal/config/validation.go#L518).
The fields should either use the same function to print a warning, or we should align the style of the warning message.

internal/config/config.go
```diff
@@ -411,8 +411,8 @@ func New(fileHandler file.Handler, name string, force bool) (*Config, error) {
 	if c.Provider.Azure != nil &&
 		(c.Provider.Azure.AppClientID != "" || c.Provider.Azure.ClientSecretValue != "") {
 		// Deprecation warning for old auth method
-		fmt.Println("DEPRECTATION WARNING: Using a service principal for authentication is deprecated and will be removed in a future release.")
-		fmt.Println("DEPRECTATION WARNING: Migrate to using a user assigned managed identity by following the migration guide: https://docs.edgeless.systems/constellation/reference/migration.")
+		fmt.Fprintf(os.Stderr, "WARNING: Using a service principal for authentication s deprecated and will be removed in an upcoming version.")
```
daniel-weisse, line 414:
```suggestion
		fmt.Fprintf(os.Stderr, "WARNING: Using a service principal for authentication is deprecated and will be removed in an upcoming version.")
```

daniel-weisse (approved):

malt3 (approved):

PR #1827, opened 2023-05-25
ci: fix aws-snp-launchmeasurement pipeline

derpsteb:
Misspelled variable name.
[testrun](https://github.com/edgelesssys/constellation/actions/runs/5077029708).


derpsteb:
same..I guess I pushed something before merging..

PR #1836, opened 2023-05-30
config: sign Azure versions on upload & verify on fetch

elchead:
### Proposed change(s)

- sign the Azure SEV SNP version file when uploading to S3
- verify the signature when fetching version data from the API
- use the new staticupload.Client to correctly invalidate the cache.

### Work for follow-up PR
- new Github action using the upload tool
- delete specific version in upload tool

see [AB#3045](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3045)
Follow-up PR from #1808

### Additional info
To test the upload tool:
```
AWS_ACCESS_KEY_ID=$ID AWS_ACCESS_KEY=$KEY bazel run //hack/configapi:upload -- --version-file $FILEPATH --cosign-pwd $PWD --private-key $KEY_PATH
```
Note: we use the `cosign-dev` key

### Checklist

- [x] Link to Milestone


elchead:
should we add documentation for the upload tool? Where would be a good place?

elchead:
should we add a delete option to the tool to purge all versions + delete specific versions (delete .json + .sig + entry in `/list`)?

daniel-weisse:
>should we add documentation for the upload tool? Where would be a good place?

Can be added as a package comment, or added tot he help command of the tool

>should we add a delete option to the tool to purge all versions + delete specific versions (delete .json + .sig + entry in /list)?

I dont think there ever is a  valid case where we would want to delete all versions.
Having the option to simply remove one version could be useful tho, if we ever happen to upload some broken data.
Since properly removing an entry requires manipulating the list file, doing this "by hand" could be a bit annoying.

internal/sigstore/sign.go
```diff
@@ -0,0 +1,23 @@
...
+	"encoding/base64"
+
+	"github.com/sigstore/cosign/v2/pkg/cosign"
+)
+
+// SignContent signs the content with the cosign encrypted private key and corresponding cosign password.
+func SignContent(password, encryptedPrivateKey, content []byte) ([]byte, error) {
+	sv, err := cosign.LoadPrivateKey(encryptedPrivateKey, password)
+	if err != nil {
+		return nil, err
```
derpsteb, line 19:
```suggestions
		return nil, fmt.Errorf("loading private key: %w", err)
```

internal/sigstore/sign_test.go
```diff
@@ -0,0 +1,26 @@
...
+package sigstore
+
+import (
+	"testing"
+
+	"github.com/sigstore/cosign/v2/pkg/cosign"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSignSignature(t *testing.T) {
```
derpsteb, line 15: Please add test inputs for (a) wrong password, (b) failing verification, i.e. missmatching keys.

internal/api/fetcher/configapi.go
```diff
@@ -35,17 +42,47 @@ func (f *ConfigAPIFetcher) FetchAzureSEVSNPVersionList(ctx context.Context, atte
...
+}
+
 // FetchLatestAzureSEVSNPVersion returns the latest versions of the given type.
 func (f *ConfigAPIFetcher) FetchLatestAzureSEVSNPVersion(ctx context.Context) (res configapi.AzureSEVSNPVersion, err error) {
 	var versions configapi.AzureSEVSNPVersionList
 	versions, err = f.FetchAzureSEVSNPVersionList(ctx, versions)
 	if err != nil {
 		return res, fmt.Errorf("fetching versions list: %w", err)
 	}
+	fmt.Println("versions", versions)
```
derpsteb, line 85: debugging print?

internal/api/fetcher/configapi.go
```diff
@@ -35,17 +42,47 @@ func (f *ConfigAPIFetcher) FetchAzureSEVSNPVersionList(ctx context.Context, atte
...
-	// TODO(elchead): follow-up PR for AB#3045 to check signature (sigstore.VerifySignature)
+	urlString, err := attestation.URL()
+	if err != nil {
+		return attestation, err
+	}
+	versionBytes, err := fetchBytesFromRawURL(ctx, urlString, f.httpc)
+	if err != nil {
+		return attestation, fmt.Errorf("fetch version %s: %w", attestation.Version, err)
+	}
+
```
derpsteb, line 53: Right now you are fetching resource A, sig(A) and then verifying that A and sig(A) match. Then you are calling another function that returns B. You are assuming that A==B. Instead you should Marshal the result of `fetch`, and check that the signature matches.

internal/api/fetcher/configapi.go
```diff
@@ -35,17 +42,47 @@ func (f *ConfigAPIFetcher) FetchAzureSEVSNPVersionList(ctx context.Context, atte
 
 // FetchAzureSEVSNPVersion fetches the version information from the config API.
 func (f *ConfigAPIFetcher) FetchAzureSEVSNPVersion(ctx context.Context, attestation configapi.AzureSEVSNPVersionGet) (configapi.AzureSEVSNPVersionGet, error) {
```
derpsteb, line 44: Ideally, signatures would be somehow mapped within the configapi. Either as their own resources or as an extra field. But I am not sure how to do that yet. This is fine for now.

internal/api/configapi/repo.go
```diff
@@ -15,41 +15,62 @@ import (
...
+
+	err = a.signAndUpload(ctx, versionBytes, filePath)
 	if err != nil {
 		return err
 	}
 	return a.addVersionToList(ctx, variant, fname)
 }
 
+// signAndUpload signs the given content and uploads it to the given filePath with the .sig suffix.
+func (a AttestationVersionRepo) signAndUpload(ctx context.Context, content []byte, filePath string) error {
```
derpsteb, line 62:
```suggestion
// uploadSignature signs the given content and uploads the signature to the given filePath with the .sig suffix.
func (a AttestationVersionRepo) uploadSignature(ctx context.Context, content []byte, filePath string) error {
```

Right now the function and it's comment suggest that `content` is also uploaded by the function. However, it is not.

hack/configapi/main.go
```diff
@@ -82,10 +93,20 @@ func main() {
...
+	handleError(myCmd.MarkPersistentFlagRequired("private-key"))
 
-	myCmd.PersistentFlags().StringVar(&awsAccessKey, "key", "", "Access key to use for AWS tests. Required for AWS KMS and storage test.")
-	handleError(myCmd.MarkPersistentFlagRequired("key"))
 	handleError(myCmd.Execute())
 }
+
+func getBytesFromFilePath(path string) []byte {
+	file, err := os.Open(path)
+	handleError(err)
```
derpsteb, line 107: pls wrap error.

internal/api/configapi/repo.go
elchead: uploadSignature also suggests only upload when it really signs and uploads.
how about this:

// signAndUpload signs the given content and uploads the signature to the given filePath with the .sig suffix.

internal/api/fetcher/fetcher.go
```diff
@@ -46,6 +48,28 @@ type apiObject interface {
 	URL() (string, error)
 }
 
+// FetchFromURL fetches the content from the given URL and returns the content as a byte slice.
+func FetchFromURL(ctx context.Context, client HTTPClient, sourceURL *url.URL) ([]byte, error) {
```
derpsteb, line 52: Please keep the function as `getFromURL` within the local scope of the pkg that needs it. If needed, copy it.

`fetcher` should only be used by implementing the `apiObject` interface and supplying the new type to `fetch`.
`FetchFromURL` is only necessary since we are not handling measurements according to that scheme, yet. I am working towards changing that right now.

hack/configapi/main.go
```diff
@@ -42,20 +46,18 @@ func handleError(err error) {
 
 func main() {
 	myCmd := &cobra.Command{
-		Use:   "upload a set of versions specific to the azure-sev-snp attestation variant to the config api",
+		Use:   fmt.Sprintf("please set the %s and %s environment variables and all flags", envAwsKeyID, envAwsKey),
 		Short: "upload a set of versions specific to the azure-sev-snp attestation variant to the config api",
 		Run: func(cmd *cobra.Command, args []string) {
 			ctx := context.Background()
```
daniel-weisse, line 52: nit: Move into its own function, instead of inlining it here

hack/configapi/main.go
```diff
@@ -42,20 +46,18 @@ func handleError(err error) {
 
 func main() {
 	myCmd := &cobra.Command{
-		Use:   "upload a set of versions specific to the azure-sev-snp attestation variant to the config api",
+		Use:   fmt.Sprintf("please set the %s and %s environment variables and all flags", envAwsKeyID, envAwsKey),
 		Short: "upload a set of versions specific to the azure-sev-snp attestation variant to the config api",
 		Run: func(cmd *cobra.Command, args []string) {
```
daniel-weisse, line 51:
```suggestion
		RunE: func(cmd *cobra.Command, args []string) error {
```
Use `RunE` to offload error handling to caller of the command.
You can then remove all calls to `panic` inside the command`. (Also helps if we want to add unit tests at some point)

internal/staticupload/upload.go
```diff
@@ -34,3 +36,15 @@ func (c *Client) Upload(
 	}
 	return output, nil
 }
+
+// Put is a convenience method to save a DEK to AWS S3 Storage by key ID.
```
derpsteb, line 40: Please keep the convencience methods local to pkg that need them. As far as I can tell this is only `internal/api/repo` so far. 

Also: please don't import `internal/kms` here. `internal/staticupload` should stay independent from our kms backend.

internal/api/fetcher/configapi.go
```diff
@@ -35,17 +42,47 @@ func (f *ConfigAPIFetcher) FetchAzureSEVSNPVersionList(ctx context.Context, atte
...
+func fetchBytesFromRawURL(ctx context.Context, urlString string, client HTTPClient) ([]byte, error) {
+	url, err := url.Parse(urlString)
+	if err != nil {
+		return nil, fmt.Errorf("parse version url %s: %w", urlString, err)
+	}
+	bytes, err := FetchFromURL(ctx, client, url)
+	if err != nil {
+		return nil, err
+	}
+	return bytes, nil
```
daniel-weisse, line 75:
```suggestion
	return FetchFromURL(ctx, client, url)
```
If there is no additional error wrapping, you may just return the result of the function directly

internal/api/fetcher/fetcher.go
daniel-weisse, line 52: Use `nil` as return value if you are returning an error, instead of `[]byte{}`

internal/api/fetcher/fetcher.go
```diff
@@ -46,6 +48,28 @@ type apiObject interface {
...
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return []byte{}, fmt.Errorf("http status code: %d", resp.StatusCode)
+	}
+	content, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return []byte{}, err
+	}
+	return content, nil
```
daniel-weisse, line 70:
```suggestion
	return io.ReadAll(resp.Body)
```
Same as above: If there is no additional error wrapping, just return the output of the function directly.

internal/staticupload/staticupload.go
```diff
@@ -57,6 +62,14 @@ type Config struct {
 	CacheInvalidationWaitTimeout time.Duration
 }
 
+// Validate checks if all necessary values are set and sets default values otherwise.
+func (c *Config) Validate() error {
```
daniel-weisse, line 66: Is this method required to implement an interface?
If not you can remove the return value, as this will always return `nil`

internal/staticupload/get.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/aws/aws-sdk-go-v2/service/s3/types"
+	"github.com/edgelesssys/constellation/v2/internal/kms/storage"
```
derpsteb, line 17: please remove the kms import.

internal/staticupload/get.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// Get is a convenience method to return a DEK from from AWS S3 Storage by key ID.
+func (s *Client) Get(ctx context.Context, keyID string) ([]byte, error) {
```
derpsteb, line 21: same as with upload.

daniel-weisse (commented):
A general question about the upload tool:
Currently it reads the versions to upload as flags from the command line.
This will most likely result in some problems once we extend the tool to upload more than just Azure SEV-SNP versions.

Would it make sense to instead read in a json/yaml file and go from there?
At the same time we could update the [Azure sev-snp version reporter](https://github.com/edgelesssys/constellation/blob/main/hack/azure-snp-report-verify/verify.go) to output a json or yaml file of the generated values.
This would make automating the tools a bit simpler.

internal/api/fetcher/configapi_test.go
```diff
@@ -16,26 +16,60 @@ import (
...
 )
 
+var testCfg = configapi.AzureSEVSNPVersion{
+	Microcode:  93,
+	TEE:        0,
+	SNP:        6,
+	Bootloader: 2,
+}
+
 func TestGetVersion(t *testing.T) {
```
derpsteb, line 28: Our usual test pattern is `TestFoo` where `Foo` is the function that is being tested.
In this case that would be `FetchLatestAzureSEVSNPVersion`.

internal/sigstore/sign.go
```diff
@@ -0,0 +1,26 @@
...
+	}
+	sig, err := sv.SignMessage(bytes.NewReader(content))
+    if err != nil {
+        return nil, fmt.Errorf("signing message: %w", err)
+    }
+	return []byte(base64.StdEncoding.EncodeToString(sig)), nil
+}
```
malt3, line 26: From a (high level) look at the cosign go package, it looks to me like this function signs the payload but doesn't upload the result to the transparency log.

I see two possible changes here:

- decide signatures created by this functions don't end up in the transparency log
  - i would make this very clear by naming the function accordingly and even adding a comment
  - I would use go stdlib code to create the signature instead. We get a large amount of new dependencies through the cosign package
-  add the transparency log upload functionality (either as a separate function or in here)

hack/configapi/main.go
```diff
@@ -42,20 +46,18 @@ func handleError(err error) {
 
 func main() {
 	myCmd := &cobra.Command{
```
malt3, line 48:
```suggestion
	cmd := &cobra.Command{
```

hack/configapi/main.go
```diff
@@ -70,6 +72,15 @@ func main() {
...
 	}
+	myCmd.PreRunE = func(cmd *cobra.Command, args []string) error {
+		if _, present := os.LookupEnv(envAwsKey); !present {
+			return fmt.Errorf("%s not set", envAwsKey)
+		}
+		if _, present := os.LookupEnv(envAwsKeyID); !present {
+			return fmt.Errorf("%s not set", envAwsKeyID)
+		}
+		return nil
+	}
```
malt3, line 83: I am very surprised that this is needed.  We have multiple go tools that instantiate various aws sdk clients without explicitly needing environment variables.
The env vars can always be used as one of several authentication mechanisms.
I am logged in with the aws cli and never need to set those, the CI uses openid for login so I don't hink a check for those environment variables can be enforced.

internal/staticupload/get.go
```diff
@@ -0,0 +1,40 @@
...
+	output, err := s.s3Client.GetObject(ctx, getObjectInput)
+	if err != nil {
+		var nsk *types.NoSuchKey
+		if errors.As(err, &nsk) {
+			return nil, storage.ErrDEKUnset
+		}
+		return nil, fmt.Errorf("downloading DEK from storage: %w", err)
+	}
+	return io.ReadAll(output.Body)
+}
```
malt3, line 35: This convenience method is very specific to one part of our code base. I would not define it here and instead make it a function in the package where it is used.

internal/staticupload/upload.go
elchead: our s3 wrapper has the same method and we want to replace all s3 instances with the new client so it will be used. Malte agreed

internal/staticupload/staticupload.go
```diff
@@ -57,6 +62,14 @@ type Config struct {
...
+	if c.DistributionID == "" {
+		c.DistributionID = DefaultDistributionID
+	}
+	return nil
+}
```
malt3, line 71:
```suggestion
// SetDefaults sets default values.
func (c *Config) SetDefaults() error {
	if c.DistributionID == "" {
		c.DistributionID = DefaultDistributionID
	}
	return nil
}
```

nit: i expect a `Validate` method to never change the underlying data.

internal/staticupload/get.go
elchead: ok

internal/staticupload/staticupload.go
```diff
@@ -24,8 +24,12 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/cloudfront"
 	cftypes "github.com/aws/aws-sdk-go-v2/service/cloudfront/types"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/google/uuid"
 )
 
+// DefaultDistributionID is the default CloudFront distribution ID to use.
+const DefaultDistributionID = "E1H77EZTHC3NE4"
```
malt3, line 31:
```suggestion
```

nit: I would move this to internal/constants/constants.go. Now you have many packages importing `staticupload` just to access this constant.

internal/sigstore/sign.go
elchead: Yes, it's not uploaded to the log. Is it okay or should I add it? But if we upload to the log, we likely need cosign as a dependency @daniel-weisse @derpsteb 
For now, I copied the function to remove the dep on cosign.

elchead: should I open a ticket in Azure and leave as is in this PR for now? Or how to proceed?

malt3: A ticket sounds like a good idea so we don't block this PR.

internal/sigstore/sign.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/theupdateframework/go-tuf/encrypted"
+)
+
+const (
+	cosignPrivateKeyPemType   = "ENCRYPTED COSIGN PRIVATE KEY"
+	sigstorePrivateKeyPemType = "ENCRYPTED SIGSTORE PRIVATE KEY"
+)
+
+// SignContent signs the content with the cosign encrypted private key and corresponding cosign password.
+func SignContent(password, encryptedPrivateKey, content []byte) ([]byte, error) {
```
daniel-weisse, line 30: Is the cosign private key encrypted in our CI?
If not (or if we can save it unencrypted, which shouldnt make a difference because saving passphrase + encrypted data is basically the same as just saving the unencrypted data) we can remove the extra steps of parsing passphrase the passphrase and decrypting the key.
This should also remove some dependencies.

internal/staticupload/get.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+)
+
+// GetObject returns an object from from AWS S3 Storage.
+func (s *Client) GetObject(ctx context.Context, params *s3.GetObjectInput, optFns ...func(*s3.Options)) (*s3.GetObjectOutput, error) {
+	return s.s3Client.GetObject(ctx, params, optFns...)
+}
```
daniel-weisse, line 18:
```suggestion
// GetObject returns an object from from AWS S3 Storage.
func (s *Client) GetObject(ctx context.Context, params *s3.GetObjectInput) (*s3.GetObjectOutput, error) {
	return s.s3Client.GetObject(ctx, params)
}
```
nit: If we dont expect to pass any optional functions to this functions, we can get away with removing them to simplify interfaces.

internal/sigstore/sign.go
malt3: > Is the cosign private key encrypted in our CI?

I think so. COSIGN_PASSWORD is a secret referenced in gh actions

internal/staticupload/get.go
malt3: I think we should just pass those along.. This ensures we have a drop-in replacement that can satisfy the same interfaces

hack/configapi/cmd/root.go
```diff
@@ -0,0 +1,115 @@
...
+	uploadCmd.PersistentFlags().StringVar(&cosignPwd, "cosign-pwd", "", "Cosign password used to decrpyt the private key.")
+
+	uploadCmd.PersistentFlags().StringVar(&privateKeyPath, "private-key", "", "File path of private key used to sign the payload.")
+	return uploadCmd
+}
+
+func runCmd(cmd *cobra.Command, _ []string) error {
+	if err := enforceRequiredFlags(cmd, "version-file", "cosign-pwd", "private-key"); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
elchead: how to handle the error then? should we panic? i wanted to avoid that. what is the benefit?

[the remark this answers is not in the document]
daniel-weisse: `MarkFlagRequired` only returns an error if the flag one is asking to mark as required [does not actually exist](https://github.com/spf13/pflag/blob/2e9d26c8c37aae03e3f9d4e90b7116f5accb7cab/flag.go#L495-L498).
Since this should never be the case if we do so directly after we define the flag, we commonly wrap the call in a function that panics on error.
See the CLI [`must` function for reference](https://github.com/edgelesssys/constellation/blob/c5e016a8e28fef42a5cc1fa758676caed773902f/cli/cmd/root.go#L99).

We want to mark the flag as required in the `NewRootCmd` function so the autocomplete directives, and error handling if not set, work before we actually execute the command.
This wont work (or shouldnt work) if we only mark them as required if the command is already executed.

internal/sigstore/sign.go
elchead: yes, both in the CI and keepass they are encrypted. There is no external dependency anymore.

hack/configapi/cmd/root.go
```diff
@@ -0,0 +1,109 @@
...
+	cfg := staticupload.Config{
+		Bucket: awsBucket,
+		Region: awsRegion,
+	}
+	privateKey, err := getBytesFromFilePath(privateKeyPath)
+	if err != nil {
+		return fmt.Errorf("reading private key: %w", err)
+	}
+
+	versionBytes, err := getBytesFromFilePath(versionFilePath)
```
daniel-weisse, line 72:
```suggestion
	versionBytes, err := os.ReadFile(versionFilePath)
```
And then you can remove the `geBytesFromFilePath` function on line 103

internal/sigstore/sign.go
daniel-weisse: We now have a dependency on `github.com/theupdateframework/go-tuf/encrypted`, but since this was already an indirect dependency previously, there shouldnt be a difference really.

If the cosign key wasnt additionally encrypted, we could skip the decrypting part in this function.
But doesnt make a big difference imo

daniel-weisse (approved):
Looks good from my side

internal/api/configapi/repo.go
derpsteb: maybe use `Sig` instead of `Signature`. your choice.

internal/sigstore/sign.go
derpsteb: please also add the ticket as a TODO to the function.

elchead:
@derpsteb could you give it another look especially the last 2 commits?

internal/config/config.go
```diff
@@ -386,18 +387,23 @@ func fromFile(fileHandler file.Handler, name string) (*Config, error) {
...
 // 4. Validate config. If `--force` is set the version validation will be disabled and any version combination is allowed.
 func New(fileHandler file.Handler, name string, force bool) (*Config, error) {
-	return NewWithClient(fileHandler, name, fetcher.NewHTTPClient(), force)
+	return NewWithFetcher(fileHandler, name, fetcher.NewConfigAPIFetcher(), force)
 }
 
-// NewWithClient is New with a custom HTTP client.
-func NewWithClient(fileHandler file.Handler, name string, client fetcher.HTTPClient, force bool) (*Config, error) {
+// NewWithFetcher is New with a custom fetcher.
+func NewWithFetcher(fileHandler file.Handler, name string, fetcher fetcher.ConfigAPIFetcher, force bool) (*Config, error) {
```
daniel-weisse, line 394: Since using `config.New()` over `config.NewWithFetcher()` is incorrect in most cases, any objections against removing the `config.NewWithFetcher()` and just requiring `config.New()` to take a ConfigAPIFetcher?

My reasoning:
* `config.New()` can only be called if we are sure that the function making the call will never be accessed in a unit test.
* In a perfect world, all code is tested somehow, so those instances should use `config.NewWithFetcher()` instead.
* A developer is likely to chose `config.New()` over `config.NewWithFetcher()` as this looks to be the default initializer for the config. However, this is most likely the wrong function for their use case, as has explained above

elchead: I agree, I was thinking the same but lost it again amidst all the refactoring 🙈 Applying the changes now

elchead: I agree. I was thinking the same but lost it again amidst all the refactoring 🙈 Applying the changes now

malt3 (approved):

derpsteb (approved):
nothing to add :P

PR #1839, opened 2023-05-31
ci: use aws s3 client that invalidates cloudfront cache

malt3:
### Proposed change(s)

- ci: use aws s3 client that invalidates cloudfront cache

### Additional info
- [e2e test](https://github.com/edgelesssys/constellation/actions/runs/5134340654)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/image/upload/internal/cmd/aws.go
+++ b/image/upload/internal/cmd/aws.go
@@ -46,10 +46,15 @@ func runAWS(cmd *cobra.Command, _ []string) error {
 	log := logger.New(logger.PlainLog, flags.logLevel)
 	log.Debugf("Parsed flags: %+v", flags)
 
-	archiveC, err := archive.New(cmd.Context(), flags.region, flags.bucket, log)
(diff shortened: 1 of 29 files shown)


internal/staticupload/staticupload.go
elchead, line 1: How about returning close() in New? This feels more intuitive that you must use this function somehow (instead of just knowing that you must call it). You could ignore it for `CacheInvalidateEager` but `CacheInvalidateBatch` seems the better strategy for most cases anyways.

malt3: That's a good idea!

image/upload/internal/cmd/aws.go
elchead, line 1: Maybe not for this PR but it seems that runProvider should be a template method which only injects the provder specific upload

malt3: I agree we should refactor that in the future

elchead:
lgtm besides the tiny nits

derpsteb (approved):

PR #1850, opened 2023-06-01
ci: remove setup-go action / disable cache where applicable

malt3:
Runners sometimes fail because they run out of disk space. One reason this happens is a change in the setup-go action@v4:

> The V4 edition of the action offers: Enabled caching by default

To combat this, we now disable the cache if it was not enabled explicitly before. Additionally, we remove setup-go where it is no longer needed.

### Proposed change(s)

- ci: remove setup-go action / disable cache where applicable

### Additional info
- [e2e test](https://github.com/edgelesssys/constellation/actions/runs/5142586524)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/azure-snp-reporter.yml
+++ b/.github/workflows/azure-snp-reporter.yml
@@ -16,11 +16,6 @@ jobs:
         with:
           ref: ${{ !github.event.pull_request.head.repo.fork && github.head_ref || '' }}
 
-      - name: Set up Go
-        uses: actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9 # v4.0.0
-        with:
-          go-version: 1.20.4
-
       - name: Build and upload azure SNP reporter container image
         id: build-and-upload
         uses: ./.github/actions/build_micro_service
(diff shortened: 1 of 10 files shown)


.github/workflows/azure-snp-reporter.yml
```diff
@@ -20,6 +20,7 @@ jobs:
         uses: actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9 # v4.0.0
         with:
           go-version: 1.20.4
+          cache: false
```
daniel-weisse, line 23: Does this job in the workflow actually require Go?
There might be a dependency on Go very deep down, but just from going over it quickly it doesnt seem like we are executing any Go commands

.github/workflows/codeql.yml
```diff
@@ -41,6 +41,7 @@ jobs:
         uses: actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9 # v4.0.0
         with:
           go-version: "1.20.4"
+          cache: false
```
daniel-weisse, line 44: Same here

daniel-weisse (approved):

.github/workflows/codeql.yml
malt3: codeql actually needs a go toolchain

.github/workflows/build-versionsapi-ci-image.yml
```diff
@@ -27,6 +27,7 @@ jobs:
         uses: actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9 # v4.0.0
         with:
           go-version: "1.20.4"
+          cache: false
```
daniel-weisse, line 31: Sorry commented on the wrong file, meant this one

PR #1853, opened 2023-06-01
Correctly deploy the AWS CCM

3u13r:
### Proposed change(s)

- Retrieve instance tags though the ec2 API and not the IMDS API
- Correctly tag the instances according to https://cloud-provider-aws.sigs.k8s.io/getting_started/ Step 5
- let coredns run before the CCM is up, since it requires DNS to lookup tags from the ec2 api
- unrelated: make debugd less noisy by respecting the interval in all cases

image: `ref/feat-aws-correct-ccm-deployment/stream/debug/v2.8.0-pre.0.20230601144711-dacd80e4372b`
new image: `ref/feat-aws-correct-ccm-deployment/stream/debug/v2.8.0-pre.0.20230602114037-b06ddeb4026c` see: https://github.com/edgelesssys/constellation/actions/runs/5155150759

Migration:
I'm not really sure how to do migration for this, since it involves basically the steps from https://cloud-provider-aws.sigs.k8s.io/getting_started/ which are intrusive since they change the e.g. the kubelet's config. I'm open to any suggestion how to handle such cases.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/cloud/aws/aws.go
+++ b/internal/cloud/aws/aws.go
@@ -52,7 +52,6 @@ type ec2API interface {
 
 type imdsAPI interface {
 	GetInstanceIdentityDocument(context.Context, *imds.GetInstanceIdentityDocumentInput, ...func(*imds.Options)) (*imds.GetInstanceIdentityDocumentOutput, error)
-	GetMetadata(context.Context, *imds.GetMetadataInput, ...func(*imds.Options)) (*imds.GetMetadataOutput, error)
 }
 
 // Cloud provides AWS metadata and API access.
--- a/internal/cloud/aws/aws.go
+++ b/internal/cloud/aws/aws.go
@@ -79,12 +79,7 @@ func New(ctx context.Context) (*Cloud, error) {
 
 // List retrieves all instances belonging to the current Constellation.
 func (c *Cloud) List(ctx context.Context) ([]metadata.InstanceMetadata, error) {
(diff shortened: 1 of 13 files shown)


internal/cloud/aws/aws.go
```diff
@@ -81,7 +79,12 @@ func New(ctx context.Context) (*Cloud, error) {
 
 // List retrieves all instances belonging to the current Constellation.
 func (c *Cloud) List(ctx context.Context) ([]metadata.InstanceMetadata, error) {
-	uid, err := readInstanceTag(ctx, c.imds, cloud.TagUID)
+	identity, err := c.imds.GetInstanceIdentityDocument(ctx, &imds.GetInstanceIdentityDocumentInput{})
+	if err != nil {
+		return nil, fmt.Errorf("retrieving instance identity: %w", err)
+	}
+
+	uid, err := c.readInstanceTag(ctx, identity.InstanceID, cloud.TagUID)
```
daniel-weisse, line 87: Since all calls to `c.readInstanceTag` require a call to `c.imds.GetInstanceIdentityDocument` first, why not move the identity document fetching into `c.readInstanceTag`?

3u13r: Good point

daniel-weisse (commented):
@msanft can you check if this fixes the Issues for the AWS driver?

malt3 (approved):
I propose that we merge this soon and give it some time to mature until our next minor release.

PR #1854, opened 2023-06-01
config: automatically upload new Azure SNP versions to API + sign version with public release key

elchead:
### Proposed change(s)

- sign version with release key and remove version from fetcher interface
- upload tool compares the latest API version with the input version (from azure-snp-reporter) and uploads a new version if the input is newer
- assign me as codeowner of upload tool
- extend azure-reporter GH action to upload updated version values to the Attestation API

### Checklist

- [x] Link to Milestone

### DIFF
--- a/hack/azure-snp-report-verify/verify.go
+++ b/hack/azure-snp-report-verify/verify.go
@@ -46,10 +46,10 @@ func (i *IsolationTEE) PrintSVNs() {
 
 func main() {
 	configAPIExportPath := flag.String("export-path", "azure-sev-snp-version.json", "Path to the exported config API file.")
-	maaJWT := flag.String("maa-jwt", "", "MAA JWT to verify")
+	maaJWT := flag.String("report", "", "MAA JWT to verify")
 	flag.Parse()
 	if *maaJWT == "" {
-		fmt.Println("Must provide --maa-jwt")
+		fmt.Println("Must provide --report")
 		return
 	}
 	report, err := getTEEReport(*maaJWT)
--- a/hack/azure-snp-report-verify/verify.go
+++ b/hack/azure-snp-report-verify/verify.go
@@ -46,12 +46,13 @@ func (i *IsolationTEE) PrintSVNs() {
 
 func main() {
 	configAPIExportPath := flag.String("export-path", "azure-sev-snp-version.json", "Path to the exported config API file.")
+	maaJWT := flag.String("maa-jwt", "", "MAA JWT to verify")
 	flag.Parse()
-	if len(os.Args) != 2 {
-		fmt.Println("Usage:", os.Args[0], "<maa-jwt>")
+	if *maaJWT == "" {
+		fmt.Println("Must provide --maa-jwt")
 		return
 	}
-	report, err := getTEEReport(os.Args[1])
+	report, err := getTEEReport(*maaJWT)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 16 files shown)


elchead:
@derpsteb in the report gh action, why do we build the container for the binary and then use `go run` in another step? I guess `go run` is good enough for both?

hack/configapi/cmd/root_test.go
```diff
@@ -0,0 +1,75 @@
...
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+package cmd
+
+import (
+	"testing"
+
+	"github.com/edgelesssys/constellation/v2/internal/api/configapi"
+	"github.com/tj/assert"
```
daniel-weisse, line 13:
```suggestion
	"github.com/stretchr/testify/assert"
```
:wink:

elchead: oh danke, was war das denn fuer ein autoimport 😲

derpsteb:
The container that is [built](https://github.com/edgelesssys/constellation/blob/main/.github/workflows/azure-snp-reporter.yml#L19) in the first job is not the tool that is run during verify. It is [used](https://github.com/edgelesssys/constellation/blob/main/.github/workflows/azure-snp-reporter.yml#L40) in the second job to get the versions from a SNP machine. The container is executed on a [Azure CVM](https://github.com/edgelesssys/constellation/blob/main/.github/workflows/azure-snp-reporter.yml#LL30) that is spawned for each run. The tool inside the container requires some extra C++ dependencies. The verification in the end is a self-contained go tool so we don't need a container, as you already pointed out.

hack/azure-snp-report-verify/verify.go
```diff
@@ -46,12 +46,13 @@ func (i *IsolationTEE) PrintSVNs() {
 
 func main() {
 	configAPIExportPath := flag.String("export-path", "azure-sev-snp-version.json", "Path to the exported config API file.")
+	maaJWT := flag.String("jwt", "", "MAA JWT key to verify version")
```
derpsteb, line 49: nit:
```suggestion
maaJWT := flag.String("jwt", "", "MAA JWT to verify version")
```

hack/configapi/cmd/root.go
```diff
@@ -61,33 +64,73 @@ func runCmd(cmd *cobra.Command, _ []string) error {
...
+		cmd.Printf("Successfully uploaded new Azure SEV-SNP version: %+v\n", inputVersion)
+	} else {
+		cmd.Printf("Input version: %+v is not newer than latest API version: %+v\n", inputVersion, latestAPIVersion)
 	}
-	cmd.Printf("Successfully uploaded new Azure SEV-SNP version: %+v\n", versions)
 	return nil
 }
 
+// IsInputNewerThanLatestAPI compares all version fields with the latest API version and returns true if any input field is newer.
+func IsInputNewerThanLatestAPI(input, latest attestationconfig.AzureSEVSNPVersion) (bool, error) {
```
derpsteb, line 107: Not entirely sure what is happening here :D, but I offer you an alternative that seems simpler to me and needs no reflections:
- Implement a method on `attestationconfig.AzureSEVSNPVersion` that is analogous to [semver.Compare](https://pkg.go.dev/golang.org/x/mod/semver#Compare). v is the object, w the argument. v is smaller than w if at least one version is smaller, v is larger than w if all versions of w are equal to v and at least one version of w is larger than w. 
- Call `input.Compare(latest)` here, sth like: `return input.Compare(latest) >= 0`.

maybe I am missing something..

derpsteb (approved):

PR #1859, opened 2023-06-02
ci: Windows cli tests

msanft:
### Proposed change(s)

- Add a very basic test for the Windows CLI by creating and initializing a test cluster without any deployments.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/e2e-windows.yml
+++ b/.github/workflows/e2e-windows.yml
@@ -73,12 +73,11 @@ jobs:
         shell: pwsh
         run: |
           .\constellation.exe init --debug
-          echo "$PWD\constellation-admin.conf"
-          setx KUBECONFIG "$PWD\constellation-admin.conf"
 
       - name: Liveness probe
         shell: pwsh
         run: |
+          setx KUBECONFIG "$PWD\constellation-admin.conf"
           $retryIntervalSeconds = 30
           $maxRetries = 50
 
--- a/.github/workflows/e2e-windows.yml
+++ b/.github/workflows/e2e-windows.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Create IAM configuration
         shell: pwsh
         run: |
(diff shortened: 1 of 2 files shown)


msanft:
[Successful run](https://github.com/edgelesssys/constellation/actions/runs/5173915800/jobs/9319928250)

.github/workflows/e2e-windows.yml
```diff
@@ -0,0 +1,93 @@
...
+          targetOS: "windows"
+          targetArch: "amd64"
+          enterpriseCLI: true
+          outputPath: "build/constellation.exe"
+
+      - name: Upload CLI artifact
+        uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
+        with:
+          path: "build/constellation.exe"
+          name: "constellation.exe"
```
malt3, line 30: This should upload container images like the other e2e tests.

.github/workflows/e2e-windows.yml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Initialize cluster
+        shell: pwsh
+        run: |
+          .\constellation.exe init --debug
+
+      - name: Terminate cluster
+        if: always()
+        shell: pwsh
+        run: |
+          .\constellation.exe terminate --debug -y
```
malt3, line 81: We should have some limited checks if the cluster is ready (number of ready nodes, deployments are ready). Otherwise, issues like unavailable container images never surface.

.github/workflows/e2e-windows.yml
msanft: Then I should also edit the config file afterwards to use the appropriate images I assume? Or what's the purpose of uploading them?

malt3: Never mind. You are correct in the sense that when using nightlies / releases, this is not necessary.

malt3 (approved):

PR #1863, opened 2023-06-02
config: Azure SNP tool can delete specific version from attestation API

elchead:
### Proposed change(s)

- add unit tests for attestationconfig.Client
- add delete command to tool
- refactor attestationconfig.Client to use the general apiclient

TODO: also make rest of attestation client testable?

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/hack/configapi/cmd/delete.go
+++ b/hack/configapi/cmd/delete.go
@@ -36,14 +36,10 @@ type deleteClient interface {
 }
 
 func (d deleteCmd) delete(cmd *cobra.Command) error {
-	defer func() {
-		if d.closeFn != nil {
-			if err := d.closeFn(cmd.Context()); err != nil {
-				fmt.Printf("close client: %s\n", err.Error())
-			}
-		}
-	}()
-	version := cmd.Flag("version").Value.String()
+	version, err := cmd.Flags().GetString("version")
+	if err != nil {
+		return err
+	}
 	return d.attestationClient.DeleteAzureSEVSNPVersion(cmd.Context(), version)
 }
 
@@ -56,9 +52,13 @@ func runDelete(cmd *cobra.Command, _ []string) error {
 	if err != nil {
 		return fmt.Errorf("create attestation client: %w", err)
 	}
+	defer func() {
+		if err := closefn(); err != nil {
+			cmd.Printf("close client: %s\n", err.Error())
+		}
+	}()
 	deleteCmd := deleteCmd{
 		attestationClient: repo,
-		closeFn:           closefn,
 	}
 	return deleteCmd.delete(cmd)
 }
--- a/hack/configapi/cmd/delete.go
+++ b/hack/configapi/cmd/delete.go
@@ -1,0 +1,61 @@
+/*
+Copyright (c) Edgeless Systems GmbH
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+package cmd
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/edgelesssys/constellation/v2/internal/api/attestationconfig/client"
+	"github.com/edgelesssys/constellation/v2/internal/staticupload"
+	"github.com/spf13/cobra"
+)
+
+// newDeleteCmd creates the delete command.
+func newDeleteCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delete",
+		Short: "delete a specific version from the config api",
+		RunE:  runDelete,
+	}
+	cmd.PersistentFlags().StringP("version", "v", "", "Name of the version to delete (without .json suffix)")
+	must(enforceRequiredFlags(cmd, "version"))
+	return cmd
+}
+
+type deleteCmd struct {
+	attestationclient client.Client
+	closefn           client.CloseFunc
+}
+
+func (d deleteCmd) delete(ctx context.Context, cmd *cobra.Command) error {
+	defer func() {
+		if d.closefn != nil {
+			if err := d.closefn(ctx); err != nil {
+				fmt.Printf("close client: %s\n", err.Error())
+			}
+		}
+	}()
+	version := cmd.Flag("version").Value.String()
+	return d.attestationclient.DeleteAzureSEVSNVersion(ctx, version)
+}
+
+func runDelete(cmd *cobra.Command, _ []string) error {
+	ctx := context.Background()
+	cfg := staticupload.Config{
+		Bucket: awsBucket,
+		Region: awsRegion,
+	}
+	repo, closefn, err := client.New(ctx, cfg, []byte(cosignPwd), []byte(privateKey))
+	if err != nil {
+		return fmt.Errorf("create attestation client: %w", err)
+	}
+	deleteCmd := deleteCmd{
+		attestationclient: repo,
+		closefn:           closefn,
+	}
+	return deleteCmd.delete(ctx, cmd)
+}
--- a/hack/configapi/cmd/delete.go
+++ b/hack/configapi/cmd/delete.go
@@ -49,7 +49,7 @@ func runDelete(cmd *cobra.Command, _ []string) error {
 		Bucket: awsBucket,
 		Region: awsRegion,
 	}
-	repo, closefn, err := client.New(ctx, cfg, []byte(cosignPwd), []byte(privateKey))
+	repo, closefn, err := client.New(ctx, cfg, []byte(cosignPwd), []byte(privateKey), false, log())
 	if err != nil {
 		return fmt.Errorf("create attestation client: %w", err)
 	}
--- a/hack/configapi/cmd/delete.go
+++ b/hack/configapi/cmd/delete.go
@@ -21,41 +21,44 @@ func newDeleteCmd() *cobra.Command {
 		Short: "delete a specific version from the config api",
 		RunE:  runDelete,
 	}
-	cmd.PersistentFlags().StringP("version", "v", "", "Name of the version to delete (without .json suffix)")
+	cmd.Flags().StringP("version", "v", "", "Name of the version to delete (without .json suffix)")
 	must(enforceRequiredFlags(cmd, "version"))
 	return cmd
 }
 
 type deleteCmd struct {
-	attestationclient client.Client
-	closefn           client.CloseFunc
+	attestationClient deleteClient
+	closeFn           client.CloseFunc
+}
+
+type deleteClient interface {
+	DeleteAzureSEVSNPVersion(ctx context.Context, versionStr string) error
 }
 
 func (d deleteCmd) delete(cmd *cobra.Command) error {
 	defer func() {
-		if d.closefn != nil {
-			if err := d.closefn(ctx); err != nil {
-				cmd.Printf("close client: %s\n", err.Error())
+		if d.closeFn != nil {
+			if err := d.closeFn(cmd.Context()); err != nil {
+				fmt.Printf("close client: %s\n", err.Error())
 			}
 		}
 	}()
 	version := cmd.Flag("version").Value.String()
-	return d.attestationclient.DeleteAzureSEVSNVersion(cmd.Context(), version)
+	return d.attestationClient.DeleteAzureSEVSNPVersion(cmd.Context(), version)
 }
 
 func runDelete(cmd *cobra.Command, _ []string) error {
-	ctx := context.Background()
 	cfg := staticupload.Config{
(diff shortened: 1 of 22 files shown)


hack/configapi/cmd/delete.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+// newDeleteCmd creates the delete command.
+func newDeleteCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delete",
+		Short: "delete a specific version from the config api",
+		RunE:  runDelete,
+	}
+	cmd.PersistentFlags().StringP("version", "v", "", "Name of the version to delete (without .json suffix)")
```
daniel-weisse, line 24:
```suggestion
	cmd.Flags().StringP("version", "v", "", "Name of the version to delete (without .json suffix)")
```
nit: If the flag is just used for this command, and not for any sub commands of `delete`, just use the regular `Flags`

hack/configapi/cmd/delete.go
```diff
@@ -0,0 +1,61 @@
...
+	defer func() {
+		if d.closefn != nil {
+			if err := d.closefn(ctx); err != nil {
+				fmt.Printf("close client: %s\n", err.Error())
+			}
+		}
+	}()
+	version := cmd.Flag("version").Value.String()
+	return d.attestationclient.DeleteAzureSEVSNVersion(ctx, version)
+}
```
daniel-weisse, line 44:
```suggestion
func (d deleteCmd) delete(cmd *cobra.Command) error {
	defer func() {
		if d.closefn != nil {
			if err := d.closefn(ctx); err != nil {
				cmd.Printf("close client: %s\n", err.Error())
			}
		}
	}()
	version := cmd.Flag("version").Value.String()
	return d.attestationclient.DeleteAzureSEVSNVersion(cmd.Context(), version)
}
```
* Don't need to pass an explicit context if the function already uses `*cobra.Command`, as the command already has a context embedded
* Use `cmd.PrintX` instead of `fmt.PrintX`

hack/configapi/cmd/delete.go
```diff
@@ -0,0 +1,61 @@
...
+
+func runDelete(cmd *cobra.Command, _ []string) error {
+	ctx := context.Background()
+	cfg := staticupload.Config{
+		Bucket: awsBucket,
+		Region: awsRegion,
+	}
+	repo, closefn, err := client.New(ctx, cfg, []byte(cosignPwd), []byte(privateKey))
```
daniel-weisse, line 52:
```suggestion
	cfg := staticupload.Config{
		Bucket: awsBucket,
		Region: awsRegion,
	}
	repo, closefn, err := client.New(cmd.Context(), cfg, []byte(cosignPwd), []byte(privateKey))
```
Same here

hack/configapi/cmd/delete.go
```diff
@@ -0,0 +1,61 @@
...
+		RunE:  runDelete,
+	}
+	cmd.PersistentFlags().StringP("version", "v", "", "Name of the version to delete (without .json suffix)")
+	must(enforceRequiredFlags(cmd, "version"))
+	return cmd
+}
+
+type deleteCmd struct {
+	attestationclient client.Client
+	closefn           client.CloseFunc
```
daniel-weisse, line 31:
```suggestion
	attestationClient client.Client
	closeFn           client.CloseFunc
```
nit: naming

hack/configapi/cmd/delete.go
```diff
@@ -0,0 +1,61 @@
...
+		Short: "delete a specific version from the config api",
+		RunE:  runDelete,
+	}
+	cmd.PersistentFlags().StringP("version", "v", "", "Name of the version to delete (without .json suffix)")
+	must(enforceRequiredFlags(cmd, "version"))
+	return cmd
+}
+
+type deleteCmd struct {
+	attestationclient client.Client
```
daniel-weisse, line 30: Instead of importing the `client.Client` interface here, define a new interface in this file that only has the relevant functions for `delete`.

Example:
```Go
type deleteClient interface {
    DeleteAzureSEVSNVersion(ctx context.Context, versionStr string) error
}

That way the implementation of delete is a lot more flexible, since we only depend on the client implementation to implement the function we actually use in the code, and writing unit test is a bit easier, since we can skip out on writing stub functions that are never called.

File: hack/configapi/cmd/delete_test.go [hunk 6]
```diff
@@ -0,0 +1,53 @@
...
+	return nil
+}
+
+func (f fakeAttestationClient) DeleteList(_ context.Context, _ variant.Variant) error {
+	return nil
+}
+
+func (f fakeAttestationClient) List(_ context.Context, _ variant.Variant) ([]string, error) {
+	return []string{}, nil
+}
```
daniel-weisse, line 53:
```suggestion
```

Can be removed if  the interface used for `attestationclient` is more restricted, see my previous comment

internal/api/attestationconfig/client/client.go
```diff
@@ -19,51 +19,60 @@ import (
...
 
-// Client manages (modifies) the version information for the attestation variants.
-type Client struct {
+// Client is the client for the attestationconfig API.
+type Client interface {
+	UploadAzureSEVSNP(ctx context.Context, versions attestationconfig.AzureSEVSNPVersion, date time.Time) error
+	List(ctx context.Context, attestation variant.Variant) ([]string, error)
+	DeleteList(ctx context.Context, attestation variant.Variant) error
+	DeleteAzureSEVSNVersion(ctx context.Context, versionStr string) error
+}
```
daniel-weisse, line 36: Unless there are a lot of places using this interface, prefer defining the interface at the place it is used, instead of the place where the implementation is.
See my previous comment for the `delete` command.

internal/api/attestationconfig/client/client.go
```diff
@@ -121,7 +130,56 @@ func (a Client) DeleteList(ctx context.Context, attestation variant.Variant) err
...
+	versionStr = versionStr + ".json"
+	variant := variant.AzureSEVSNP{}
+	// delete version file
+	filePath := fmt.Sprintf("%s/%s/%s", constants.CDNAttestationConfigPrefixV1, variant.String(), versionStr)
+	ops = append(ops, deleteCmd{
+		path: filePath,
+	})
+
+	// delete signature file
+	sigPath := fmt.Sprintf("%s/%s/%s", constants.CDNAttestationConfigPrefixV1, variant.String(), versionStr+".sig")
```
daniel-weisse, line 143:
```suggestion
	filePath := path..Join(constants.CDNAttestationConfigPrefixV1, variant.String(), versionStr)
	ops = append(ops, deleteCmd{
		path: filePath,
	})

	// delete signature file
	sigPath :=path..Join(constants.CDNAttestationConfigPrefixV1, variant.String(), versionStr+".sig")
```
I don't have a strong preference of using `fmt.Sprintf` over `path.Join` (or the other way around), but we should try to keep the functions used to generate the url paths consistent, e.g. [line 159](https://github.com/edgelesssys/constellation/blob/ed8fb1c89cd9f13db110e009e1fbde3d88409378/internal/api/attestationconfig/client/client.go#L159) uses `path.Join` instead.

internal/api/attestationconfig/client/client.go
```diff
@@ -166,14 +224,58 @@ func put(ctx context.Context, client s3Client, bucket, path string, data []byte)
...
+	) (*s3.DeleteObjectOutput, error)
 }
 
 // CloseFunc is a function that closes the client.
 type CloseFunc func(ctx context.Context) error
+
+func removeVersion(versions attestationconfig.AzureSEVSNPVersionList, versionStr string) (removedVersions attestationconfig.AzureSEVSNPVersionList, err error) {
+	for i, v := range versions {
+		if v == versionStr {
+			removedVersions = append(versions[:i], versions[i+1:]...)
```
daniel-weisse, line 255: Need to make sure `versions[i+1]` actually exists

internal/api/attestationconfig/client/client_test.go
```diff
@@ -1,82 +1,51 @@
...
-	flag.Parse()
-	if *cosignPwd == "" || *privateKeyPath == "" {
-		flag.Usage()
-		fmt.Println("Required flags not set: --cosign-pwd, --private-key. Skipping tests.")
-		os.Exit(1)
-	}
-	if _, present := os.LookupEnv(envAwsKey); !present {
-		fmt.Printf("%s not set. Skipping tests.\n", envAwsKey)
-		os.Exit(1)
+// TODO replace with unit testable version
```
daniel-weisse, line 17: Fix TODO or add ticket reference and/or add person responsible to implement this

daniel-weisse (commented):
Just some style things

internal/api/attestationconfig/client/client_test.go
elchead: i wanted to get feedback on this way of testing. You think we should use it in all parts?

internal/api/attestationconfig/azure.go
```diff
@@ -34,14 +34,14 @@ type AzureSEVSNPVersion struct {
 	Microcode uint8 `json:"microcode"`
 }
 
-// AzureSEVSNPVersionGet is the request to get the version information of the specific version in the config api.
-type AzureSEVSNPVersionGet struct {
+// AzureSEVSNPVersionAPI is the request to get the version information of the specific version in the config api.
+type AzureSEVSNPVersionAPI struct {
```
derpsteb, line 38: Why did you rename this? And what is the rationale behind the name? 
The previous name seemed pretty intuitive to me and follows the pattern of other objects within the Resource API.

elchead: The Resource can be used for upload and get. Do u have a better naming suggestion?

derpsteb: Makes sense :+1: Maybe adapt the comment?
I guess it should just be named `AzureSEVSNPVersion`. But we already have that. I guess we can fix that when refactoring this to use `AttestationCfg`.

hack/configapi/cmd/root.go
```diff
@@ -94,7 +96,7 @@ func runCmd(cmd *cobra.Command, _ []string) error {
 	}
 	if isNewer {
 		fmt.Printf("Input version: %+v is newer than latest API version: %+v\n", inputVersion, latestAPIVersion)
```
daniel-weisse, line 98:
```suggestion
		cmd.Printf("Input version: %+v is newer than latest API version: %+v\n", inputVersion, latestAPIVersion)
```

internal/api/attestationconfig/client/client.go
```diff
@@ -9,49 +9,49 @@ import (
...
 		cosignPwd:     cosignPwd,
 		privKey:       privateKey,
+		fetcher:       fetcher.New(),
 	}
-	repoClose := func(ctx context.Context) error {
-		return repo.Close(ctx)
-	}
-	return repo, repoClose, nil
+	return repo, repo.Close, nil
 }
```
daniel-weisse, line 55: Why not just return the close function returned by `apiclient.NewClient`?

internal/api/attestationconfig/client/client.go
```diff
@@ -99,60 +99,142 @@ func (a Client) createAndUploadSignature(ctx context.Context, content []byte, fi
...
+	Upload(
+		ctx context.Context, input *s3.PutObjectInput, opts ...func(*s3manager.Uploader),
+	) (*s3manager.UploadOutput, error)
+	DeleteObject(ctx context.Context, params *s3.DeleteObjectInput,
+		optFns ...func(*s3.Options),
+	) (*s3.DeleteObjectOutput, error)
+}
+
+// CloseFunc is a function that closes the client.
+type CloseFunc func(ctx context.Context) error
```
daniel-weisse, line 193: Doesnt need to be exported (or defined at all probably).

internal/api/attestationconfig/client/client_test.go
daniel-weisse: I prefer having unit tests over integration tests that need to be ran manually in packages.
That being said, it might still be helpful to have an integration test to test the functionality of a package against a real API.
If you want to keep a test like that, I suggest moving it into a separate directory and package, e.g. `internal/api/attestationconfig/client/test/integration_test.go

derpsteb: I like the idea for testing, but would not extend it to the other tests within this PR.

derpsteb (approved):

PR #1864, opened 2023-06-02
cli: skip k8s upgrade in case of outdated version

derpsteb:
### Proposed change(s)

- CLI version `v2.7.1` supports k8s version `v1.25.7`. CLI version `v2.8` will support `v1.25.10` as oldest k8s version. Therefore a user might not be able to use their old config with the new CLI. To address this, this PR allows invalid k8s-patch versions during config validation and checks k8s versions later, in the relevant commands. `upgrade apply` will ask for confirmation upon encountering invalid k8s versions. If the user confirms, k8s upgrades will be skipped.
- Commit message includes code level summary what is changed.
- Added unit tests fail without the other changes. 

### Additional info
- [Azure, 1+1, k8s only](https://github.com/edgelesssys/constellation/actions/runs/5157869490)


cli/internal/cmd/upgradeapply.go
```diff
@@ -269,6 +274,23 @@ func (u *upgradeApplyCmd) parseUpgradeVars(cmd *cobra.Command, conf *config.Conf
 	}
 }
 
+// handleInvalidK8sPatchVersion checks if the Kubernetes patch version is supported and asks for confirmation if not.
+func handleInvalidK8sPatchVersion(cmd *cobra.Command, version string, yes bool) error {
+	valid := versions.NewValidK8sVersion(version, true) == ""
+
+	if !valid && !yes {
+		confirmed, err := askToConfirm(cmd, fmt.Sprintf("WARNING: The Kubernetes patch version %s is not supported. Do you want to continue anyway?", version))
```
malt3, line 282:
```suggestion
		confirmed, err := askToConfirm(cmd, fmt.Sprintf("WARNING: The Kubernetes patch version %s is not supported. If you continue, Kubernetes upgrades will be skipped. Do you want to continue anyway?", version))
```

I would try to explain with a little more detail what saying "yes" actually entails. Feel free to phrase this in a different way!

internal/versions/versions.go
```diff
@@ -40,15 +40,35 @@ func SupportedK8sVersions() []string {
 type ValidK8sVersion string
 
 // NewValidK8sVersion validates the given string and produces a new ValidK8sVersion object.
-func NewValidK8sVersion(k8sVersion string) (ValidK8sVersion, error) {
-	if IsSupportedK8sVersion(k8sVersion) {
-		return ValidK8sVersion(k8sVersion), nil
+// Returns an empty string if the given version is invalid.
+// strict controls whether the patch version is checked or not.
+func NewValidK8sVersion(k8sVersion string, strict bool) ValidK8sVersion {
```
malt3, line 45: We should maybe refactor this a little in a follow up PR. Personally I'm not happy with the fact that the return type `ValidK8sVersion` can actually be an invalid (empty string) version.
A proper solution (for a separate PR) could be return an error in this function and decide how to handle that on the caller side. 
That's more of a longer term thing and shouldn't block this PR.

malt3 (approved):
A nit about wording in the confirmation request and some code we should refactor later. Otherwise LGTM.

malt3 (approved):

PR #1867, opened 2023-06-05
api: use correct error types in versions/cli

derpsteb:
### Proposed change(s)

- Imported wrong error types during refactoring. This leads to checks like [this](https://github.com/edgelesssys/constellation/blob/main/internal/api/versions/cli/rm.go#L162) to not recognize errors correctly as the error returned is `client.NotFoundError`.


malt3 (approved):
Thanks for fixing!

PR #1870, opened 2023-06-05
staticupload: don't request empty invalidation

derpsteb:
### Proposed change(s)

- If no files have been touched, do not initiate an invalidation. Otherwise, the AWS API is unhappy.

### DIFF
--- a/internal/staticupload/staticupload.go
+++ b/internal/staticupload/staticupload.go
@@ -14,6 +14,7 @@ package staticupload
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


internal/staticupload/staticupload.go
```diff
@@ -171,6 +181,10 @@ func (c *Client) invalidateCacheForKeys(ctx context.Context, keys []string) (str
 		return "", InvalidationError{inner: fmt.Errorf("too many keys to invalidate: %d", len(keys))}
 	}
 
+	if len(keys) == 0 {
+		return "", InvalidationError{inner: errors.New("no keys to invalidate")}
+	}
+
```
daniel-weisse, line 187: I would assume the cdn client errors anyway if we pass an empty list of keys?
In that case we wouldnt need to manually return an error here. 
Of course this depends on if the error has side effects in the cdn or similar

daniel-weisse (approved):
Can you add tests for empty lists of invalidation keys?

malt3 (approved):

PR #1878, opened 2023-06-05
config: move all config types into file config.go

derpsteb:
### Proposed change(s)

- regression. docgen only includes doc comments from one single file in it's output. Therefore all config types need to be located in config.go


daniel-weisse (approved):
If our goal is to keep the `config.go` file clean, we could move all the struct methods to separate files, but doesnt need to happen now.

malt3 (approved):

elchead:
maltes change could be considered but lgtm

PR #1882, opened 2023-06-06
config: disable user-facing version fetch for Azure SEV SNP

elchead:
### Proposed change(s)
don't allow "latest" value and disable user-facing version fetcher for Azure SEV SNP.
REASON: v2.8 release.
Revert after release

Co-authored-by: @derpsteb

### Related issue
[AB#3036](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3036)

### Checklist

- [ ] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)
- [x] Add labels (e.g., for changelog category)
- [x] Link to Milestone

PR #1899, opened 2023-06-08
config: enable azure snp version fetcher again + minimum age for latest version

elchead:
### Proposed change(s)

- revert #1882 since v2.8 was released
- add a minimum age for the version value because the azure-sev-snp-report only gets the values on one machine. However, Azure does not provide official information on version rollouts, so we add a safety buffer of 2 weeks before we accept a new version update
- uploaded the current latest version with a timestamp of 2023-01-01 to be able to fetch latest versions now

### Related issue
#1882
[AB#3195](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3195)
[AB#3190](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3190)

### Additional info
- we should contact Azure to try to get official information on the version rollout

### Checklist

- [x] Link to Milestone

### DIFF
--- a/hack/configapi/cmd/root.go
+++ b/hack/configapi/cmd/root.go
@@ -56,7 +56,7 @@ func newRootCmd() *cobra.Command {
 	}
 	rootCmd.Flags().StringVarP(&versionFilePath, "version-file", "f", "", "File path to the version json file.")
 	rootCmd.Flags().BoolVar(&force, "force", false, "force to upload version regardless of comparison to latest API value.")
-	rootCmd.Flags().StringP("upload-date", "d", "", "force to upload version regardless of comparison to latest API value.")
+	rootCmd.Flags().StringP("upload-date", "d", "", "upload a version with this date as version name. Setting it implies --force.")
 	must(enforceRequiredFlags(rootCmd, "version-file"))
 	rootCmd.AddCommand(newDeleteCmd())
 	return rootCmd
@@ -98,6 +98,7 @@ func runCmd(cmd *cobra.Command, _ []string) error {
 		}
 	} else {
 		uploadDate = time.Now()
+		force = true
 	}
 
 	doUpload := false
@@ -111,11 +112,7 @@ func runCmd(cmd *cobra.Command, _ []string) error {
 		if err != nil {
 			return fmt.Errorf("comparing versions: %w", err)
 		}
-		if isNewer {
-			cmd.Printf("Input version: %+v is newer than latest API version: %+v\n", inputVersion, latestAPIVersion)
-		} else {
-			cmd.Printf("Input version: %+v is not newer than latest API version: %+v\n", inputVersion, latestAPIVersion)
-		}
+		cmd.Print(versionComparisonInformation(isNewer, inputVersion, latestAPIVersion.AzureSEVSNPVersion))
 		doUpload = isNewer
 	} else {
 		doUpload = true
@@ -139,6 +136,13 @@ func runCmd(cmd *cobra.Command, _ []string) error {
 		cmd.Printf("Successfully uploaded new Azure SEV-SNP version: %+v\n", inputVersion)
 	}
 	return nil
+}
+
+func versionComparisonInformation(isNewer bool, inputVersion attestationconfigapi.AzureSEVSNPVersion, latestAPIVersion attestationconfigapi.AzureSEVSNPVersion) string {
+	if isNewer {
+		return fmt.Sprintf("Input version: %+v is newer than latest API version: %+v\n", inputVersion, latestAPIVersion)
+	}
+	return fmt.Sprintf("Input version: %+v is not newer than latest API version: %+v\n", inputVersion, latestAPIVersion)
 }
 
 // isInputNewerThanLatestAPI compares all version fields with the latest API version and returns true if any input field is newer.
--- a/hack/configapi/cmd/root.go
+++ b/hack/configapi/cmd/root.go
@@ -56,6 +56,7 @@ func newRootCmd() *cobra.Command {
 	}
 	rootCmd.Flags().StringVarP(&versionFilePath, "version-file", "f", "", "File path to the version json file.")
 	rootCmd.Flags().BoolVar(&force, "force", false, "force to upload version regardless of comparison to latest API value.")
+	rootCmd.Flags().StringP("upload-date", "d", "", "force to upload version regardless of comparison to latest API value.")
 	must(enforceRequiredFlags(rootCmd, "version-file"))
 	rootCmd.AddCommand(newDeleteCmd())
 	return rootCmd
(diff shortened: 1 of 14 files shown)


internal/api/attestationconfigapi/fetcher.go
```diff
@@ -71,16 +76,34 @@ func (f *fetcher) FetchAzureSEVSNPVersion(ctx context.Context, azureVersion Azur
...
+	res, err = f.FetchAzureSEVSNPVersion(ctx, getVersionRequest)
 	if err != nil {
 		return res, fmt.Errorf("fetching version: %w", err)
 	}
-	return get, nil
+	return
+}
+
+// assumes that list is sorted reversely alphanumerically.
+func getLatestVersionOlderThanMinimumAge(list AzureSEVSNPVersionList, now time.Time, minimumAgeVersion time.Duration) (AzureSEVSNPVersionAPI, error) {
```
derpsteb, line 97: why have `minimumAgeVersion` as arg?

derpsteb:
we contacted azure via email.

hack/configapi/cmd/root.go
```diff
@@ -56,6 +56,7 @@ func newRootCmd() *cobra.Command {
 	}
 	rootCmd.Flags().StringVarP(&versionFilePath, "version-file", "f", "", "File path to the version json file.")
 	rootCmd.Flags().BoolVar(&force, "force", false, "force to upload version regardless of comparison to latest API value.")
+	rootCmd.Flags().StringP("upload-date", "d", "", "force to upload version regardless of comparison to latest API value.")
```
derpsteb, line 59: the description does not match the flag. also: I feel like the flag should imply `force` being set.

hack/configapi/cmd/root.go
```diff
@@ -85,21 +86,43 @@ func runCmd(cmd *cobra.Command, _ []string) error {
...
+		isNewer, err := isInputNewerThanLatestAPI(inputVersion, latestAPIVersion.AzureSEVSNPVersion)
+		if err != nil {
+			return fmt.Errorf("comparing versions: %w", err)
+		}
+		if isNewer {
 			cmd.Printf("Input version: %+v is newer than latest API version: %+v\n", inputVersion, latestAPIVersion)
+		} else {
+			cmd.Printf("Input version: %+v is not newer than latest API version: %+v\n", inputVersion, latestAPIVersion)
 		}
+		doUpload = isNewer
```
derpsteb, line 119:
```suggestion
out := fmt.Sprintf("Input version: %+v is newer than latest API version: %+v", inputVersion, latestAPIVersion)
		if !isNewer {
			out = fmt.Sprintf("Input version: %+v is not newer than latest API version: %+v", inputVersion, latestAPIVersion)
		}
		cmd.Println(out)
```

hack/configapi/cmd/root.go
```diff
@@ -85,21 +86,43 @@ func runCmd(cmd *cobra.Command, _ []string) error {
...
+	} else {
+		doUpload = true
+		cmd.Println("Forcing upload of new version")
+	}
+
+	if doUpload {
```
derpsteb, line 125:
```suggestion
		doUpload = isNewer
	} 

	if doUpload || force {
```

if you want the print statement you could add a `if force {cmd.Println...}`

trying to remove a few `else` blocks.

internal/api/attestationconfigapi/fetcher_test.go
```diff
@@ -10,68 +10,111 @@ import (
...
+		fetcherVersions []string
+		timeAtTest      time.Time
+		wantErr         bool
+		want            AzureSEVSNPVersionAPI
 	}{
-		"get version with valid signature": {
-			signature: []byte("MEQCIBPEbYg89MIQuaGStLhKGLGMKvKFoYCaAniDLwoIwulqAiB+rj7KMaMOMGxmUsjI7KheCXSNM8NzN+tuDw6AywI75A=="), // signed with release key
-			want:      testCfg,
+		"get latest version if older than 2 weeks": {
+			signature:       []byte("MEQCIBPEbYg89MIQuaGStLhKGLGMKvKFoYCaAniDLwoIwulqAiB+rj7KMaMOMGxmUsjI7KheCXSNM8NzN+tuDw6AywI75A=="), // signed with release key
```
derpsteb, line 42: we should find a way of using the dev key here.

derpsteb:
a few minor suggestions. only point I would want to address are the test signatures. otherwise, lgtm :+1:

derpsteb (approved):

internal/api/attestationconfigapi/fetcher.go
```diff
@@ -63,24 +73,42 @@ func (f *fetcher) FetchAzureSEVSNPVersion(ctx context.Context, azureVersion Azur
...
+		dateStr := strings.TrimSuffix(v, ".json")
+		versionDate, err := time.Parse("2006-01-01-01-01", dateStr)
+		if err != nil {
+			return AzureSEVSNPVersionAPI{}, fmt.Errorf("parsing version date %s: %w", dateStr, err)
+		}
+		if now.Sub(versionDate) > minimumAgeVersion {
+			return AzureSEVSNPVersionAPI{Version: v}, nil
+		}
+	}
+	return AzureSEVSNPVersionAPI{}, fmt.Errorf("no version older than %s found", minimumAgeVersion)
```
daniel-weisse, line 113: While in practice this error should not happen, I think it can be a bit confusing to read as a user.
If I understand the code correctly this should occur when a user has version set to `latest` in their config, but the API doesnt have a "latest" version available.
A user now gets the message "no version older than <date> found", which imo is the opposite of what `latest` would suggest.

Maybe a simple "no valid version found" would work here.

internal/config/config.go
```diff
@@ -981,8 +974,8 @@ func (c AzureSEVSNP) EqualTo(old AttestationCfg) (bool, error) {
 }
 
 // FetchAndSetLatestVersionNumbers fetches the latest version numbers from the configapi and sets them.
-func (c *AzureSEVSNP) FetchAndSetLatestVersionNumbers(fetcher attestationconfigapi.Fetcher) error {
-	versions, err := fetcher.FetchAzureSEVSNPVersionLatest(context.Background())
+func (c *AzureSEVSNP) FetchAndSetLatestVersionNumbers(fetcher attestationconfigapi.Fetcher, now time.Time) error {
+	versions, err := fetcher.FetchAzureSEVSNPVersionLatest(context.Background(), now)
```
daniel-weisse, line 978: Can we change this function to accept a context?

daniel-weisse (commented):
This PR does not affect when we upload new version to the API, correct?
If a new version is detected it is uploaded immediately.
The validation if a version in the API is a valid "latest" version happens only on the client side?

How high do you estimate the effort of only adding a version to the `/list` endpoint if it is a valid "latest" version?

This would ensure that anyone fetching information from the `/list` endpoints receives a list of references to valid version that *should* work to verify Azure VMs.

In any case, we should update the RFC to reflect how/when a version for Azure SEV can be considered as "latest"

PR #1900, opened 2023-06-08
attestation: add `awsSEVSNP` as new variant

derpsteb:
### Proposed change(s)

- Add `awsSEVSNP` as attestation variant. This does not include the attestation implementation. Instead, the standard awsNitroTPM attestation is used. A warning is printed during init and the config includes a warning comment.
- Enable AMD-SEV-SNP upon selecting the new variant
- Move `internal/variant` to `internal/attestation/variant`
- Change the default region for AWS in our e2e tests. Clear measurements for awsSEVSNP during e2e tests.

### Additional info
- [testrun](https://github.com/edgelesssys/constellation/actions/runs/5211769903) :green_circle: with image: `ref/feat-aws-snp-attestation/stream/debug/v2.8.0-pre.0.20230608071806-582d3a0b5009`
- Will add feature label to the PR that adds attestation, as the feature should be completed with that PR.

### Checklist

- [x] Link to Milestone


internal/config/config.go
```diff
@@ -751,6 +764,47 @@ func (c *Config) WithOpenStackProviderDefaults(openStackProvider string) *Config
...
 }
 
+// AWSSEVSNP is the configuration for AWS SEV-SNP attestation.
+type AWSSEVSNP struct {
+	// description: |
+	//   Expected TPM measurements.
+	Measurements measurements.M `json:"measurements" yaml:"measurements" validate:"required,no_placeholders"`
+	// description: |
+	//   Expected launch measurement in SNP report.
+	LaunchMeasurement measurements.Measurement `json:"launchMeasurement" yaml:"launchMeasurement" validate:"required"`
```
daniel-weisse, line 774: In the [attestation config RFC](https://github.com/edgelesssys/constellation/blob/main/rfc/attestation-config.md#aws), we specified the AWS SNP launch-measurement to be a list of acceptable values, this wont work if we use `measurement.Measurement` for the measurement in the config.
Instead, we should use a structure similar to the one used for ID key digest on Azure.

daniel-weisse (changes requested):

internal/attestation/aws/snp/snp.go
```diff
@@ -0,0 +1,54 @@
...
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+/*
+--------- WARNING! ---------
+THIS PACKAGE DOES CURRENTLY NOT IMPLEMENT ANY SNP ATTESTATION.
+It exists to implement required interfaces while implementing other parts of the AWS SNP attestation variant within Constellation.
+----------------------------
+
+# SNP
```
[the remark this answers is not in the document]
derpsteb: I removed some Azure specific abbreviation from the bottom. Not sure what else you mean. Would be great if you could highlight that on your next pass if you still see something.

internal/config/config.go
derpsteb: If it is alright with you I would keep this as is for now. We are still not 100% sure how the final object might look like. Because we might write a port of [ibm/sev-snp-measure](https://github.com/IBM/sev-snp-measure). Or we might hit some roadblock while doing so. 

The upcoming attestation PR will also ignore the field in the beginning so we have something working quickly.

daniel-weisse (approved):

PR #1906, opened 2023-06-09
config: drop support for deprecated Azure's service principal authentication

elchead:
### Context
This is the last step to completely deprecate the old method of handling Azure service accounts (see more on the migration [here](https://docs.edgeless.systems/constellation/reference/migration#migrating-from-azures-service-principal-authentication-to-managed-identity-authentication)). With the next release the CLI enforces the user to update its configuration to the new unified UAMI.

### Proposed change(s)

- err when setting app registration fields config file and warn when setting ENV variable
- update docs

### Additional info
[AB#2967](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/2967)

### Checklist

- [ ] Link to Milestone


internal/config/config.go
```diff
@@ -145,8 +145,8 @@ type AzureConfig struct {
 	//   Authorize spawned VMs to access Azure API.
 	UserAssignedIdentity string `yaml:"userAssignedIdentity" validate:"required"`
 	// description: |
-	//    Application client ID of the Active Directory app registration.
-	AppClientID string `yaml:"appClientID,omitempty" validate:"omitempty,uuid"`
+	//  DEPRECATED:  Application client ID of the Active Directory app registration.
+	AppClientID string `yaml:"appClientID,omitempty" validate:"omitempty,deprecated,deprecatedAppClientID"`
```
derpsteb, line 149: We need to remove the config key from the struct completely. Otherwise, each newly generated config also includes the key. 

I think a way to provide an extra error message for the field being set is to iterate over the ValidationErrors [1] and check the field name [2]. Keys that are not defined in the Config struct produce an error already. Then you can wrap the validation error or sth similar.

[1] https://github.com/edgelesssys/constellation/blob/main/internal/config/config.go#L731
[2] https://pkg.go.dev/github.com/go-playground/validator/v10#FieldError

cli/internal/cmd/verify_test.go
```diff
@@ -183,21 +179,25 @@ func TestVerify(t *testing.T) {
...
 				assert.Error(err)
 			} else {
 				assert.NoError(err)
 				assert.Contains(out.String(), "OK")
 				assert.Equal(tc.wantEndpoint, tc.protoClient.endpoint)
+
+				var validationErr *config.ValidationError
+				if errors.As(err, &validationErr) {
+					t.Log(validationErr.LongMessage())
+				}
```
derpsteb, line 200: I guess those are debug prints that can be removed?

internal/config/config.go
```diff
@@ -407,7 +424,7 @@ func New(fileHandler file.Handler, name string, fetcher attestationconfigapi.Fet
 	// Read secrets from env-vars.
 	clientSecretValue := os.Getenv(constants.EnvVarAzureClientSecretValue)
 	if clientSecretValue != "" && c.Provider.Azure != nil {
-		c.Provider.Azure.ClientSecretValue = clientSecretValue
+		return nil, fmt.Errorf("unsupported environment variable: %s : %w", constants.EnvVarAzureClientSecretValue, UnsupportedAppRegistrationError{})
```
derpsteb, line 427: We shouldn't error in case an env variable is set that we no longer support. Informing the user about the deprecation seems like a nice idea though. A warning should be enough imo.

internal/config/config.go
```diff
@@ -381,11 +376,33 @@ func fromFile(fileHandler file.Handler, name string) (*Config, error) {
...
 		}
+		if isAppClientIDError(err) {
+			return nil, UnsupportedAppRegistrationError{}
+		}
 		return nil, fmt.Errorf("could not load config from file %s: %w", name, err)
 	}
 	return &conf, nil
 }
 
+func isAppClientIDError(err error) bool {
```
derpsteb, line 387: Do you think there could be a unittest that asserts there is only one such error? To prevent a situation in the future where this recognizes an unrelated key.

derpsteb:
Looks a lot cleaner now. I like it! :+1:

internal/config/config.go
```diff
@@ -424,7 +424,7 @@ func New(fileHandler file.Handler, name string, fetcher attestationconfigapi.Fet
 	// Read secrets from env-vars.
 	clientSecretValue := os.Getenv(constants.EnvVarAzureClientSecretValue)
 	if clientSecretValue != "" && c.Provider.Azure != nil {
-		return nil, fmt.Errorf("unsupported environment variable: %s : %w", constants.EnvVarAzureClientSecretValue, UnsupportedAppRegistrationError{})
+		fmt.Printf("WARNING: the environment variable %s is no longer used %s", constants.EnvVarAzureClientSecretValue, UnsupportedAppRegistrationError{}.Error())
```
derpsteb, line 427: nit: warnings should go to stderr.

dev-docs/workflows/github-actions.md
```diff
@@ -83,25 +84,4 @@ In addition, you need to create a Service Account which Constellation itself is
...
-
-Afterward, you need to define a few secrets either as Github Action Secrets or in a secrets file for *act* as described before.
-
-The following secrets need to be defined:
-
-* `AZURE_E2E_CREDENTIALS`: The output of `az ad sp ...`
-* `AZURE_E2E_CLIENT_SECRET`: The client secret value for the registered app on Azure (which is defined as `appClientID`).
-
-For information on how to achieve this, refer to the [First steps](https://docs.edgeless.systems/constellation/getting-started/first-steps) in the documentation for Constellation.
+see [here](https://docs.edgeless.systems/constellation/workflows/config#creating-iam-credentials)
```
derpsteb, line 87:
```suggestion
See [here](https://docs.edgeless.systems/constellation/workflows/config#creating-iam-credentials).
```

derpsteb (approved):
looks good from my side, maybe @3u13r wants to take a quick look. he is more familiar with the topic.

Please update the PR description to give more context. This should be understandable by a user who clicked the PR link while reading the changelog.

malt3 (approved):
LGTM

derpsteb:
Please add some more information to the PR description for users who are redirected from the changelog. They can't see the linked ticket. We probably describe the topic in the docs somewhere.

PR #1897, opened 2023-06-07
cli: change generate-config flag to update-config flag

miampf:
### Proposed change(s)

- change `--generate-config` flag to `--update-config` in the iam create subcommand
- rename `--write-config` flag to `--update-config` in the upgrade check subcommand

### Context

SInce we'd need to support the growing number of flags in `constellation config generate` for the `--generate-config` flag, we should change the workflow to scale better while preserving flexibility.

### Additional info
- This changes the workflow to require `constellation config generate` before `constellation iam create`

### Checklist

- [X] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)
- [X] Add labels (e.g., for changelog category)
- [X] Link to Milestone

PR #1913, opened 2023-06-13
config: iam create aws check zone contains availability zone

elchead:
**Why?**
```sh
constellation create --control-plane-nodes 1 --worker-nodes 1

The following Constellation cluster will be created:
1 control-plane node of type m6a.xlarge will be created.
1 worker node of type m6a.xlarge will be created.
Do you want to create this cluster? [y/n]: y
Creating

Error: creating EC2 Subnet: InvalidParameterValue: Value (eu-central-1) for parameter availabilityZone is invalid. Subnets can currently only be created in the following availability zones: eu-central-1a, eu-central-1b, eu-central-1c.
	status code: 400, request id: 225f1583-b2d8-4278-aacd-23113226d00d

  with module.public_private_subnet.aws_subnet.private,
  on modules/public_private_subnet/main.tf line 15, in resource "aws_subnet" "private":
  15: resource "aws_subnet" "private" {


Error: creating EC2 Subnet: InvalidParameterValue: Value (eu-central-1) for parameter availabilityZone is invalid. Subnets can currently only be created in the following availability zones: eu-central-1a, eu-central-1b, eu-central-1c.
	status code: 400, request id: 1c1e558f-9c14-46bd-96e7-60af26b8d965

  with module.public_private_subnet.aws_subnet.public,
  on modules/public_private_subnet/main.tf line 22, in resource "aws_subnet" "public":
  22: resource "aws_subnet" "public" {


Attempting to roll back.
```
fails because config was generated without availability zone.

(Problem still persists with new `--update-config`)

### Proposed change(s)
- info that `--update-config` also updates the zone/region if it's not yet written in the file

- check for AWS suffix when parsing the flag
- for `--update-config` check if config provider and iam provider are the same and that the config zone is either empty or the same as the flag value

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/iamcreate.go
+++ b/cli/internal/cmd/iamcreate.go
@@ -79,11 +79,10 @@ func newIAMCreateAWSCmd() *cobra.Command {
 
 	cmd.Flags().String("prefix", "", "name prefix for all resources (required)")
 	must(cobra.MarkFlagRequired(cmd.Flags(), "prefix"))
-	// TODO: leave the availability zone out and should be called region if only necessary for iam creation?
 	cmd.Flags().String("zone", "", "AWS availability zone the resources will be created in, e.g. us-east-2a (required)\n"+
 		"Find available zones here: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-availability-zones. "+
 		"Note that we do not support every zone / region. You can find a list of all supported regions in our docs.")
-
+	must(cobra.MarkFlagRequired(cmd.Flags(), "zone"))
 	return cmd
 }
 
@@ -100,9 +99,9 @@ func newIAMCreateAzureCmd() *cobra.Command {
 	cmd.Flags().String("resourceGroup", "", "name prefix of the two resource groups your cluster / IAM resources will be created in (required)")
 	must(cobra.MarkFlagRequired(cmd.Flags(), "resourceGroup"))
 	cmd.Flags().String("region", "", "region the resources will be created in, e.g. westus (required)")
+	must(cobra.MarkFlagRequired(cmd.Flags(), "region"))
 	cmd.Flags().String("servicePrincipal", "", "name of the service principal that will be created (required)")
 	must(cobra.MarkFlagRequired(cmd.Flags(), "servicePrincipal"))
-
 	return cmd
 }
 
@@ -118,6 +117,7 @@ func newIAMCreateGCPCmd() *cobra.Command {
 
 	cmd.Flags().String("zone", "", "GCP zone the cluster will be deployed in (required)\n"+
 		"Find a list of available zones here: https://cloud.google.com/compute/docs/regions-zones#available.")
+	must(cobra.MarkFlagRequired(cmd.Flags(), "zone"))
 	cmd.Flags().String("serviceAccountID", "", "ID for the service account that will be created (required)\n"+
 		"Must match ^[a-z](?:[-a-z0-9]{4,28}[a-z0-9])$.")
 	must(cobra.MarkFlagRequired(cmd.Flags(), "serviceAccountID"))
--- a/cli/internal/cmd/iamcreate.go
+++ b/cli/internal/cmd/iamcreate.go
@@ -588,8 +588,8 @@ func awsZoneToRegion(zone string) (string, error) {
 	return fmt.Sprintf("%s-%s-%c", parts[0], parts[1], parts[2][0]), nil
 }
 
-// ValidateConfigWithFlagCompatibility checks if the config is compatible with the flags.
-func ValidateConfigWithFlagCompatibility(iamProvider cloudprovider.Provider, cfg config.Config, flags iamFlags) error {
+// validateConfigWithFlagCompatibility checks if the config is compatible with the flags.
+func validateConfigWithFlagCompatibility(iamProvider cloudprovider.Provider, cfg config.Config, flags iamFlags) error {
 	if !hasProviderCfg(iamProvider, cfg) {
 		return fmt.Errorf("iam provider %q seems different from config provider", iamProvider)
 	}
--- a/cli/internal/cmd/iamcreate.go
+++ b/cli/internal/cmd/iamcreate.go
@@ -235,8 +235,7 @@ func (c *iamCreator) create(ctx context.Context) error {
 		if err = c.fileHandler.ReadYAML(flags.configPath, &conf); err != nil {
 			return fmt.Errorf("error reading the configuration file: %w", err)
 		}
-		err := validateConfigWithFlagCompatibility(c.provider, conf, flags)
-		if err != nil {
+		if err := validateConfigWithFlagCompatibility(c.provider, conf, flags); err != nil {
 			return err
 		}
 		c.cmd.Printf("The configuration file %q will be automatically updated with the IAM values and zone/region information.\n", flags.configPath)
--- a/cli/internal/cmd/iamcreate.go
+++ b/cli/internal/cmd/iamcreate.go
@@ -230,16 +230,15 @@ func (c *iamCreator) create(ctx context.Context) error {
 
 	var conf config.Config
 	if flags.updateConfig {
-		c.cmd.Printf("The configuration file %q will be automatically updated and populated with the IAM values.\n", flags.configPath)
 		c.log.Debugf("Parsing config %s", flags.configPath)
 		if err = c.fileHandler.ReadYAML(flags.configPath, &conf); err != nil {
 			return fmt.Errorf("error reading the configuration file: %w", err)
 		}
-
 		err := ValidateConfigWithFlagCompatibility(c.provider, conf, flags)
 		if err != nil {
 			return err
 		}
+		c.cmd.Printf("The configuration file %q will be automatically updated with the IAM values and zone/region information.\n", flags.configPath)
 	}
 
 	c.spinner.Start("Creating", false)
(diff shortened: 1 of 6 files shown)


cli/internal/cmd/iamcreate.go
```diff
@@ -396,6 +396,10 @@ func (c *awsIAMCreator) parseFlagsAndSetupConfig(cmd *cobra.Command, flags iamFl
 	if err != nil {
 		return iamFlags{}, fmt.Errorf("parsing zone string: %w", err)
 	}
+	// check that zone has availability zone suffix
+	if !strings.HasSuffix(zone, "a") && !strings.HasSuffix(zone, "b") && !strings.HasSuffix(zone, "c") {
+		return iamFlags{}, fmt.Errorf("missing availability zone in %s flag. Availability zone is needed to support the default attestation variant SEV-SNP. please refer to our docs and https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-availability-zones", zone)
+	}
```
msanft, line 402: Should we 

a) Make this comment more generic (not specify SEV-SNP) 

or 

b) Check this only if SEV-SNP attestation is used (which should be the default soon?)

I'm not completely sure about the state of the attestation variants on AWS, but as we will keep supporting NitroTPM attestation, this error may be misunderstood when a user creates a cluster with NitroTPM (which should work without specifying a zone iirc)

cli/internal/cmd/iamcreate.go
```diff
@@ -396,6 +396,10 @@ func (c *awsIAMCreator) parseFlagsAndSetupConfig(cmd *cobra.Command, flags iamFl
 	if err != nil {
 		return iamFlags{}, fmt.Errorf("parsing zone string: %w", err)
 	}
+	// check that zone has availability zone suffix
+	if !strings.HasSuffix(zone, "a") && !strings.HasSuffix(zone, "b") && !strings.HasSuffix(zone, "c") {
+		return iamFlags{}, fmt.Errorf("missing availability zone in %s flag. Availability zone is needed to support the default attestation variant SEV-SNP. please refer to our docs and https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-availability-zones", zone)
```
msanft, line 401: Also I think we had some "policy" on personal addressing of users (e.g. "please refer"), but I'm not entirely sure anymore. @thomasten should know best

thomasten: We follow the Microsoft Style Guide. (You may add this to the dev-docs). Rules that apply here:
https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/p/please (related: https://learn.microsoft.com/en-us/style-guide/top-10-tips-style-voice#be-brief)
https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/w/we

cli/internal/cmd/iamcreate.go
elchead: > Should this check be at config validation instead of here?

I agree in principle to better have logic at one place. However, there is no validation check for this cmd and the current validator is imo not nice to use. I think this is sth to improve in the future in general. As we can see the validation checks are not enforced everywhere. I think it's better to fail early in the code and config validation would happen already quite late in the code path.

I would keep as is, also in sight of soonish support of multiple availability zones in the soon future according to @malt3. What do you think?

msanft: Don't have any strong opinion here, but just as a heads-up; It would probably make sense to sync this with #1897, as the config update should probably work both with specifying the zone as a flag and having it in the config file that will be updated through `iam create --update-config` before executing the command. (i.e. users can specify the zone ad-hoc on `iam create` or specify it in their Constellation config beforehand)

elchead: thanks, didn't know about the relevant PR

derpsteb: I would like to keep validation in the config validation for now. I am happy to make up a plan on how we can improve the config validation. And once we have that plan I am also all for implementing it incrementally. But I don't want to introduce random deviations from our current way of doing this. 

This check seems like a validation that should happen on _all_ configs. Not just the ones created by `iam create`.

re "late in the code path": as far as I can see it, the Config would be created/validated before any slow operations (the IAM creation) are triggered.

elchead:
waiting for #1897 to be merged

elchead:
can't do validation in config since zone cannot be specified through a flag in `config generate` atm. Should we change the workflow to move the zone specification to `config generate`?  
Makes more sense to specify there IMHO anyways. What do you think? @daniel-weisse

thomasten:
I think `config generate` shouldn't have flags for simple fields. Seems more convenient to edit the file afterwards.
(That said, what's the --kubernetes flag for? Seems to be easy to edit it in the file.)

elchead:
> I think you can still do zone validation in config. The next command using the config will validate it. I think that's how it works for other fields, too.

That's true, I missed that. we only validate when we read the config. In that case the user would have to manually edit this every time. A flag seems more convenient since it is absolutely necessary to specify this every time, unlike most other options

daniel-weisse:
> Ah I see, because zone is passed to `iam create`. We could also remove the flags from that command and require to edit the config.

Because `iam create` can also be used without a config, I think we shouldn't remove the flag from that command.
One thing we could do is making the `zone` and `config` mutually exclusive, i.e. if the user doesn't specify a zone, the zone from the config is used.

I don't think adding a flag for zone to `config generate` to remove the need for user interaction with the config file is a good argument, since afaik we will require users to interact with their config once the multi zone feature is fully implemented before they can execute `constellation create`.

elchead:
So I conclude, we don't add a flag and keep as is. 
I make `zone` and `config` mutually exclusive as @daniel-weisse suggested.

cli/internal/cmd/iamcreate.go
```diff
@@ -582,3 +587,57 @@ func awsZoneToRegion(zone string) (string, error) {
...
+	if !hasProviderCfg(iamProvider, cfg) {
+		return fmt.Errorf("iam provider %q seems different from config provider", iamProvider)
+	}
+
+	return checkIfCfgZoneAndFlagZoneDiffer(iamProvider, flags, cfg)
+}
+
+func checkIfCfgZoneAndFlagZoneDiffer(iamProvider cloudprovider.Provider, flags iamFlags, cfg config.Config) error {
+	flagZone := flagZoneOrAzRegion(iamProvider, flags)
+	configZone := configZoneOrAzRegion(iamProvider, cfg)
```
thomasten, line 602:
```suggestion
	configZone := cfg.GetZone()
```
this one doesn't exist yet, but there's already a similar func GetRegion

cli/internal/cmd/iamcreate_test.go
```diff
@@ -829,3 +858,94 @@ func TestIAMCreateGCP(t *testing.T) {
...
+			assert.NoError(err)
+		})
+	}
+}
+
+func createFSWithConfig(cfg config.Config) func(require *require.Assertions, provider cloudprovider.Provider, existingConfigFiles []string, existingDirs []string) afero.Fs {
+	return func(require *require.Assertions, provider cloudprovider.Provider, existingConfigFiles []string, existingDirs []string) afero.Fs {
+		fs := afero.NewMemMapFs()
+		fileHandler := file.NewHandler(fs)
+		fmt.Println("CFG", cfg.Provider.AWS)
```
thomasten, line 942:
```suggestion
```

docs/docs/workflows/config.md
```diff
@@ -88,7 +88,7 @@ See also Constellation's [Kubernetes support policy](../architecture/versions.md
 ## Creating an IAM configuration
 
 You can create an IAM configuration for your cluster automatically using the `constellation iam create` command.
-If you already have a constellation configuration file, you can add the `--update-config` flag to the command. This writes the needed IAM fields into your configuration.
+If you already have a constellation configuration file, you can add the `--update-config` flag to the command. This writes the needed IAM fields into your configuration. Furthermore, the flag updates the zone/region of the configuration when it's not yet set.
```
thomasten, line 91:
```suggestion
If you already have a Constellation configuration file, you can add the `--update-config` flag to the command. This writes the needed IAM fields into your configuration. Furthermore, the flag updates the zone/region of the configuration if it hasn't been set yet.
```
nit: grammar
(and I also missed the lower case Constellation in the other PR)

cli/internal/cmd/iamcreate.go
thomasten, line 1: I think we need (additional) config validation for catching the error you describe in "why?" for cases where users don't go the "iam create+update config" route.

thomasten (dismissed):

cli/internal/cmd/iamcreate.go
elchead: Yes there is no input validation on the fields zone and region in the config, just if they are set. It's actually confusing to me that the field with the zone contains the entire region + availability zone. I will add a validation for that

```
provider:
    # Configuration for AWS as provider.
    aws:
        region: eu-central-1 # AWS data center region. See: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions
        zone: eu-central-1b # AWS data center zone name in defined region. See: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-availability-zones
```

elchead: I'm wondering if we should invest more into thorough input validation. A lot of potential bugs can be traced back to not intended usage, I think

cli/internal/cmd/iamcreate.go
```diff
@@ -373,6 +375,9 @@ func (c *awsIAMCreator) parseFlagsAndSetupConfig(cmd *cobra.Command, flags iamFl
 		zone:   zone,
 	}
 
+	if !strings.HasSuffix(zone, "a") && !strings.HasSuffix(zone, "b") && !strings.HasSuffix(zone, "c") {
+		return iamFlags{}, fmt.Errorf("missing availability zone in %s flag. for more information see the docs and https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-availability-zones", zone)
+	}
 	flags.aws.region, err = awsZoneToRegion(zone)
```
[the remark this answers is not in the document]
elchead: What do you find cleaner about it? Actually putting a regex as a global package variable with a generic name, but provider specific usage does not seem very clean to me.

docs/docs/workflows/config.md
thomasten, line 91:
```suggestion
If you already have a Constellation configuration file, you can add the `--update-config` flag to the command. This writes the needed IAM fields into your configuration. Furthermore, the flag updates the zone/region of the configuration if it hasn't been set yet.
```
this got lost somehow

cli/internal/cmd/iamcreate.go
```diff
@@ -232,11 +231,15 @@ func (c *iamCreator) create(ctx context.Context) error {
 
 	var conf config.Config
 	if flags.updateConfig {
-		c.cmd.Printf("The configuration file %q will be automatically updated and populated with the IAM values.\n", flags.configPath)
 		c.log.Debugf("Parsing config %s", flags.configPath)
 		if err = c.fileHandler.ReadYAML(flags.configPath, &conf); err != nil {
 			return fmt.Errorf("error reading the configuration file: %w", err)
 		}
+		err := validateConfigWithFlagCompatibility(c.provider, conf, flags)
+		if err != nil {
```
thomasten, line 239:
```suggestion
		if err := validateConfigWithFlagCompatibility(c.provider, conf, flags); err != nil {
```
nit

thomasten (approved):

daniel-weisse (approved):

PR #1914, opened 2023-06-13
ci: add automated tests for reproducible builds

malt3:
### Proposed change(s)

- ci: add automated tests for reproducible builds

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/reproducible-builds.yml
+++ b/.github/workflows/reproducible-builds.yml
@@ -1,0 +1,99 @@
+# Build Constellation CLI and check for reproducible builds
+name: Reproducible Builds
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "45 06 * * 1" # Every Monday at 6:45am
+
+jobs:
+  build-binaries:
+    strategy:
+        fail-fast: false
+        matrix:
+          target:
+            - "cli_enterprise_darwin_amd64"
+            - "cli_enterprise_darwin_arm64"
+            - "cli_enterprise_linux_amd64"
+            - "cli_enterprise_linux_arm64"
+            - "cli_enterprise_windows_amd64"
+          runner: ["ubuntu-22.04", "macos-13"]
+    env:
+        bazel_target: "//cli:${{ matrix.target }}"
+        binary: "${{ matrix.target }}-${{ matrix.runner }}"
+    runs-on: ${{ matrix.runner }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+        with:
+          ref: ${{ !github.event.pull_request.head.repo.fork && github.head_ref || '' }}
+
+      - name: Setup bazel
+        uses: ./.github/actions/setup_bazel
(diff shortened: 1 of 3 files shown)


.github/workflows/reproducible-builds.yml
```diff
@@ -0,0 +1,104 @@
...
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - "release/**"
+  schedule:
+    - cron: "45 06 * * 1" # Every Monday at 6:45am
+  pull_request: # TODO: remove before merging
```
elchead, line 12: remove before merge

.github/workflows/reproducible-builds.yml
```diff
@@ -0,0 +1,104 @@
+# Build Constellation CLI and check for reproducible builds
+name: Reproducible Builds
+
+on:
+  workflow_dispatch:
+  push:
```
elchead, line 6: do we run on every push and on scheduled time? I think scheduled is enough

.github/workflows/reproducible-builds.yml
```diff
@@ -0,0 +1,104 @@
...
+          shell: bash
+          if: runner.os == 'Linux'
+          run: sha256sum cli_enterprise*
+
+        - name: Compare binaries
+          shell: bash
+          run: |
+            # shellcheck disable=SC2207,SC2116
+            list=($(echo "cli_enterprise*"))
+            diff -s --to-file="${list[0]}" "${list[@]:1}" | tee "${GITHUB_STEP_SUMMARY}"
```
elchead, line 104: Should we delete the artifacts after the run?

elchead:
Not sure if you already knew, but you can also manually trigger the workflow:
```sh
gh workflow run reproducible-builds.yml --ref feat/ci/reproducible-builds
```
I did it just know

.github/workflows/reproducible-builds.yml
malt3: Agreed

.github/workflows/reproducible-builds.yml
malt3: GitHub actions runners are ephemeral by default and the workspace directory in particular is always clean when starting a new job (even if you use self-hosted runners), so I don't see a need to delete files on the runner.

If you are talking about workflow artifacts that I can access via the webinterface after the run has completed: These are persisted by design since

- They are free
- They are super helpful when debugging diffs in the binaries (I can just download them from github and analyse the differences)
- They will automatically be deleted after some time

elchead:
> Not sure if you already knew, but you can also manually trigger the workflow:
> 
> ```shell
> gh workflow run reproducible-builds.yml --ref feat/ci/reproducible-builds
> ```
> 
> I did it just know

It failed due to failure to fetch aws credentials:
gh run view 5334894423 --log-failed

redoing now

.github/workflows/reproducible-builds.yml
elchead: that makes sense, thx for the explanation

elchead (approved):
lgtm with comments

PR #1916, opened 2023-06-13
attestation: add SNP-based attestation for aws-sev-snp 

derpsteb:
### Proposed change(s)

- Add attestation for the recently merged AWS SNP variant, #1900. The attestation currently only verifies the SNP report's signature. The hardcoded AMD root key is the trust anchor. The intermediate signing key is fetched from the AMD key service (KDS). The final signing key is part of the report. The attestation key can not be verified. The TPM can also not be verified since there is not endorsement key for AWS' TPM. The TPM is not in the confidential context.
- `launchMeasurement` is ignored for now. `launchMeasurement` has a different value, depending on the number of vCPUs of the issuing machine. Therefore we will have to write a port of [ibm/sev-snp-measure](https://github.com/IBM/sev-snp-measure) to precalculate these values in the CLI. To have most of the attestation tested in our e2e tests, I am proposing this PR without the launchMeasurement verification and will add `launchMeasurement` verification in later PR.

### Additional info
- go-tpm-tools includes the SNP report in the attestation document. We can only start leveraging that on AWS once proper VLEK support is added to go-sev-guest.
- The tldr for https://github.com/google/go-sev-guest/pull/47: AWS depends on a new-ish SNP feature for the report signing. They use a certificate called VLEK, Azure/GCP use the VCEK. The PR is about patching the lib so it doesn't crash when reading a VLEK. 
- The testcode relies on some real-world testdata files. This seemed like the simplest way to have some meaningful smoke testing, while keeping execution time low. Happy to hear other ideas.
- EDIT: **see comment at the bottom**

### Checklist

- [x] Link to Milestone


internal/attestation/aws/snp/testdata/ask.pem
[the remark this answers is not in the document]
derpsteb: Kept the attestationDocument in internal/vtpm as I am unsure how to put that into a constant nicely. Happy to change it if you know how.

internal/attestation/aws/snp/validator.go
```diff
@@ -55,47 +65,95 @@ func getTrustedKey(_ context.Context, attDoc vtpm.AttestationDocument, _ []byte)
...
 }
 
-func getEC2Client(ctx context.Context, region string) (awsMetadataAPI, error) {
-	client, err := awsConfig.LoadDefaultConfig(ctx, awsConfig.WithRegion(region))
+type kdsClient struct{}
+
+// GetASK requests the current certificate chain from the AMD KDS API and returns the ASK.
+func (k kdsClient) GetASK(ctx context.Context) (*x509.Certificate, error) {
+	// Hardcoding "Milan" (the CPU generation) might break in the future. But this is too far down the road to invest time right now.
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://kdsintf.amd.com/vlek/v1/Milan/cert_chain", nil)
```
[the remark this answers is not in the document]
derpsteb: I don't see this as something the user should have to configure. If at some point there are multiple CPU generations we can determine the correct API to call based on the SNP report. I will update the comment.

internal/attestation/aws/snp/validator.go
```diff
@@ -55,47 +65,95 @@ func getTrustedKey(_ context.Context, attDoc vtpm.AttestationDocument, _ []byte)
...
 
-// tpmEnabled verifies if the virtual machine has the tpm2.0 feature enabled.
-func (v *Validator) tpmEnabled(attestation vtpm.AttestationDocument, _ *attest.MachineState) error {
-	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/verify-nitrotpm-support-on-ami.html
-	// 1. Get the vm's ami (from IdentiTyDocument.imageId)
-	// 2. Check the value of key "TpmSupport": {"Value": "v2.0"}"
-	ctx := context.Background()
+// validateSNPReport validates the report by checking if it has a valid VLEK signature.
+// The certificate chain ARK -> ASK -> VLEK is also validated.
+// No CRLs are checked.
```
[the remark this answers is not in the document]
derpsteb: I extended the comment and added more info to one of our internal tickets.

https://dev.azure.com/Edgeless/Edgeless/_backlogs/backlog/Edgeless%20Team/Features?showParents=false&workitem=3205

internal/attestation/aws/snp/testdata/ask.pem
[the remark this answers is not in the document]
daniel-weisse: You could just put backticks around the string, no?

daniel-weisse (approved):
Code looks good to me now.
Before we finalize the feature we should think about how trust is established from the SNP report.
I know we are still missing a couple features from AWS to properly do so, but from my understanding, right now, we just verify that there is a valid SNP report added to the tpm attestation.
From what I can tell we never connect the two in any way.
Is there maybe a possibility to verify the SNP report using the user data that was used to generate the report?
That way we at least know that the SNP report was issued for this attestation.

derpsteb:
Had a short discussion on the phone:
I will move the content of `validateSNPReport` into `getTrustedKey`. Then we can validate that the userdata of the SNP report matches the userdata of the aTLS connection. This does not prevent an attacker capable of injecting a faked report from doing so. But it binds the report to the current aTLS connection. So at least the fake report has to be generated for each aTLS connection.

internal/attestation/aws/snp/testdata/ask.pem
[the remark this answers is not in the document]
derpsteb: done.

internal/attestation/aws/snp/validator.go
```diff
@@ -65,9 +73,17 @@ func getTrustedKey(_ context.Context, attDoc vtpm.AttestationDocument, _ []byte)
...
+	validate(attestation vtpm.AttestationDocument, kdsClient askGetter, ark *x509.Certificate, userData []byte) error
+}
+
+type awsValidator struct{}
+
+// validate the report by checking if it has a valid VLEK signature.
 // The certificate chain ARK -> ASK -> VLEK is also validated.
-func (v *Validator) validateSNPReport(attestation vtpm.AttestationDocument, _ *attest.MachineState) error {
+// Checks that the report's userData matches the connection's userData.
+func (awsValidator) validate(attestation vtpm.AttestationDocument, kdsClient askGetter, ark *x509.Certificate, userData []byte) error {
```
daniel-weisse, line 86: Since this function is called from `getTrustedKey`, which takes a context as argument, we can hand it down to this function and avoid creating a `context.Background()` on the fly

internal/attestation/aws/snp/validator.go
```diff
@@ -31,10 +32,12 @@ import (
 type Validator struct {
 	variant.AWSSEVSNP
 	*vtpm.Validator
-	// amd root key
+	// AMD root key.
 	ark *x509.Certificate
-	// kdsClient is used to query the AMD KDS API.
+	// kdsClient is required for testing.
```
daniel-weisse, line 37:
```suggestion
	// kdsClient is used to query the AMD KDS API.
```
I would keep the old description, its not relevant for someone reading this code to know that we use this for testing, but its relevant to know what this client does

internal/attestation/aws/snp/validator.go
```diff
@@ -46,16 +49,21 @@ func NewValidator(cfg *config.AWSSEVSNP, log attestation.Logger) *Validator {
...
-		getTrustedKey,
-		v.validateSNPReport,
+		v.getTrustedKey,
+		returnNil,
 		log,
 	)
 	return v
 }
 
+func returnNil(attestation vtpm.AttestationDocument, state *attest.MachineState) error { return nil }
```
daniel-weisse: Could be inlined as anonymous function, but can also be kept as is

internal/attestation/aws/snp/validator.go
derpsteb: The type's comment has the description you want. It describes what the type does. I think it is useful to tell the reader why that property exists. I can't find where I read that but found that style of comment very helpful while reading new code. I extended the comments, to include the information you are asking for.

internal/attestation/aws/snp/validator.go
```diff
@@ -7,45 +7,63 @@ SPDX-License-Identifier: AGPL-3.0-only
...
-	v.getDescribeClient = getEC2Client
 	return v
 }
 
 // getTrustedKeys return the public area of the provides attestation key.
-// Normally, here the trust of this key should be verified, but currently AWS does not provide this feature.
-func getTrustedKey(_ context.Context, attDoc vtpm.AttestationDocument, _ []byte) (crypto.PublicKey, error) {
+// Normally, the key should be verified here, but currently AWS does not provide means to do so.
+func (v *Validator) getTrustedKey(ctx context.Context, attDoc vtpm.AttestationDocument, userData []byte) (crypto.PublicKey, error) {
+	if err := v.reportValidator.validate(ctx, attDoc, v.kdsClient, v.ark, userData); err != nil {
```
thomasten, line 64: I think it makes more sense to put the TPM pubkey in the SNP report user data so we have a chain
SNP report -> verifies -> TPM -> verifies -> actual userData
similar how it works on Azure. Do you agree @derpsteb @daniel-weisse ? Or am I missing something?

internal/attestation/aws/snp/issuer_test.go
```diff
@@ -40,79 +36,3 @@ func TestGetAttestationKey(t *testing.T) {
 	// if everything worked fine, tpmAk and getAk are the same key
 	assert.Equal(tpmAk, getAk)
 }
-
-func TestGetInstanceInfo(t *testing.T) {
```
elchead, line 44: Should we try to keep testing this function? Happy to help out

daniel-weisse: This functions was removed

internal/attestation/aws/snp/validator.go
daniel-weisse: Technically there shouldn't be a difference in security (yet), but I agree that the story makes a bit more sense that way.

derpsteb: I agree with Daniel's conclusion. Nevertheless, I changed the code to include the sha512 hash of the AK (pub) in the SNP report. The validator calculates the hash of the AK it received in the attestation statement and compares it to the userdata.

thomasten (approved):

derpsteb:
[Smoketest](https://github.com/edgelesssys/constellation/actions/runs/5321159336)

derpsteb:
[Smoketest nr 2](https://github.com/edgelesssys/constellation/actions/runs/5324222253)

internal/attestation/aws/snp/errors.go
daniel-weisse, line 1: The errors and function in this file can be kept private, I think

derpsteb:
We are currently seeing an issue on AWS where some hosts are not able to generate SNP reports. Attestation fails for those hosts. As we don't know how quickly Amazon will fix the issue Daniel & I decided to make NitroTPM attestation the default on AWS again. This way we can start testing SNP manually but e2e tests will not fail.

We contacted AWS about the issue via email.

PR #1928, opened 2023-06-14
ci: replace katexochen with elchead in assignee list

derpsteb:
### Proposed change(s)

- Update list of assignees to include Adrian. Removed Paul as he is currently not actively working on Constellation.


katexochen (approved):
Thanks! Would love to also find some replacement in Codeowners at least for the bigger pieces (`cli`,`.github`,...)

PR #1929, opened 2023-06-14
cli: store upgrade files in versioned folders

msanft:
### Proposed change(s)

- Implement versioning of Constellation upgrades, so that a user can perform multiple upgrades without cleaning the `constellation-upgrade` directory. Each upgrade is marked by a timestamp string and a unique identifier and the corresponding files get placed into a unique subdirectory for their upgrade ID. (e.g. `constellation-upgrade/upgrade-20230614105831-976b92d0`)

### Additional info
- [Upgrade e2e test](https://github.com/edgelesssys/constellation/actions/runs/5265731418)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/upgradeapply.go
+++ b/cli/internal/cmd/upgradeapply.go
@@ -62,7 +62,7 @@ func runUpgradeApply(cmd *cobra.Command, _ []string) error {
 	defer log.Sync()
 
 	fileHandler := file.NewHandler(afero.NewOsFs())
-	upgrader, err := kubernetes.NewUpgrader(cmd.Context(), cmd.OutOrStdout(), log)
+	upgrader, err := kubernetes.NewUpgrader(cmd.Context(), cmd.OutOrStdout(), log, false)
 	if err != nil {
 		return err
 	}
--- a/cli/internal/cmd/upgradeapply.go
+++ b/cli/internal/cmd/upgradeapply.go
@@ -149,7 +149,7 @@ func (u *upgradeApplyCmd) migrateTerraform(cmd *cobra.Command, file file.Handler
 		return fmt.Errorf("checking workspace: %w", err)
 	}
 
-	targets, vars, err := u.parseUpgradeVars(cmd, conf, fetcher)
+	targets, vars, err := parseTerraformUpgradeVars(cmd, conf, fetcher)
 	if err != nil {
 		return fmt.Errorf("parsing upgrade variables: %w", err)
 	}
@@ -200,7 +200,8 @@ func (u *upgradeApplyCmd) migrateTerraform(cmd *cobra.Command, file file.Handler
 	return nil
 }
 
-func (u *upgradeApplyCmd) parseUpgradeVars(cmd *cobra.Command, conf *config.Config, fetcher imageFetcher) ([]string, terraform.Variables, error) {
+// parseTerraformUpgradeVars parses the variables required to execute the Terraform script with.
+func parseTerraformUpgradeVars(cmd *cobra.Command, conf *config.Config, fetcher imageFetcher) ([]string, terraform.Variables, error) {
 	// Fetch variables to execute Terraform script with
 	provider := conf.GetProvider()
 	attestationVariant := conf.GetAttestationConfig().GetVariant()
--- a/cli/internal/cmd/upgradeapply.go
+++ b/cli/internal/cmd/upgradeapply.go
@@ -62,7 +62,7 @@ func runUpgradeApply(cmd *cobra.Command, _ []string) error {
 	defer log.Sync()
 
 	fileHandler := file.NewHandler(afero.NewOsFs())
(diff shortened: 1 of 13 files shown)


cli/internal/kubernetes/upgrade.go
```diff
@@ -85,10 +86,22 @@ type Upgrader struct {
...
 	log              debugLog
+	upgradeID        string
 }
 
 // NewUpgrader returns a new Upgrader.
-func NewUpgrader(ctx context.Context, outWriter io.Writer, log debugLog) (*Upgrader, error) {
+// If checkOnly is true, the upgrader will only create the resources necessary for upgrade checking. (i.e. no Terraform client)
+// Not creating the Terraform client in the check-case makes sense as the initialization of the Terraform client copies the
+// executable to the upgrade directory, which is not necessary for the check-only case.
+func NewUpgrader(ctx context.Context, outWriter io.Writer, log debugLog, checkOnly bool) (*Upgrader, error) {
```
derpsteb, line 96: I agree that we might not always want to initialize the terraform client here. But if we don't, we should guard all access to `tfUpgrade` with a null-pointer check. `if u.tfUpgrade == nil {...}`. Alternatively you could omit the flag, put the initialization into a function and call that init function in case `u.tfUpgrade` is nil.

One could also argue that our `upgrade check` is incomplete, as it does not check if there are migrations that would need to be applied. If I am not mistaken we are applying migrations during `upgrade apply` unconditionally. Meaning that it makes no difference if the user is only upgrading services, image or k8s - migrations are applied in any case. Because of that, `upgrade check` could do a dry-run `terraform apply` to inform the user about the upcoming migration. I don't expect that you do this right now. I just had this thought and would like to hear what others think.

derpsteb:
Looks great. Just that one comment.

Started another [e2e testrun](https://github.com/edgelesssys/constellation/actions/runs/5275294976/jobs/9540610808) as I think the one you posted didn't upgrade the services and thus didn't execute the changed code.

cli/internal/cmd/upgradeapply.go
```diff
@@ -62,7 +62,7 @@ func runUpgradeApply(cmd *cobra.Command, _ []string) error {
 	defer log.Sync()
 
 	fileHandler := file.NewHandler(afero.NewOsFs())
-	upgrader, err := kubernetes.NewUpgrader(cmd.Context(), cmd.OutOrStdout(), log)
+	upgrader, err := kubernetes.NewUpgrader(cmd.Context(), cmd.OutOrStdout(), log, false)
```
katexochen, line 65: This is hard to read. Use a constant value to give the boolean some meaning.

cli/internal/kubernetes/upgrade.go
msanft: re. `upgrade check`:
We could certainly do this! However I think we should not add the "cloud resources" as a singular upgradable component such as microservices, k8s, etc. - Since the CLI gets upgraded either way, it might expect the cloud resources to exist no matter which component got upgraded. I think both `upgrade check` and `upgrade apply` should always consider the Terraform migrations.

msanft: re. nil check:
The main reason for not initializing the Terraform client here was that in the case of an `upgrade check`, the Terraform upgrade directories were created as well, as they get initialized with the Terraform client. If we dry-run the migrations in `upgrade check`, this would be obsolete.

msanft: As it probably needs some more code changes which are out of scope for this PR, I track this in [AB#3226](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3226).

msanft:
See #1942

derpsteb (approved):
lgtm.

cli/internal/kubernetes/upgrade.go
```diff
@@ -85,10 +96,23 @@ type Upgrader struct {
...
 	log              debugLog
+	upgradeID        string
 }
 
 // NewUpgrader returns a new Upgrader.
-func NewUpgrader(ctx context.Context, outWriter io.Writer, log debugLog) (*Upgrader, error) {
+func NewUpgrader(ctx context.Context, outWriter io.Writer, log debugLog, upgradeCmdKind UpgradeCmdKind) (*Upgrader, error) {
+	upgradeID := "upgrade-" + time.Now().Format("20060102150405") + "-" + strings.Split(uuid.New().String(), "-")[0]
+	if upgradeCmdKind == UpgradeCmdKindCheck {
+		upgradeID += "-check"
```
derpsteb, line 106: recap from our short conversation online: 
- change to "-tmp"
- add comment why we need to handle those folder separately.

derpsteb:
If you could update the PR title to something slightly more descriptive. That would be great. E.g. "cli: store backups during upgrade in timestamped folders". Thanks!

PR #1930, opened 2023-06-14
fix: Azure SEV-SNP version always gets overwritten by latest API versions

elchead:
We found this bug earlier and fixed it but somehow it appeared again.

**Why?**
When the user sets an attestation version, the value gets overwritten by the latest value. 

### Proposed change(s)

- check if latest is set before setting latest numerical value

### Checklist

- [x] Link to Milestone


internal/config/config_test.go
```diff
@@ -77,17 +78,68 @@ func TestReadConfigFile(t *testing.T) {
...
+			assert.Equal(tc.wantResult, result)
+		})
+	}
+}
+
+func modifyConfigForAzureToPassValidate(c *Config) {
+	c.RemoveProviderAndAttestationExcept(cloudprovider.Azure)
+	c.Image = "v" + constants.VersionInfo()
+	c.Provider.Azure.SubscriptionID = "f4278079-288c-4766-a98c-ab9d5dba01a5"
+	c.Provider.Azure.TenantID = "d4ff9d63-6d6d-4042-8f6a-21e804add5aa"
```
derpsteb, line 126: can we use values like `11111111-1111-1111-1111-111111111111` here?

internal/config/azure.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return err
+	}
+	// set number and keep isLatest flag
+	c.mergeWithLatestVersion(versions.AzureSEVSNPVersion)
+	return nil
+}
+
+func (c *AzureSEVSNP) mergeWithLatestVersion(latest attestationconfigapi.AzureSEVSNPVersion) {
+	if c.BootloaderVersion.IsLatest {
```
derpsteb, line 84: The `IsLatest` name seems a bit counterintuitive to me. At least in this place. `WantLatest` would represent the intention of the variable more accurately, wouldn't it?

elchead: Want is better,  I agree. naming is hard :)

derpsteb: it is! :D

derpsteb (approved):
lgtm :+1:

PR #1933, opened 2023-06-15
cli: improve user warning / information

elchead:
### Context

Currently there is no useful output that indicates success to the user:
```

./constellation config fetch-measurements --insecure

```
We also discussed printing a diff of the updated measurements, but comparing the `measurement.M` object seems tricky and is not worth the effort.

---

We should issue a warning on cluster creation with debug image and `debugCluster: false`:
```
../constellation create --control-plane-nodes 1 --worker-nodes 1 -y

Creating
15.92s user 5.14s system 16% cpu 2:06.29s total
```
Otherwise the user will get stuck at `init`.

### Proposed change(s)

- print success on fetch-measurements
- print warning for isDebugImg && !isDebugCluster 
- 

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -36,6 +36,7 @@ import (
 	"gopkg.in/yaml.v3"
 
 	"github.com/edgelesssys/constellation/v2/internal/api/attestationconfigapi"
+	"github.com/edgelesssys/constellation/v2/internal/api/versionsapi"
 	"github.com/edgelesssys/constellation/v2/internal/attestation/idkeydigest"
 	"github.com/edgelesssys/constellation/v2/internal/attestation/measurements"
 	"github.com/edgelesssys/constellation/v2/internal/attestation/variant"
@@ -544,7 +545,11 @@ func (c *Config) IsReleaseImage() bool {
 
 // IsDebugImage checks whether image name looks like a debug image.
 func (c *Config) IsDebugImage() bool {
-	return strings.Contains(c.Image, "debug")
+	v, err := versionsapi.NewVersionFromShortPath(c.Image, versionsapi.VersionKindImage)
+	if err != nil {
+		return false
+	}
+	return v.Stream == "debug"
 }
 
 // GetProvider returns the configured cloud provider.
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -543,8 +543,8 @@ func (c *Config) IsReleaseImage() bool {
 	return strings.HasPrefix(c.Image, "v")
 }
 
-// IsDebugImage checks whether image name looks like a debug image.
-func (c *Config) IsDebugImage() bool {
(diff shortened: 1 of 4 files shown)


internal/config/config.go
```diff
@@ -542,6 +542,11 @@ func (c *Config) IsReleaseImage() bool {
 	return strings.HasPrefix(c.Image, "v")
 }
 
+// IsDebugImage checks whether image name looks like a debug image.
+func (c *Config) IsDebugImage() bool {
+	return strings.Contains(c.Image, "debug")
+}
```
malt3, line 548: we may want to parse the image using `github.com/edgelesssys/constellation/v2/internal/api/versionsapi.NewVersionFromShortPath` and compare the `version.Stream` to `debug`.

internal/config/config.go
```diff
@@ -542,6 +543,15 @@ func (c *Config) IsReleaseImage() bool {
 	return strings.HasPrefix(c.Image, "v")
 }
 
+// IsDebugImage checks whether image name looks like a debug image.
+func (c *Config) IsDebugImage() bool {
```
thomasten, line 547: This func should be renamed to make clear that this isn't a reliable check. Otherwise, the use of it can be dangerous depending on what the caller is doing based on the info.

elchead: what about it is dangerous? you are talking about a case where a customer uses a custom image with exactly our version format: `^ref/([a-zA-Z0-9-]+)/stream/([a-zA-Z0-9-]+)/([a-zA-Z0-9.-]+)$`? and uses debug after `/stream` without it meaning debug image?
That seems unlikely to me, but what name would you suggest then?

thomasten: I'm talking about the opposite case where a debug image is not named like this. This func returning false doesn't mean it's a secure production image.
One name would be the very precise one: `IsNamedLikeDebugImage`. Callers then know that they can use the result of this func for things like printing warnings, but not for checking whether the image has (no) debug capabilities.

elchead: oki will apply. but just to be sure: debug images are just intended for us developers not for external right? so creating a  debug image is not officially supported anyways right?

thomasten: We might use debug images in PoCs. We might have already done that in the past, not sure.
Creating a debug image from source yourself isn't supported in terms of "Enterprise support", but is valid for external OSS users or external contributors.

IMO it's reason enough that some people may infer so. For example, if you previously used an SGX SDK, you may be used to the meaning of Debug=0 == production enclave.
Apart from the security aspect, functions should not lie/guess. IsDebugImage would need to return "don't know" if NewVersionFromShortPath failed. IsNamedLikeDebugImage can return false without guessing.

malt3 (approved):

PR #1940, opened 2023-06-16
cli: `upgrade apply --force` skips all compatibility checks

elchead:
### Context

We want to allow more flexibility to users by skipping safety checks when using `--force`. In the past not having this skip option during experimentation led to situations that were difficult to get out off for customers.

### Proposed change(s)

- skip all compatibility checks (`compatibility.IsValidUpgrade`)

### Related issue
- [AB#3183](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3183)

### Checklist

- [x] Link to Milestone


malt3:
Thanks for tackling this! Adding myself as a reviewer since I would like to test this in some scenarios / edge cases I've encountered recently.

malt3:
We need to add the force override in more places (any callers of `github.com/edgelesssys/constellation/v2/internal/compatibility.IsValidUpgrade` that come from upgrade apply).

Instances I identified:

- https://github.com/edgelesssys/constellation/blob/3fc8a5eff96caff5f99a1fb53268c0d7908be410/cli/internal/kubernetes/upgrade.go#L387

- https://github.com/edgelesssys/constellation/blob/3fc8a5eff96caff5f99a1fb53268c0d7908be410/cli/internal/helm/client.go#L78

The helm code is a bit tricky: Even if the force flag is used, we probably want to skip calling `helm upgrade` if the installed helm charts already match the desired helm charts.

elchead:
@malt3 added the checks and skips all `compatibility.IsValidUpgrade` calls with force now

cli/internal/helm/client.go
```diff
@@ -75,7 +75,7 @@ func NewClient(client crdClient, kubeConfigPath, helmNamespace string, log debug
 	return &Client{kubectl: client, fs: fileHandler, actions: actions{config: actionConfig}, log: log}, nil
 }
 
-func (c *Client) shouldUpgrade(releaseName, newVersion string) error {
+func (c *Client) shouldUpgrade(releaseName, newVersion string, force bool) error {
```
malt3, line 78: The helm chart logic is still not doing what I would expect it to do:

What I want to happen:

if currentVersion == newVersion => already at target version, do nothing (do not perform an upgrade)
if (currentVersion != newVersion) && force => perform upgrade
if (currentVersion != newVersion) && isValidUpgrade(currentVersion, newVersion) => perform upgrade
else: not compatible (do not perform an upgrade)

What this PR would end up doing:

if force: perform upgrade
if !force && isValidUpgrade(currentVersion, newVersion) => perform upgrade
else: not compatible (do not perform an upgrade)

elchead:
fixed error.As without double ptr as tested here:
https://go.dev/play/p/iYKluOm1blr

malt3 (approved):
LGTM

elchead:
@derpsteb should any other label be added instead?

PR #1941, opened 2023-06-19
terraform: gcp node groups

malt3:
### Context

Similar PRs will be opened for other CSPs.
In follow up PRs, we will address the bootstrapper-related code and make it ready for multiple node groups in different zones.
Please note that this PR should not change any behavior on its own.
Existing clusters can be migrated and (for now) we still create one group for control-plane nodes and one for workers.

### Proposed change(s)
- terraform: gcp node groups

### How to test

- Check that an existing gcp cluster on v2.8 can be migrated to this terraform state without any resources being deleted.
- Manually add an additional node group with different settings (zone, instance type) to the terraform.tfvars file. It should be created successfully using `terraform apply` (although it can't join the Constellation yet).

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/upgradeapply.go
+++ b/cli/internal/cmd/upgradeapply.go
@@ -258,15 +258,28 @@ func (u *upgradeApplyCmd) parseUpgradeVars(cmd *cobra.Command, conf *config.Conf
 		targets := []string{}
 
 		vars := &terraform.GCPClusterVariables{
-			CommonVariables: commonVariables,
-			Project:         conf.Provider.GCP.Project,
-			Region:          conf.Provider.GCP.Region,
-			Zone:            conf.Provider.GCP.Zone,
-			CredentialsFile: conf.Provider.GCP.ServiceAccountKeyPath,
-			InstanceType:    conf.Provider.GCP.InstanceType,
-			StateDiskType:   conf.Provider.GCP.StateDiskType,
-			ImageID:         imageRef,
-			Debug:           conf.IsDebugCluster(),
+			Name: conf.Name,
+			NodeGroups: map[string]terraform.GCPNodeGroup{
+				"control_plane_default": {
+					Role:            "ControlPlane",
+					StateDiskSizeGB: conf.StateDiskSizeGB,
+					Zone:            conf.Provider.GCP.Zone,
+					InstanceType:    conf.Provider.GCP.InstanceType,
+					DiskType:        conf.Provider.GCP.StateDiskType,
+				},
+				"worker_default": {
+					Role:            "Worker",
+					StateDiskSizeGB: conf.StateDiskSizeGB,
+					Zone:            conf.Provider.GCP.Zone,
+					InstanceType:    conf.Provider.GCP.InstanceType,
+					DiskType:        conf.Provider.GCP.StateDiskType,
+				},
+			},
+			Project: conf.Provider.GCP.Project,
+			Region:  conf.Provider.GCP.Region,
+			Zone:    conf.Provider.GCP.Zone,
+			ImageID: imageRef,
+			Debug:   conf.IsDebugCluster(),
 		}
 		return targets, vars, nil
 	default:
--- a/cli/internal/cmd/upgradeapply.go
+++ b/cli/internal/cmd/upgradeapply.go
@@ -172,6 +172,7 @@ func (u *upgradeApplyCmd) migrateTerraform(cmd *cobra.Command, file file.Handler
 
 	if hasDiff {
 		// If there are any Terraform migrations to apply, ask for confirmation
+		fmt.Fprintln(cmd.OutOrStdout(), "The upgrade requires a migration of Constellation cloud resources by applying an updated Terraform template. Please manually review the suggested changes below.")
 		if !flags.yes {
 			ok, err := askToConfirm(cmd, "Do you want to apply the Terraform migrations?")
 			if err != nil {
--- a/cli/internal/terraform/terraform_test.go
+++ b/cli/internal/terraform/terraform_test.go
@@ -24,7 +24,12 @@ import (
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/goleak"
 )
+
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m)
+}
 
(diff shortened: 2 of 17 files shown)


cli/internal/terraform/variables_test.go
```diff
@@ -0,0 +1,288 @@
...
+constellation_cmdline = "console=ttyS0,115200n8"
+vcpus = 4
+memory = 8192
+metadata_api_image = "example.com/metadata-api:latest"
+metadata_libvirt_uri = "qemu:///system"
+nvram = "/usr/share/OVMF/constellation_vars.production.fd"
+firmware = "/usr/share/OVMF/OVMF_CODE.fd"
+`
+	got := vars.String()
+	t.Logf("got:\n%s", got)
```
daniel-weisse, line 286:
```suggestion
```

malt3:
> Are there any images built for this, or can I just use the latest debug image from main?

There are no changes to code running in the image so no new image is needed.

>  DO I just create a cluster on v2.8 and then run upgrade apply with a CLI from this branch?

Yes. That should work. You should manually verify that the new terraform files end up in the `constellation-terraform` folder.

cli/internal/cmd/upgradeapply.go
```diff
@@ -258,15 +258,28 @@ func (u *upgradeApplyCmd) parseUpgradeVars(cmd *cobra.Command, conf *config.Conf
...
+					InstanceType:    conf.Provider.GCP.InstanceType,
+					DiskType:        conf.Provider.GCP.StateDiskType,
+				},
+			},
+			Project: conf.Provider.GCP.Project,
+			Region:  conf.Provider.GCP.Region,
+			Zone:    conf.Provider.GCP.Zone,
+			ImageID: imageRef,
+			Debug:   conf.IsDebugCluster(),
 		}
```
elchead, line 283: DRY and also use in create.go:

```suggestion
	pkg terraform
	func NewGCPClusterVariables(conf config.Config) *terraform.GCPClusterVariables {
		return &terraform.GCPClusterVariables{
			Name: conf.Name,
			NodeGroups: map[string]terraform.GCPNodeGroup{
				"control_plane_default": {
					Role:            "ControlPlane",
					StateDiskSizeGB: conf.StateDiskSizeGB,
					Zone:            conf.Provider.GCP.Zone,
					InstanceType:    conf.Provider.GCP.InstanceType,
					DiskType:        conf.Provider.GCP.StateDiskType,
				},
				"worker_default": {
					Role:            "Worker",
					StateDiskSizeGB: conf.StateDiskSizeGB,
					Zone:            conf.Provider.GCP.Zone,
					InstanceType:    conf.Provider.GCP.InstanceType,
					DiskType:        conf.Provider.GCP.StateDiskType,
				},
			},
			Project: conf.Provider.GCP.Project,
			Region:  conf.Provider.GCP.Region,
			Zone:    conf.Provider.GCP.Zone,
			ImageID: imageRef,
			Debug:   conf.IsDebugCluster(),
		}
	}
```

cli/internal/terraform/terraform_test.go
```diff
@@ -24,8 +24,13 @@ import (
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/goleak"
 )
 
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m)
+}
```
elchead, line 32: why do we need this? Should we add this to our testing principles that tests of concurrent code should always check this?

malt3: We should add this to every package that contains tests.
Even if we don't use concurrent code directly, it sometimes uncovers concurrency bugs in code that we import.

> Should we add this to our testing principles...

I think this should be added to our testing principles.

cli/internal/cmd/upgradeapply.go
elchead: sounds good

cli/internal/terraform/terraform_test.go
elchead: create a new ticket

daniel-weisse:
Adding new node group and creating them with `terraform apply` works.

> although it can't join the Constellation yet

Is this only a restriction for new control-plane groups or creating the group in a different zone?
I added a new worker group with a different instance type and the nodes were able to join just fine.
The nodes are also properly removed by `constellation terminate`

malt3:
> Is this only a restriction for new control-plane groups or creating the group in a different zone?

Yes. I'm already working on proper support for multiple zones on GCP for a follow up. It's a bit more involved to do this efficiently.

cli/internal/cmd/upgradeapply.go
malt3: Also as a general rule, I think we should follow 

> [A little copying is better than a little dependency.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s)

-- https://go-proverbs.github.io/

daniel-weisse:
Terraform migrations for existing clusters also work as expected.
I was able to add new instance groups to a cluster created with v2.8 CLI.

Small nitpick:
When running `constellation upgrade apply`, a user is immediately prompted by terraform to perform an action:
```sh
$ ./constellation upgrade apply

Terraform used the selected providers to generate the following execution
plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:
```
IMO, this could be bit confusing if users don't know Constellation using terraform under the hood to manage cloud infrastructure.
Maybe a small hint about Constellation needing to migrate/update is terraform configuration would be helpful here?

malt3:
> IMO, this could be bit confusing if users don't know Constellation using terraform under the hood to manage cloud infrastructure.
Maybe a small hint about Constellation needing to migrate/update is terraform configuration would be helpful here?

I agree that this is confusing, but it was actually introduced in #1685 and the UX of terraform migrations need more fine tuning in general. See also AB#3216 for ideas to improve this. 

I'll add a simple message in this PR but I do think this needs more work in general.

cli/internal/cmd/upgradeapply.go
malt3: Actually, since `upgradeapply.go` lives in `cli/internal/cmd` and `create.go` lives in  `cli/internal/cloudcmd`, I don't see an easy way to share this code.

daniel-weisse (approved):

PR #1942, opened 2023-06-19
cli: dry-run Terraform migrations on `upgrade check`

msanft:
### Context
- The `upgrade check` command used not to take the Terraform migrations into account that would happen within the checked upgrade. This should be changed in this PR.

### Proposed change(s)

- The `upgrade check` command should perform a dry-run of the Terraform migrations that will be applied within the update, to show the user which actions on cloud resources will be taken upon the update.

### Additional info
- Right now, the `parseTerraformUpgradeVars` function is shared between both the `upgrade check` and the `upgrade apply` command. This is certainly not a great dependency model and it should likely be factored out. I was unsure about what an appropriate place might be.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/upgradecheck.go
+++ b/cli/internal/cmd/upgradecheck.go
@@ -106,7 +106,7 @@ func parseUpgradeCheckFlags(cmd *cobra.Command) (upgradeCheckFlags, error) {
 	}
 	force, err := cmd.Flags().GetBool("force")
 	if err != nil {
-		fmt.Errorf("parsing force bool: %w", err)
+		return upgradeCheckFlags{}, fmt.Errorf("parsing force bool: %w", err)
 	}
 	writeConfig, err := cmd.Flags().GetBool("write-config")
 	if err != nil {
--- a/cli/internal/cmd/upgradecheck.go
+++ b/cli/internal/cmd/upgradecheck.go
@@ -18,6 +18,8 @@ import (
 	"github.com/edgelesssys/constellation/v2/cli/internal/featureset"
 	"github.com/edgelesssys/constellation/v2/cli/internal/helm"
 	"github.com/edgelesssys/constellation/v2/cli/internal/kubernetes"
+	"github.com/edgelesssys/constellation/v2/cli/internal/terraform"
+	"github.com/edgelesssys/constellation/v2/cli/internal/upgrade"
 	"github.com/edgelesssys/constellation/v2/internal/api/attestationconfigapi"
 	"github.com/edgelesssys/constellation/v2/internal/api/fetcher"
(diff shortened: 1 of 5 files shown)


elchead:
Could you please add the Context section of the new PR template to the description?

elchead:
I would suggest adding a flag `--dry-run` rather than adding a new subcommand as done in `helm`:
https://helm.sh/docs/chart_template_guide/debugging/

cli/internal/upgrade/terraform.go
```diff
@@ -120,8 +120,7 @@ func (u *TerraformUpgrader) PlanTerraformMigrations(ctx context.Context, opts Te
 // aborted by the user.
 func (u *TerraformUpgrader) CleanUpTerraformMigrations(fileHandler file.Handler, upgradeID string) error {
 	cleanupFiles := []string{
-		filepath.Join(constants.UpgradeDir, upgradeID, constants.TerraformUpgradeBackupDir),
-		filepath.Join(constants.UpgradeDir, upgradeID, constants.TerraformUpgradeWorkingDir),
+		filepath.Join(constants.UpgradeDir, upgradeID),
```
elchead, line 123: is it safe to not specify the subdirs anymore? not saying it is not, but just want to raise potential implications if there could be more directories. We still seem to follow explicit listing in l.97f

cli/internal/cmd/upgradecheck.go
```diff
@@ -661,16 +717,20 @@ func (v *versionCollector) filterCompatibleCLIVersions(ctx context.Context, cliP
...
-	stream      string
+	configPath        string
+	force             bool
+	writeConfig       bool
+	ref               string
+	stream            string
+	terraformLogLevel terraform.LogLevel
 }
 
 type upgradeChecker interface {
```
elchead, line 728: the interface seems to get more away from what the name suggests: "checking" and instead has side-effects. 
PlanTerraformMigrations and CleanUpTerraformMigrations don't seem to belong there.
By seperating this we could maybe also avoid passing the same checker object twice in `runUpgradeCheck` for cmd creation?

elchead:
Happy to jump on a call to discuss together

cli/internal/upgrade/terraform.go
msanft: Yeah, I should've changed this when we implemented the upgrade versioning. Before we had the 2 subfolders that needed to be cleaned individually. Now we have 1 folder `{constants.UpgradeDir}/{upgradeID}`

derpsteb:
We had a short discussion on the `--dry-run` flag and came to the conclusion that this is not relevant for this PR.

PR #1943, opened 2023-06-19
cloud: fix discovery of GCP nodes across zones

malt3:
### Context

This is part of the larger effort to support multiple node groups.
When creating a GCP cluster across zones, newly joining nodes need to auto-discover the Constellation control-plane in order to obtain a join ticket.
Before this PR, only nodes in the same zone could be discovered. With this PR, instances across all zones of the same region can be discovered.

### Proposed change(s)
- cloud: fix discovery of GCP nodes across multiple zones

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/cloud/gcp/gcp.go
+++ b/internal/cloud/gcp/gcp.go
@@ -23,6 +23,8 @@ import (
 	"path"
 	"regexp"
 	"strings"
+	"sync"
+	"time"
 
 	compute "cloud.google.com/go/compute/apiv1"
 	"cloud.google.com/go/compute/apiv1/computepb"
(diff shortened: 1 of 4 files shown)


internal/cloud/gcp/gcp.go
```diff
@@ -399,6 +445,66 @@ func (c *Cloud) initSecretHash(ctx context.Context, project, zone, instanceName
...
+	return region, nil
+}
+
+// zones retrieves all zones that are within a region.
+func (c *Cloud) zones(ctx context.Context, project, region string) ([]string, error) {
+	c.cacheMux.Lock()
+	defer c.cacheMux.Unlock()
+	if cachedZones, ok := c.zoneCache[region]; ok && time.Since(cachedZones.zoneCacheTime) < maxCacheAgeInvariantResource {
+		return cachedZones.zones, nil
+	}
```
3u13r, line 474: A few strategically placed newlines would improve readibility here.

3u13r (approved):

PR #1944, opened 2023-06-19
ref: little improvement of general api client

elchead:
### Proposed change(s)

-
-

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/internal/api/client/client.go
+++ b/internal/api/client/client.go
@@ -5,7 +5,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 */
 
 /*
-Package client provides a client for the versions API.
+Package client provides a general client to be used in for the several APIs.
 
 The client needs to be authenticated with AWS. It should be used in internal
 development and CI tools for administrative tasks. For just fetching information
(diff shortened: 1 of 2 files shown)


katexochen:
Please provide a PR description why you are doing this change. In the same way, neither your commit message nor the PR title says anything meaningful about the change.

internal/api/client/client.go
```diff
@@ -56,37 +56,8 @@ type Client struct {
 	Log *logger.Logger
 }
 
-// NewReadOnlyClient creates a new read-only client.
-// This client can be used to fetch objects but cannot write updates.
-func NewReadOnlyClient(ctx context.Context, region, bucket, distributionID string,
```
katexochen, line 61: I prefer the approach with an explicit constructor. IMO a ReadOnlyClient has a fundamental difference to a normal client, what is expressed by a separate constructor. I think such fundamental change in behavior shouldn't be made in a function parameter.

If you despite want to use boolean values as parameter, please define constant values in the same package so a caller can pass a meaningful value.

katexochen: If you just want to reduce code duplication (which would be a liitle overengineerin IMO), you could define a private function that is called by both constructors to hide the boolean parameter from callers.

PR #1945, opened 2023-06-20
csi: aws csi driver policies

msanft:
### Context

- To use the AWS CSI driver, the corresponding storage permissions are necessary.

### Proposed change(s)

- Add the required storage permissions.

### Additional info
- Should be merged with the [driver PR](https://github.com/edgelesssys/constellation-aws-ebs-csi-driver/pull/1)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/terraform/terraform/iam/aws/main.tf
+++ b/cli/internal/terraform/terraform/iam/aws/main.tf
@@ -197,6 +197,10 @@ resource "aws_iam_role_policy_attachment" "attach_worker_node_policy" {
   policy_arn = aws_iam_policy.worker_node_policy.arn
 }
 
+resource "aws_iam_role_policy_attachment" "attach_worker_node_csi_policy" {
+  role       = aws_iam_role.worker_node_role.name
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonEBSCSIDriverPolicy"
+}
 
 // Add all permissions here, which are needed by the bootstrapper
 resource "aws_iam_policy" "constellation_bootstrapper_policy" {
--- a/cli/internal/terraform/terraform/iam/aws/main.tf
+++ b/cli/internal/terraform/terraform/iam/aws/main.tf
@@ -232,7 +232,7 @@ resource "aws_iam_role_policy_attachment" "csi_driver_policy_worker" {
   policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonEBSCSIDriverPolicy"
 }
 
-// TODO(msanft): incorporate this into the custom worker node policy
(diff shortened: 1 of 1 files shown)


cli/internal/terraform/terraform/iam/aws/main.tf
```diff
@@ -197,6 +197,11 @@ resource "aws_iam_role_policy_attachment" "attach_worker_node_policy" {
   policy_arn = aws_iam_policy.worker_node_policy.arn
 }
 
+// TODO(msanft): incorporate this into the custom worker node policy
+resource "aws_iam_role_policy_attachment" "attach_worker_node_csi_policy" {
```
3u13r, line 201: This is superfluous as it is the same resource as the first one below, right?

cli/internal/terraform/terraform/iam/aws/main.tf
```diff
@@ -232,11 +226,13 @@ resource "aws_iam_role_policy_attachment" "attach_bootstrapper_policy_control_pl
   policy_arn = aws_iam_policy.constellation_bootstrapper_policy.arn
 }
 
+// TODO(msanft): incorporate this into the custom worker node policy
 resource "aws_iam_role_policy_attachment" "csi_driver_policy_worker" {
   role       = aws_iam_role.worker_node_role.name
   policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonEBSCSIDriverPolicy"
 }
 
+// TODO(msanft): incorporate this into the custom worker node policy
```
daniel-weisse, line 235:
```suggestion
// TODO(msanft): incorporate this into the custom control-plane node policy
```

daniel-weisse (approved):

cli/internal/terraform/terraform/iam/aws/main.tf
daniel-weisse: By "incorporate into custom policy", do you mean to extract the permissions from the managed AWS-CSI role and add the permissions to our role?

msanft: Exactly. I think we should add this to the minimal permission set we define for all CSPs

msanft: The one that we document, that's used in the CI, etc.

daniel-weisse: Do you think it is something worth doing because the managed role provides excess permissions, or is it just to combine permissions into one?
If its the later I would argue against it and keep the current way of attaching the role, as this guarantees the role to always have the correct permissions for CSI

malt3 (approved):

derpsteb:
Slight title change for changelog readability.

PR #1955, opened 2023-06-21
terraform: azure node groups

elchead:
### Context

Similar to #1941.

Additionally, we found a bug in the current terraform migration behavior that allows any target when the target slice is empty.

### Proposed change(s)

- terraform: azure node groups
- add a check to skip terraform migration when the targets are empty

### How to test
as in #1941:

Check that an existing azure cluster on v2.8 can be migrated to this terraform state without any resources being deleted.
Manually add an additional node group with different settings (zone, instance type) to the terraform.tfvars file. It should be created successfully using terraform apply (although it can't join the Constellation yet).

Correct creation and deletion using the PR-CLI was also verified.

### Checklist

- [x] Link to Milestone


cli/internal/terraform/terraform/azure/modules/scale_set/main.tf
```diff
@@ -34,8 +57,9 @@ resource "azurerm_linux_virtual_machine_scale_set" "scale_set" {
   upgrade_mode                    = "Manual"
   secure_boot_enabled             = var.secure_boot
   source_image_id                 = var.image_id
-  tags                            = var.tags
-
+  tags                            = local.tags
+  zones                           = var.zones
+  // TODO add
```
3u13r, line 62: This need to be a proper `TODO` or be cleaned-up

cli/internal/terraform/terraform/azure/modules/scale_set/variables.tf
```diff
@@ -1,7 +1,31 @@
...
+  description = "The role of the instance group."
+  validation {
+    condition     = contains(["ControlPlane", "Worker"], var.role)
+    error_message = "The role has to be 'ControlPlane' or 'Worker'."
+  }
+}
+
+variable "tags" {
+  type        = map(string)
+  description = "Cluster tag."
```
3u13r, line 22: In the description you use `tag` as singular while the variable name and type signal that this (can) contain multiple `tags`.

cli/internal/cmd/upgradeapply.go
```diff
@@ -157,10 +158,10 @@ func (u *upgradeApplyCmd) migrateTerraform(cmd *cobra.Command, file file.Handler
...
 
 	opts := upgrade.TerraformUpgradeOptions{
-		LogLevel:   flags.terraformLogLevel,
-		CSP:        conf.GetProvider(),
-		Vars:       vars,
-		Targets:    targets,
+		LogLevel: flags.terraformLogLevel,
+		CSP:      conf.GetProvider(),
+		Vars:     vars,
+		// Targets:    targets,
```
3u13r, line 164: This is also a left-over from debugging, right?

cli/internal/terraform/terraform/azure/modules/scale_set/main.tf
```diff
@@ -11,6 +11,29 @@ terraform {
...
+  # migration: allow the old node group names to work since they were created without the uid
+  # and without multiple node groups in mind
+  # node_group: worker_default => name == "<base>-1-worker"
+  # node_group: control_plane_default => name:  "<base>-control-plane"
+  # new names:
+  # node_group: foo, role: Worker => name == "<base>-worker-<uid>"
+  # node_group: bar, role: ControlPlane => name == "<base>-control-plane-<uid>"
+  group_uid = random_id.uid.hex
+  maybe_uid = (var.node_group_name == "control_plane_default" || var.node_group_name == "worker_default") ? "" : "-${local.group_uid}"
+  #maybe_one = var.node_group_name == "worker_default" ? "-1" : "" # TODO adding "-1" forces replaces of resource on upgrade
```
3u13r, line 30: Move the `# TODO` comment in a new line, if it's still needed. If so, also adhere to our `TODO` guidelines.

cli/internal/terraform/terraform/azure/modules/scale_set/main.tf
```diff
@@ -11,6 +11,29 @@ terraform {
...
 }
 
+locals {
+  role_dashed = var.role == "ControlPlane" ? "control-plane" : "worker"
+  tags = merge(
+    var.tags,
+    { constellation-role = local.role_dashed },
+    { constellation-node-group = var.node_group_name },
+  )
+  # migration: allow the old node group names to work since they were created without the uid
```
3u13r, line 21: The migration probably has a start version (from which version on do we need this migration/do we start migrating)  and an end version (when have we migrated / can expect the current infrastructure to fit the new assumptions). Then we can more easily remove this migration when it is no longer needed.
Please add that information here or in other parts of the migration and refer to it by e.g. referring to the same ticket.

cli/internal/cmd/upgradeapply.go
```diff
@@ -244,16 +245,32 @@ func parseTerraformUpgradeVars(cmd *cobra.Command, conf *config.Config, fetcher
...
-			InstanceType:         conf.Provider.Azure.InstanceType,
-			StateDiskType:        conf.Provider.Azure.StateDiskType,
 			ImageID:              imageRef,
-			SecureBoot:           *conf.Provider.Azure.SecureBoot,
-			CreateMAA:            conf.GetAttestationConfig().GetVariant().Equal(variant.AzureSEVSNP{}),
-			Debug:                conf.IsDebugCluster(),
+			NodeGroups: map[string]terraform.AzureNodeGroup{
+				"control_plane_default": {
+					Role:          "ControlPlane",
+					InstanceCount: 1, // TODO
```
3u13r, line 255: Left-over `TODO` ?

cli/internal/cmd/upgradeapply.go
```diff
@@ -244,16 +245,32 @@ func parseTerraformUpgradeVars(cmd *cobra.Command, conf *config.Config, fetcher
...
+					InstanceType:  conf.Provider.Azure.InstanceType,
+					DiskSizeGB:    conf.StateDiskSizeGB,
+					DiskType:      conf.Provider.Azure.StateDiskType,
+					// Location:      conf.Provider.Azure.Location,
```
3u13r, line 259: Also left-over from debugging?

3u13r (changes requested):

cli/internal/terraform/terraform/azure/modules/scale_set/main.tf
elchead: the distinction of the current naming will only become obsolete once all cluster with a version <v2.9 are destroyed. Not sure at what point we want users to force to switch. It was also done like this in the GCP PR.
The existing name cannot be changed because it would force a recreate. 
Any suggestion?

3u13r: Hm, then maybe the name "migration" is a bit misleading an "compatibility" is more correct? Since we want to upload compatibility with the old version and don't change (aka. migrate) any resources, as you mentioned.

@malt3 What do you think of this? 

How long do we expect this piece of code to be there? I'd want to get rid of it _eventually_ (in 3-4 versions?). We can always have migrations with downtime, as long as they can be prepared for.

cli/internal/cmd/upgradeapply.go
3u13r: Just that we are on the same page here, the current code provides no solution which should be merged into main, as it disables the Terraform migration for all CSPs, right? 

Since I'm reading the code for the first time, maybe @malt3 @derpsteb, or @msanft can chime in on how to best handle the migration part here. My question would be, why this did not come up during the GCP version of this PR.

elchead: yes the current commit cannot be merged. I also wonder why it didn't cause problems in GCP

cli/internal/cmd/upgradeapply.go
elchead: in the past we didn't set this during an upgrade, so it assumed the type default of 0 and it worked.. maybe 0 just assumes 1? I actually wonder why this was not a problem if we created a cluster with a count higher than 1

cli/internal/cmd/upgradeapply.go
malt3: I believe this was an oversight on my side. I was pretty sure that this worked in my tests but looking at the code, we will have to either drop targets altogether or include the full list of moved resources in the list of targets.

3u13r: From what I see is that GCP should work correctly, since it does not migrate any resources. But as soon as we want to migrate a resource Terraform needs to migrate _ALL_ moved resources. 
> Resource instances in your current state have moved to new addresses in the
latest configuration. Terraform must include those resource instances while
planning in order to ensure a correct result, but your -target=... options to
not fully cover all of those resource instances.

cli/internal/terraform/terraform/azure/modules/scale_set/main.tf
daniel-weisse: afaik we can't ever remove this backwards compatibility option because it will break upgrade compatibility with any version of Constellation created before the version this features is released in.

malt3: I thought about this naming change for some time now.
The whole migration / backwards compatibility for the scale set name can be skipped (and I can also remove it from the gcp template), as long as we get the lifecycle block correct.
This means:

- Existing clusters can have scale sets with the legacy names, they will not be replaced because of the lifecycle ignore annotations
- New clusters will use the new naming scheme even for the two default scale sets

elchead: so we remove the move block and add the old resources to ignore_changes?

elchead: I'm wondering if the benefit of not having the migration name check outweighs that we are then unable to change the legacy resource at all

malt3: No. The move block stays. This is about the following naming thing:

```
  group_uid = random_id.uid.hex
  maybe_uid = (var.node_group_name == "control_plane_default" || var.node_group_name == "worker_default") ? "" : "-${local.group_uid}"
  #maybe_one = var.node_group_name == "worker_default" ? "-1" : "" # TODO adding "-1" forces replaces of resource on upgrade
  name = "${var.base_name}-${local.role_dashed}${local.maybe_uid}"
```

And I would like to simplify it to always use the new naming scheme:

```
  group_uid = random_id.uid.hex
  name = "${var.base_name}-${local.role_dashed}${local.group_uid}"
```

Additionally, you need to explicitly ignore changes of the name for the scale set in the lifecycle block:

```hcl
resource "azurerm_linux_virtual_machine_scale_set" "scale_set" {

    # removed for brevity [...]

  lifecycle {
    ignore_changes = [
      name, # required. Allow legacy scale sets to keep their old names
      instances,       # required. autoscaling modifies the instance count externally
      source_image_id, # required. update procedure modifies the image id externally
    ]
  }

}
```

elchead: gotcha, thanks!

elchead:
@malt3 , @msanft made the changes as discussed:
- skip migration for empty targets
- add explicit targets for Azure migration

cli/internal/cloudcmd/create.go
```diff
@@ -18,6 +18,7 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
```
malt3, line 21:
```suggestion
```

nit: since this is CSP-agnostic code, it's a little bit cleaner to not import an Azure specific package here.
Since we have generics, you can simply define this at the bottom of the file (or in a separate file of the same package):

```go
func toPtr[T any](v T) *T {
	return &v
}
```

cli/internal/terraform/terraform/azure/main.tf
```diff
@@ -221,58 +221,47 @@ resource "azurerm_network_security_group" "security_group" {
...
-  name            = "${local.name}-control-plane"
-  instance_count  = var.control_plane_count
-  state_disk_size = var.state_disk_size
-  state_disk_type = var.state_disk_type
-  resource_group  = var.resource_group
-  location        = var.location
-  instance_type   = var.instance_type
-  confidential_vm = var.confidential_vm
-  secure_boot     = var.secure_boot
+// add role as label, old name is base_name-dashed_role
```
malt3, line 224:
```suggestion
```

nit: looks like a leftover comment for you

cli/internal/cloudcmd/create.go
```diff
@@ -212,25 +213,35 @@ func (c *Creator) createGCP(ctx context.Context, cl terraformClient, opts Create
...
 	vars := terraform.AzureClusterVariables{
-		CommonVariables: terraform.CommonVariables{
-			Name:               opts.Config.Name,
-			CountControlPlanes: opts.ControlPlaneCount,
-			CountWorkers:       opts.WorkerCount,
-			StateDiskSizeGB:    opts.Config.StateDiskSizeGB,
+		Name: opts.Config.Name,
+		NodeGroups: map[string]terraform.AzureNodeGroup{
+			"control_plane_default": {
+				Role:          "control-plane",
```
malt3, line 219:
```suggestion
				Role:          "ControlPlane",
```

cli/internal/cloudcmd/create.go
```diff
@@ -212,25 +213,35 @@ func (c *Creator) createGCP(ctx context.Context, cl terraformClient, opts Create
...
+				InstanceCount: to.Ptr(opts.ControlPlaneCount),
+				InstanceType:  opts.InsType,
+				DiskSizeGB:    opts.Config.StateDiskSizeGB,
+				DiskType:      opts.Config.Provider.Azure.StateDiskType,
+				Zones:         nil, // TODO(elchead): support zones AB#3225
+			},
+			"worker_default": {
+				Role:          "worker",
```
malt3, line 227:
```suggestion
				Role:          "Worker",
```

cli/internal/cmd/upgradeapply.go
```diff
@@ -236,24 +241,36 @@ func parseTerraformUpgradeVars(cmd *cobra.Command, conf *config.Config, fetcher
...
 			UserAssignedIdentity: conf.Provider.Azure.UserAssignedIdentity,
-			InstanceType:         conf.Provider.Azure.InstanceType,
-			StateDiskType:        conf.Provider.Azure.StateDiskType,
 			ImageID:              imageRef,
-			SecureBoot:           *conf.Provider.Azure.SecureBoot,
-			CreateMAA:            conf.GetAttestationConfig().GetVariant().Equal(variant.AzureSEVSNP{}),
-			Debug:                conf.IsDebugCluster(),
+			NodeGroups: map[string]terraform.AzureNodeGroup{
+				"control_plane_default": {
+					Role:         "control-plane",
```
malt3, line 258:
```suggestion
					Role:         "ControlPlane",
```

cli/internal/cmd/upgradeapply.go
```diff
@@ -236,24 +241,36 @@ func parseTerraformUpgradeVars(cmd *cobra.Command, conf *config.Config, fetcher
...
+					InstanceType: conf.Provider.Azure.InstanceType,
+					DiskSizeGB:   conf.StateDiskSizeGB,
+					DiskType:     conf.Provider.Azure.StateDiskType,
+				},
+				"worker_default": {
+					Role:         "worker",
```
malt3, line 264:
```suggestion
					Role:         "Worker",
```

cli/internal/terraform/terraform/azure/modules/scale_set/variables.tf
```diff
@@ -1,7 +1,31 @@
...
+}
+
+variable "role" {
+  type        = string
+  description = "The role of the instance group."
+  validation {
+    condition     = contains(["control-plane", "worker"], var.role)
+    error_message = "The role has to be 'control-plane' or 'worker'."
+  }
+}
```
malt3, line 18:
```suggestion
variable "role" {
  type        = string
  description = "The role of the instance group."
  validation {
    condition     = contains(["ControlPlane", "Worker"], var.role)
    error_message = "The role has to be 'ControlPlane' or 'Worker'."
  }
}
```

cli/internal/terraform/terraform/azure/modules/scale_set/main.tf
```diff
@@ -11,6 +11,19 @@ terraform {
   }
 }
 
+locals {
+  tags = merge(
+    var.tags,
+    { constellation-role = var.role },
```
malt3, line 17: This should perform the transformation `ControlPlane` => `control-plane` and `Worker`=>`worker`

cli/internal/terraform/terraform/azure/variables.tf
```diff
@@ -1,54 +1,44 @@
...
+    instance_count = optional(number)
+    instance_type  = string
+    disk_size      = number
+    disk_type      = string
+    zones          = optional(list(string))
+  }))
+  description = "A map of node group names to node group configurations."
+  validation {
+    condition     = can([for group in var.node_groups : group.role == "control-plane" || group.role == "worker"])
+    error_message = "The role has to be 'control-plane' or 'worker'."
```
malt3, line 18:
```suggestion
    condition     = can([for group in var.node_groups : group.role == "ControlPlane" || group.role == "Worker"])
    error_message = "The role has to be ControlPlane' or 'Worker'."
```

cli/internal/terraform/terraform/azure/main.tf
```diff
@@ -221,58 +221,47 @@ resource "azurerm_network_security_group" "security_group" {
...
+  instance_type             = each.value.instance_type
+  confidential_vm           = var.confidential_vm
+  secure_boot               = var.secure_boot
+  resource_group            = var.resource_group
   user_assigned_identity    = var.user_assigned_identity
+  image_id                  = var.image_id
   network_security_group_id = azurerm_network_security_group.security_group.id
   subnet_id                 = azurerm_subnet.node_subnet.id
-  backend_address_pool_ids = [
+  backend_address_pool_ids = each.value.role == "control-plane" ? [
```
malt3, line 250:
```suggestion
  backend_address_pool_ids = each.value.role == "ControlPlane" ? [
```

malt3 (commented):
Currently, the interface between go and terraform (terraform.tfvars) uses "ControlPlane" and "Worker" instead of "control-plane" and "worker". The transformation is done in the provider.
This is already the case on other CSPs where the names are expected to have one format in the go code and another when used as labels / tags on resources.

You can either perform the changes as suggested or we strictly specify that terraform.tfvars will use `control-plane` and `worker` as role names and we change the implementation in all terraform templates. Not sure what's easier / cleaner.

cli/internal/cloudcmd/create.go
elchead: I was wondering why we have camelCase if we really need dash notation in TF. Any reasoning beside it being Go convention?

cli/internal/cmd/upgradeapply.go
elchead: We decided to remove targets in AB#3238

malt3 (approved):
LGTM. I'll create a follow up for myself to use TFString / the terraform role names in all terraform templates and also ensure the naming on GCP uses the new format.

PR #1956, opened 2023-06-21
ci: download bootlogs in correct aws region

derpsteb:
### Context
We changed the default region on AWS for our e2e tests since we are using SNP machines on AWS in the past. They are currently only available in `eu-west-1` in Europe. 

### Proposed change(s)
- Update region argument for bootlog download script to be the same region as the cluster is created in.

### Checklist

- [x] Link to Milestone


elchead:
nit: change Context description to:

We need to change the default region on AWS for our e2e tests since we want to use SNP machines, which are currently only available in `eu-west-1` in Europe.

malt3 (approved):

PR #1958, opened 2023-06-22
operators: infrastructure autodiscovery

malt3:
### Context

This change is part of the larger effort to support arbitrary node groups.
When new node groups are added or old node groups are deleted, the cluster needs to learn about those changes in order to correctly control the cluster-autoscaler and perform upgrades safely.
This PR adds two new controllers in order to react to infrastructure changes:

- NodeJoinWacher watches for nodes that join and leave the cluster. It's used to trigger a reconciliation of the other new controller
- ExternalScalingGroupReconciler finds scaling groups in cloud provider infrastructure and dynamically creates / deletes ScaleSet CRs

### Proposed change(s)

- operators: infrastructure autodiscovery
- docs: document scaling configuration when using multiple node groups

### How to test

- create a cluster on GCP
- manually add an additional worker node group in the terraform.tfvars file and `terraform apply` the changes
- wait for new nodes to join
- follow new docs and see if autoscaling works with all worker node groups

### Checklist

- [x] Link to Milestone


operators/constellation-node-operator/internal/infrastructure/infrastructure.go
```diff
@@ -0,0 +1,11 @@
...
+Copyright (c) Edgeless Systems GmbH
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+// Package infrastructure contains a controllers that reconcile on cloud provider infrastructure.
+// It is used to create, delete and update the spec of infrastructure-related k8s resources based on the
+// actual state of the infrastructure.
+// It uses polling (with additional triggers) to check the state of the infrastructure.
+package infrastructure
```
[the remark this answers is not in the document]
malt3: My idea was making it extra clear what we are reconciling on (external resources vs k8s resources). This controller would also interfere with the other envtests, since it creates (and deletes) resources that the other envtests manually create / destroy for testing purposes and are supposed to reconcile on in peace.
Not sure what's the easiest way to resolve this if we were to place them all together.

malt3:
I'll need to refactor the executor / executor_test some more. Moving back to draft for now.

operators/constellation-node-operator/internal/cloud/aws/client/scalinggroup.go
```diff
@@ -153,22 +156,67 @@ func (c *Client) ListScalingGroups(ctx context.Context, uid string) (controlPlan
...
-				} else if *tag.Value == "worker" {
-					workerGroupIDs = append(workerGroupIDs, *group.AutoScalingGroupName)
+			if tag.Key == nil || tag.Value == nil {
+				continue
+			}
+			key := *tag.Key
+			value := strings.ToLower(*tag.Value)
+			switch key {
+			case "constellation-role":
+				if value == "control-plane" {
```
elchead, line 177:
```suggestion
				if value == role.ControlPlane.TFString() {
```

operators/constellation-node-operator/internal/cloud/aws/client/scalinggroup.go
```diff
@@ -153,22 +156,67 @@ func (c *Client) ListScalingGroups(ctx context.Context, uid string) (controlPlan
...
+					role = updatev1alpha1.ControlPlaneRole
+				} else if value == "worker" {
```
elchead, line 179:
```suggestion
				} else if value == role.Worker.TFString() {
```

operators/constellation-node-operator/internal/infrastructure/scalinggroup_controller.go
```diff
@@ -0,0 +1,137 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: strings.ToLower(config.resourceName),
+		},
+		Spec: updatev1alpha1.ScalingGroupSpec{
+			NodeVersion:         mainconstants.NodeVersionResourceName,
+			GroupID:             config.groupID,
+			AutoscalerGroupName: config.autoscalingGroupName,
+			NodeGroupName:       config.nodeGroupName,
+			Min:                 1,
+			Max:                 10,
```
elchead, line 105: should this be configurable / use constant?

malt3: > should this be configurable

This is just a sensible default with some minimun, some maximum and autoscaling disabled (false).
The docs explain how users can configure this after the resource is created.

> use constant

Sounds good

operators/constellation-node-operator/internal/infrastructure/scalinggroup_controller.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+		}
+	}
+
+	logr.Info("external scaling groups reconciled")
+
+	return executor.Result{}, nil
+}
+
+func createScalingGroupIfNotExists(ctx context.Context, config newScalingGroupConfig) error {
```
elchead, line 92: just a nit how we could ease unit testing in the future IMHO. I'd separate the k8s client from the config and restrain the interface to only Create(). 
Could have saved you some trouble writing the complete fake. But would just keep as is now

operators/constellation-node-operator/internal/executor/executor_test.go
```diff
@@ -0,0 +1,221 @@
...
+func TestMain(m *testing.M) {
+	goleak.VerifyTestMain(m)
+}
+
+func TestNew(t *testing.T) {
+	testCases := map[string]struct {
+		config               Config
+		wantPollingFrequency time.Duration
+	}{
+		"default": {
```
elchead, line 29:
```suggestion
		"applies default polling frequency": {
```

operators/constellation-node-operator/internal/executor/executor_test.go
```diff
@@ -0,0 +1,221 @@
...
+			config:               Config{},
+			wantPollingFrequency: defaultPollingFrequency,
+		},
+		"custom": {
```
elchead, line 33:
```suggestion
		"custom polling frequency": {
```

operators/constellation-node-operator/main.go
```diff
@@ -196,6 +219,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	wg := &sync.WaitGroup{}
+	exec.Start(context.Background(), wg)
+	defer wg.Wait()
```
elchead, line 224: avoid deferring correct call order to caller. instead encapsulate inside start() (see suggestion)

operators/constellation-node-operator/main.go
```diff
@@ -196,6 +219,11 @@ func main() {
...
+	defer exec.Stop()
```
elchead, line 225: Stop would make sure to only return when the Starter routines have finished
```suggestion
	waitFn := exec.Start(context.Background())
	defer exec.Stop(waitFn)
```

operators/constellation-node-operator/internal/executor/executor.go
```diff
@@ -0,0 +1,164 @@
...
+			requeueAfter := e.pollingFrequency // default polling frequency
+			if res.Requeue {
+				requeueAfter = res.RequeueAfter // requeue after the given duration
+			}
+
+			nextScheduledReconcile <- requeueAfter
+		}
+	}()
+}
+
```
elchead, line 129:
```suggestion
type waitFn func()

func (e *Executor) Start(ctx context.Context) waitFn {
	wg *sync.WaitGroup

	e.mux.Lock()
	defer e.mux.Unlock()
	if e.started {
		return
	}
	e.started = true

	logr := log.FromContext(ctx)
	e.externalTrigger = make(chan struct{}, 1)
	e.stop = make(chan struct{}, 1)
	// execute is used by the go routines below to communicate
	// that a reconciliation should happen
	execute := make(chan struct{}, 1)
	// nextScheduledReconcile is used to communicate the next scheduled reconciliation time
	nextScheduledReconcile := make(chan time.Duration, 1)
	// trigger a reconciliation on startup
	nextScheduledReconcile <- 0

	wg.Add(2)

	// timer routine is responsible for triggering the reconciliation after the timer expires
	// or when triggered externally
	go func() {
		defer wg.Done()
		defer close(execute)
		defer logr.Info("Timer stopped")
		for {
			nextScheduledReconcileAfter := <-nextScheduledReconcile
			timer := *time.NewTimer(nextScheduledReconcileAfter)
			select {
			case <-e.stop:
				return
			case <-ctx.Done():
				return
			case <-e.externalTrigger:
			case <-timer.C:
			}
			execute <- struct{}{}
		}
	}()

	// executor routine is responsible for executing the reconciliation
	go func() {
		defer wg.Done()
		defer close(nextScheduledReconcile)
		defer logr.Info("Executor stopped")

		for {
			_, ok := <-execute
			// execute channel closed. executor should stop
			if !ok {
				return
			}
			res, err := e.controller.Reconcile(ctx)
			if err != nil {
				logr.Error(err, "reconciliation failed")
				nextScheduledReconcile <- 0 // requeue immediately
				continue
			}
			requeueAfter := e.pollingFrequency // default polling frequency
			if res.Requeue {
				requeueAfter = res.RequeueAfter // requeue after the given duration
			}

			nextScheduledReconcile <- requeueAfter
		}
	}()
	return wg.Wait()
}

```

operators/constellation-node-operator/main.go
```diff
@@ -196,6 +219,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	wg := &sync.WaitGroup{}
```
elchead, line 222: is there a reason to only start the executor here? could it theoretically be possible that NewNodeJoinWatcher triggers the controller before it is started?

operators/constellation-node-operator/internal/executor/executor_test.go
```diff
@@ -0,0 +1,221 @@
...
+	// we cannot assert the exact number of reconciliations here, because the executor might
+	// select the stop case or the timer case first.
+	assertBetween(assert, 10, 11, ctrl.counter)
+}
+
+type stubController struct {
+	stopped        bool
+	stop           chan struct{}
+	reconcileEvent chan struct{}
+	counter        int
```
elchead, line 189:
```suggestion
	reconcilationCounter        int
```

operators/constellation-node-operator/internal/executor/executor_test.go
```diff
@@ -0,0 +1,221 @@
...
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+			exec := New(nil, tc.config)
+			assert.Equal(tc.wantPollingFrequency, exec.pollingFrequency)
+		})
+	}
+}
+
+func TestStart(t *testing.T) {
```
elchead, line 50:
```suggestion
func TestStartTriggersImmediateReconcilation
```

operators/constellation-node-operator/internal/executor/executor_test.go
```diff
@@ -0,0 +1,221 @@
...
+	assert := assert.New(t)
+	ctrl := newStubController(Result{}, nil)
+	exec := Executor{
+		controller:       ctrl,
+		pollingFrequency: time.Hour * 24 * 365, // 1 year. Should be high enough to not trigger the timer in the test.
+	}
+	// on start, the executor should trigger a reconciliation
+	wg := sync.WaitGroup{}
+	exec.Start(context.Background(), &wg)
+	<-ctrl.reconcileEvent
```
elchead, line 60:
```suggestion
	<-ctrl.reconcileEvent // makes sure to wait until reconcile was called (maybe name attribute to .WaitUntilReconciled ?)
```

operators/constellation-node-operator/internal/executor/executor_test.go
```diff
@@ -0,0 +1,221 @@
...
+func (s *stubController) Reconcile(ctx context.Context) (Result, error) {
+	if s.stopped {
+		return Result{}, errors.New("controller stopped")
+	}
+	s.counter++
+	select {
+	case <-s.stop:
+		s.stopped = true
+	case s.reconcileEvent <- struct{}{}:
+	case <-ctx.Done():
```
elchead, line 212: i struggle to understand what this does. could you explain? :)

operators/constellation-node-operator/internal/executor/executor_test.go
```diff
@@ -0,0 +1,221 @@
...
+	exec.Stop()
+	ctrl.stop <- struct{}{}
+
+	wg.Wait()
+
+	// initial trigger
+	assert.Equal(1, ctrl.counter)
+}
+
+func TestError(t *testing.T) {
```
elchead, line 70:
```suggestion
func TestErrorTriggersImmediateReconcilation(t *testing.T) {
```

operators/constellation-node-operator/internal/executor/executor.go
```diff
@@ -0,0 +1,164 @@
...
+		for {
+			_, ok := <-execute
+			// execute channel closed. executor should stop
+			if !ok {
+				return
+			}
+			res, err := e.controller.Reconcile(ctx)
+			if err != nil {
+				logr.Error(err, "reconciliation failed")
+				nextScheduledReconcile <- 0 // requeue immediately
```
elchead, line 117: should we limit the number of failure retries to avoid an infinite retry loop?

operators/constellation-node-operator/internal/executor/executor_test.go
```diff
@@ -0,0 +1,221 @@
...
+	ctrl.stop <- struct{}{}
+
+	wg.Wait()
+
+	// we cannot assert the exact number of reconciliations here, because the executor might
+	// select the stop case or the timer case first.
+	assertBetween(assert, 10, 11, ctrl.counter)
+}
+
+func TestResultRequeue(t *testing.T) {
```
elchead, line 93:
```suggestion
func TestRequeueAfterResultRequeInterval(t *testing.T) {
```

operators/constellation-node-operator/internal/executor/executor_test.go
```diff
@@ -0,0 +1,221 @@
...
+	ctrl.stop <- struct{}{}
+
+	wg.Wait()
+
+	// we cannot assert the exact number of reconciliations here, because the executor might
+	// select the stop case or the timer case first.
+	assertBetween(assert, 10, 11, ctrl.counter)
+}
+
+func TestExternalTrigger(t *testing.T) {
```
elchead, line 119: add a test case for multiple simulataneous triggers.

operators/constellation-node-operator/internal/executor/executor_test.go
```diff
@@ -0,0 +1,221 @@
...
+	// canceling the context should stop the executor without blocking
+	cancel()
+
+	wg.Wait()
+
+	// initial trigger
+	assert.Equal(1, ctrl.counter)
+}
+
+func TestPollingFrequency(t *testing.T) {
```
elchead, line 163:
```suggestion
func TestRequeAfterPollingFrequency(t *testing.T) {
```

operators/constellation-node-operator/internal/executor/executor_test.go
```diff
@@ -0,0 +1,221 @@
...
+func newStubController(result Result, err error) *stubController {
+	return &stubController{
+		reconcileEvent: make(chan struct{}),
+		stop:           make(chan struct{}, 1),
+		result:         result,
+		err:            err,
+	}
+}
+
+func (s *stubController) Reconcile(ctx context.Context) (Result, error) {
```
elchead, line 203:
```suggestion
func (s *stubController) Reconcile(_ context.Context) (Result, error) {
```

operators/constellation-node-operator/internal/cloud/gcp/client/scalinggroup.go
```diff
@@ -145,18 +148,49 @@ func (c *Client) ListScalingGroups(ctx context.Context, uid string) (controlPlan
 
 			groupID, err := c.canonicalInstanceGroupID(ctx, *grpManager.SelfLink)
 			if err != nil {
-				return nil, nil, fmt.Errorf("normalizing instance group ID: %w", err)
+				return nil, fmt.Errorf("normalizing instance group ID: %w", err)
 			}
 
+			var role updatev1alpha1.NodeRole
 			switch strings.ToLower(template.Properties.Labels["constellation-role"]) {
 			case "control-plane", "controlplane":
```
elchead, line 156: should we use the role.Role type throughout?

operators/constellation-node-operator/internal/deploy/deploy_test.go
```diff
@@ -437,24 +358,35 @@ func (g *stubScalingGroupGetter) GetAutoscalingGroupName(scalingGroupID string)
...
-			workerGroupIDs = append(workerGroupIDs, item.groupID)
-		}
+		scalingGroups = append(scalingGroups, cspapi.ScalingGroup{
+			Name:                 item.name,
+			NodeGroupName:        item.nodeGroupName,
+			GroupID:              item.groupID,
+			AutoscalingGroupName: item.autoscalingGroupName,
+			Role: func() updatev1alpha1.NodeRole {
+				if item.isControlPlane {
+					return "ControlPlane"
```
elchead, line 372: role.Role?

operators/constellation-node-operator/internal/cloud/gcp/client/scalinggroup.go
elchead: TODO remove controlplane when migrated?

operators/constellation-node-operator/config/crd/bases/update.edgeless.systems_scalinggroups.yaml
```diff
@@ -56,6 +56,10 @@ spec:
                   (used by cluster-autoscaler).
                 format: int32
                 type: integer
+              nodeGroupName:
```
elchead, line 59: what happens with existing objects without this field? it's just empty?

elchead:
i will test it later on Azure

elchead (commented):
looks good and clean, like the design of the new controllers. I wonder how important it is that the scaling group is up-to-date when performing an upgrade. should we add an annotation when upgrading to trigger a reconcile?
the PR should probably have `no changelog` since the change is not yet user-facing?

operators/constellation-node-operator/internal/infrastructure/infrastructure.go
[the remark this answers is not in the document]
malt3: Will do this late in the review process to make the diff view easier to look at

operators/constellation-node-operator/internal/infrastructure/scalinggroup_controller.go
malt3: Agreed. If we touch this in the future, this would be a good improvement.

malt3:
>  I wonder how important it is that the scaling group is up-to-date when performing an upgrade.

The worst that can happen is that we are unable to make progress with replacing nodes until we discover the new scale set.
This would slow down the upgrade but still reconcile correctly eventually.

operators/constellation-node-operator/config/crd/bases/update.edgeless.systems_scalinggroups.yaml
malt3: Good question. The field would remain empty. I'll need to implement a migration to patch the field in cases where the resource already exists. Thanks!

operators/constellation-node-operator/internal/executor/executor_test.go
malt3: What this would do is wait until either of the events happen:

- a reconcile event is sent via the channel
- the context is cancelled

I needed this in an earlier iteration of the test `TestContextCancel`. I noticed that this is not actually required anymore so I'll remove it for simplicity.

operators/constellation-node-operator/main.go
malt3: No particular reason. That's theoretically possible. I'll move it close to `executor.New`

operators/constellation-node-operator/internal/executor/executor.go
malt3: This uses the same retry logic as kubernetes operators: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler.
While looking into this a bit more, I learned that the controller-runtime offers configurable rate limiting strategies:

- https://github.com/kubernetes-sigs/controller-runtime/blob/a784ee78d04bdf216006adfbc313328a0aded88a/pkg/controller/controller.go#L57-L60
- https://github.com/kubernetes-sigs/controller-runtime/blob/a784ee78d04bdf216006adfbc313328a0aded88a/pkg/ratelimiter/ratelimiter.go#L21-L30
- https://pkg.go.dev/k8s.io/client-go/util/workqueue?utm_source=godoc#RateLimiter
- https://github.com/kubernetes/client-go/blob/master/util/workqueue/default_rate_limiters.go#L77-L83

The default seems to be exponential backoff with a cap of retrying once every 1000 seconds (~16 minutes).

So while I wouldn't want to set a limit, using a rate limiting strategy (preferably exponential backoff) sounds very reasonable. Wdyt?

malt3:
Mostly done with the remarks. Still need to implement some form of backoff / retry limit logic and need to handle the migration if existing scalinggroup CRs with an empty `nodeGroupName` field.

elchead:
changes lgtm, waiting for final changes to test e2e on azure

operators/constellation-node-operator/internal/executor/executor.go
elchead: i thought of exponential backoff as first solution too. If not too hard to implement using the default k8s limiter and adding a check rateLimiter.When(struct{}) before requeueing seems enough right?

elchead:
Tested on Azure. Upscaling works as expected.

malt3:
>  Not sure if AWS should be tested too.

I'll test AWS one more time after #1983 is merged and I can rebase onto it.

3u13r:
There's another (maybe simpler) architecture to achieve that the `ExternalScalinggroupController` reconciles similar to the current arch but without the need for a custom `Executor` implementation and without the `NodeJoinWatcher`. 

The idea boils down to:
* Move the `Watches` of the node join/leave events from the `NodeJoinWaiter` to the `ExternalScalinggroupController` 
* The request for the `ExternalScalinggroupController` is a sentinel token (basically the "tick" of the `Executor`)
* Every `return` statement in the `ExternalScalinggroupController` re-queues the token. On error it is re-queued immediately, if the reconciliation was successful we reconcile after 15 minutes (the current `defaultPollingFrequency`)

malt3:
As discussed I renamed the package to make it clear the the `ExternalScalingGroupReconciler` is not a k8s controller.

operators/constellation-node-operator/internal/executor/executor.go
```diff
@@ -0,0 +1,215 @@
...
+		defer wg.Wait()
+		e.Stop()
+	}
+
+	e.mux.Lock()
+	defer e.mux.Unlock()
+	if e.running {
+		return stopWait
+	}
+	e.running = true
```
3u13r, line 85: The `e.mux` looks to me like it just guards the `e.running` boolean. Should we rename `mux` to `runnningMux` or even use an atomic boolean?

3u13r (changes requested):

operators/constellation-node-operator/internal/executor/executor.go
malt3: Atomic boolean is the clearest

operators/constellation-node-operator/internal/executor/executor.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+}
+
+// Config is the configuration for the executor.
+type Config struct {
+	PollingFrequency time.Duration
+	RateLimiter      RateLimiter
+}
+
+func (c *Config) applyDefaults() {
```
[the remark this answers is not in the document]
malt3: The reason I like this pattern is that the user can use a composite literal to initialize the struct which gives you named fields:

```go
conf := Config{
    PollingFrequency: 10 * time.Second,
    RateLimiter: customratelimiter.New(),
}
```

... and any uninitialized fields automatically get initialized by calling `applyDefaults()` as a consumer of the config.

`New` for config / options structs is a weird in my opinion since you lose the named fields:

```go
conf := New(10 * time.Second, customratelimiter.New())
```

And if you don't call an explicit constructor (`New` / `NewWithDefaults`) as a user, you may have default (nil) initialized interfaces that can lead to panics.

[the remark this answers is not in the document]
3u13r: I see your point. Normally, I'd suggest to simply use a config for `New()` to retain the naming, but we don't want to have configs all the way down, I suppose. 
I like your suggestion for the middle ground.

3u13r (approved):
LGTM

PR #1957, opened 2023-06-22
attestation: revert awsevsnp attestation to nitrotpm

derpsteb:
### Context
AWS currently has a bug in it's SNP implementation where SNP report generation sometimes fails. This keeps nodes from joining Constellation clusters. In order to release AWS SNP support as soon as possible we decided to revert our AWS SNP attestation implementation until the bug is fixed.

### Proposed change(s)

- Revert #1916 . This means specifying `awsSEVSNP` as attestation config spawns SNP machines, but uses the same attestation mechanism as nitroTPM.
- ~remove launchMeasurement from config~

### Additional info
- [Smoketest](https://github.com/edgelesssys/constellation/actions/runs/5344092915) :green_circle: 

### Checklist

~- [x] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)~
- [x] Add labels (e.g., for changelog category)
- [x] Link to Milestone

**EDIT: Moved config changes and docs changes to a second PR. This PR only reverts the SNP attestation. This is done to ensure we can easily do `revert(revert(snp-attestation))` once SNP is fixed on AWS. #1959**

PR #1959, opened 2023-06-22
attestation: docs and config changes for SNP attestation

derpsteb:
This PR consists of two commits originally put for review in #1957. See the description of that PR to learn more.

Above PR needs to be merged first.


derpsteb:
[Smoketest](https://github.com/edgelesssys/constellation/actions/runs/5347313274)

internal/config/config.go
```diff
@@ -358,7 +357,7 @@ func Default() *Config {
 		// AWS uses aws-nitro-tpm as attestation variant
 		// AWS will have aws-sev-snp as attestation variant
 		Attestation: AttestationConfig{
-			AWSSEVSNP:          &AWSSEVSNP{Measurements: measurements.DefaultsFor(cloudprovider.AWS, variant.AWSSEVSNP{}), LaunchMeasurement: measurements.WithAllBytes(0x00, measurements.Enforce, measurements.PCRMeasurementLength)},
+			AWSSEVSNP:          &AWSSEVSNP{Measurements: measurements.DefaultsFor(cloudprovider.AWS, variant.AWSSEVSNP{})},
```
thomasten, line 360: please weaken the "attestation not production ready" warning a bit

docs/docs/getting-started/first-steps.md
```diff
@@ -64,7 +64,13 @@ If you encounter any problem with the following steps, make sure to use the [lat
 
     If you require the OS image to be available in another region, [let us know](https://github.com/edgelesssys/constellation/issues/new?assignees=&labels=&template=feature_request.md&title=Support+new+AWS+image+region:+xx-xxxx-x).
 
-    You can find a list of all [regions in AWS's documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions).
+    You can find a list of all [regions in AWS' documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions).
```
thomasten, line 67: I think it was correct before https://www.google.com/search?q="aws's"+site%3Aamazon.com

docs/docs/getting-started/first-steps.md
```diff
@@ -64,7 +64,13 @@ If you encounter any problem with the following steps, make sure to use the [lat
...
+
+    If you want to use AMD SEV-SNP machines you will have to use on of the following regions:
+     * `eu-west-1`
+     * `us-east-2`
```
thomasten, line 71: use us-east-2a in the iam create commands on this page and on the config page

docs/docs/workflows/config.md
```diff
@@ -64,6 +64,18 @@ By default, Constellation uses `n2d-standard-4` VMs (4 vCPUs, 16 GB RAM) to crea
 
 By default, Constellation uses `m6a.xlarge` VMs (4 vCPUs, 16 GB RAM) to create your cluster. Optionally, you can switch to a different VM type by modifying **instanceType** in the configuration file. Supported are all nitroTPM-enabled machines with a minimum of 4 vCPUs (`xlarge` or larger). Refer to the [list of nitroTPM-enabled instance types](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enable-nitrotpm-prerequisites.html) or run `constellation config instance-types` to get the list of all supported options.
 
+Constellation also support AMD SEV-SNP machines. The list of AWS instance types that support SEV-SNP can be found in the [AWS documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/snp-requirements.html). Please mind the region restrictions mentioned in the [Getting started](../getting-started/first-steps.md#create-a-cluster) section.
+
+To enable AMD SEV-SNP in your cluster you will have to select the `awsSEVSNP` attestation config. You can also use the `--attestation` flag of the [config generate](../reference/cli.md#constellation-config-generate) command.
+
+:::Caution
```
thomasten, line 71:
```suggestion
:::caution
```

docs/docs/workflows/config.md
```diff
@@ -64,6 +64,18 @@ By default, Constellation uses `n2d-standard-4` VMs (4 vCPUs, 16 GB RAM) to crea
...
+Due to a bug in AWS' SNP implementation, SNP report generation may fail in unforeseeable circumstances.
+Therefore, even if you select attestation type `awsSEVSNP`, Constellation still uses NitroTPM-based attestation.
+Nonetheless, memory encryption is enabled.
```
thomasten, line 74:
```suggestion
Nonetheless, runtime encryption is enabled.
```
I think that's the term we want to use user-facing

docs/docs/workflows/config.md
```diff
@@ -64,6 +64,18 @@ By default, Constellation uses `n2d-standard-4` VMs (4 vCPUs, 16 GB RAM) to crea
 
 By default, Constellation uses `m6a.xlarge` VMs (4 vCPUs, 16 GB RAM) to create your cluster. Optionally, you can switch to a different VM type by modifying **instanceType** in the configuration file. Supported are all nitroTPM-enabled machines with a minimum of 4 vCPUs (`xlarge` or larger). Refer to the [list of nitroTPM-enabled instance types](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enable-nitrotpm-prerequisites.html) or run `constellation config instance-types` to get the list of all supported options.
 
+Constellation also support AMD SEV-SNP machines. The list of AWS instance types that support SEV-SNP can be found in the [AWS documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/snp-requirements.html). Please mind the region restrictions mentioned in the [Getting started](../getting-started/first-steps.md#create-a-cluster) section.
+
+To enable AMD SEV-SNP in your cluster you will have to select the `awsSEVSNP` attestation config. You can also use the `--attestation` flag of the [config generate](../reference/cli.md#constellation-config-generate) command.
```
thomasten, line 69: as awsSEVSNP is the default, I think this needs to be rephrased/rearranged

docs/docs/getting-started/first-steps.md
thomasten, line 71: this probably, too

thomasten (changes requested):

docs/docs/getting-started/first-steps.md
derpsteb: I feel like people at AWS aren't sure themself? :D
Taken from your search: 
exhibit A, both versions used: https://sustainability.aboutamazon.com/environment/the-cloud?energyType=true
exhibit B, both versions used: https://aws.amazon.com/security/

I am happy to change it back. My motivation for changing it: "With a plural noun ending in -s, you only need to add an apostrophe to the end of the word to make it possessive." 
https://www.britannica.com/dictionary/eb/qa/How-to-Make-Possessives-in-English

sidenote: while searching "aws possessive form" on google, I found a .docx document on possessive forms on a s3 bucket. unfortunately the document is not by AWS :D.

thomasten: Interesting that they don't use it uniformly. I remembered having seen it as "AWS's" and just did the quick Google check without further validation. Then I actually don't care.

Additional thoughts:
* I don't know if AWS is considered a plural noun. You think because the S means services? And if it would mean service, you'd write AWS's?
* chatgpt says: The possessive form of AWS (Amazon Web Services) would be "AWS's."

derpsteb: Changed it back.

docs/docs/workflows/config.md
```diff
@@ -62,7 +62,22 @@ By default, Constellation uses `n2d-standard-4` VMs (4 vCPUs, 16 GB RAM) to crea
 </tabItem>
 <tabItem value="aws" label="AWS">
 
-By default, Constellation uses `m6a.xlarge` VMs (4 vCPUs, 16 GB RAM) to create your cluster. Optionally, you can switch to a different VM type by modifying **instanceType** in the configuration file. Supported are all nitroTPM-enabled machines with a minimum of 4 vCPUs (`xlarge` or larger). Refer to the [list of nitroTPM-enabled instance types](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enable-nitrotpm-prerequisites.html) or run `constellation config instance-types` to get the list of all supported options.
+By default, Constellation uses `m6a.xlarge` VMs (4 vCPUs, 16 GB RAM) to create your cluster.
+Optionally, you can switch to a different VM type by modifying **instanceType** in the configuration file.
+If you are using the default attestation type `awsSEVSNP` you can use the instance types described in [AWS's AMD SEV-SNP docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/snp-requirements.html).
+Please mind the region restrictions mentioned in the [Getting started](../getting-started/first-steps.md#create-a-cluster) section.
```
thomasten, line 68: not optimal to point from the more reference-like config page to the more guide-like page for further info

docs/docs/workflows/config.md
```diff
@@ -62,7 +62,22 @@ By default, Constellation uses `n2d-standard-4` VMs (4 vCPUs, 16 GB RAM) to crea
...
+
+If you are using the attestation type `awsNitroTPM` you can choose from a larger pool of node types as mentioned in this [list of nitroTPM-enabled instance types](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enable-nitrotpm-prerequisites.html).
```
thomasten, line 70:
```suggestion
If you are using the attestation variant `awsNitroTPM`, you can choose any of the [nitroTPM-enabled instance types](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enable-nitrotpm-prerequisites.html).
```

thomasten (changes requested):

docs/docs/getting-started/first-steps.md
```diff
@@ -50,11 +50,19 @@ If you encounter any problem with the following steps, make sure to use the [lat
...
     ```bash
-    constellation iam create aws --zone=eu-central-1a --prefix=constellTest --generate-config
+    constellation iam create aws --zone=us-east-2a --prefix=constellTest --generate-config
     ```
 
-    This command creates IAM configuration for the AWS zone `eu-central-1a` using the prefix `constellTest` for all named resources being created. It also creates the configuration file `constellation-conf.yaml` in your current directory with the IAM values filled in.
+    This command creates IAM configuration for the AWS zone `us-east-2a` using the prefix `constellTest` for all named resources being created. It also creates the configuration file `constellation-conf.yaml` in your current directory with the IAM values filled in.
 
+    Depending on the attestation type you select during config generation different regions are available:
+    AMD SEV-SNP machines (attestation type set to `awsSEVSNP`) are currently available in the following regions:
```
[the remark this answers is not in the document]
derpsteb: I'd go with the name in the config as that is harder to discover. The help text for the flag prints the possible values.

docs/docs/workflows/config.md
derpsteb: Not sure what to do here. I don't want to copy the information as I assume the information will diverge at some point. But I also want to point the user to the limitations if they aren't aware. 
The workflows section also feels pretty guide-like to me?

thomasten: I'm also not sure what to do here. I thought maybe you would came up with an idea :D
Yes, reference-like is probably not correct. But I would feel more comfortable when the more-intro-level first-steps would reference the more-extensive-level config page. However, I also don't want to need to lookup something in the config workflow to be able to complete the first-steps.
So conclusion is probably to keep it as is.

thomasten (approved):

PR #1960, opened 2023-06-22
terraform: always use uniform role names

malt3:
### Context

terraform.tfvars should use the following role names to refer to roles:

"control-plane", "worker"

Go code uses a `Role` type that has the string constants `ControlPlane` and `Worker`. From now on, the CLI should convert the role names when writing the terraform.tfvars file.

### Proposed change(s)

- terraform: always use uniform role names

### Checklist

- [x] Link to Milestone
- [x] Perform manual tests to ensure nothing breaks


daniel-weisse (approved):
Did not test this in a cluster, but code lgtm

PR #1961, opened 2023-06-22
terraform: QEMU node groups

elchead:
### Context

Similar to #1941.

### Proposed change(s)

- terraform: qemu node groups

### How to test
E2E-test-mini was used to test.

might want to extend testing to as in #1941

### Checklist

- [x] Link to Milestone


elchead:
triggered E2E: https://github.com/edgelesssys/constellation/actions/runs/5354411606

cli/internal/cmd/upgradeapply.go
```diff
@@ -303,6 +303,7 @@ func parseTerraformUpgradeVars(cmd *cobra.Command, conf *config.Config, fetcher
 			Debug:   conf.IsDebugCluster(),
 		}
 		return targets, vars, nil
+	// TODO support for QEMU?
```
daniel-weisse, line 306: To support upgrades on QEMU we would need to implement a lot of custom logic and solutions.
Especially things that run "outside" of the Constellation context, i.e. not a service that runs in the Kubernetes cluster, or as part of the CLI, but some service that runs on the host and interacts with QEMU/libvirt to scale and update resources.

This is something we probably don't want to do, since the effort for implementing and keeping it functional would be a lot higher than the use we get out of it.
QEMU clusters should be used for testing only, for more integrated on-prem solutions,  we support Constellation on OpenStack (although I think OpenStack also does not support upgrades at the current time)

cli/internal/terraform/terraform/qemu/modules/instance_group/main.tf
```diff
@@ -53,24 +56,46 @@ resource "libvirt_domain" "instance_group" {
     type        = "pty"
     target_port = "0"
   }
+  lifecycle {
+    ignore_changes = [
+      name, # required. Allow legacy scale sets to keep their old names
+    ]
+  }
```
daniel-weisse, line 63: Since QEMU doesn't support upgrades, we can drop the backwards compatibility code here.

cli/internal/cmd/upgradeapply.go
elchead: thx for clarifying

elchead:
Naming seems identical to old pattern and firmware is optional, but still there is a permission error to mount. Not sure what it is

elchead:
testing E2E on main: https://github.com/edgelesssys/constellation/actions/runs/5376090677

elchead:
> testing E2E on main: https://github.com/edgelesssys/constellation/actions/runs/5376090677

also fails on main so fixing to version of last green run

elchead:
new run after fixing qemu on main: https://github.com/edgelesssys/constellation/actions/runs/5388202941

cli/internal/terraform/variables.go
```diff
@@ -271,77 +271,77 @@ func (v *OpenStackClusterVariables) String() string {
 	writeLinef(b, "debug = %t", v.Debug)
 
 	return b.String()
+	// f := hclwrite.NewEmptyFile()
+	// gohcl.EncodeIntoBody(v, f.Body())
+	// return string(f.Bytes())
```
daniel-weisse, line 276:
```suggestion
```

e2e/miniconstellation/main.tf
```diff
@@ -135,7 +135,7 @@ resource "azurerm_linux_virtual_machine" "main" {
     publisher = "Canonical"
     offer     = "0001-com-ubuntu-server-jammy-daily"
     sku       = "22_04-daily-lts"
-    version   = "latest"
+    version   = "22.04.202305020" // latest versions have permission error : Error: error creating libvirt domain: internal error: Process exited prior to exec: libvirt: QEMU Driver error : failed to umount devfs on /dev: Permission denied
```
daniel-weisse, line 138: Is this still true with the downgrade qemu version?

elchead: no it's not, good catch. i'm reverting it

cli/internal/terraform/terraform/qemu/main.tf
```diff
@@ -139,3 +119,13 @@ locals {
...
+
+moved {
+  from = module.control_plane
+  to   = module.instance_group["control_plane_default"]
+}
+
+moved {
+  from = module.worker
+  to   = module.instance_group["worker_default"]
+}
```
malt3, line 131:
```suggestion
```

can be removed since we never upgrade existing QEMU clusters

cli/internal/terraform/terraform/qemu/modules/instance_group/main.tf
```diff
@@ -6,10 +6,9 @@ terraform {
     }
   }
 }
-
 resource "libvirt_domain" "instance_group" {
-  name     = "${var.name}-${var.role}-${count.index}"
   count    = var.amount
+  name     = "${var.base_name}-${var.role}-${count.index}"
```
malt3, line 11:
```suggestion
  name     = "${var.base_name}-${var.role}-${local.group_uid}-${count.index}"
```

name should probably also have a uid suffix (similar to GCP and Azure) so different groups use different VM names

cli/internal/terraform/terraform/qemu/modules/instance_group/main.tf
```diff
@@ -56,20 +55,29 @@ resource "libvirt_domain" "instance_group" {
 }
 
 resource "libvirt_volume" "boot_volume" {
-  name           = "constellation-${var.role}-${count.index}-boot"
   count          = var.amount
+  name           = "constellation-${var.role}-${count.index}-boot"
```
malt3, line 59: See above

cli/internal/terraform/terraform/qemu/variables.tf
```diff
@@ -1,3 +1,20 @@
+variable "node_groups" {
+  type = map(object({
+    role           = string
+    instance_count = number // number of instances in the node group
+    disk_size      = number // size of state disk (GiB)
+    vcpus          = number
+    memory         = number // amount of memory per instance (MiB)
+    machine        = string // machine type. use 'q35' for secure boot and 'pc' for non secure boot. See 'qemu-system-x86_64 -machine help'
```
malt3, line 8:
```suggestion
```

machine has to be uniform for all node groups and should remain a global option (instead of being chosen per group).

elchead:
running another E2E test with the requested changes

elchead:
@malt3 lgty? E2E passed after feedback inclusion

malt3 (approved):

PR #1962, opened 2023-06-23
ci: parse ovmf binaries from metadata

derpsteb:
### Context
AWS SNP attestation should verify the launch measurement from the SNP report. The launch measurement should be configurable by the user through the constellation config. As the launch measurement differs depending on the number of vCPUs of a machine we would have to provide multiple launch measurements in the config, one for each possible VM size. To avoid this we will soon start using [this](https://github.com/edgelesssys/sev-snp-measure-go) library to pre-calculate launchmeasurements during runtime. The input for that calculation is produced as part of the pipeline changes in this PR.

### Proposed change(s)
- Build sev-snp-measure-go.
- Run equivalence test between sev-snp-measure-go and it's [upstream](https://github.com/IBM/sev-snp-measure).
- Generate MetadataWrapper object.

### Additional info
- Saving the files to the API will be added in a second PR.
- [Testrun](https://github.com/edgelesssys/constellation/actions/runs/5353628310/jobs/9709705054) :green_circle: 

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/aws-snp-launchmeasurement.yml
+++ b/.github/workflows/aws-snp-launchmeasurement.yml
@@ -54,6 +54,12 @@ jobs:
         ref: main
         path: sev-snp-measure
 
+    - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # tag=v3.5.2
+      with:
+        repository: edgelesssys/sev-snp-measure-go.git
+        ref: main
+        path: sev-snp-measure-go
+
     - name: Run sev-snp-measure
       shell: bash
       run: |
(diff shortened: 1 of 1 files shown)


derpsteb:
Converting to draft again. Please don't review yet. Still have to check if we can hardcode the produced metadata.

derpsteb:
Seems like we can't hardcode the values: https://github.com/virtee/sev-snp-measure/issues/27

.github/workflows/aws-snp-launchmeasurement.yml
```diff
@@ -66,3 +72,21 @@ jobs:
         done
 
         jq < measurements.json
+        popd || exit 1
+
+    - name: Test equivalence sevsnpmeasure & sev-snp-measure
+      shell: bash
+      run: |
+        pushd sev-snp-measure-go/e2e || exit 1
+        go test --expected-values ../../sev-snp-measure/measurements.json --ovmf ${{ steps.build-uefi.outputs.ovmfPath }}
```
daniel-weisse, line 81: mostly a nit, and affects the sev-snp-measure repo, not this PR:
I would prefer having this as an actual command, instead of an e2e test

derpsteb: Because constellation users might want to verify? Or what motivation would there be for such a command?
Like this `sev-snp-measure compare measurements.json ovmf.fd`?

derpsteb: For visibility: we decided to move test to sev-snp-measure-go repo.

daniel-weisse (approved):

PR #1964, opened 2023-06-23
helm: enable volume snapshot support if CSI drivers are deployed to the cluster

daniel-weisse:
### Context

Kubernetes supports [Volume Snapshots](https://kubernetes.io/docs/concepts/storage/volume-snapshots/) for CSI drivers, with the feature being GA since Kubernetes v1.20.
The original deployment files and source code for the snapshotter can be found on [GitHub](https://github.com/kubernetes-csi/external-snapshotter).
To make use of this feature installation of special snapshot CRDs and a snapshot controller are required.
Creating the snapshots and restoring PVCs is then handled by the individual CSI driver implementations.

Currently, if a user wants to use pvc snapshotting, they will have to manually install the CRDs and deploy the controller.

### Proposed change(s)
- Add helm charts for the snapshot CRDs and for the snapshot controller (+ validation webhook)
- Install the snapshot CRDs and deploy the controller on `constellation init`

### Additional info
Quick e2e tests to verify deployment works (The snapshot feature was additionally manually tested on both Azure and GCP):

* Azure: https://github.com/edgelesssys/constellation/actions/runs/5357424147
* GCP: https://github.com/edgelesssys/constellation/actions/runs/5357427037

### Checklist

- [x] Link to Milestone


daniel-weisse:
The first two commits are form this PR alone.
The last 3 were mostly already reviewed as part of https://github.com/edgelesssys/constellation/pull/1973, but I would appreciate another look at them before merging

cli/internal/helm/README.md
```diff
@@ -9,15 +35,24 @@ Because upgrades should be a CLI-only operation and we want to avoid the behavio
...
+* CRDs are updated by cilium-operator.
+
+### cert-manager
+
+* installCRDs flag is set during upgrade. This flag is managed by cert-manager. cert-manager is in charge of correctly upgrading the CRDs.
+* WARNING: upgrading cert-manager might break other installations of cert-manager in the cluster, if those other installation are not on the same version as the Constellation-manager installation. This is due to the cluster-wide CRDs.
+
+### Operators
+
+* Manually update CRDs before upgrading the chart. Update by running applying the CRDs found in the `operators/crds/` folder.
```
derpsteb, line 49:
```suggestion
* Manually update CRDs before upgrading the chart. Update by applying the CRDs found in the `operators/crds/` folder.
```

cli/internal/helm/loader.go
```diff
@@ -644,3 +474,18 @@ func loadChartsDir(efs embed.FS, dir string) (*chart.Chart, error) {
 
 	return loader.LoadFiles(files)
 }
+
+func controlPlaneTolerations() []map[string]any {
```
3u13r, line 478: Could this be a variable? I think this could even be scoped to just the function where it is used currently (if there are no plans to re-use this in the near future of course).

cli/internal/helm/loader.go
```diff
@@ -146,21 +155,18 @@ func (i *ChartLoader) loadRelease(info chartInfo, helmWaitMode helm.WaitMode) (h
 
 	switch info.releaseName {
 	case ciliumInfo.releaseName:
-		values, err = i.loadCiliumValues()
+		values = ciliumVals[i.csp.String()]
```
3u13r, line 158: If we decide to remove the check for non-existent entries should we add a check that the map contains values for each `cloudprovider.Provider` value? I worry about e.g. foregtting to add a new CSP to the cilium values and not having a clear error output.

3u13r (changes requested):
Just 2 minor things

cli/internal/helm/loader.go
daniel-weisse: Changed to variable and moved to values.go

3u13r (changes requested):

3u13r (approved):
LGTM

PR #1975, opened 2023-06-27
config: improve usage and meaning of validate

elchead:
### Context

We want to make sure that a consumer of the config can rely on it being in a valid state (as verified in `config.Validate`).
Currently, the usage is not consistent and better documentation should discourage mutation of a config object without subsequent validation.

### Proposed change(s)

- discourage use of config.Default + mutation
- discuss violating usage in miniup

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/miniup.go
+++ b/cli/internal/cmd/miniup.go
@@ -143,7 +143,7 @@ func (m *miniUpCmd) prepareConfig(cmd *cobra.Command, fileHandler file.Handler, 
 			return nil, errors.New("not overwriting existing config")
 		}
 	}
-	if featureset.CurrentEdition != featureset.EditionEnterprise {
+	if !featureset.CanUseEmbeddedMeasurmentsAndImage {
 		cmd.PrintErrln("Generating a valid default config is not supported in the OSS build of the Constellation CLI. Consult the documentation for instructions on where to download the enterprise version.")
 		return nil, errors.New("cannot create a mini cluster without a config file in the OSS build")
 	}
--- a/cli/internal/cmd/miniup.go
+++ b/cli/internal/cmd/miniup.go
@@ -143,15 +143,16 @@ func (m *miniUpCmd) prepareConfig(cmd *cobra.Command, fileHandler file.Handler, 
 		}
 	}
 
-	config := config.Default()
-	config.Name = constants.MiniConstellationUID
-	config.RemoveProviderAndAttestationExcept(cloudprovider.QEMU)
-	config.StateDiskSizeGB = 8
-
+	// TODO: Should we allow this at all?  It seems more logical to follow the usage pattern for the other providers: constellation config generate mini -> constellation init. But maybe it's more user friendly to not persist a config file in the workspace for a testing cluster?
+	config := config.MiniDefault()
 	// only release images (e.g. v2.7.0) use the production NVRAM
 	if !config.IsReleaseImage() {
 		config.Provider.QEMU.NVRAM = "testing"
 	}
+	// TODO: this validation would fail since no image nor measurements are set; we create a cluster with what we call invalid?
+	//if err := config.Validate(flags.force); err != nil {
+	//	return nil, fmt.Errorf("validating config: %w", err)
+	//}
 
 	m.log.Debugf("Prepared configuration")
 
--- a/cli/internal/cmd/miniup.go
+++ b/cli/internal/cmd/miniup.go
@@ -13,6 +13,7 @@ import (
 	"net"
 
 	"github.com/edgelesssys/constellation/v2/cli/internal/cloudcmd"
+	"github.com/edgelesssys/constellation/v2/cli/internal/featureset"
 	"github.com/edgelesssys/constellation/v2/cli/internal/libvirt"
 	"github.com/edgelesssys/constellation/v2/cli/internal/terraform"
 	"github.com/edgelesssys/constellation/v2/internal/api/attestationconfigapi"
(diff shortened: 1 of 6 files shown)


cli/internal/cmd/miniup.go
```diff
@@ -143,15 +143,16 @@ func (m *miniUpCmd) prepareConfig(cmd *cobra.Command, fileHandler file.Handler,
 		}
 	}
 
-	config := config.Default()
-	config.Name = constants.MiniConstellationUID
-	config.RemoveProviderAndAttestationExcept(cloudprovider.QEMU)
-	config.StateDiskSizeGB = 8
-
+	// TODO: Should we allow this at all?  It seems more logical to follow the usage pattern for the other providers: constellation config generate mini -> constellation init. But maybe it's more user friendly to not persist a config file in the workspace for a testing cluster?
+	config := config.MiniDefault()
```
daniel-weisse, line 147: The mini constellation command was created to explicitly skip the whole config-generate -> create -> init flow and provide one single command to handle all of this.

cli/internal/cmd/miniup.go
```diff
@@ -143,15 +143,16 @@ func (m *miniUpCmd) prepareConfig(cmd *cobra.Command, fileHandler file.Handler,
...
 	// only release images (e.g. v2.7.0) use the production NVRAM
 	if !config.IsReleaseImage() {
 		config.Provider.QEMU.NVRAM = "testing"
 	}
+	// TODO: this validation would fail since no image nor measurements are set; we create a cluster with what we call invalid?
+	//if err := config.Validate(flags.force); err != nil {
+	//	return nil, fmt.Errorf("validating config: %w", err)
+	//}
```
daniel-weisse, line 155: The config created for mini constellation should be valid (at least when executing the command on a release CLI), also I'm not sure if it makes sense to validate the config we just created from a default function (it should be valid).
To fix the problems here, maybe disable this command (or the option to run it without passing an explicit config path) if the CLI was built with the wrong build-flags similar to how its done in [fetch-measurements](https://github.com/edgelesssys/constellation/blob/main/cli/internal/cmd/configfetchmeasurements.go#L85)?

msanft (commented):
LGTM except for what Daniel already mentioned

cli/internal/cmd/miniup.go
```diff
@@ -142,16 +143,18 @@ func (m *miniUpCmd) prepareConfig(cmd *cobra.Command, fileHandler file.Handler,
 			return nil, errors.New("not overwriting existing config")
 		}
 	}
-
-	config := config.Default()
-	config.Name = constants.MiniConstellationUID
-	config.RemoveProviderAndAttestationExcept(cloudprovider.QEMU)
-	config.StateDiskSizeGB = 8
-
+	if featureset.CurrentEdition != featureset.EditionEnterprise { // TODO: or !featureset.CanFetchMeasurements? image is also needed..
```
daniel-weisse, line 146: The measurements for the image should already be embedded in the CLI.
Checking for the edition should be enough.
Alternatively you could add another feature flag, e.g. `CanMiniUp` if you want to explicitly control access to this command.

cli/internal/cmd/miniup.go
```diff
@@ -142,16 +143,18 @@ func (m *miniUpCmd) prepareConfig(cmd *cobra.Command, fileHandler file.Handler,
...
+		return nil, errors.New("cannot create a mini cluster without a config file in the OSS build")
+	}
+	config := config.MiniDefault()
 	// only release images (e.g. v2.7.0) use the production NVRAM
 	if !config.IsReleaseImage() {
 		config.Provider.QEMU.NVRAM = "testing"
 	}
+	if err := config.Validate(flags.force); err != nil {
+		return nil, fmt.Errorf("validating config: %w", err)
+	}
```
daniel-weisse, line 157: Maybe move both of these checks into `config.MiniDefault()`?
We probably always want to conditionally check what NVRAM to use, and we also always want to make sure the config we just generated is valid (which should be the case if it was generated with an enterprise CLI).

cli/internal/cmd/miniup.go
malt3: If we want this nice message, I agree that it should use a separate feature `CanMiniUp` or `CanMini`

malt3 (approved):
otherwise LGTM

cli/internal/cmd/miniup.go
elchead: I don't like `canMini` because it implies that the command doesn't work, when it really only means that it doesn't work without config flag. Why is Enterprise not good enough? I think it won't change that generating a complete config is only supported on Enterprise?

malt3: I would like to remove the actual edition from the features if possible.
How about: `CanUseEmbeddedMeasurements` or something similar? That sounds like we may be able to reuse it in other places in a meaningful way.

internal/config/config_test.go
```diff
@@ -376,6 +376,15 @@ func TestValidate(t *testing.T) {
 				return cnf
 			}(),
 		},
+		"miniup default config is not valid because image and measurements are missing in OSS": {
+			cnf: func() *Config {
+				config, err := MiniDefault()
+				require.Error(t, err)
+				return config
```
daniel-weisse, line 383: nit: checking for an error here is redundant since this test is designed to check for validation errors anyway.
What would make more sense is to check if the config returned by `MiniDefault` is not `nil`

daniel-weisse (approved):

cli/internal/cmd/miniup.go
elchead: agree with not coupling edition and a specific feature. changing it

PR #1976, opened 2023-06-27
terraform: aws node groups

malt3:
### Context

Refactor terraform code for AWS provider to go from two node groups (1x control plane, 1x worker) to arbitrary node groups.
On AWS, this is especially hard to do, since most network resources are zone-scoped, which means that we have to duplicate the networking resource for every zone.
This is handled differently for two categories of resources:

- Cheap resources are created for every zone in the chosen region. This ensures we can depend on those resources for loadbalancing and can add / remove node groups in every region without downtime.
- Expensive resources (nat gateways) are only created in a region when there is at least one node group in that region.

Additionally, some terraform migrations have to be performed by the cli directly. [Terraform does not yet support moving resources where the target name is dynamic](https://github.com/hashicorp/terraform/issues/31335), so moving `aws_eip.lb` -> `aws_eip[var.zone]` is not possible with `moved` blocks.

#### Future migrations

The bootstrapper needs a fix to support loadbalancers in multiple regions. This fix is included in this PR. We need to wait one release so that every bootstrapper in existing clusters can handle multiple loadbalancers before we can add every zone of the regions as a target zone of the loadbalancer. The terraform code has a TODO that explains this as well.
A final issue I still need to solve is that the upgrade where we actually want to enable multiple zones needs to recreate the loadbalancer.
This will lead to a short time where the kubernetes API is unreachable. This is especially problematic since it leads to failed upgrade steps after the terraform migration whenever a connection to the cluster is used.

### Proposed change(s)

- terraform: aws node groups
- manual terraform state transitions for AWS

### How to test (upgrade)

- Create an AWS cluster on v2.8
- Upgrade using the CLI on this branch
- You should see some new resources and replacement of the loadbalancer
- Otherwise, nothing should be replaced (most important: the existing scale sets stay intact)

### How to test (multi-zone)

- Create an AWS **debug** cluster using the CLI on this branch
- Manually add a node group to the terraform.tfvars file
- run `terraform apply` in the constellation-terraform folder
- check that the new node group can join the cluster

### Checklist

- [x] Link to Milestone
- [x] Make AWS metadata code work with multiple AZs

### TODOs for future PRs

- [ ] Enable multi-zone support (including a change to the LB to target every availability zone of the chosen region)
- [ ] Work around control plane being unreachable for a short time when enabling multi-zone LB

### DIFF
--- a/cli/internal/cloudcmd/create.go
+++ b/cli/internal/cloudcmd/create.go
@@ -130,16 +130,27 @@ func (c *Creator) Create(ctx context.Context, opts CreateOptions) (clusterid.Fil
 
 func (c *Creator) createAWS(ctx context.Context, cl terraformClient, opts CreateOptions) (idFile clusterid.File, retErr error) {
 	vars := terraform.AWSClusterVariables{
-		CommonVariables: terraform.CommonVariables{
-			Name:               opts.Config.Name,
-			CountControlPlanes: opts.ControlPlaneCount,
-			CountWorkers:       opts.WorkerCount,
-			StateDiskSizeGB:    opts.Config.StateDiskSizeGB,
+		Name: opts.Config.Name,
+		NodeGroups: map[string]terraform.AWSNodeGroup{
+			"control_plane_default": {
+				Role:            role.ControlPlane.TFString(),
+				StateDiskSizeGB: opts.Config.StateDiskSizeGB,
+				InitialCount:    opts.ControlPlaneCount,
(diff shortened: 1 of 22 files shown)


elchead:
the enable_cpu config also solves AB#3235 right?
I.e. if not SNP -> the value will always remain nil
if SNP -> always true

elchead:
tested creation: new node group joins as expected

lgtm

elchead (approved):

PR #1977, opened 2023-06-28
bootstraper: delete helm chart on installation failure before retrying installation

daniel-weisse:
### Context

When installing helm charts during `constellation init`, the bootstrapper will retry an installation on error.
Because of how the retry loop and installation is set up, the chart is not cleaned up if the error didn't occur immediately, but later during the installation, e.g. a startup check fails after 3 minutes into the installation.
This in turn will cause an error when the installation is retried because the name of the installation is still in use.

For example, the installation of cert-manager may fail due to timeouts during the start up check.
The retry will then fail with the following error:

```text
installing cert-manager: helm install: cannot re-use a name that is still in use
```

### Proposed change(s)

- Install all charts with the `Atomic` flag set
  -  This causes the chart to be uninstalled on failure, e.g. install context cancelled or start up checks failed
- Move manual setting of the `Wait` flag from the various `InstallXXX` functions to the constructor of the helm client, since it was set for all functions
- Log installation failures in the retrier

### DIFF
--- a/bootstrapper/internal/helm/helm.go
+++ b/bootstrapper/internal/helm/helm.go
@@ -214,7 +214,7 @@ func (i installDoer) Do(ctx context.Context) error {
 	i.log.With(zap.String("chart", i.chart.Name())).Infof("Trying to install Helm chart")
 
 	if _, err := i.client.RunWithContext(ctx, i.chart, i.values); err != nil {
-		i.log.With(zap.Error(err)).Errorf("Helm chart installation failed")
+		i.log.With(zap.Error(err), zap.String("chart", i.chart.Name())).Errorf("Helm chart installation failed")
 		return err
 	}
 
--- a/bootstrapper/internal/helm/helm.go
+++ b/bootstrapper/internal/helm/helm.go
@@ -57,6 +57,8 @@ func New(log *logger.Logger) (*Client, error) {
 	action := action.NewInstall(actionConfig)
 	action.Namespace = constants.HelmNamespace
 	action.Timeout = timeout
+	action.Atomic = true
+	action.Wait = true
 
 	return &Client{
 		action,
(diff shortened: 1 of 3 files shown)


bootstrapper/internal/helm/helm.go
```diff
@@ -67,7 +69,6 @@ func New(log *logger.Logger) (*Client, error) {
 // InstallConstellationServices installs the constellation-services chart. In the future this chart should bundle all microservices.
 func (h *Client) InstallConstellationServices(ctx context.Context, release helm.Release, extraVals map[string]any) error {
 	h.ReleaseName = release.ReleaseName
-	h.Wait = release.Wait
```
thomasten, line 70: Wasn't this false before as set here?
https://github.com/edgelesssys/constellation/blob/9dae948f9d84af337abfb08bd0dde603f839dfe6/cli/internal/helm/loader.go#L171

bootstrapper/internal/helm/helm.go
```diff
@@ -215,7 +213,10 @@ type installDoer struct {
 func (i installDoer) Do(ctx context.Context) error {
 	i.log.With(zap.String("chart", i.chart.Name())).Infof("Trying to install Helm chart")
 
-	_, err := i.client.RunWithContext(ctx, i.chart, i.values)
+	if _, err := i.client.RunWithContext(ctx, i.chart, i.values); err != nil {
+		i.log.With(zap.Error(err)).Errorf("Helm chart installation failed")
```
thomasten, line 217: Should this log message also have the chart name as tag?

daniel-weisse: Good idea, will add

bootstrapper/internal/helm/helm.go
thomasten: Ok if you want to keep the wait, you should either remove the `Wait` field from the `helm.Release` struct or revert your change and set `Wait: true` in loader.go.

thomasten (approved):

derpsteb:
Slightly touching title & description to make the fixed bug more obvious when reading changelog.

PR #1985, opened 2023-06-29
rfc: canonical endpoint / custom dns

malt3:
### Context

This is a required change to enable true multi-zone support for AWS. It also has other benefits discussed in the RFC.
As for any RFC, feel free to comment on any aspect including:

- soundness of the general idea
- edge cases that are not reflected
- improvements
- clarifications

If you feel like this RFC can be implemented in the way it is phrased, approve it.
I'll leave it open for a few days.

### Proposed change(s)

- rfc: canonical endpoint / custom dns

### Checklist

- [x] Link to Milestone

### DIFF
--- a/rfc/canonical-endpoint-dns.md
+++ b/rfc/canonical-endpoint-dns.md
@@ -1,0 +1,76 @@
+# Canonical endpoint / Constellation with custom DNS name
+
+Customers may want to use a DNS name for the cluster endpoint instead of the public ip address.
+The public ip may be bound to a zonal loadbalancer (AWS), limiting the availability of the cluster to a single zone.
+Additionally, the public ip address may change if the ip is freed and re-allocated accidentally.
+For branding or organizational reasons, customers may want to use a custom DNS name instead of a bare ip address or csp-provided hostname.
+Additionally, a dual-stack setup with IPv4 and IPv6 can be achieved more easily with a DNS name.
+
+## State of loadbalancing / DNS on the relevant CSPs
+
+### AWS
+
+On AWS, each loadbalancer [has a public dns name](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-internet-facing-load-balancers.html) in the format `name-1234567890.region.elb.amazonaws.com`.
+This hostname can be used as a default endpoint for Constellation for interactions with the bootstrapper, cluster joining and the kubernetes API.
+
+Additionally, customers may register custom domains and point them to the loadbalancer hostname.
+
+### Azure
+
+On Azure, each public ip [can have a public dns name](https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/public-ip-addresses#dns-name-label) in the format `name-1234567890.location.cloudapp.azure.com`.
+This hostname can be used as a default endpoint for Constellation for interactions with the bootstrapper, cluster joining and the kubernetes API.
+
+Additionally, customers may register custom domains and point them to either the public ip hostname or address.
+
+### GCP
+
+On Google Cloud, neither loadbalancers nor public ips have a default, public dns name.
+
+We have to use the (single) public ipv4 address as the default endpoint for Constellation for interactions with the bootstrapper, cluster joining and the kubernetes API.
+
+Additionally, customers may register custom domains and point them to the public ip address.
+
+### Conclusion
+
+**Fallback endpoint**:
+
+- loadbalancer dns name on AWS (`name-1234567890.region.elb.amazonaws.com`)
+- public ip dns name on Azure (`name-1234567890.location.cloudapp.azure.com`)
+- public ip on GCP (`203.0.113.1`)
+
+**(Optional) customer provided dns**:
+
+- any custom domain that points to the fallback endpoint
+
+## Steps for enabling DNS (aka: "canonical endpoint")
+
+1. start adding the fallback endpoint (and the custom domain, if set) to the SAN field of the kube-apiserver certificates
+  - SAN should still contain everything it contains before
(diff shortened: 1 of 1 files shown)


rfc/canonical-endpoint-dns.md
```diff
@@ -0,0 +1,76 @@
...
+  - optional customer provided hostname (if set)
+2. once every apiserver uses the extended SAN field, the cluster-internal advertise-address for the apiserver can be switched to the fallback endpoint
+  - no-op on GCP (this is already the status quo)
+  - ensure every existing kubelet uses the new endpoint
+  - ensure newly joining kubelet use the new endpoint
+  - k8s configmap uses new endpoint (`kube-system/kubeadm-config` -> `.data.ClusterConfiguration`)
+3. switch id file `constellation-id.json` to use either fallback endpoint or customer endpoint (Q: which is better if both are available?)
+4. switch setting in kubernetes config (`constellation-admin.conf`) to use either fallback endpoint or customer endpoint (Q: which is better if both are available?)
+  - in existing clusters: patch config
+  - in new clusters: write correct endpoint on `constellation init`
```
daniel-weisse, line 62:
```suggestion
1. start adding the fallback endpoint (and the custom domain, if set) to the SAN field of the kube-apiserver certificates
    - SAN should still contain everything it contains before
      - legacy public ip we use before switching to dns
      - `kubernetes`, `kubernetes.default`, `kubernetes.default.svc`, `kubernetes.default.svc.cluster.local`
      - 127.0.0.1, ::1, localhost
    - fallback endpoint (as specified above)
    - optional customer provided hostname (if set)
2. once every apiserver uses the extended SAN field, the cluster-internal advertise-address for the apiserver can be switched to the fallback endpoint
    - no-op on GCP (this is already the status quo)
    - ensure every existing kubelet uses the new endpoint
    - ensure newly joining kubelet use the new endpoint
    - k8s configmap uses new endpoint (`kube-system/kubeadm-config` -> `.data.ClusterConfiguration`)
3. switch id file `constellation-id.json` to use either fallback endpoint or customer endpoint (Q: which is better if both are available?)
4. switch setting in kubernetes config (`constellation-admin.conf`) to use either fallback endpoint or customer endpoint (Q: which is better if both are available?)
    - in existing clusters: patch config
    - in new clusters: write correct endpoint on `constellation init`
```
nit: you need to add extra indentation, otherwise the markdown rendering of the nested lists is broken (see GitHub preview of the file)

daniel-weisse: As for the question about whether to use fallback or custom endpoints for kube-config/id-file, I would suggest using the fallback, as this should be considered the "safer" endpoint.
As this is not user controlled, it should be more resilient to user error as well.
It is unlikely a CSP accidentally removes the default DNS name of a loadbalancer, while I can see a user doing so.

rfc/canonical-endpoint-dns.md
```diff
@@ -0,0 +1,76 @@
...
+Additionally, a dual-stack setup with IPv4 and IPv6 can be achieved more easily with a DNS name.
+
+## State of loadbalancing / DNS on the relevant CSPs
+
+### AWS
+
+On AWS, each loadbalancer [has a public dns name](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-internet-facing-load-balancers.html) in the format `name-1234567890.region.elb.amazonaws.com`.
+This hostname can be used as a default endpoint for Constellation for interactions with the bootstrapper, cluster joining and the kubernetes API.
+
+Additionally, customers may register custom domains and point them to the loadbalancer hostname.
```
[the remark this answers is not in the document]
malt3: AWS will resolve the dns name to **all** ip addresses when using this name:

> The DNS servers resolve the DNS name of your load balancer to the public IP addresses of the load balancer nodes for your load balancer.

This is the result of using the AWS loadbalancer DNS name:

```mermaid
flowchart TD
    subgraph dns
        record["A name-1234567890.region.elb.amazonaws.com 203.0.113.1,203.0.113.2,203.0.113.3"]
    end
    subgraph global
        lb["lb (distributes)"]
    end
    subgraph zoneafront[eu-central-1a]
        fronta["frontend (203.0.113.1)"]-->lb
    end
    subgraph zonebfront[eu-central-1b]
        frontb["frontend (203.0.113.2)"]-->lb
    end
    subgraph zonecfront[eu-central-1c]
        frontc["frontend (203.0.113.3)"]-->lb
    end
    subgraph zoneaback[eu-central-1a]
        subneta["subnet (192.168.178.0/24)"]
    end
    subgraph zonebback[eu-central-1b]
        subnetb["subnet (192.168.179.0/24)"]
    end
    subgraph zonecback[eu-central-1c]
        subnetc["subnet (192.168.180.0/24)"]
    end
    record-->fronta
    record-->frontb
    record-->frontc
    lb-->subneta
    lb-->subnetb
    lb-->subnetc
```

3u13r:
We should also try to sketch the steps for migrating an existing cluster more explicitly. From what I see we need the following steps:
 1. One-time upgrade migration: discover and add the fallback endpoint to the `kubeadm-config`
 2. During the upgrade each control-plane node will be replaced by a node with the new config
 3. The advertise address is a node-specific parameter which will be set inside the `JoinCluster` function
 4. One-time upgrade migration: after a successful upgrade we can also replace the endpoint values in the `constellation-id.json` and `constellation-admin.conf` of the current workspace

Note that step 2 happens automatically and step 3 is part of the general implementation of this feature and not migration specific.

3u13r (approved):
LGTM

daniel-weisse (approved):

PR #1991, opened 2023-06-30
ci: update aws asg resource selector

msanft:
### Context

The introduction of node groups also brought a naming change (i.e. change of explicit module instances to a map of an arbitrary amount of module instances) which broke the seleciton of the control plane and worker instance group name in the boot log collection. 

### Proposed change(s)

- Fix the boot log collection resource selectors.

### Checklist

- [x] Link to Milestone


malt3 (approved):
LGTM except for this not being tidy.

katexochen:
@msanft I couldn't reproduce any conflict between shellcheck and shfmt. I've applied the changes that tidy suggests and the CI passes without problems, see https://github.com/edgelesssys/constellation/actions/runs/5454750258.

PR #1992, opened 2023-06-30
helm: fix upgrade command skipping all service upgrades

daniel-weisse:
### Context
`constellation upgrade apply` should go over a list of all available charts and only skip a single chart if it is already up to date. Other charts should still be upgraded if one doesn't need to.

### Proposed change(s)

- correctly use `errors.As` to not skip other service upgrades during `upgrade apply`
- correctly use `errors.As` to not return an error (only print it) if the node version is the same as the upgrade target

### Additional info
- should be merged before v2.9 to not break upgrades

### DIFF
--- a/cli/internal/helm/client.go
+++ b/cli/internal/helm/client.go
@@ -85,10 +85,6 @@ func (c *Client) shouldUpgrade(releaseName, newVersion string, force bool) error
 
 	// This may break for cert-manager or cilium if we decide to upgrade more than one minor version at a time.
 	// Leaving it as is since it is not clear to me what kind of sanity check we could do.
-	if currentVersion == newVersion {
-		return NoUpgradeRequiredError{}
-	}
-
 	if !force {
 		if err := compatibility.IsValidUpgrade(currentVersion, newVersion); err != nil {
 			return err
@@ -105,20 +101,12 @@ func (c *Client) shouldUpgrade(releaseName, newVersion string, force bool) error
 	return nil
 }
 
-// NoUpgradeRequiredError is returned if the current version is the same as the target version.
-type NoUpgradeRequiredError struct{}
-
-func (e NoUpgradeRequiredError) Error() string {
-	return "no upgrade required since current version is the same as the target version"
-}
-
 // Upgrade runs a helm-upgrade on all deployments that are managed via Helm.
 // If the CLI receives an interrupt signal it will cancel the context.
 // Canceling the context will prompt helm to abort and roll back the ongoing upgrade.
 func (c *Client) Upgrade(ctx context.Context, config *config.Config, timeout time.Duration, allowDestructive, force bool, upgradeID string) error {
 	upgradeErrs := []error{}
 	upgradeReleases := []*chart.Chart{}
-	invalidUpgrade := &compatibility.InvalidUpgradeError{}
 
 	for _, info := range []chartInfo{ciliumInfo, certManagerInfo, constellationOperatorsInfo, constellationServicesInfo} {
 		chart, err := loadChartsDir(helmFS, info.path)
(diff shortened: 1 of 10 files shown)


cli/internal/helm/client.go
```diff
@@ -85,10 +85,6 @@ func (c *Client) shouldUpgrade(releaseName, newVersion string, force bool) error
 
 	// This may break for cert-manager or cilium if we decide to upgrade more than one minor version at a time.
 	// Leaving it as is since it is not clear to me what kind of sanity check we could do.
-	if currentVersion == newVersion {
```
elchead, line 88: This was the comment to skip helm upgrade on the same version: 
https://github.com/edgelesssys/constellation/pull/1940#issuecomment-1597132318

internal/staticupload/staticupload_test.go
```diff
@@ -108,12 +108,13 @@ func TestUpload(t *testing.T) {
...
-				assert.ErrorAs(err, &InvalidationError{})
+				assert.ErrorAs(err, &invalidationErr)
 				return
 			}
 			if tc.wantErr {
-				assert.False(errors.As(err, &InvalidationError{}))
+				assert.False(errors.As(err, &invalidationErr))
 				assert.Error(err)
 				return
 			}
```
elchead, line 120:
```suggestion
			var invalidationErr *InvalidationError
			if tc.wantCacheInvalidationErr {
				assert.ErrorIs(err, &invalidationErr)
				return
			}
			if tc.wantErr {
				assert.ErrorIs(err, &invalidationErr))
				assert.Error(err)
				return
			}
```

elchead: ok we have it in several places replicated.. i guess it doesn't matter

elchead:
lets enforce error ptr types everywhere:
```
// Unwrap returns the inner error, which is nil in this case.
func (e *InvalidUpgradeError) Unwrap() error {
	return e.innerErr
}

// Error returns the String representation of this error.
func (e *InvalidUpgradeError) Error() string {
	return fmt.Sprintf("upgrading from %s to %s is not a valid upgrade: %s", e.from, e.to, e.innerErr)
}
```

```
func (e *UnsupportedAppRegistrationError) Error() string {
	return "Azure app registrations are not supported since v2.9. Migrate to using a user assigned managed identity by following the migration guide: https://docs.edgeless.systems/constellation/reference/migration.\nPlease remove it from your config and from the Kubernetes secret in your running cluster. Ensure that the UAMI has all required permissions."
}
```

internal/staticupload/staticupload_test.go
daniel-weisse: I'm not fully understanding the suggestion?
The current code makes sure that if `tc.wantErr` is true, but `tc.wantCacheInvalidationErr` is false, then we check that the returned error **is not** of type `*InvalidationError`, your suggestion would change that to always checking that it is

cli/internal/helm/client.go
daniel-weisse: @malt3 could you elaborate on why you want to skip upgrades to the same version even when the force flag is set?
Seems like a valid use case to me to actually force an upgrade, no matter what

elchead:
> > lets enforce error ptr types everywhere
> 
> I think I caught all of the cases where basic structs were used. Not sure if we can enforce this with something from

I think this was the only one left I saw. I couldn't find anything quickly to enforce it automatically, but I think for now it's good if all developers now it and keep an eye during review :)

elchead:
lgtm once maltes point is discussed

cli/internal/helm/client.go
malt3: The logic we are implementing is actually just converging to a desired state.
If actual == desired, we should not do anything.
The helm upgrade has side effects (like writing a huge folder with backups of custom resources).
The helm client itself likely doesn't do anything if actual == desired, so it has no useful side effects either.

daniel-weisse: This however is assuming that localVersionNumber == clusterVersionNumber is equal to desiredChart == actualChart, which might for some reason not be true.
If I am using a flag that tells me it is "disabling version compatibility checks", I would assume it actually disables all of them, which we would not do in this case (skip the upgrade if versions are equal)

malt3: Sounds like a reasonable request. I'm fine with always running the helm upgrade if force is set.

elchead (approved):

malt3 (approved):

PR #1993, opened 2023-06-30
bootstrapper: skip waiting for helm deployments

malt3:
### Context

Not all helm charts we deploy can be deployed successfully with only one control-plane node and no workers.
Disable waiting for helm deployments to be ready until we fixed the individual deployments.

### Proposed change(s)

- bootstrapper: skip waiting for helm deployments

### Checklist

- [x] Link to Milestone

### DIFF
--- a/bootstrapper/internal/helm/helm.go
+++ b/bootstrapper/internal/helm/helm.go
@@ -57,8 +57,10 @@ func New(log *logger.Logger) (*Client, error) {
 	action := action.NewInstall(actionConfig)
 	action.Namespace = constants.HelmNamespace
 	action.Timeout = timeout
-	action.Atomic = true
-	action.Wait = true
+	// TODO(malt3): Enable "Wait" once every service gets ready with only
+	// one control-plane node and no worker nodes.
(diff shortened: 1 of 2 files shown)


bootstrapper/internal/helm/helm.go
```diff
@@ -57,8 +57,10 @@ func New(log *logger.Logger) (*Client, error) {
 	action := action.NewInstall(actionConfig)
 	action.Namespace = constants.HelmNamespace
 	action.Timeout = timeout
-	action.Atomic = true
-	action.Wait = true
+	// TODO(malt3): Enable "Wait" once every service gets ready with only
+	// one control-plane node and no worker nodes.
+	action.Atomic = false
+	action.Wait = false
```
thomasten, line 63: does Atomic also have to be reverted?

malt3: Yes. If you follow the usage of Wait into helms source code, you'll find that somewhere there is a statement like `Wait = Wait | Atomic`. I'm on mobile but will update this with a link to the source.

PR #1999, opened 2023-07-03
docs: how to set up MiniConstellation on Azure

elchead:
### Context

The goal is to make the first use of Constellation easier for people who DON'T already have a Linux machine.
The easiest setup is MiniConstellation since it does not involve any creation of IAM resources or resource quotas to spin up a full cluster.

As of now MiniConstellation is only supported on Linux. We cannot expect everyone to use Linux locally, especially product managers or other higher level people.

We don't document AWS, since their supported bare-metal machines are very big and expensive.

### Proposed change(s)

- document how to set up Azure to run MiniConstellation

### Related issue
- [AB#3201](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3201)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/docs/docs/getting-started/first-steps-local.md
+++ b/docs/docs/getting-started/first-steps-local.md
@@ -8,100 +8,7 @@ You have two options:
 
 Both options use virtualization to create a local cluster with control-plane nodes and worker nodes. They **don't** require hardware with Confidential VM (CVM) support. For attestation, they currently use a software-based vTPM provided by KVM/QEMU.
 
-## Quick setup
-
-### Prerequisites
-
-* [Install Terraform](https://developer.hashicorp.com/terraform/downloads)
-* [Install the Azure CLI](https://learn.microsoft.com/de-de/cli/azure/install-azure-cli)
-  to authenticate with Azure:
-
-  ```sh
-  az login
-  ```
-
-### Instructions
-
-Through the Terraform template for Azure it's easy to set up a MiniConstellation cluster on a remote VM.
-
-1. Clone the Constellation repository:
-
-    ```sh
-    git clone https://github.com/edgelesssys/constellation.git
-    ```
-
-2. Set up the remote Azure VM through Terraform:
-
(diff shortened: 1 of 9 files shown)


docs/docs/getting-started/first-steps-local.md
```diff
@@ -2,32 +2,101 @@
...
+* machine requirements:
   * An x86-64 CPU with at least 4 cores (6 cores are recommended)
   * At least 4 GB RAM (6 GB are recommended)
   * 20 GB of free disk space
-  * Hardware virtualization enabled in the BIOS/UEFI (often referred to as Intel VT-x or AMD-V/SVM)
-  * `iptables` rules configured to not drop forwarded packages.
-    If running the following command returns no error, please follow [the troubleshooting guide](#vms-have-no-internet-access):
+  * Hardware virtualization enabled in the BIOS/UEFI (often referred to as Intel VT-x or AMD-V/SVM) / nested-virtualization support when using a VM
+* OS / library requirements:
+  * recommended: Ubuntu 22.04 LTS
```
m1ghtym0, line 72: Are those requirements below automatically installed with Ubuntu 22.04?

m1ghtym0 (commented):
Thanks for building all that Terraform tooling:-)
I'm unsure if `setup-miniconstellation.sh` should be hidden behind a script.
Users might have the experience of executing a bunch of scripts but don't really know what's going on.
Ultimately, it's probably all about interacting with Constellation so it might be fine.

docs/docs/getting-started/first-steps-local.md
elchead: I agree it's confusing. Maybe we just limit ourselves to supporting Ubuntu and list all packages that are required (see dev-docs/miniconstellation/azure-terraform/cloud-init.yaml)?

elchead: I agree that the part about using constellation in `setup-miniconstellation.sh` should not be hidden. So I put that in the  first-steps and just refer to the script as a quick way?

elchead:
When you follow the Terraform route, let me know if you had to start the docker service manually, since I had to manually start it on one test machine

elchead:
Etel and I agreed that it doesn't really make sense to add the quick setup to the official docs. Not even costwise does it make sense to spin up a mini cluster in the cloud vs just spawning a regular cloud cluster. We keep the setup in dev-docs though for whenever someone might still want to use it.

m1ghtym0 (approved):

PR #2000, opened 2023-07-04
ci: use strict semver for gcp guest agent image

katexochen:
### Context

Renovate bot isn't picking up updates of the gcp-guest-agent container image for some time.

### Proposed change(s)

- Making the version a valid semver should help renovate to handle it

### Checklist

### DIFF
--- a/.github/workflows/build-gcp-guest-agent.yml
+++ b/.github/workflows/build-gcp-guest-agent.yml
@@ -30,7 +30,7 @@ jobs:
           beforeDot="${semver%%.*}"
           afterDot="${semver#*.}"
           afterDotEvaluated=$((afterDot))
-          semver="$beforeDot.$afterDotEvaluated"
+          semver="v${beforeDot}.${afterDotEvaluated}.0"
           echo "Semver tag of guest-agent is $semver"
           echo "latest=$semver" | tee -a "$GITHUB_OUTPUT"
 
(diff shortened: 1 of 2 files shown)


katexochen:
I've retagged the old image with the new format and used that tag in versions.go to validate that renovate can indeed update to the newer one.

.github/workflows/build-gcp-guest-agent.yml
```diff
@@ -100,7 +100,7 @@ jobs:
       - name: Log in to the Container registry
         id: docker-login
         if: steps.needs-build.outputs.out == 'true'
-        uses: ./.github/actions/container_registry_login
+        uses: ./constellation/.github/actions/container_registry_login
```
3u13r, line 103: I don't get this change. We use the old version everywhere else.

katexochen: We're checking out side-by-side with guest-agent source, see https://github.com/edgelesssys/constellation/blob/feat/ci/gcpguestagent/.github/workflows/build-gcp-guest-agent.yml#L82

3u13r: Ah I see, thanks for the pointer.

katexochen: Bug introduced in https://github.com/edgelesssys/constellation/commit/025d34a259cff3a0fca8d219ffab3b3ef7abbe8c

3u13r (approved):
LGTM

PR #2011, opened 2023-07-04
docs: explain the role of PCR[10] and why it is not reproducible

malt3:
### Context

Explain why PCR[10] is not reproducible and what components are measured into PCR[10] (Linux IMA).

### Proposed change(s)

- docs: explain the role of PCR[10] and why it is not reproducible

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/docs/docs/architecture/attestation.md
+++ b/docs/docs/architecture/attestation.md
@@ -144,7 +144,7 @@ The latter means that the value can be generated offline and compared to the one
 | 7           | Secure Boot State                                                | Azure, Constellation Bootloader | No                          |
 | 8           | -                                                                | -                               | -                           |
 | 9           | initramfs                                                        | Linux Kernel                    | Yes                         |
-| 10          | Reserved                                                         | -                               | No                          |
+| 10          | Userspace[^1]                                                    | Linux IMA[^1]                   | No[^1]                      |
 | 11          | Reserved for Unified Kernel Image components                     | (Constellation Bootloader)      | Yes                         |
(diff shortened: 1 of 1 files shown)


docs/docs/architecture/attestation.md
```diff
@@ -308,3 +308,9 @@ flowchart LR
   D["Public key"]-- "verifies" -->E["Runtime measurements"]
   E["Runtime measurements"]-- "verify" -->F["Constellation cluster"]
 ```
+
+## References
+
+[^1]: Linux IMA produces runtime measurements of user space binaries.
+However, these measurements aren't deterministic and thus, PCR[10] can't be compared to a constant value.
```
thomasten, line 315:
```suggestion
However, these measurements aren't deterministic and thus, PCR\[10] can't be compared to a constant value.
```
nit: escape opening bracket

docs/docs/architecture/attestation.md
```diff
@@ -144,7 +144,7 @@ The latter means that the value can be generated offline and compared to the one
 | 7           | Secure Boot State                                                | Azure, Constellation Bootloader | No                          |
 | 8           | -                                                                | -                               | -                           |
 | 9           | initramfs                                                        | Linux Kernel                    | Yes                         |
-| 10          | Reserved                                                         | -                               | No                          |
+| 10          | Userspace[^1]                                                    | Linux IMA[^1]                   | No[^1]                      |
```
thomasten, line 147: nit: I have a slight preference for adding the reference only in the last column. But keep it as is if you prefer.

thomasten (approved):

PR #2061, opened 2023-07-07
cli: add "--skip-helm-wait" flag

malt3:
### Context

This flag can be used to disable the atomic and wait flags during helm install. This is useful when debugging a failing `constellation init`, since the user gains access to the cluster even if one of the deployments is never in a ready state.

### Proposed change(s)

- cli: add "--skip-helm-wait" flag
- bootstrapper: fix deadlock during helm install

### Additional info
I tested this manually.
You can see the difference by either creating  a broken deployment on purpose or by measuring the time difference between skipping and waiting.
The skip flags makes `constellation init` a lot faster.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/bootstrapper/internal/helm/helm.go
+++ b/bootstrapper/internal/helm/helm.go
@@ -69,6 +69,9 @@ func New(log *logger.Logger) (*Client, error) {
 // InstallConstellationServices installs the constellation-services chart. In the future this chart should bundle all microservices.
 func (h *Client) InstallConstellationServices(ctx context.Context, release helm.Release, extraVals map[string]any) error {
 	h.ReleaseName = release.ReleaseName
+	if err := h.setWaitMode(release.WaitMode); err != nil {
+		return err
+	}
 
 	mergedVals := helm.MergeMaps(release.Values, extraVals)
 
@@ -79,6 +82,9 @@ func (h *Client) InstallConstellationServices(ctx context.Context, release helm.
 func (h *Client) InstallCertManager(ctx context.Context, release helm.Release) error {
 	h.ReleaseName = release.ReleaseName
 	h.Timeout = 10 * time.Minute
+	if err := h.setWaitMode(release.WaitMode); err != nil {
+		return err
+	}
 
 	return h.install(ctx, release.Chart, release.Values)
 }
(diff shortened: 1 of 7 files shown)


internal/deploy/helm/helm.go
```diff
@@ -12,6 +12,7 @@ type Release struct {
 	Chart       []byte
 	Values      map[string]any
 	ReleaseName string
+	WaitMode    WaitMode
```
thomasten, line 15: I can't see this being set anywhere

cli/internal/helm/loader.go
```diff
@@ -101,7 +101,7 @@ func NewLoader(csp cloudprovider.Provider, k8sVersion versions.ValidK8sVersion)
 }
 
 // Load the embedded helm charts.
-func (i *ChartLoader) Load(config *config.Config, conformanceMode bool, masterSecret, salt []byte) ([]byte, error) {
+func (i *ChartLoader) Load(config *config.Config, conformanceMode bool, helmWaitMode helm.WaitMode, masterSecret, salt []byte) ([]byte, error) {
```
thomasten, line 104: helmWaitMode is unused

thomasten (changes requested):
A crucial part of this change seems to be missing.

bootstrapper/internal/helm/helm.go
daniel-weisse, line 1: Would it make sense to also disable the retrier loop if `WaitMode` is anything other than `Atomic`?
Otherwise broken deployments may just result in an error for the user instead of gaining cluster access  because the retrier will try to install an already existing chart. (Retry loop doesn't clean up failed installations without the atomic flag)

malt3: I'll perform some more tests to ensure this works correctly

malt3:
I found an additional bug in helm: letting the ctx we pass into [`install.RunWithContext`](https://pkg.go.dev/helm.sh/helm/v3/pkg/action#Install.RunWithContext) expire actually leads to a deadlock sometimes.
[This issues seems to be known upstream.](https://github.com/helm/helm/issues/11971)
To counteract this, I decided to limit the number of retries instead of setting a global 10 minute timeout.

daniel-weisse (approved):

thomasten (approved):

PR #2069, opened 2023-07-07
docs: suggest changes for first-steps-local

thomasten:
### Proposed change(s)

- Add a sentence about the option to use a VM. It's a bit redundant to the prerequisites, but I think it's better to have it more prominent.
- Minor reordering of the preqrequisites
- Install commands
  - Add some sudos (I guess you copied that from the cloud-init ;)
  - Simplify some commands
  - Remove start of docker service. I think this works reliably when you install it on a shell. But I can re-add it if you'd prefer that.
- Other small changes


elchead:
improvement on #1999 , lgtm

elchead (approved):

PR #2079, opened 2023-07-10
cli: remove old migrations

derpsteb:
### Context
Depending on the changes in a release we need migration code during upgrades to change existing cluster resources, etc. Since we force users to upgrade one minor version at a time we can remove old migration code. This PR does that.

### Proposed change(s)
- Remove migrations that are not required for upgrading from 2.8.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/helm/client.go
+++ b/cli/internal/helm/client.go
@@ -8,13 +8,11 @@ package helm
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
 	"time"
 
-	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
 	"github.com/edgelesssys/constellation/v2/internal/compatibility"
 	"github.com/edgelesssys/constellation/v2/internal/config"
 	"github.com/edgelesssys/constellation/v2/internal/constants"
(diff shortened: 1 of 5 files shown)


cli/internal/helm/client.go
```diff
@@ -327,49 +325,19 @@ func (c *Client) applyMigrations(ctx context.Context, releaseName string, values
 	}
 
 	if currentV.Major == 2 && currentV.Minor == 7 {
```
daniel-weisse, line 327:
```suggestion
	if currentV.Major == 2 && currentV.Minor == 8 {
```
This branch should run if the current cluster is running v2.8

cli/internal/helm/client.go
```diff
@@ -327,49 +325,19 @@ func (c *Client) applyMigrations(ctx context.Context, releaseName string, values
...
-		return fmt.Errorf("marshalling attestation config: %w", err)
-	}
-	joinServiceVals["attestationConfig"] = string(attestationConfigJSON)
-
+// migrateFrom2_9 is currently a no-op that is kept for documentation purposes.
+// If you have to implement the function please make sure to update the below comment to your situation.
+// migrateFrom2_9 applies the necessary migrations for upgrading from v2.9.x to v2.10.x.
+// migrateFrom2_9 should be applied for v2.9.x --> v2.10.x.
+// migrateFrom2_9 should NOT be applied for v2.9.0 --> v2.9.x.
+func migrateFrom2_9(ctx context.Context, values map[string]any, conf *config.Config, kubeclient crdClient) error {
```
daniel-weisse, line 340:
```suggestion
// migrateFrom2_8 is currently a no-op that is kept for documentation purposes.
// If you have to implement the function please make sure to update the below comment to your situation.
// migrateFrom2_8 applies the necessary migrations for upgrading from v2.8.x to v2.9.x.
// migrateFrom2_8 should be applied for v2.8.x --> v2.9.x.
// migrateFrom2_8 should NOT be applied for v2.9.0 --> v2.9.x.
func migrateFrom2_8(ctx context.Context, values map[string]any, conf *config.Config, kubeclient crdClient) error {
```
Since the new release is v2.9, and this function migrates from the previous release, it should be named `migrateFrom2_8`

daniel-weisse (approved):

PR #2080, opened 2023-07-10
ci: logcollection to OpenSearch in non-debug clusters

msanft:
### Context
Previously, Filebeat and Logstash (our utilities for logcollection to OpenSearch) were _only_ deployed through the `debugd`, which only got deployed in debug-clusters. We can also deploy Filebeat and Logstash through Helm as a Kubernetes deployment, allowing us to also deploy it in non-debug clusters. This should also be helpful in resolving [this bug](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3227), as we are able to retrieve systemd-logs _after_ the `debugd` runs. (i.e. when Kubernetes runs)

### Proposed change(s)

- Create a hack-script `hack/logcollector` which retrieves and templates the configuration files into the format that the Elastic Helm charts need. This seems to be very complex for a small problem, but it will allow us to keep a consistent log format between non-debug and debug cluster logs, as we only specify the output format in `debugd/{filebeat,logstash}` and don't have duplication.
- Create a workflow which deploys Filebeat and Logstash in e2e runs. (temporarily deploy without any condition, i.e. also in debug clusters, to resolve the above mentioned bug)

### TODO

- ~Write documentation on how to deploy in non-CI clusters~
- ~Use a "real" CI condition for the deployment rather than deploying to all clusters~ TODO added in code, for now deploy both in all runs to resolve ["missing OpenSearch logs"-bug](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3227) - during this time, this will lead to duplicated logs in non-buggy debug clusters.
- Potentially do modifications to filebeat config if https://discuss.elastic.co/t/filebeat-k8s-deployment-duplicated-filestream-id/338008 turns out not to be a bug
- ~Test CI container build of debugd-{filebeat,logstash}~ [Workflow run](https://github.com/edgelesssys/constellation/actions/runs/5528589110/jobs/10085559696)
- ~Possibly extend metadata-fields as viable (e.g. csp)~

### Checklist

- [x] Link to Milestone


elchead:
sry for the late review. having a look today. did you already run an E2E test to test if logs are uploaded on OS?

msanft:
> sry for the late review. having a look today. did you already run an E2E test to test if logs are uploaded on OS?

Yes, they are :-)

elchead:
@msanft you can rebase to fix govulncheck

elchead:
nit: remove "Previously, " from description. Otherwise nice description :)

hack/logcollector/cmd/template.go
```diff
@@ -0,0 +1,116 @@
...
+
+	templateCmd.Flags().String("dir", "", "Directory to place the templated configurations in (required)")
+	templateCmd.MarkFlagRequired("dir")
+	templateCmd.MarkFlagDirname("dir")
+	templateCmd.Flags().String("username", "", "OpenSearch username (required)")
+	templateCmd.MarkFlagRequired("username")
+	templateCmd.Flags().String("password", "", "OpenSearch password (required)")
+	templateCmd.MarkFlagRequired("password")
+	templateCmd.Flags().Int("port", 5045, "Logstash port")
+	templateCmd.Flags().StringToString("info", nil, "Additional fields for the Logstash pipeline in the format --info key1=value1,key2=value2,...")
```
elchead, line 37: can you explain a little more what Logstash pipeline refers to? I guess these are metadata attributes (like tags) attached to the log entries?

hack/logcollector/main.go
```diff
@@ -0,0 +1,13 @@
...
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+package main
+
+import "github.com/edgelesssys/constellation/v2/hack/logcollector/cmd"
+
+func main() {
+	cmd.Execute()
+}
```
elchead, line 13: Even though it's a hack, can you write a very short README.md with purpose and example usage?

I could imagine it being useful even during local development and it would be nice to have a quick reference how to deploy the charts (kind of like in the GH action)

hack/logcollector/cmd/template.go
msanft: Exactly, the fields supplied here will be added as "tags" to the log entries. See https://www.elastic.co/guide/en/logstash/current/configuration.html and [us using them in the CI to set GitHub metadata](https://github.com/edgelesssys/constellation/pull/2080/files#diff-30c1e389a577e5a166b88e4dfd7b0c611f3305bf9bb9994ac2a79c5079d54978R25-R37)

hack/logcollector/main.go
msanft: I added docs [here](https://github.com/edgelesssys/constellation/pull/2080/files#diff-a2e3245686df61d5cc57d3bdabb74eb4c4fde1000f868577d6aef52ccf08469b)

dev-docs/workflows/logcollection.md
```diff
@@ -0,0 +1,53 @@
...
+        ...
+    ```
+
+    This will place the templated configuration in the current directory.
+    OpenSearch user credentials can be created by any admin in OpenSearch. Logging in with your company CSP accounts should grant you sufficient permissions to [create a user](https://opensearch.org/docs/latest/security/access-control/users-roles/#create-users) and [grant him the required `all_access` role](https://opensearch.org/docs/latest/security/access-control/users-roles/#map-users-to-roles).
+    One can add additional key-value pairs to the configuration by appending `--info key=value` to the command. These key-value pairs will be attached to the log entries and can be used to filter them in OpenSearch.
+2. Deploy Logstash
+
+    ```bash
+    cd Logstash
```
elchead, line 31:
```suggestion
    # assuming at repo root
    cd hack/logcollector/internal/templates/logstash/
```

dev-docs/workflows/logcollection.md
```diff
@@ -0,0 +1,53 @@
...
+    make add
+    make install
+    cd ..
+    ```
+
+    This will add the required Logstash Helm charts and deploy them to your cluster.
+3. Deploy Filebeat
+
+    ```bash
+    cd Filebeat
```
elchead, line 41:
```suggestion
    # assuming at repo root
    cd hack/logcollector/internal/templates/filebeat/
```

dev-docs/workflows/logcollection.md
```diff
@@ -0,0 +1,53 @@
...
+1. Template the deployment configuration through the `loco` CLI.
+
+    ```bash
+    bazel run //hack/logcollector template -- \
+        --dir="." --username="<OPENSEARCH_USERNAME>" \
+        --password="<OPENSEARCH_PW>" \
+        --info key=value \
+        --info key2=value2
+        ...
+    ```
```
elchead, line 23:
```suggestion
    bazel run //hack/logcollector template -- --dir="." --username="<OPENSEARCH_USERNAME>" --password="<OPENSEARCH_PW>"
    --info test=MY_TESTNAME
```
add a tag to easily find own logs

dev-docs/workflows/logcollection.md
msanft: This would head into the templates directory, which is not correct. We want to get into the directory where the templated things get placed. If the command from step 1 has been executed (`--dir="."`), `cd logstash` is correct

dev-docs/workflows/logcollection.md
msanft: same as above

dev-docs/workflows/logcollection.md
elchead: yeah just `logstash` now its working

elchead:
```suggestion
    cd logstash
```

.github/actions/e2e_test/action.yml
```diff
@@ -241,6 +241,16 @@ runs:
         existingConfig: ${{ steps.constellation-iam-create.outputs.existingConfig }}
         artifactNameSuffix: ${{ steps.create-prefix.outputs.prefix }}
 
+    - name: Deploy logcollection
+      id: deploy-logcollection
+      # TODO(msanft):temporarily deploy in debug clusters too to resolve "missing logs"-bug
+      # see https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3227
+      # if: inputs.isDebugImage == 'false'
```
elchead, line 248: the reason for keeping the old debug log collection is to collect them even before the cluster is running?
I.e. logstash only collects new events that it sees and not prior ones? If this was the case, how could this help us see how the debug filebeat container failed?

debugd/internal/debugd/logcollector/logcollector.go
```diff
@@ -99,6 +100,20 @@ func NewStartTrigger(ctx context.Context, wg *sync.WaitGroup, provider cloudprov
...
+			logger.Infof("Getting logstash config template")
+			tmpl, err = getFilebeatTemplate(ctx, logger)
+			if err != nil {
+				logger.Errorf("Getting filebeat config template: %v", err)
+				return
+			}
+			filebeatConf := filebeatConfInput{
+				LogstashHost: "localhost:5044",
+			}
+			if err := writeFilebeatConf(tmpl, filebeatConf); err != nil {
```
elchead, line 112: can you include in the proposed changes why we add filebeat here?

elchead: could you test that the changes in the debugd logcollector are working?

elchead:
otherwise, looks good to me. Very clean work :)

.github/actions/e2e_test/action.yml
msanft: Filebeat collects systemd-logs (which should include why the podman container from debugd fails) - this will help us resolve the bug, hopefully

debugd/internal/debugd/logcollector/logcollector.go
msanft: > could you test that the changes in the debugd logcollector are working?

yes, doing so already

msanft: > can you include in the proposed changes why we add filebeat here?

what do you mean?

msanft:
Running the [daily e2e test](https://github.com/edgelesssys/constellation/actions/runs/5531041571), which should deploy both the pre-bootstrapper and K8s logcollection, I noticed that systemd tags (e.g. `systemd-unit`) seem to be missing. This is very weird, as the pre-bootstrapper logcollection (podman container) still used the [same image as before](https://github.com/edgelesssys/constellation/blob/e536ff6df19de44469ad32e9a01cfe6d55662717/internal/versions/versions.go#L115-L118). Even though, there are systemd logs existing:
image

Furthermore, I didn't find duplicated logs, which theoretically should be the case, since Filebeat should also collect logs that were made prior to its deployment. This leads me to believe that the pre-bootstrapper logcollection didn't get deployed correctly, but the e2e logs don't give me any indication of an error. This might be the missing-logs problem that we already have a ticket for, but it would be a weird coincidence.

I don't know if or how the missing tags are related to the missing duplicated logs. If anyone has an idea, feel free to let me know

debugd/internal/debugd/logcollector/logcollector.go
elchead: > > can you include in the proposed changes why we add filebeat here?
> 
> what do you mean?

nvm, saw that we deployed filebeat equally before

.github/actions/e2e_test/action.yml
msanft: Yes. Filebeat also collects logs made prior to its deployment

msanft:
[e2e run with new container images
](https://github.com/edgelesssys/constellation/actions/runs/5539741769)[corresponding logs](https://search-e2e-logs-y46renozy42lcojbvrt3qq7csm.eu-central-1.es.amazonaws.com/_dashboards/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-1d,to:now))&_a=(columns:!(message),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'74517cf0-6442-11ed-acf1-47dda8fdfbbb',key:metadata.github.run-id,negate:!f,params:(query:'5539741769'),type:phrase),query:(match_phrase:(metadata.github.run-id:'5539741769')))),index:'74517cf0-6442-11ed-acf1-47dda8fdfbbb',interval:auto,query:(language:kuery,query:''),sort:!(!('@timestamp',asc))))

elchead:
> [e2e run with new container images ](https://github.com/edgelesssys/constellation/actions/runs/5539741769)[corresponding logs](https://search-e2e-logs-y46renozy42lcojbvrt3qq7csm.eu-central-1.es.amazonaws.com/_dashboards/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-1d,to:now))&_a=(columns:!(message),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'74517cf0-6442-11ed-acf1-47dda8fdfbbb',key:metadata.github.run-id,negate:!f,params:(query:'5539741769'),type:phrase),query:(match_phrase:(metadata.github.run-id:'5539741769')))),index:'74517cf0-6442-11ed-acf1-47dda8fdfbbb',interval:auto,query:(language:kuery,query:''),sort:!(!('@timestamp',asc))))

maybe it's useful to add the jobname to the metadata and not just the workflow name? Makes it easier to filter

msanft:
> > [e2e run with new container images ](https://github.com/edgelesssys/constellation/actions/runs/5539741769)[corresponding logs](https://search-e2e-logs-y46renozy42lcojbvrt3qq7csm.eu-central-1.es.amazonaws.com/_dashboards/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-1d,to:now))&_a=(columns:!(message),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'74517cf0-6442-11ed-acf1-47dda8fdfbbb',key:metadata.github.run-id,negate:!f,params:(query:'5539741769'),type:phrase),query:(match_phrase:(metadata.github.run-id:'5539741769')))),index:'74517cf0-6442-11ed-acf1-47dda8fdfbbb',interval:auto,query:(language:kuery,query:''),sort:!(!('@timestamp',asc))))
> 
> maybe it's useful to add the jobname to the metadata and not just the workflow name? Makes it easier to filter

It's unfortunately not exposed by GitHub: https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

.github/actions/e2e_test/action.yml
```diff
@@ -241,6 +241,16 @@ runs:
...
+      uses: ./.github/actions/deploy_logcollection
+      with:
```
elchead, line 250: can we inject the test name here and insert it as input?

msanft:
Turning this back into a draft until missing journald logs are resolved

.github/actions/deploy_logcollection/action.yml
```diff
@@ -0,0 +1,66 @@
...
+         --info provider="${{ inputs.provider }}" \
+         --info deployment-type="k8s"
+
+    - name: Deploy Logstash
+      id: deploy-logstash
+      shell: bash
+      env:
+        KUBECONFIG: ${{ inputs.kubeconfig }}
+      run: |
+        cd logstash
```
katexochen, line 54: use `step.workdir` instead.

.github/actions/deploy_logcollection/action.yml
```diff
@@ -0,0 +1,66 @@
...
+    - name: Template Logcollection Helm Values
+      id: template
+      shell: bash
+      run: |
+        bazel run //hack/logcollector template -- \
+         --dir $(realpath .) \
+         --username ${{ inputs.opensearch-user }} \
+         --password ${{ inputs.opensearch-pwd }} \
+         --port ${{ inputs.logstash-port }} \
+         --info is-debug-cluster=false \
```
katexochen, line 36: As you want to deploy it for debug cluster to debug log collection, this should come from input.

.github/workflows/e2e-test-daily.yml
```diff
@@ -84,6 +84,9 @@ jobs:
           azureIAMCreateCredentials: ${{ secrets.AZURE_E2E_IAM_CREDENTIALS }}
           registry: ghcr.io
           githubToken: ${{ secrets.GITHUB_TOKEN }}
+          awsOpenSearchDomain: ${{ secrets.AWS_OPENSEARCH_DOMAIN }}
+          awsOpenSearchUsers: ${{ secrets.AWS_OPENSEARCH_USER }}
+          awsOpenSearchPwd: ${{ secrets.AWS_OPENSEARCH_PWD }}
```
katexochen, line 89: Don't we need this for the other e2e workflows as well?

debugd/internal/debugd/logcollector/logcollector.go
```diff
@@ -152,6 +167,51 @@ func getTemplate(ctx context.Context, logger *logger.Logger) (*template.Template
...
+		return nil, fmt.Errorf("removing filebeat template container: %w\n%s", err, out)
+	}
+
+	tmpl, err := template.ParseFiles("/run/filebeat/templates/filebeat.yml")
+	if err != nil {
+		return nil, fmt.Errorf("parsing filebeat template: %w", err)
+	}
+
+	return tmpl, nil
+}
```
katexochen, line 213: maybe you could reuse the `getLogstashTemplate` function and pass the paths you want to copy? Seems pretty generic...

debugd/internal/debugd/logcollector/logcollector.go
```diff
@@ -212,12 +273,35 @@ func startPod(ctx context.Context, logger *logger.Logger) error {
...
+	if err := templ.Execute(file, in); err != nil {
+		return fmt.Errorf("executing filebeat config template: %w", err)
+	}
+
+	return nil
+}
+
+type filebeatConfInput struct {
+	LogstashHost string
+}
```
katexochen, line 302: Same here, pass paths, reuse function.

dev-docs/workflows/logcollection.md
```diff
@@ -0,0 +1,54 @@
...
+        --username <OPENSEARCH_USERNAME> \
+        --password <OPENSEARCH_PW> \
+        --info key=value \
+        --info key2=value2
+        ...
+    ```
+
+    This will place the templated configuration in the current directory.
+    OpenSearch user credentials can be created by any admin in OpenSearch. Logging in with your company CSP accounts should grant you sufficient permissions to [create a user](https://opensearch.org/docs/latest/security/access-control/users-roles/#create-users) and [grant him the required `all_access` role](https://opensearch.org/docs/latest/security/access-control/users-roles/#map-users-to-roles).
+    One can add additional key-value pairs to the configuration by appending `--info key=value` to the command. These key-value pairs will be attached to the log entries and can be used to filter them in OpenSearch. For example, it might be helpful to add a `test=<xyz>` tag to be able to filter out logs from a specific test run.
```
katexochen, line 28: Break long lines.

hack/logcollector/cmd/template.go
```diff
@@ -0,0 +1,122 @@
...
+	return nil
+}
+
+func (m infoMap) Extend(other infoMap) infoMap {
+	for k, v := range other {
+		m[k] = v
+	}
+	return m
+}
+
```
katexochen, line 74: Please move under the type.

hack/logcollector/cmd/template.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/edgelesssys/constellation/v2/hack/logcollector/internal"
+	"github.com/spf13/cobra"
+)
+
+type infoMap map[string]string
+
+var defaultInfoMap = infoMap{
+	"is-debug-cluster": "false",
+}
```
katexochen, line 19: Better don't use magic defaults.

hack/logcollector/cmd/template.go
```diff
@@ -0,0 +1,122 @@
...
+
+	templateCmd.Flags().String("dir", "", "Directory to place the templated configurations in (required)")
+	must(templateCmd.MarkFlagRequired("dir"))
+	must(templateCmd.MarkFlagDirname("dir"))
+	templateCmd.Flags().String("username", "", "OpenSearch username (required)")
+	must(templateCmd.MarkFlagRequired("username"))
+	templateCmd.Flags().String("password", "", "OpenSearch password (required)")
+	must(templateCmd.MarkFlagRequired("password"))
+	templateCmd.Flags().Int("port", 5045, "Logstash port")
+	templateCmd.Flags().StringToString("info", nil, "Additional fields for the Logstash pipeline in the format --info key1=value1,key2=value2,...")
```
katexochen, line 37: Please validate that flags have valid keys defined in `fields.go`

debugd/logstash/assets.go
```diff
@@ -0,0 +1,15 @@
...
+
+package logstash
+
+import "embed"
+
+// Assets are the exported Logstash template files.
+//
+//go:embed config/*
+//go:embed templates/*
+var Assets embed.FS
```
katexochen, line 15: Where is this used?

hack/logcollector/main.go
katexochen: Your os.Exit is in the cmd package, which is an antipattern. Either move it up into main package, or move cmd -> cmd/logcolletor and rename package to main.

katexochen (changes requested):

hack/logcollector/main.go
katexochen: _Sorry for hijacking this comment, I should have clicked a line above._

katexochen:
When using this to debug the missing logs, we should write the logs into a separate index to not get confused and enable smooth cleanup.

.github/workflows/e2e-test-daily.yml
msanft: All other workflows should already have it

msanft: update: added it to release e2e test where it was missing

debugd/internal/debugd/logcollector/logcollector.go
msanft: I don't really see the value here. I think a generic implementation would be way more complicated to read, as we would have a lot of function args, considering that we template the files before copying them. Also, we change individual yaml fields of the files. If you have an idea for an easy implementation im up to do it though

debugd/logstash/assets.go
msanft: https://github.com/edgelesssys/constellation/blob/660c9041174767232d762b0579ac06bafad0ff54/hack/logcollector/internal/filebeat.go#L23

msanft: https://github.com/edgelesssys/constellation/blob/660c9041174767232d762b0579ac06bafad0ff54/hack/logcollector/internal/logstash.go#L23

hack/logcollector/cmd/template.go
msanft: As its in debugd's internal package right now and thus can't be imported in `/hack`, should we maybe take the time and build a top-level internal package for this?

katexochen: Or move it to hack and import in debugd? Otherwise ok for to move the pkg.

msanft:
We can't import from `hack` in `debugd` as `hack` imports constellation. 
Probably the best way to go is to do a full refactor of the logcollection into its own internal package

msanft:
Manual test worked, running CI now:

[stable image](https://github.com/edgelesssys/constellation/actions/runs/5680034099)
[debug image](https://github.com/edgelesssys/constellation/actions/runs/5680036266)

If the tests complete successfully (i.e. correct logs in OS), I think we can merge this

hack/logcollector/internal/filebeat.go
malt3, line 1: This feels very complex for what it does. Could we replace this logic with a simple terraform template that uses the helm provider and that performs the necessary templating?

malt3 (commented):
I think the installation of the helm charts is overly complicated. Otherwise this looks good to me.

hack/logcollector/internal/filebeat.go
msanft: Had a quick look at the template provider and think that it might work, however I am not sure as we have to do nested templating (i.e. template file 1, use file 1 as a template value in file 2)

I would prefer if we merge this as-is so that [AB#3227](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3227) can be resolved and to address the complex templating logic in a separate PR

msanft:
Tested this manually with the rebase & latest changes

Just triggered another [CI run](https://github.com/edgelesssys/constellation/actions/runs/5888079950)

msanft:
[e2e](https://github.com/edgelesssys/constellation/actions/runs/5901397232)
[corresponding logs](https://search-e2e-logs-y46renozy42lcojbvrt3qq7csm.eu-central-1.es.amazonaws.com/_dashboards/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15m,to:now))&_a=(columns:!(_source),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:'74517cf0-6442-11ed-acf1-47dda8fdfbbb',key:metadata.github.run-id,negate:!f,params:(query:'5901397232'),type:phrase),query:(match_phrase:(metadata.github.run-id:'5901397232')))),index:'74517cf0-6442-11ed-acf1-47dda8fdfbbb',interval:auto,query:(language:kuery,query:''),sort:!()))

elchead (approved):
code is functional, lgtm

malt3 (approved):

PR #2083, opened 2023-07-10
disk-mapper: set LUKS2 token to allow reusing unintialized state disks

daniel-weisse:
### Context
Constellation encrypts its statefull disks using cryptsetup.
For this, a node will first generate a completely random password during the initramfs stage, and then update the password for the disk once it joins a cluster using a password derived from the master secret.

If during boot a node detects its state disk to already be an encrypted LUKS2 disk, it will request a decryption key from a running cluster, instead of formatting the disk.
This allows a node to easily rejoin a cluster upon reboot.
However, this also means that a node which restarted after having first formatted its disk, but before joining a cluster for the first time, will be stuck in initramfs trying to decrypt the disk.

### Proposed change(s)

Use the cryptsetup [`token_json`](https://mbroz.fedorapeople.org/libcryptsetup_API/group__crypt-tokens.html#gadcc2116c40c3bb5a01a7f51d2998423e) functions to add additional context to our encrypted state disks:

* On boot, check if the state disk has a token set
  * If not, format the disk and set a token with custom field `diskIsInitialized=false`
  * If token is set, check the `diskIsInitialized` field
    * If `diskIsInitialized=false`, reformat the disk
    * If `diskIsInitialized=true`, request a decryption key from a running cluster
* On cluster join, update the passphrase of the disk, and update the token's `diskIsInitialized` to `true`

### Additional info
How to test:

* Start a Constellation using the following image:
  ```
  ref/feat-disk-mapper-reusable-disks/stream/debug/v2.10.0-pre.0.20230717121326-51159d780840
  ```
* Reboot a node after initramfs, but before `constellation init`/it joined the cluster and make sure it can successfully boot back into userland
* Reboot a node after it joined the cluster and make sure it successfully rejoins the cluster

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/bazel/toolchains/go_module_deps.bzl
+++ b/bazel/toolchains/go_module_deps.bzl
@@ -2199,8 +2199,8 @@ def go_dependencies():
         build_file_generation = "on",
         build_file_proto_mode = "disable_global",
         importpath = "github.com/goccy/go-json",
-        sum = "h1:CrxCmQqYDkv1z7lO7Wbh2HN93uovUHgrECaO5ZrCXAU=",
-        version = "v0.10.2",
+        sum = "h1:/pAaQDLHEoCq/5FFmSKBswWmK6H0e8g4159Kc/X/nqk=",
+        version = "v0.9.11",
     )
 
     go_repository(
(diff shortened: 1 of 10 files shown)


malt3 (approved):
I looked through all changes and validated that the integration test works. LGTM.

PR #2084, opened 2023-07-10
ci: combine node count inputs into one

daniel-weisse:
### Context
workflow_dispatch inputs are limited to maximum of 10 user configurable inputs.
Therefore we need to combine some of them into a single input to no go over this limit.
In the past we used a combined CLI version and image version field, which imo was somewhat hard to read, write, and fully understand the format of.
The goal of this PR is to combine simpler inputs into one

### Proposed change(s)

- Remove combination of cli and image version field
- Combine worker and control-plane node counts into one

### Additional info
- Once node groups are fully implemented, we probably want to revisit this to allow setting of multiple node groups
  - This could use a similar format, e.g. controlPlaneCount="1:3:2", with a mapping of machineType="large:small:medium"
- The `keepMeasurements` can likely also be removed, since any images built will now have measurements uploaded, meaning we can run `fetch-measurements` for debug images 

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/.github/workflows/e2e-test-manual.yml
+++ b/.github/workflows/e2e-test-manual.yml
@@ -3,14 +3,10 @@ name: e2e test manual
 on:
   workflow_dispatch:
     inputs:
-      workerNodesCount:
-        description: "Number of worker nodes to spawn."
-        default: "2"
-        required: true
-      controlNodesCount:
-        description: "Number of control-plane nodes to spawn."
-        default: "3"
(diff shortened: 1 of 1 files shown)


daniel-weisse:
Ran an e2e test just in case: https://github.com/edgelesssys/constellation/actions/runs/5509815898

malt3 (approved):

PR #2088, opened 2023-07-11
config: make deployCSIDriver backwards compatible

derpsteb:
### Context
We added the field in 2.9 but can only require it in 2.10.

### Proposed change(s)

- Fallback to false and warn the user if the field is not specified.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -130,7 +130,7 @@ type AWSConfig struct {
 	IAMProfileWorkerNodes string `yaml:"iamProfileWorkerNodes" validate:"required"`
 	// description: |
 	//   Deploy Persistent Disk CSI driver with on-node encryption. For details see: https://docs.edgeless.systems/constellation/architecture/encrypted-storage
-	DeployCSIDriver *bool `yaml:"deployCSIDriver" validate:"required"`
+	DeployCSIDriver *bool `yaml:"deployCSIDriver"`
 }
 
 // AzureConfig are Azure specific configuration values used by the CLI.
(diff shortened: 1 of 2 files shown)


internal/config/config.go
```diff
@@ -130,7 +130,7 @@ type AWSConfig struct {
 	IAMProfileWorkerNodes string `yaml:"iamProfileWorkerNodes" validate:"required"`
 	// description: |
 	//   Deploy Persistent Disk CSI driver with on-node encryption. For details see: https://docs.edgeless.systems/constellation/architecture/encrypted-storage
-	DeployCSIDriver *bool `yaml:"deployCSIDriver" validate:"required"`
+	DeployCSIDriver *bool `yaml:"deployCSIDriver"`
```
daniel-weisse, line 133:
```suggestion
	DeployCSIDriver *bool `yaml:"deployCSIDriver"` # TODO after v2.9 release: re-enable "required" validation
```
Maybe add a TODO here to re-enable the validation

internal/config/config.go
```diff
@@ -452,10 +452,14 @@ func New(fileHandler file.Handler, name string, fetcher attestationconfigapi.Fet
...
-	// Backwards compatibility: configs without the field `microserviceVersion` are valid in version 2.6.
-	// In case the field is not set in an old config we prefill it with the default value.
-	if c.MicroserviceVersion == "" {
-		c.MicroserviceVersion = Default().MicroserviceVersion
+	// Backwards compatibility: configs without the field `deployCSIDriver` are valid in version 2.8.
+	// TODO (msanft): v2.9. Remove after v2.9 release.
+	if c.Provider.AWS != nil && c.Provider.AWS.DeployCSIDriver == nil {
+		fmt.Fprintln(
+			os.Stderr,
+			"WARNING: 'provider.aws.deployCSIDriver' will be required in an upcoming version. Setting to 'false'.",
```
daniel-weisse, line 460:
```suggestion
			"WARNING: 'provider.aws.deployCSIDriver' not set. The key will be required in v2.10. Defaulting to 'false'.",
```
We have clear plans on when we want to enforce this value. Maybe add that here?

daniel-weisse (approved):

msanft (approved):
Thanks for noticing!

PR #2090, opened 2023-07-12
aws: use new LB controller to fix SecurityGroup cleanup on K8s service deletion 

elchead:
### Context

The AWS cloud controller manager (CCM), which is currently responsible for LB creation, has issues on cleaning up security groups on K8s service deletion.  
This leads to problems of `constellation terminate` (see #2105).
We addressed this issue upstream (see https://github.com/kubernetes/cloud-provider-aws/issues/566), but did not get any help.

There is a newer controller called `aws-load-balancer-controller` which has some responsibility overlap with the CCM, but its distinction is unfortunately not well documented (https://github.com/kubernetes/cloud-provider-aws/issues/203).
The official docs shortly mention that its usage is preferred over CCM, which is not actively developed anymore (see https://docs.aws.amazon.com/eks/latest/userguide/network-load-balancing.html#:~:text=We%20recommend%20that%20you%20use,the%20AWS%20Load%20Balancer%20Controller.).

### Proposed change(s)
IMPORTANT: we disable the iam upgrade because of permission issues and discuss a solution on how to best upgrade iam permissions. The upgrade still works when the user manually sets the new permissions

- use the newer, dedicated `aws-load-balancer-controller` for managing the LBs
  - disable the LB service from the CCM
  - on `create` install the a new helm chart `aws-load-balancer-controller` which needs to be installed after `certManager`
  - add new IAM permissions for the new controller
- on `upgrade apply`
  - apply updated IAM profile
  - install `aws-load-balancer-controller` if not yet installed, otherwise upgrade 
 - document the AWS external LB service creation

~TODO~: it's not easily visible when the DNS endpoint from the LB is reachable. The K8s service publishes the endpoint way before the LB is provisioned. How to check that the service is externally reachable?

-> Amazon suggests to go the AWS portal and wait for the LB targets to be ready before pinging the DNS endpoint. So our current doc seems ok. https://docs.aws.amazon.com/eks/latest/userguide/network-load-balancing.html#:~:text=Wait%20until%20the%20status%20of%20all%20targets%20is%20healthy%20before%20continuing.%20It%20might%20take%20several%20minutes%20before%20all%20targets%20are%20healthy.

~TODO~: how to deal with .helmignore issue? Our implementation is unable to apply rules to files in subdirectories (e.g. `crds/kustomization.yaml`); But the fix from upstream in the helm sdk (using `symwalk`) causes problems when loading cilium in production (no idea why, see TODO in code)

-> keep current implementation and just add an annotation to the function about the caveat

### How to test
- check upgrade (iam roles and terraform resources should update, + new aws-lb chart)
- lb-test should work
- delete lb service, then delete the cluster

### Additional info
[E2E lb run](https://github.com/edgelesssys/constellation/actions/runs/5590440541); new run after changes https://github.com/edgelesssys/constellation/actions/runs/5629928898/job/15255421418 ✅
[E2e upgrade run](https://github.com/edgelesssys/constellation/actions/runs/5607978039/jobs/10259768263) ✅ from v2.9.0; tested manually locally after changes

co-authored by @malt3 

### Checklist

- [x] Link to Milestone


elchead:
tagging as `breaking change` to put emphasis on new service annotation for external LB

cli/internal/helm/loader.go
```diff
@@ -405,6 +431,14 @@ func (i *ChartLoader) loadConstellationServicesValues() (map[string]any, error)
 		values["tags"] = map[string]any{
 			"AWS": true,
 		}
+		values["aws-load-balancer-controller"] = map[string]any{
+			// TODO: nodeSelector is already set in (customized) values.yaml. Should we have a custom values.yaml instead of configuring here for easier testing?
```
derpsteb, line 435: It seems like `nodeSelector` always has the same value. Thus it should be placed in values.yaml. 
So far we have kept all static values in values.yaml.

cli/internal/kubernetes/upgrade.go
```diff
@@ -96,21 +97,37 @@ type Upgrader struct {
 	upgradeID  string
 }
 
+// NewUpgradeID returns a new upgrade ID.
+// is private function to avoid construction outside this function.
+//
+//nolint:all
```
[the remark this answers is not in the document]
elchead: the comment is above ;) I wanted to enforce no usage outside this function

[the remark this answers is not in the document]
daniel-weisse: I don't think hitting the constructor with a `nolint:all` because the returned type is private is the correct way of going about this.
We have plenty data types that use constructors (and should only created using those Constructors).

Enforcing the creation of an `upgradeID` using `NewUpgradeID` is already given by the fields of the data type being private.

cli/internal/helm/loader.go
```diff
@@ -51,25 +52,27 @@ var (
...
 
 // ChartLoader loads embedded helm charts.
 type ChartLoader struct {
 	csp                          cloudprovider.Provider
 	joinServiceImage             string
 	keyServiceImage              string
-	ccmImage                     string
-	cnmImage                     string
+	ccmImage                     string // cloud controller manager image
+	cnmImage                     string // Azure cloud node manager image
```
3u13r, line 64:
```suggestion
	azureCNMImage                     string // Azure cloud node manager image
```
Either as an addition or replacement for the comment.

cli/internal/helm/update-aws-load-balancer-chart.sh
```diff
@@ -0,0 +1,53 @@
...
+  --depth 1 \
+  --branch="${branch}" \
+  "${chart_url}" "${repo_tmp_dir}"
+
+git sparse-checkout add "${chart_dir}"
+git checkout
+cd "${callDir}"
+
+# remove values.yaml from upstream chart
+rm "${repo_tmp_dir}/${chart_dir}/values.yaml"
```
daniel-weisse, line 40: Why do we remove the upstream values file and keep our own copy?
I think this could result in some annoying issues if the upstream chart introduces changes to the values file and the deployment templates.

Is there any particular reason for doing this, instead of "dynamically" updating values similar to how we do it for other charts, e.g. Cilium or cert-manager?

daniel-weisse (commented):
Reviewed mostly the helm parts.
I think with some effort we can get rid of the helm package in the bootstrapper, which should also make installation of charts for other parts easier, e.g. in https://github.com/edgelesssys/constellation/pull/2080

cli/internal/helm/loader.go
```diff
@@ -157,9 +169,11 @@ func (i *ChartLoader) loadRelease(info chartInfo, helmWaitMode helm.WaitMode) (h
 		updateVersions(chart, compatibility.EnsurePrefixV(constants.VersionInfo()))
 
 		values, err = i.loadConstellationServicesValues()
+	case awsLBControllerInfo.releaseName:
+		values, err = i.loadAWSLoadBalancerControllerValues()
 	}
 
-	if err != nil {
+	if err != nil || values == nil {
```
3u13r, line 176: I don't think it's good practice to check the return value if there is also an error returned. All functions above should return an error if they don't return the values. So essentially we check if any of the functions was called. Therefore, I think that a default statement which returns an error is clearer.

3u13r (changes requested):

cli/internal/cmd/upgradeapply.go
```diff
@@ -448,6 +461,61 @@ type cloudUpgrader interface {
...
+	executeMigration(cmd *cobra.Command, file file.Handler, migrateCmd upgrade.MigrationCmd, flags upgradeApplyFlags) error
+}
+
+type migrationCmdExecutor struct {
+	log debugLog
+}
+
+// planMigration checks for Terraform migrations and asks for confirmation if there are any. The user input is returned as confirmedDiff.
+// adapted from migrateTerraform().
+func (u *migrationCmdExecutor) planMigration(cmd *cobra.Command, file file.Handler, migrateCmd upgrade.MigrationCmd) (hasDiff bool, err error) {
```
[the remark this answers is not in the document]
elchead: what do you suggest? The method was designed as a template, i.e. it being agnostic to migrating `terraform resources` or `iam`. is `planTfMigration` better?

cli/internal/helm/update-aws-load-balancer-chart.sh
elchead: https://github.com/edgelesssys/constellation/pull/2090#discussion_r1266948181 it makes it easier to test with `helm install` by keeping static values in the yaml and only injecting dynamic values.
On the downside it could cause issues with chart updates. Can inject dynamically if better

internal/deploy/helm/install.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/edgelesssys/constellation/v2/internal/retry"
+	"go.uber.org/zap"
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/chart/loader"
+	"helm.sh/helm/v3/pkg/cli"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+// TODO clenaup once agreed on design.
```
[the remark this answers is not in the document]
elchead: ok, will add my name next time, but these TODOs are not meant to stay there. I just wanted to raise awareness during review.

cli/internal/kubernetes/upgrade.go
```diff
@@ -96,21 +97,36 @@ type Upgrader struct {
...
 		// Therefore, add a tmp-suffix to the upgrade ID to indicate that the directory will be cleared after the check.
-		upgradeID += "-tmp"
+		str += "-tmp"
 	}
+	return upgradeID{str}
+}
+
+type upgradeID struct {
+	string
+}
```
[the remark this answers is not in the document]
elchead: a type alias allows implicit conversion so it has no benefit over using a string directly (I don't get why Go allows this)
I explicitly wanted to enforce that `upgradeID` is not just a string. Any better suggestion?

cli/internal/helm/update-aws-load-balancer-chart.sh
daniel-weisse: While I agree with the comment by @derpsteb , the AWS loadbalancer chart is not controlled by us, so we shouldn't add to the values.yaml directly.

cli/internal/kubernetes/upgrade.go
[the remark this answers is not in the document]
elchead: So you suggest to export the type? @daniel-weisse 
Don't mean to be pedantic
```
	upgradeID := kubernetes.UpgradeID{"g"}
```
this would still work but I guess it's obvious that this is wrong usage?

bootstrapper/internal/helm/helm.go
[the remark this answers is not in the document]
elchead: looked into it and seems that the chart is only loaded like this during install (some values are not set during upgrade?). Should it still be moved to the new shared helm pkg?

cli/internal/upgrade/iammigrate.go
```diff
@@ -0,0 +1,112 @@
...
+// IAMMigrateCmd is a terraform migration command for IAM.
+type IAMMigrateCmd struct {
+	tf        tfIAMClient
+	upgradeID string
+	csp       cloudprovider.Provider
+	logLevel  terraform.LogLevel
+}
+
+// NewIAMMigrateCmd creates a new IAMMigrateCmd.
+func NewIAMMigrateCmd(ctx context.Context, upgradeID string, csp cloudprovider.Provider, logLevel terraform.LogLevel) (*IAMMigrateCmd, error) {
```
3u13r, line 40: Regarding https://github.com/edgelesssys/constellation/pull/2090#discussion_r1267848167 and https://github.com/edgelesssys/constellation/pull/2090#discussion_r1266949703.
If we expect `upgradeID string` here we loose all gurantees of creating a variable of type `upgradeID` in the first place. Exporing it and expecting `kubernetes.UpgradeID` here creates a very likely unwanted dependency. Since we always call `kubernetes.NewUpgradeID(...)` directly before calling `kubernetes.NewUpgrader(...)` maybe we can create the upgrateID directly in `kubernetes.NewUpgrader(...)` and then have a getter, have the field be public, or simply return it as a string from the constructor. This has the same gurantees as the current solution while completely removing the need of enforing a type.

cli/internal/kubernetes/upgrade.go
```diff
@@ -152,6 +153,22 @@ func NewUpgrader(ctx context.Context, outWriter io.Writer, fileHandler file.Hand
...
+func (u *Upgrader) GetTerraformUpgrader(ctx context.Context, terraformDir string) (*terraform.Client, error) {
+	tfClient, err := terraform.New(ctx, filepath.Join(constants.UpgradeDir, u.upgradeID, terraformDir))
+	if err != nil {
+		return nil, fmt.Errorf("setting up terraform client: %w", err)
+	}
+	return tfClient, nil
+}
+
+// GetUpgradeID returns the upgrade ID.
+// TODO remove.
```
derpsteb, line 167: open todo. either remove, fix or add your name + AB issue ID.  same in L157

bootstrapper/internal/helm/helm.go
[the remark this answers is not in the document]
3u13r: Why is this resolved? I am also in favor of deprecating the whole `bootstrapper/internal/helm` package. I think that the functions and setting of variables can be moved to the `bootstrapper/internal/kubernetes` package.

docs/docs/getting-started/install.md
```diff
@@ -278,84 +278,89 @@ The built-in `AdministratorAccess` policy is a superset of these permissions.
 
 To [create a Constellation cluster](../workflows/create.md#the-create-step), you need the following permissions:
 
+# TODO should we refer to the file instead of manually copying the changes.. Nobody seems to update it
```
3u13r, line 281: I think the idea of having this copied over here is that the versioned docs still provide the correct values. For referencing the correct file, we'd need to point to an not-yet-existing release branch.

3u13r (changes requested):

docs/sidebars.js
```diff
@@ -143,6 +143,11 @@ const sidebars = {
           label: 'Upgrade your cluster',
           id: 'workflows/upgrade',
         },
+        {
+          type: 'doc',
+          label: 'Deploy an application',
```
3u13r, line 148: I'd align this with the heading of the actual page and change this to `Expose a service`.

docs/docs/getting-started/install.md
elchead: But does everyone really think of updating it? Do we have any process? We could still refer to the main.tf file with the release tag.

elchead: > I think the idea of having this copied over here is that the versioned docs still provide the correct values. For referencing the correct file, we'd need to point to an not-yet-existing release branch.

But does everyone really think of updating it? Do we have any process? We could still refer to the main.tf file with the release tag.

3u13r (changes requested):

cli/internal/clusterid/id.go
```diff
@@ -63,3 +64,8 @@ func (f *File) Merge(other File) *File {
 
 	return f
 }
+
+// GetClusterName returns the name of the cluster.
+func GetClusterName(cfg *config.Config, idFile File) string {
```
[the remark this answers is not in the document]
elchead: I can see your point about package boundary from a purist point of view, but I honestly think that it's okay to entangle them since they are `internal` + clusterid is a tiny pkg with with one struct. But of course I can make it an interface :) I'm just wondering what we gain. Are we ever wanting to inject a another config? `clusterid` is also a very tiny dependency to import

.github/runners/azure-cvm/azure-function/requirements.txt
```diff
@@ -199,26 +199,30 @@ charset-normalizer==3.1.0 \
     --hash=sha256:f8303414c7b03f794347ad062c0516cee0e15f7a612abd0ce1e25caf6ceb47df \
```
3u13r: For the sake of keeping the history clean, please don't merge the main into feature branches but just rebase it. See point 4. from our developer onboarding.

docs/docs/workflows/lb.md
```diff
@@ -1 +1,16 @@
...
+# Expose a service
+Constellation integrates the native load balancers of each CSP. Therefore, to expose a service simply [create a service of type `LoadBalancer`](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer).
+
+
+
+## Internet-facing LB service on AWS
+
+To expose your application service externally you might want to use a Kubernetes Service of type `LoadBalancer`. On AWS, load-balancing is achieved through the [AWS Load Balancing Controller](https://kubernetes-sigs.github.io/aws-load-balancer-controller) as in the managed EKS.
+
+Since recent versions, the controller deploy an internal LB by default requiring to set an annotation `service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing` in order to have an internet-facing LB. For more details, see the [official docs](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/service/nlb/).
```
daniel-weisse, line 10:
```suggestion
Since recent versions, the controller deploy an internal LB by default requiring to set an annotation `service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing` to have an internet-facing LB. For more details, see the [official docs](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/service/nlb/).
```
From vale: Consider using 'to' instead of 'in order to'.

internal/deploy/helm/install.go
```diff
@@ -0,0 +1,159 @@
...
+	return &Installer{
+		action,
+		log,
+	}, nil
+}
+
+// InstallChart is the generic install function for helm charts.
+// When timeout is nil, the default timeout is used.
+func (h *Installer) InstallChart(ctx context.Context, release Release) error {
+	return h.InstallChartWithValues(ctx, release, release.Values)
```
daniel-weisse, line 64:
```suggestion
	return h.InstallChartWithValues(ctx, release,  nil)
```
InstallChartWithValues already uses the release values, this makes the MergeMaps function more efficient.

internal/deploy/helm/install.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// InstallChartWithValues is the generic install function for helm charts with custom values.
+func (h *Installer) InstallChartWithValues(ctx context.Context, release Release, extraValues map[string]any) error {
```
daniel-weisse, line 68: Thoughts about having just one function, since calling `InstallChartWithValues` with `extraValues=nil` is valid usage, and does the same thing as `InstallChart`?

daniel-weisse (commented):
Some final notes, otherwise looks good to me

internal/deploy/helm/install.go
elchead: I generally don't like dangling args and flags in functions, but have no strong opinion. I think it doesn't hurt to have both?

cli/internal/upgrade/iammigrate.go
```diff
@@ -0,0 +1,111 @@
...
+	if err := fileHandler.RemoveAll(constants.TerraformIAMWorkingDir); err != nil {
+		return fmt.Errorf("removing old terraform directory: %w", err)
+	}
+	if err := fileHandler.CopyDir(filepath.Join(constants.UpgradeDir, c.upgradeID, constants.TerraformIAMUpgradeWorkingDir), constants.TerraformIAMWorkingDir); err != nil {
+		return fmt.Errorf("replacing old terraform directory with new one: %w", err)
+	}
+
+	if err := fileHandler.RemoveAll(filepath.Join(constants.UpgradeDir, c.upgradeID, constants.TerraformIAMUpgradeWorkingDir)); err != nil {
+		return fmt.Errorf("removing terraform upgrade directory: %w", err)
+	}
```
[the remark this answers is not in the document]
elchead: good catch, is fixed

cli/internal/helm/values.go
```diff
@@ -194,6 +194,8 @@ var ciliumVals = map[string]map[string]any{
 	},
 }
 
+var nodeSelectorForControlPlane = map[string]any{"node-role.kubernetes.io/control-plane": ""}
```
daniel-weisse, line 197:
```suggestion
var controlPlaneNodeSelector = map[string]any{"node-role.kubernetes.io/control-plane": ""}
```
Nit: goes a bit smoother and keeps in line with naming of `controlPlaneTolerations`

daniel-weisse (approved):

PR #2097, opened 2023-07-12
operators: correctly update launch template default version on AWS image upgrade

malt3:
### Context

When upgrading the default launch template version in the operator, we would pass a `*int64` to `Sprintf("%d", VAR)`.
This results in the address of the variable being used as a launch template version when setting the default version.
This prevents the operator from reconciling correctly.

### Proposed change(s)

- operators: correctly update launch template default version on AWS image upgrade

### Special thanks

Kudos to @elchead for finding the root cause.

### Checklist

- [x] Link to Milestone


derpsteb (approved):
That's one subtle bug fixed right there.

PR #2098, opened 2023-07-12
ci: remove PCR[7]

malt3:
### Context

PCR[7] is unstable, since the measurement depends on the Secure Boot configuration provided by earlier boot stages.
Since the security of Constellation depends on measured boot, we do not have to validate the secure boot state of the system.

### Proposed change(s)

- ci: remove PCR[7]

### Checklist

- [x] Link to Milestone


derpsteb:
Included in https://github.com/edgelesssys/constellation/pull/2100

PR #2107, opened 2023-07-17
cli: write Terraform migration output directly to `constellation-id.json`

msanft:
### Context

Previously, `constellation upgrade` commands that performed Terraform migrations would write their output to a `terraform-migration-output.json` file, which had to be manually merged with `constellation-id.json` to pertain state in one place between multiple updates. If the user failed to do so, upgrading a cluster twice would try to overwrite the existing `terraform-migration-output.json` file and hence fail.

### Proposed change(s)

- Create a backup of the old `constellation-id.json` before an upgrade.
- Write Terraform migration output directly into `constellation-id.json` by merging Terraform migration output into the old existing file.

### Additional info
- Manual testing on AWS and Azure migrations done, works as expected.
- [e2e test](https://github.com/edgelesssys/constellation/actions/runs/5573453159)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/clusterid/BUILD.bazel
+++ b/cli/internal/clusterid/BUILD.bazel
@@ -1,4 +1,5 @@
 load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("//bazel/go:go_test.bzl", "go_test")
 
 go_library(
     name = "clusterid",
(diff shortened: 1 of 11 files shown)


derpsteb (approved):
Looking good.
Maybe remove the bugfix label to show the change under "Other changes"? Seems more like a general improvement than a bugfix to me. Your choice.

PR #2108, opened 2023-07-17
bootstrapper: add fallback endpoint and custom endpoint to SAN field

malt3:
### Context

This is part of the [canonical endpoint RFC](https://github.com/edgelesssys/constellation/blob/main/rfc/canonical-endpoint-dns.md). We start adding the fallback endpoint and the custom endpoint to the apiserver certificate SAN field.
After this is rolled out, we can use the new endpoints.

### Proposed change(s)

- bootstrapper: add fallback endpoint to SAN field
- bootstrapper: add custom endpoint to SAN field
- cli: migrate old clusters by adding endpoints to SAN in ClusterConfiguration of running cluster

### Checklist

- [x] Link to Milestone


cli/internal/terraform/terraform.go
```diff
@@ -103,53 +103,69 @@ func (c *Client) PrepareUpgradeWorkspace(path, oldWorkingDir, newWorkingDir, bac
...
+	apiServerCertSANsUntyped, ok := apiServerCertSANsOutput.Value.([]any)
+	if !ok {
+		return ApplyOutput{}, fmt.Errorf("invalid type in api_server_cert_sans output: %s is not a list of elements", apiServerCertSANsOutput.Type.FriendlyName())
+	}
+	apiServerCertSANs := make([]string, len(apiServerCertSANsUntyped))
+	for i, v := range apiServerCertSANsUntyped {
+		apiServerCertSANs[i], ok = v.(string)
+		if !ok {
+			return ApplyOutput{}, fmt.Errorf("invalid type in api_server_cert_sans output: item at index %v of list is not a string", i)
+		}
```
derpsteb, line 150: nit: could be moved into sth like `func toStringSlice(in []any) ([]string, err)`. Just to make it obvious what is happening here.

cli/internal/kubernetes/upgrade.go
```diff
@@ -343,6 +347,71 @@ func (u *Upgrader) GetClusterAttestationConfig(ctx context.Context, variant vari
 	return existingAttestationConfig, existingConf, nil
 }
 
+// ExtendClusterConfigCertSANs extends the ClusterConfig stored under "kube-system/kubeadm-config" with the given SANs.
+// Existing SANs are preserved.
+func (u *Upgrader) ExtendClusterConfigCertSANs(ctx context.Context, alternativeNames []string) error {
```
derpsteb, line 352: Could separate getting/updating of cluster resources from data processing for easier unit testing.
Not sure if this would need a unit test. I am leaning towards "no".

derpsteb (approved):
Doing another manual test run on all 3 CSPs. But code looks good to me :)

internal/cloud/azure/azure.go
```diff
@@ -157,40 +159,34 @@ func (c *Cloud) GetCCMConfig(ctx context.Context, providerID string, cloudServic
...
+		return nil, fmt.Errorf("retrieving DNS wildcard: %w", err)
 	}
-	if resp.Properties == nil || resp.Properties.IPAddress == nil {
-		return "", fmt.Errorf("could not resolve public IP address reference for load balancer")
+	customEndpoint, err := c.getCustomEndpoint(ctx)
+	var hasCustomEndpoint bool
+	if err == nil && len(customEndpoint) > 0 {
+		hasCustomEndpoint = true
+	} else {
+		log.Printf("No custom endpoint found: %v", err)
```
daniel-weisse, line 183:
```suggestion
```
Usage of `log` package.
Consider using our logging package, or remove the log statement

internal/cloud/azure/azure.go
```diff
@@ -409,6 +405,119 @@ func (c *Cloud) getVMInterfaces(ctx context.Context, vm armcompute.VirtualMachin
...
+	parts := strings.SplitN(dnsName, ".", 2)
+	if len(parts) != 2 {
+		return "", fmt.Errorf("invalid dns name %q", dnsName)
+	}
+	return fmt.Sprintf("*.%s", parts[1]), nil
+}
+
+// getLoadBalancerDNSName retrieves the dns name of the load balancer.
+// On Azure, the DNS name is the DNS name of the public IP address of the load balancer.
+func (c *Cloud) getLoadBalancerDNSName(ctx context.Context) (string, error) {
```
daniel-weisse, line 463: This function is very similar to `getLoadBalancerPublicIP`.
Maybe combine some of the functionality into one function?

internal/cloud/gcp/gcp.go
```diff
@@ -124,6 +124,7 @@ func (c *Cloud) Close() {
 }
 
 // GetLoadBalancerEndpoint returns the endpoint of the load balancer.
+// Warning: on GCP, this returns a hostname:port, not just a hostname.
```
daniel-weisse, line 127: From our dev-docs: [endpoint: host+port](https://github.com/edgelesssys/constellation/blob/main/dev-docs/conventions.md#network), so returning hostname:port should be the expected outcome.
If this works differently for other CSPs, the warning should be moved there.

internal/cloud/azure/azure.go
malt3: Thanks! Good find. This was a left over thing from debugging.

internal/cloud/azure/azure.go
malt3: I'm planning to remove `getLoadBalancerPublicIP` soon, as part if the second half of the canonical endpoint rfc. This is why most of the logic is duplicated.

cli/internal/kubernetes/upgrade.go
malt3: I think the current implementation makes it easier to refactor this to use retryOnConflict later.
The retry loop always needs to get, patch locally, update.
So I'd leave this as-is for now.

internal/cloud/gcp/gcp.go
malt3: Currently, only GCP returns a valid endpoint (host+port).
Maybe we should just fix the other CSPs to also return host+port?

daniel-weisse (approved):

internal/cloud/cloud.go
```diff
@@ -33,4 +33,7 @@ const (
 	TagUID = "constellation-uid"
 	// TagInitSecretHash is the tag/label key used to identify the hash of the init secret.
 	TagInitSecretHash = "constellation-init-secret-hash"
+	// TagCustomEndpoint is the tag/label key used to identify the custom endpoint
+	// or dns name that should be added to tls cert SANs.
+	TagCustomEndpoint = "constellation-custom-endpoint"
```
3u13r, line 38: I'm thinking about moving this to `InitRequest` in `init.proto` since we don't need this value pre-init, which is different than the other values here.

internal/cloud/aws/aws.go
```diff
@@ -126,46 +131,152 @@ func (c *Cloud) InitSecretHash(ctx context.Context) ([]byte, error) {
 }
 
 // GetLoadBalancerEndpoint returns the endpoint of the load balancer.
-func (c *Cloud) GetLoadBalancerEndpoint(ctx context.Context) (string, error) {
+func (c *Cloud) GetLoadBalancerEndpoint(ctx context.Context) (host, port string, err error) {
+	// try new architecture first
```
3u13r, line 135: Maybe add a `TODO(malt3)` to remove the fallback, if possible.

3u13r (changes requested):

internal/cloud/cloud.go
malt3: Yeah that's possible and not a bad solution. I feel like the current approach is cleaner for the consumer of `GetAPIServerCertSANs`, since all fields in the list of SANs can be retrieved from the CSP metadata implementation. Otherwise, the consumer has to combine info from metadata package and init call.
But if you have a strong preference for the init.proto, I am happy to change it.

cli/internal/kubernetes/upgrade.go
```diff
@@ -343,6 +346,68 @@ func (u *Upgrader) GetClusterAttestationConfig(ctx context.Context, variant vari
...
+		return fmt.Errorf("marshaling ClusterConfiguration: %w", err)
+	}
+
+	kubeadmConfig.Data[constants.ClusterConfigurationKey] = string(newConfigYAML)
+	u.log.Debugf("Triggering kubeadm config update now")
+	if _, err = u.stableInterface.UpdateConfigMap(ctx, kubeadmConfig); err != nil {
+		return fmt.Errorf("setting new kubeadm config: %w", err)
+	}
+
+	fmt.Fprintln(u.outWriter, "Successfully extended the cluster's apiserver SAN field")
```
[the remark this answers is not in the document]
malt3: I'll do a follow up PR

internal/cloud/cloud.go
3u13r: Hm, I think that the most obvious drawback is the limitation set by the CSP. I just checked and all 3 major CSPs support a superset of DNS names. I don't think that there is future in which we support more than one custom DNS name, right?
On the other hand, I think that we'll at some point send over large(r) parts of the Kubernetes config in an attempt to make it configurable. So the list of values we send over from the cli and that are needed in the K8s bootstrapping will increase.

I'm not 100% sure, maybe there's a third person wanting share their opinion?

malt3: >  I don't think that there is future in which we support more than one custom DNS name, right?

Agreed. I don't see a use case for more than one custom DNS name.

> On the other hand, I think that we'll at some point send over large(r) parts of the Kubernetes config in an attempt to make it configurable.

If we do that, we can also get rid of most of the cloud provider metadata package, no? So it feels to me like we should either send everything via init or try to get everything from csp metadata but avoid splitting this into two sources.

> maybe there's a third person wanting share their opinion?

Maybe @daniel-weisse since you know the init.proto quite well?

daniel-weisse: I don't see any immediate downsides to sending this as part of init.
Also seems a bit more intuitive than setting the DNS names as some metadata tag.

I'm not sure how new nodes receive this value currently. If they also rely on metadata tags, we need to take some care in adjusting that.

malt3: > I'm not sure how new nodes receive this value currently. If they also rely on metadata tags, we need to take some care in adjusting that.

new nodes receive this value via the kubeadm configmap.

You convinced me. I'll just move this logic to the initproto.

internal/cloud/aws/aws.go
```diff
@@ -126,46 +130,98 @@ func (c *Cloud) InitSecretHash(ctx context.Context) ([]byte, error) {
...
 	}
 
-	// TODO(malt3): ideally, we would use DNS here instead of IP addresses.
-	// Requires changes to the infrastructure.
-	return *output.LoadBalancers[0].AvailabilityZones[0].LoadBalancerAddresses[0].IpAddress, nil
+	return *loadbalancer.AvailabilityZones[0].LoadBalancerAddresses[0].IpAddress, nil
+}
+
+// TODO(malt3): uncomment and use as soon as we switch the primary endpoint to DNS.
+// func (c *Cloud) getLoadBalancerDNSName(ctx context.Context) (string, error) {
```
3u13r, line 191: I'd favor commenting this out via multi-line comments `/**/`. This way the latest `git blame` for those lines stays with the latest changes to the code itself.

internal/cloud/azure/azure.go
```diff
@@ -409,6 +382,84 @@ func (c *Cloud) getVMInterfaces(ctx context.Context, vm armcompute.VirtualMachin
...
+	if resp.Properties == nil || resp.Properties.IPAddress == nil {
+		return "", fmt.Errorf("could not resolve public IP address reference for load balancer")
+	}
+	return *resp.Properties.IPAddress, nil
+}
+
+// TODO(malt3): uncomment and use as soon as we switch the primary endpoint to DNS.
+// // getLoadBalancerDNSName retrieves the dns name of the load balancer.
+// // On Azure, the DNS name is the DNS name of the public IP address of the load balancer.
+// func (c *Cloud) getLoadBalancerDNSName(ctx context.Context) (string, error) {
```
3u13r, line 426: I'd favor commenting this out via multi-line comments `/**/`. This way the latest `git blame` for those lines stays with the latest changes to the code itself.

3u13r (changes requested):

3u13r (approved):
LGTM

PR #2109, opened 2023-07-17
ci: keep embedded measurements if stable image is used

msanft:
### Context

Previously, the daily e2e test would overwrite the CLI's expected PCR values in the `constellation-conf.yml` independent of the image used (i.e. on both debug- and release images). This caused us to notice changes in PCR values of CSPs with a delay.

### Proposed change(s)

- Keep the CLI's default measurements in the daily e2e test if a stable image is used.

### Additional info
- [e2e run prior to this fix](https://github.com/edgelesssys/constellation/actions/runs/5550089676), notice PCR[15] = `000...` even though a stable image is used.
- [e2e run with the fix](https://github.com/edgelesssys/constellation/pull/2109#issue-1807536801) TODO see unpatched PCR values
- [e2e after @malt3 applied his proposed fix](https://github.com/edgelesssys/constellation/actions/runs/5751553291)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/actions/constellation_create/action.yml
+++ b/.github/actions/constellation_create/action.yml
@@ -119,7 +119,7 @@ runs:
       if: inputs.fetchMeasurements
       shell: bash
       run: |
-        config fetch-measurements --debug --insecure
+        constellation config fetch-measurements --debug --insecure
 
     - name: Set image
       id: setImage
--- a/.github/actions/constellation_create/action.yml
+++ b/.github/actions/constellation_create/action.yml
@@ -115,13 +115,6 @@ runs:
           yq eval -i "(.kubernetesVersion) = \"${{ inputs.kubernetesVersion }}\"" constellation-conf.yaml
         fi
 
-    - name: Update measurements for non-stable images
-      if: inputs.fetchMeasurements
-      shell: bash
-      run: |
-        constellation config fetch-measurements --debug --insecure
-        cat constellation-conf.yaml
-
     - name: Set image
       id: setImage
       shell: bash
@@ -137,6 +130,12 @@ runs:
 
         yq eval -i "(.image) = \"${imageInput}\"" constellation-conf.yaml
         echo "image=${imageInput}" | tee -a "$GITHUB_OUTPUT"
+
+    - name: Update measurements for non-stable images
+      if: inputs.fetchMeasurements
+      shell: bash
+      run: |
+        constellation config fetch-measurements --debug --insecure
 
     - name: Set instanceType
       if: inputs.machineType && inputs.machineType != 'default'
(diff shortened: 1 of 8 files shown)


derpsteb:
removing bugfix label as it's conflicting with no changelog. Not sure what takes precedence in changelog generation.

daniel-weisse:
Since we updated the image build pipeline to generate (and upload) measurements for all streams, I think we can completely get rid of the deletion of measurements from the config.
E.g. keep embedded measurements if stream == stable, run `config fetch-measurements` for all other streams to update the config with the expected values.
This also allows us to remove one of the (limited) run inputs from the workflow

.github/actions/constellation_create/action.yml
```diff
@@ -115,52 +115,11 @@ runs:
           yq eval -i "(.kubernetesVersion) = \"${{ inputs.kubernetesVersion }}\"" constellation-conf.yaml
         fi
 
-    - name: Remove embedded measurements
-      if: inputs.keepMeasurements == 'false'
+    - name: Update measurements for non-stable images
+      if:  inputs.fetchMeasurements
```
daniel-weisse, line 119:
```suggestion
      if: inputs.fetchMeasurements
```

.github/workflows/e2e-test-release.yml
```diff
@@ -205,7 +204,7 @@ jobs:
           azureIAMCreateCredentials: ${{ secrets.AZURE_E2E_IAM_CREDENTIALS }}
           registry: ghcr.io
           githubToken: ${{ secrets.GITHUB_TOKEN }}
-
+
```
daniel-weisse, line 207:
```suggestion

```
Remove whitespace

daniel-weisse (approved):

msanft:
Successful after adding `--insecure` flag https://github.com/edgelesssys/constellation/actions/runs/5584456110/jobs/10205916088

msanft:
PCR values in s3 seem off for debug images, forwarded to @malt3

.github/workflows/e2e-test-manual.yml
```diff
@@ -245,7 +236,6 @@ jobs:
           gcpInClusterServiceAccountKey: ${{ secrets.GCP_CLUSTER_SERVICE_ACCOUNT }}
           test: ${{ inputs.test }}
           kubernetesVersion: ${{ inputs.kubernetesVersion }}
-          keepMeasurements: ${{ inputs.keepMeasurements }}
```
malt3, line 248: the manual e2e has to set `fetchMeasurements` if a debug image is used. Otherwise PCR9 is taken from the embedded nightly image, instead of fetching fresh images for the debug image.

malt3:
>  PCR values in s3 seem off for debug images

I am pretty sure that this is a bug in the e2 test manual: When using a debug image, you now have to fetch measurements to replace the embedded measurements.

PR #2113, opened 2023-07-18
cli: set Azure ConfidentialVM option in terraform vars when migrating

malt3:
### Context

The `ConfidentialVM` option was never set when applying terraform migrations, leading to it being default initialized as `false`.
This in turn leads to the upgrade trying to change the settings for node groups on Azure from confidential VMs to non-confidential VMs.

### Proposed change(s)

- cli: set Azure ConfidentialVM option in terraform vars when migrating

### Checklist

- [x] Link to Milestone


derpsteb (approved):

derpsteb:
For visibility: there will be another PR to handle Terraform variables similarly in create and upgrade.

PR #2123, opened 2023-07-20
upgrade: retry when node image update fails due to conflict error

elchead:
### Context

During upgrade apply the write to our NodeVersion custom resource may fail because the constellation-operator updated the status field of the resource between read and write.
Our operator solves this by retrying the write if it encounters a write conflict [1]. The CLI's upgrade apply code [2] should use the same pattern.

[1] https://github.com/edgelesssys/constellation/blob/main/operators/constellation-node-operator/controllers/joiningnode_controller.go#L65-L78
[2] https://github.com/edgelesssys/constellation/blob/main/cli/internal/kubernetes/upgrade.go#L253

### Proposed change(s)

- add a retry of the update

### Related issue
- [AB#3280](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3280)

### Checklist

- [x] Link to Milestone


derpsteb (approved):
Lgtm :+1: 
I like the mock pkg!
Maybe run an upgrade test before merging.

PR #2125, opened 2023-07-24
cli: use Semver type to represent microservice versions

derpsteb:
### Context

Up until now we used strings to represent microservice version information. This invited bugs where the input validation was missing or the "v" prefix was missing/duplicate. Our internal Semver datatype is supposed to solve this issue and embed all necessary validation. 

### Proposed change(s)

- use semver.Semver as type for microservice version in config and in all places where that version is consumed
- minor changes to semver pkg. see history.

### Related issue
- [AB#2956](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/2956)

### Additional info
- Updating ValidK8sVersion will happen in a separate PR

### Checklist

- [x] Link to Milestone


cli/internal/cmd/upgradecheck.go
```diff
@@ -560,7 +559,7 @@ func (v *versionUpgrade) buildString() (string, error) {
 		fmt.Fprintln(&upgradeMsg, "")
 	}
 
-	if v.newServices != "" {
+	if v.newServices != (consemver.Semver{}) {
```
daniel-weisse, line 562:
```suggestion
	if v.newServices != consemver.Semver{} {
```
The parenthesis shouldn't be needed

cli/internal/cmd/upgradecheck.go
```diff
@@ -572,12 +571,12 @@ func (v *versionUpgrade) buildString() (string, error) {
 	}
 
 	// no upgrades available
-	if v.newServices == "" && len(v.newImages) == 0 {
+	if v.newServices == (consemver.Semver{}) && len(v.newImages) == 0 {
```
daniel-weisse, line 574:
```suggestion
	if v.newServices == consemver.Semver{} && len(v.newImages) == 0 {
```
Same here

cli/internal/cmd/upgradecheck.go
```diff
@@ -589,7 +588,7 @@ func (v *versionUpgrade) buildString() (string, error) {
 func (v *versionUpgrade) writeConfig(conf *config.Config, fileHandler file.Handler, configPath string) error {
 	// can't sort image map because maps are unsorted. services is only one string, k8s versions are sorted.
 
-	if v.newServices != "" {
+	if v.newServices != (consemver.Semver{}) {
```
daniel-weisse, line 591:
```suggestion
	if v.newServices != consemver.Semver{} {
```
And here

cli/internal/cmd/upgradecheck.go
```diff
@@ -685,16 +684,16 @@ type versionFetcher interface {
...
 	}
 	list := versionsapi.List{
 		Ref:         v.flags.ref,
 		Stream:      v.flags.stream,
 		Granularity: versionsapi.GranularityMajor,
-		Base:        fmt.Sprintf("v%d", cliVersion.Major),
+		Base:        fmt.Sprintf("v%d", cliVersion.Major()),
 		Kind:        versionsapi.VersionKindCLI,
 	}
 	minorList, err := v.versionsapi.FetchVersionList(ctx, list)
```
daniel-weisse, line 699: Can we rework this method to return a slice of `Semver` instead?
I think this could simplify a lot of the error handling made here.

cli/internal/cmd/create.go
```diff
@@ -21,6 +21,8 @@ import (
 	"github.com/edgelesssys/constellation/v2/internal/constants"
 	"github.com/edgelesssys/constellation/v2/internal/file"
 	"github.com/edgelesssys/constellation/v2/internal/semver"
+	consemver "github.com/edgelesssys/constellation/v2/internal/semver"
```
elchead, line 24: why not change to semver using the import above?

cli/internal/cmd/create.go
```diff
@@ -301,7 +303,7 @@ func must(err error) {
 }
 
 // validateCLIandConstellationVersionAreEqual checks if the image and microservice version are equal (down to patch level) to the CLI version.
-func validateCLIandConstellationVersionAreEqual(cliVersion, imageVersion, microserviceVersion string) error {
+func validateCLIandConstellationVersionAreEqual(cliVersion consemver.Semver, imageVersion string, microserviceVersion consemver.Semver) error {
```
elchead, line 306: nit: should we make conf.Image a semver type too?

internal/config/config.go
```diff
@@ -75,7 +75,8 @@ type Config struct {
 	KubernetesVersion string `yaml:"kubernetesVersion" validate:"required,supported_k8s_version"`
 	// description: |
 	//   Microservice version to be installed into the cluster. Defaults to the version of the CLI.
-	MicroserviceVersion string `yaml:"microserviceVersion" validate:"required,version_compatibility"`
+	// TODO (derpsteb): add unmarshal.
```
daniel-weisse, line 78: TODO already implemented?

internal/versions/versions.go
```diff
@@ -18,10 +18,22 @@ import (
 
 	"github.com/edgelesssys/constellation/v2/internal/compatibility"
 	"github.com/edgelesssys/constellation/v2/internal/constants"
+	consemver "github.com/edgelesssys/constellation/v2/internal/semver"
 	"github.com/edgelesssys/constellation/v2/internal/versions/components"
 	"golang.org/x/mod/semver"
 )
 
+// CLIVersion returns the version of this CLI.
+func CLIVersion() consemver.Semver {
```
elchead, line 27: why don't we replace constants.VersionInfo with the new method?

internal/semver/semver.go
```diff
@@ -6,24 +6,38 @@ SPDX-License-Identifier: AGPL-3.0-only
 
 /*
 Package semver provides functionality to parse and process semantic versions, as they are used in multiple components of Constellation.
+
+The official semantic versioning specification [1] disallows leading "v" prefixes.
+However, the Constellation config uses the "v" prefix for versions to make version strings more recognizable.
+This package bridges the gap between Go's semver pkg (doesn't allow "v" prefix) and the Constellation config (requires "v" prefix).
+
+[1] https://semver.org/.
```
daniel-weisse, line 14:
```suggestion
The official [semantic versioning specification] disallows leading "v" prefixes.
However, the Constellation config uses the "v" prefix for versions to make version strings more recognizable.
This package bridges the gap between Go's semver pkg (doesn't allow "v" prefix) and the Constellation config (requires "v" prefix).

[semantic versioning specification]: https://semver.org/
```
This properly embeds the link.
See the [Go doc comments page](https://tip.golang.org/doc/comment#links)

internal/semver/semver.go
```diff
@@ -47,18 +61,72 @@ func New(version string) (Semver, error) {
...
-		Patch:      patch,
-		Prerelease: pre,
+		major:      major,
+		minor:      minor,
+		patch:      patch,
+		prerelease: pre,
 	}, nil
 }
 
+// NewFromInt constructs a new Semver from three integers: MAJOR.MINOR.PATCH.
```
daniel-weisse, line 71:
```suggestion
// NewFromInt constructs a new Semver from three integers and prerelease string: MAJOR.MINOR.PATCH-PRERELEASE.
```

cli/internal/cmd/create.go
derpsteb: I didn't because it behaves differently in a few situations. Namely, whenever it is a shortpath. I think the abstractions don't match very well and we should have a `ImageVersion`, `ShortPath` or similar type.

internal/semver/semver.go
```diff
@@ -115,3 +218,20 @@ func (v *Semver) UnmarshalJSON(data []byte) error {
 	*v = version
 	return nil
 }
+
+// ByVersion implements [sort.Interface] for sorting semantic version strings.
+// Copied from Go's semver pkg with minimal modification.
+// https://cs.opensource.google/go/x/mod/+/master:semver/semver.go
+type ByVersion []Semver
```
elchead, line 225: nit: `SortByVersion` to indicate sorting functionality?

internal/semver/semver.go
```diff
@@ -87,12 +166,36 @@ func (v Semver) CompatibleWithBinary() bool {
...
 	}
 
-	return v.Compare(binaryVersion) == 0 || binaryVersion.IsUpgradeTo(v)
+	isUpgrade := binaryVersion.IsUpgradeTo(v) == nil
+
+	return v.Compare(binaryVersion) == 0 || isUpgrade
 }
 
-// NextMinor returns the next minor version in the format "vMAJOR.MINOR".
+// NextMinor returns the next minor version in the format "vm.MINOR".
```
daniel-weisse, line 174:
```suggestion
// NextMinor returns the next minor version in the format "vMAJOR.MINOR+1".
```

internal/semver/semver_test.go
```diff
@@ -7,20 +7,22 @@ SPDX-License-Identifier: AGPL-3.0-only
 package semver
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v2"
```
daniel-weisse, line 15:
```suggestion
	"gopkg.in/yaml.v3"
```
Wrong import

internal/semver/semver_test.go
```diff
@@ -218,71 +220,70 @@ func TestCanUpgrade(t *testing.T) {
 	testCases := map[string]struct {
 		version1 Semver
 		version2 Semver
-		want     bool
 		wantErr  bool
```
elchead, line 223: wantErr is misleading naming. the orignal `want` was used to indicate if we accept an upgrade. 
the assertion actually checks the reverse of what the naming suggests:
`			assert.Equal(tc.wantErr, tc.version2.IsUpgradeTo(tc.version1) == nil)`
i.e. wantErr needs to be true when it has not error.

I'd rename to `wantUpgrade`.

internal/semver/semver_test.go
```diff
@@ -218,71 +220,70 @@ func TestCanUpgrade(t *testing.T) {
...
 			assert := assert.New(t)
 
-			assert.Equal(tc.want, tc.version2.IsUpgradeTo(tc.version1))
+			assert.Equal(tc.wantErr, tc.version2.IsUpgradeTo(tc.version1) == nil)
 		})
 	}
 }
 
-func TestNextMinor(t *testing.T) {
+func TestNextminor(t *testing.T) {
```
elchead, line 286:
```suggestion
func TestNextMinor(t *testing.T) {
```

internal/semver/semver_test.go
```diff
@@ -304,3 +305,75 @@ func TestNextMinor(t *testing.T) {
...
+		t.Run(name, func(t *testing.T) {
+			marshalled, err := yaml.Marshal(tc.version)
+
+			require.NoError(t, err)
+			require.Equal(t, tc.want, string(marshalled))
+		})
+	}
+}
+
+func TestVersionUnmarshalYAML(t *testing.T) {
```
elchead, line 347: i'd make the test marshal + unmarshal to test if they are compatible

internal/semver/semver.go
```diff
@@ -70,13 +138,24 @@ func (v Semver) Compare(other Semver) int {
...
 func (v Semver) MajorMinorEqual(other Semver) bool {
-	return v.Major == other.Major && v.Minor == other.Minor
+	return v.major == other.major && v.minor == other.minor
 }
 
 // IsUpgradeTo returns if a version is an upgrade to another version.
 // It checks if the version of v is greater than the version of other and allows a drift of at most one minor version.
-func (v Semver) IsUpgradeTo(other Semver) bool {
-	return v.Compare(other) > 0 && v.Major == other.Major && v.Minor-other.Minor <= 1
+func (v Semver) IsUpgradeTo(other Semver) error {
```
elchead, line 146: I think that this logic should not be part of the type, since we support different upgrades for the different version types (microService, node image).
Instead, i'd have a function 
`compatibility.IsMicroServiceUpgrade()` , the alternative of having seperate types for each version type seems overkill

internal/versions/versions.go
derpsteb: Originally didn't to not import semver into constants, keep `CompatibleWithBinary` and to not replace all callsites of constants.VersionInfo. But I guess it's fine to use CLIVersion.

internal/semver/semver.go
derpsteb: I made the type private so it's not used outside of semver.Sort.
When using this in sort.Sort the call would look like `sort.Sort(SortByVersion(foo))`. I think `sort.Sort(ByVersion(foo))` is more elegant?

cli/internal/cmd/upgradecheck.go
derpsteb: Couldn't make the compiler happy without them.

internal/constants/constants.go
```diff
@@ -231,9 +234,15 @@ b92PDCpM7FZAINQF88s1TZS/HmRXYk62UJ4eqPduvUnJmXhNikhLbMi6fw==
...
 
-// VersionInfo returns the version of a binary.
-func VersionInfo() string {
-	return versionInfo
+// BinaryVersion returns the version of this Binary.
+func BinaryVersion() semver.Semver {
+	version, err := semver.New(versionInfo)
+	if err != nil {
+		// This is not user input, unrecoverable, should never happen.
+		panic(fmt.Sprintf("parsing embedded CLI version: %s", err))
```
daniel-weisse, line 242:
```suggestion
		panic(fmt.Sprintf("parsing embedded version information: %s", err))
```
This function is also called by the bootstrapper, and other Constellation binaries.

internal/config/migration/migration.go
```diff
@@ -28,11 +29,11 @@ const (
 // Config defines configuration used by CLI.
 type Config struct {
 	Version             string         `yaml:"version" validate:"eq=v2"`
-	Image               string         `yaml:"image" validate:"required,version_compatibility"`
+	Image               string         `yaml:"image" validate:"required,image_compatibility"`
```
daniel-weisse, line 32: I think you can just ignore the validation tags here.
They are mostly leftover because I simply copied the structs from the previous Config definition.

daniel-weisse (approved):

cli/internal/cmd/upgradecheck.go
derpsteb: From private discussions: I agree that this would be a nice change. However, this would require quite some changes in the versionsapi pkg to make the change consistent. Only changing the type of List.Versions is not enough. This could be done in a separate PR if we decide that this is a useful improvement.

PR #2126, opened 2023-07-24
bazel: check for docker config and fix cli usage in container

katexochen:
### Context

The bazel container enables running bazel in situations where bazel cannot executed directly on the host.

### Proposed change(s)

- Check for container config file when starting container. Without the file container images can't be pushed.
- Use the injected paths to the bazel cache when executing binaries that were previously build (cli). The created symlinks within the container are pointing to host fs paths and won't work.

### Checklist


msanft:
works🚀

msanft (approved):
works 🚀

PR #2128, opened 2023-07-24
cli: output CSI driver versions on `status`

daniel-weisse:
### Context
Individual versions of the CSI chart were not printed when running `constellation status`

### Proposed change(s)
- Print the installed versions of the CSI chart
- Only print aws-load-balancer-controller version if it is installed

### DIFF
--- a/cli/internal/helm/client.go
+++ b/cli/internal/helm/client.go
@@ -221,23 +221,27 @@ func (c *Client) Versions() (ServiceVersions, error) {
 	if err != nil {
 		return ServiceVersions{}, fmt.Errorf("getting %s version: %w", constellationServicesInfo.releaseName, err)
 	}
-	awsLBVersion, err := c.currentVersion(awsLBControllerInfo.releaseName)
-	if err != nil && !errors.Is(err, errReleaseNotFound) {
-		return ServiceVersions{}, fmt.Errorf("getting %s version: %w", awsLBControllerInfo.releaseName, err)
+	csiVersions, err := c.csiVersions()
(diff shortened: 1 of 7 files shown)


cli/internal/helm/client.go
```diff
@@ -240,43 +249,31 @@ func (c *Client) currentVersion(release string) (string, error) {
...
 	}
-}
 
-// Cilium returns the version of the Cilium release.
-func (s ServiceVersions) Cilium() string {
-	return s.cilium
-}
+	// No CSI driver installed
+	if len(packedChartRelease) == 0 {
+		return nil, nil
```
3u13r, line 260: `return nil, nil` always looks weird. Should we maybe return a "ChartNotInstalledErr" that we gracefully handle in the caller?

daniel-weisse: Returning an error here makes no sense, since we simply use the output to fill a struct.
The "returned value" is handled [here](https://github.com/edgelesssys/constellation/blob/b0ef84e69abfbf95e837cf4fd476a04d070be7ab/cli/internal/helm/serviceversion.go#L37).

3u13r: While I think it might be ok-ish in this situation, it also opens up the possibility of people orienting themselves by this code. I had a look at a more general refactoring along the lines of making this function private and moving it to / calling it from `.String()` and `ConstellationServices()`. We don't seem to any other operation on the returned values than calling those functions (also only in 1 place each).

elchead (approved):
code lgtm

PR #2129, opened 2023-07-24
bootstrapper: refactor coredns and cilium setup

daniel-weisse:
### Context
Cilium was one of the first deployments we installed via helm. The installation requires special setup not required by other helm deployments.
Over time we refactored the helm usage in both CLI and bootstrapper, which slowly resulted in a drift between how we install Cilium, and how we install all other charts.
The Cilium installation also performed setup of other components, mainly CoreDNS.
This PR aims to address these issues by aligning the implementation of Cilium closer with other deployments.

### Proposed change(s)
- Decouple the extra set up of CoreDNS from the Cilium installation steps
- Align Cilium chart installation with other deployments by dynamically creating additional values and merging them with `helm.InstallChartWithValues` instead of manipulating the chart release directly using different functions depending on CSP

### Additional info
Test runs:
* [Azure](https://github.com/edgelesssys/constellation/actions/runs/5646120826) ✔️
* [GCP](https://github.com/edgelesssys/constellation/actions/runs/5646122281) ✔️
* [AWS](https://github.com/edgelesssys/constellation/actions/runs/5646123698) ✔️


3u13r (approved):
LGTM

PR #2133, opened 2023-07-25
cli: don't backup CRs that cannot be found

daniel-weisse:
### Context

A CRD can define multiple versions for a Custom Resource, not all of these versions have to be installed in the cluster.
During the CR backup during `constellation upgrade apply`, this currently results in an error for the CSI chart, which defines the versions `v1` and `v1beta1`, but only installs `v1`.

### Proposed change(s)

- Fix errors when creating backups by skipping not-found custom resources

### Checklist

- [ ] Link to Milestone


cli/internal/helm/backup.go
```diff
@@ -56,11 +57,23 @@ func (c *Client) backupCRs(ctx context.Context, crds []apiextensionsv1.CustomRes
 	c.log.Debugf("Starting CR backup")
 	for _, crd := range crds {
 		c.log.Debugf("Creating backup for resource type: %s", crd.Name)
+
+		// Iterate over all versions of the CRD
+		// TODO: Consider iterating over crd.Status.StoredVersions instead
+		// Currently, we have to ignore not-found errors, because a CRD might define
+		// a version that is not installed in the cluster.
```
derpsteb, line 64:
```suggestion
		// a version for which no resources are installed.
```

derpsteb (approved):

cli/internal/helm/backup.go
daniel-weisse: No, it is not "no resources installed".
The old backup code runs fine for CRs where no actual resource was ever deployed to the cluster.
The issue here is that the CSI snapshot CRD defines version in the spec (`v1`, and `v1beta1`), but one of those versions (`v1beta1`) is not installed in the cluster, i.e. afaik you also can't create any CRs at version `v1beta1`.

PR #2137, opened 2023-07-26
bazel: fix container versioning

katexochen:
### Proposed change(s)

- The v perfix was missing in the workflow, and the -0 suffix was to much in the script. :man_shrugging:


3u13r (approved):
LGTM

PR #2142, opened 2023-07-31
api: add functions to transparently handle signatures upon API interaction

derpsteb:
### Context

We currently need to add more objects to the attestationcfg api (AWS SNP). These objects need to be signed. For Azure SNP we are already creating signed objects. Same for image measurements. However, both of these implementation work differently. For AWS SNP we would have to copy the code from Azure SNP 1:1. We will have to add more objects in the future (TDX, GCP SNP, ...). Which will again require copying the code. To prevent this code duplication and potentially harmonize image measurement signing, I want to add functions to the API client/fetcher that transparently handle signatures. 

### Proposed change(s)

- make properties of `internal/api/versionsapi/version.go:Version` private. This allows us to remove the dependency on `internal/api` from `internal/sigstore`.
- move public key from `internal/sigstore/verify.go:VerifySignature` into property of CosignVerifier. Until now we had to pass a CosignVerifier object _and_ the public key. By combining them we only need to pass one. It also separate public key parsing and signature verification. Main point being that the number of arguments for `FetchAndVerify` is increased by 1 instead of 2. 
- Add `FetchAndVerify`, `SignAndUpload` and `DeleteWithSignature` functions to fetcher/client. If we can migrate all API interactions to these methods we will also remove the need for passing public keys directly and can only rely on CosignVerifier.

### Additional info
- An upcoming PR will use the new functions in the Azure SNP implementation
- An upcoming PR will use the new functions to add AWS SNP objects

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/upgradecheck.go
+++ b/cli/internal/cmd/upgradecheck.go
@@ -35,6 +35,7 @@ import (
 	"github.com/edgelesssys/constellation/v2/internal/kubernetes/kubectl"
 	consemver "github.com/edgelesssys/constellation/v2/internal/semver"
 	"github.com/edgelesssys/constellation/v2/internal/sigstore"
+	"github.com/edgelesssys/constellation/v2/internal/sigstore/keyselect"
 	"github.com/edgelesssys/constellation/v2/internal/versions"
 	"github.com/siderolabs/talos/pkg/machinery/config/encoder"
 	"github.com/spf13/afero"
@@ -52,7 +53,7 @@ func newUpgradeCheckCmd() *cobra.Command {
 	}
 
 	cmd.Flags().BoolP("update-config", "u", false, "update the specified config file with the suggested versions")
-	cmd.Flags().String("ref", constants.ReleaseRef, "the reference to use for querying new versions")
+	cmd.Flags().String("ref", versionsapi.ReleaseRef, "the reference to use for querying new versions")
 	cmd.Flags().String("stream", "stable", "the stream to use for querying new versions")
 
 	return cmd
@@ -351,7 +352,7 @@ func (v *versionCollector) newMeasurements(ctx context.Context, csp cloudprovide
 		v.log.Debugf("Fetching measurements for image: %s", version)
 		shortPath := version.ShortPath()
 
-		publicKey, err := sigstore.CosignPublicKeyForVersion(version)
+		publicKey, err := keyselect.CosignPublicKeyForVersion(version)
 		if err != nil {
 			return nil, fmt.Errorf("getting public key: %w", err)
 		}
@@ -649,7 +650,12 @@ func getCompatibleImageMeasurements(ctx context.Context, writer io.Writer, clien
 		return nil, fmt.Errorf("fetching measurements: %w", err)
 	}
 
-	if err = sigstore.VerifyWithRekor(ctx, version, rekor, hash); err != nil {
+	pubkey, err := keyselect.CosignPublicKeyForVersion(version)
+	if err != nil {
+		return nil, fmt.Errorf("getting public key: %w", err)
+	}
+
+	if err = sigstore.VerifyWithRekor(ctx, pubkey, rekor, hash); err != nil {
 		if _, err := fmt.Fprintf(writer, "Warning: Unable to verify '%s' in Rekor.\nMake sure measurements are correct.\n", hash); err != nil {
 			return nil, fmt.Errorf("writing to buffer: %w", err)
 		}
--- a/cli/internal/cmd/upgradecheck.go
+++ b/cli/internal/cmd/upgradecheck.go
@@ -86,7 +86,6 @@ func runUpgradeCheck(cmd *cobra.Command, _ []string) error {
 			verListFetcher: versionfetcher,
 			fileHandler:    fileHandler,
 			client:         http.DefaultClient,
-			cosign:         &sigstore.CosignVerifier{},
 			rekor:          rekor,
 			flags:          flags,
 			cliVersion:     constants.BinaryVersion(),
@@ -338,7 +337,6 @@ type versionCollector struct {
 	verListFetcher versionListFetcher
 	fileHandler    file.Handler
 	client         *http.Client
-	cosign         cosignVerifier
 	rekor          rekorVerifier
 	flags          upgradeCheckFlags
 	versionsapi    versionFetcher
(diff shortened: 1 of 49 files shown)


internal/sigstore/verify.go
```diff
@@ -12,37 +12,50 @@ import (
...
-	sigRaw := base64.NewDecoder(base64.StdEncoding, bytes.NewReader(signature))
+// CosignVerifier wraps a public key that can be used for verifying signatures.
+type CosignVerifier struct {
+	publicKey crypto.PublicKey
+}
 
-	pubKeyRaw, err := cryptoutils.UnmarshalPEMToPublicKey(publicKey)
+// SetPublicKey parses the pem encoded public key, validates it's parameters and updates the object if necessary.
+func (c *CosignVerifier) SetPublicKey(pem []byte) error {
+	pubkey, err := cryptoutils.UnmarshalPEMToPublicKey(pem)
```
daniel-weisse, line 32: Why have this as a separate function instead of using a constructor which takes a public key as argument?
The `CosignVerifier` does not work without this function being called first, so it would make sense to use it as a constructor for it, no?
Additionally I currently don't see any use cases where we would want to change the used public key during a verification process, so this function should only ever be called once.

elchead: I agree, setters are always dangerous when they need to be called in order for some other function to work.

internal/api/attestationconfigapi/fetcher.go
```diff
@@ -43,7 +43,13 @@ func NewFetcher() Fetcher {
 
 // NewFetcherWithClient returns a new fetcher with custom http client.
 func NewFetcherWithClient(client apifetcher.HTTPClient) Fetcher {
-	return newFetcherWithClientAndVerifier(client, sigstore.CosignVerifier{})
+	verifier := sigstore.CosignVerifier{}
+	err := verifier.SetPublicKey([]byte(cosignPublicKey))
+	if err != nil {
+		// This relies on an embedded public key. If this key can not be validated, there is no way to recover from this.
+		panic(fmt.Errorf("creating cosign verifier: %w", err))
```
elchead, line 50: Library code should not panic. It seems irrelevant in our current use case, but as a library you can rarley know if a panic is approrpiate (there are some edge cases: https://stonecode.ca/idiomatic-panics/ ). Imagine this running concurrently with other code and panicing would likely stop it from terminating gracefully.

internal/sigstore/verify.go
```diff
@@ -55,11 +68,8 @@ func (CosignVerifier) VerifySignature(content, signature, publicKey []byte) erro
 }
 
 // CosignPublicKeyForVersion returns the public key for the given version.
-func CosignPublicKeyForVersion(ver versionsapi.Version) ([]byte, error) {
-	if err := ver.Validate(); err != nil {
-		return nil, fmt.Errorf("selecting public key: invalid version %s: %w", ver.ShortPath(), err)
-	}
-	if ver.Ref == versionsapi.ReleaseRef && ver.Stream == "stable" {
+func CosignPublicKeyForVersion(v version) ([]byte, error) {
+	if v.Ref() == constants.ReleaseRef && v.Stream() == "stable" {
```
elchead, line 72: If we intend to make sigstore general and not specfic to version signing, we should not have this function in this package? I don't see the benefit of having an interface if we don't have another implementation making use of it. Otherwise we decouple from the specific versionapi import, but couple through the constant and are spreading versionapi logic. Could you maybe make AzureSNP use your logic

I think that the right way to do it will emerge when we actually implement the signing logic for sth other than version.
I see that for Attestationconfigapi we are hardcoding the key in the pkg. I think we should do the same for version then.

internal/api/client/client.go
```diff
@@ -249,6 +251,50 @@ func Update(ctx context.Context, c *Client, obj APIObject) error {
 	return nil
 }
 
+// SignAndUpdate signs the given apiObject and updates the object and it's signature in the API.
+// This function should be used in favor of manually managing signatures.
+// The signing is specified as part of the signer argument.
+func SignAndUpdate(ctx context.Context, c *Client, obj APIObject, signer sigstore.Signer) error {
```
elchead, line 257: can you use the new code inside the attestationconfigapi client?

internal/api/fetcher/fetcher.go
```diff
@@ -71,6 +73,32 @@ func Fetch[T apiObject](ctx context.Context, c HTTPClient, obj T) (T, error) {
 	return newObj, nil
 }
 
+// FetchAndVerify fetches the given apiObject, checks if it can fetch an accompanying signature and verifies if the signature matches the found object.
+// The public key used to verify the signature is embedded in the verifier argument.
+// FetchAndVerify uses a generic to return a new object of type T.
+// Otherwise the caller would have to cast the interface type to a concrete object, which could fail.
+func FetchAndVerify[T apiObject](ctx context.Context, c HTTPClient, obj T, cosignVerifier sigstore.Verifier) (T, error) {
```
elchead, line 80: can we use it in attestationconfigapi?

internal/attestation/measurements/measurements.go
```diff
@@ -153,7 +148,7 @@ func (m *M) FetchAndVerify(
 // The hash of the fetched measurements is returned.
 func (m *M) fetchAndVerify(
```
elchead, line 149: should we use the new api client here too?

internal/sigstore/verify.go
derpsteb: This function is necessary for image measurements. 
It is not required to sign version objects in the API. There are no such objects.
The interface is called `version` because image versions are specified by ref and stream.
The interface is necessary to remove the import of `internal/api/versionsapi`, so `internal/api` can import `internal/sigstore` without a cyclic import.

Do you have a location in mind that you see more fitting?

internal/api/fetcher/fetcher.go
derpsteb: see above.

internal/attestation/measurements/measurements.go
derpsteb: I would like to migrate that use as well, but I think this may be more involved. I am currently focused on migrating Azure SNP and creating the AWS objects.

internal/api/client/client.go
elchead: Ok, but I think it makes sense to test the design of an API by actually using it. How do we know that the new code even works?

derpsteb: We don't. But it's easier to review this way. Since the code isn't used I don't see a problem with that. I will start using the code and thus test it shortly after it's merged.

internal/sigstore/verify.go
derpsteb: for visibility: we agreed that there is currently no concrete need for refactoring measurement handling (as in: upcoming feature that require changes to that code). 

adding patch with subpkg.

cli/internal/cmd/upgradecheck.go
derpsteb, line 1: necessary for unittesting getCompatibleImageMeasurements. uncovered that the unittest was not working as expected.

daniel-weisse (approved):

PR #2146, opened 2023-07-31
cli: install cilium in cli instead of bootstrapper

elchead:
### Context

Cilium is the last chart inside the bootstrapper to be moved to the CLI (`constellation init`)
### Proposed change(s)

-

### Related issue
- [AB#3301](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3301)

### Checklist

- [x] Link to Milestone


elchead:
testing AWS: https://github.com/edgelesssys/constellation/actions/runs/5716586953
GCP and Azure were tested manually

elchead:
@daniel-weisse you mentioned to merge helmsuiteinstaller (all chart installer) and the single chart installer. I think there is value in having a helm wrapper around the external dependency for mocking (later on that today) and hence not having to mock the entire helm installation logic (suite installer). Happy to discuss later

elchead:
debugging why GCP is not working properly. putting back to draft

elchead:
testing GCP with dirty fix: https://github.com/edgelesssys/constellation/actions/runs/5731260557

cli/internal/helm/suite_install.go
derpsteb, line 1: Some thoughts on naming. I know the naming wasn't introduced in this PR, but it seems like a good point to clear it up. Currently we have:
- `helmInstallationClient` implementing `SuiteInstaller`. Used during `init` to do the initial installation.
- `Installer` implementing `helmInstaller`. Used during `init` (by `helmInstallationClient`) to install charts.
- `Client`. Used during `upgrade check`, `upgrade apply` and `status`.

I propose we rename:
- `helmInstallationClient` to `InitializationClient`
- `SuiteInstaller` to `Initializer`
- `Client` to `UpgradeClient`

The names `Installer`, `helmInstallationClient` and `Client` are all very close to each other.

cli/internal/helm/suite_install.go
```diff
@@ -0,0 +1,268 @@
...
+			return nil, fmt.Errorf("patching GCP node: %w", err)
+		}
+		vals["ipv4NativeRoutingCIDR"] = output.GCP.IPCidrPod
+		vals["strictModeCIDR"] = output.GCP.IPCidrPod
+	}
+	return vals, nil
+}
+
+// setupMicroserviceVals returns the values for the microservice chart.
+func setupMicroserviceVals(provider cloudprovider.Provider, measurementSalt []byte, uid, serviceAccURI string, output terraform.ApplyOutput) (map[string]any, error) {
```
derpsteb, line 165: Is there a reason why this couldn't be integrated with `extendConstellationServicesValues`?

elchead: no, we could unify all chart setup now, it's a good idea. But would do so in a separate PR

cli/internal/helm/suite_install.go
elchead: Good point, naming should be changed.
I apply your renaming. I hesitate with `Initialization` because it suggests that it only prepares some charts when in fact it does everything. Maybe `HelmSuiteInstaller` ?

elchead:
autoscaler GCP: https://github.com/edgelesssys/constellation/actions/runs/5736778847
sonobuoy quick GCP: https://github.com/edgelesssys/constellation/actions/runs/5736781694

cli/internal/helm/suite_install.go
derpsteb: I explicitly would like to prevent the repeated use of `Install`/`Installer`. 
For me `InitializationClient` only implies that this client is used in the beginning of some process. Comments and method names/signatures should clarify the usage.

Opinions @3u13r @daniel-weisse ?

daniel-weisse: The goal would be to, at some point, have a single Client that just runs an `Apply` method on the cluster.
For now, I would stick with Initializer/InitClient and Upgrader/UpgradeClient, simply to signal their usage for the CLI `init` and `upgrade` commands.

cli/internal/helm/suite_install.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+// lb_port: everywhere hardcoded in TF 6443, for host use ip
+// TODO(malt3): switch over to DNS name on AWS and Azure
+// soon as every apiserver certificate of every control-plane node
+// has the dns endpoint in its SAN list.
+func setupCiliumVals(provider cloudprovider.Provider, output terraform.ApplyOutput) map[string]any {
+	vals := map[string]any{
+		"k8sServiceHost": output.IP,
+		"k8sServicePort": 6443, // TODO take from tf?
```
derpsteb, line 146: unresolved todo.

cli/internal/terraform/terraform/gcp/outputs.tf
```diff
@@ -30,3 +30,8 @@ output "ip_cidr_nodes" {
 output "ip_cidr_pods" {
   value = local.cidr_vpc_subnet_pods
 }
+
+// TODO(elchead): should use instead of hardcoded ports?
+output "lb_port_kube" {
```
derpsteb, line 35: Seems unused. Can be replaced by constant?

cli/internal/helm/suite_install.go
derpsteb, line 1: Can be renamed to `init.go`/`initialize.go`/`...`.

derpsteb (approved):
LGTM! :+1:

PR #2148, opened 2023-08-01
cli: add `--workspace` flag to set base directory for Constellation workspace

daniel-weisse:
### Context

The goal of this PR is to let users specify the [Constellation workspace](https://docs.edgeless.systems/constellation/architecture/orchestration#workspaces) using flags.

For the longest time, our CLI allowed setting a path for the config and master secret files. Other files and folders however, were always assumed to be in the current working directory.
This PR aims to address this issue by adding the `--workspace` flag, which allows a user to set the workspace for a CLI command.
All actions on files will then be run from the given workspace instead from the directory the command is executed from.

### Proposed change(s)
- Remove `--config` flag
- Remove `--master-secret` flag
- Add `--workspace` flag
  - Replaces `--config` and `--master-secret` flags by allowing a user to specify the base directory context from where CLI commands are executed
  - All files are created and accessed relative to the path given by the flag
  - If omitted, uses the current working directory

### Additional info
Not sure if we can properly test the flag using the e2e tests, so I request the reviewers to try out their workflows with the new flag and report any issues that may occur.

I already did some manual testing on GCP. Currently missing testing for IAM migration, general updates and manual tests on Azure/AWS.

[e2e test Azure](https://github.com/edgelesssys/constellation/actions/runs/5727536494/job/15520160471)
[e2e test AWS](https://github.com/edgelesssys/constellation/actions/runs/5727540574/job/15520174774)

### Checklist

- [x] Link to Milestone


elchead:
nit: change the title to "cli: add `--workspace` flag in favor of `--config` and `--master-secret`". seems to make it easier for the user what the change means in the changelog. you agree?

In the description I'd add:
The new flag makes the CLI consistent with the workspace concept (https://docs.edgeless.systems/constellation/architecture/orchestration#workspaces)

cli/internal/cloudcmd/create.go
```diff
@@ -23,15 +23,14 @@ import (
...
-	"github.com/edgelesssys/constellation/v2/internal/constants"
 	"github.com/edgelesssys/constellation/v2/internal/imagefetcher"
 )
 
 // Creator creates cloud resources.
 type Creator struct {
 	out                io.Writer
 	image              imageFetcher
-	newTerraformClient func(ctx context.Context) (tfResourceClient, error)
+	newTerraformClient func(ctx context.Context, workspace string) (tfResourceClient, error)
```
elchead, line 33: why are you using the big client again? I think we should try to get rid of it. My idea was to actually get rid of this bigger client and separate them (tfResource / tfIAM). A cluster `create` should not be able to access IAM functionality.

cli/internal/upgrade/terraform.go
```diff
@@ -129,91 +123,71 @@ func (u *TerraformUpgrader) PlanTerraformMigrations(ctx context.Context, opts Te
...
-	Plan(ctx context.Context, logLevel terraform.LogLevel, planFile string) (bool, error)
+	ShowPlan(ctx context.Context, logLevel terraform.LogLevel, output io.Writer) error
+	Plan(ctx context.Context, logLevel terraform.LogLevel) (bool, error)
 }
 
 // tfResourceClient is a Terraform client for managing cluster resources.
 type tfResourceClient interface {
-	PrepareUpgradeWorkspace(path, oldWorkingDir, newWorkingDir, backupDir string, vars terraform.Variables) error
+	PrepareUpgradeWorkspace(embeddedPath, oldWorkingDir, newWorkingDir, backupDir string, vars terraform.Variables) error
 	CreateCluster(ctx context.Context, provider cloudprovider.Provider, logLevel terraform.LogLevel) (terraform.ApplyOutput, error)
```
elchead, line 191: Is there a reason to undo the recent terraform changes whith stronger type safety? Maybe a mistake during merge

cli/internal/cloudcmd/iam.go
```diff
@@ -19,26 +19,26 @@ import (
 	"github.com/edgelesssys/constellation/v2/cli/internal/terraform"
 	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
 	"github.com/edgelesssys/constellation/v2/internal/cloud/gcpshared"
```
elchead, line 21: same here

cli/internal/cmd/configfetchmeasurements.go
elchead, line 1: This file change also undoes the recent changes by Otto?

cli/internal/cloudcmd/create.go
daniel-weisse: I'm not sure what you are talking about?
I simply added a string as argument to the function. The goal of this PR is not to refactor any client usage. Those were left untouched.

cli/internal/upgrade/terraform.go
daniel-weisse: Again, I'm not sure what changes you are talking about in this case.
All that is happening here is removal of constant paths in favor of using dynamic workspace paths.
In some cases I simplified some logic, but that has nothing to do with undoing terraform changes for type safety, does it?

cli/internal/cloudcmd/iam.go
daniel-weisse: The goal here was to align the usage of the cloudcmd IAM Terraform clients with the cloud resource terraform client.
Again, this shouldn't have undone any recent changes, but simply simplified logic in some places and moved the Terraform client creation from the constructor to the methods of the client (Similar to the cloud resource creator).

cli/internal/cmd/configfetchmeasurements.go
daniel-weisse: Could you highlight the changes you are talking about?
From what I can tell this only updates the use of the config flag to the workspace flag, and removes returning a pointer to an uninitialized struct in favor of just returning `nil`

cli/internal/cmd/cmd.go
```diff
@@ -10,3 +10,53 @@ Package cmd provides the Constellation CLI.
 It is responsible for the interaction with the user.
 */
 package cmd
+
+import (
```
elchead, line 14: nit: does it make sense to put this in `workspace.go`. Seems more intuitive to look for anything there.
Even more intuitive a type alias `workspace` with these methods? But I see that it might be too much work. Just a thought.

cli/internal/cmd/configfetchmeasurements_test.go
```diff
@@ -78,17 +77,17 @@ func TestParseFetchMeasurementsFlags(t *testing.T) {
 			require := require.New(t)
 
 			cmd := newConfigFetchMeasurementsCmd()
-			cmd.Flags().String("config", constants.ConfigFilename, "") // register persistent flag manually
-			cmd.Flags().Bool("force", false, "")                       // register persistent flag manually
+			cmd.Flags().String("workspace", "", "") // register persistent flag manually
```
elchead, line 80: nit: nothing to be changed here but it smells to me that we need to do that. Just an idea. Can we test the CLI from a user perspective?
Like here: https://gianarb.it/blog/golang-mockmania-cli-command-with-cobra
In our case we would wrap the cobra command to be able to mock the external dependencies of the run function.

cli/cmd/root.go
```diff
@@ -43,13 +42,13 @@ func NewRootCmd() *cobra.Command {
 	// Set output of cmd.Print to stdout. (By default, it's stderr.)
 	rootCmd.SetOut(os.Stdout)
 
-	rootCmd.PersistentFlags().String("config", constants.ConfigFilename, "path to the configuration file")
-	must(rootCmd.MarkPersistentFlagFilename("config", "yaml"))
-
+	rootCmd.PersistentFlags().String("workspace", "", "path to the Constellation workspace")
```
malt3, line 45:
```suggestion
	rootCmd.PersistentFlags().StringP("workspace", "C", "", "path to the Constellation workspace")
```

The short flag `-C` is a convention used by many tools (including make, tar, [`go`](https://github.com/golang/go/issues/50332) and many others).

cli/internal/cmd/configfetchmeasurements_test.go
daniel-weisse: Possible, and maybe something we want to do in the future, but I think that would be way out of scope for this PR

malt3:
If I understand this PR correctly, you update every use of file paths and prepend the workspace path to it.
While this works and is stable, I wonder why we do not follow the convention of other tools and perform `[os.Chdir(workspace)](https://pkg.go.dev/os#Chdir)` and keep all paths relative to the CWD?
I'm just trying to understand the tradeoffs and why this implementation was chosen.

elchead:
Had a rough look over the changes and lgtm. Maybe we can test the flag explicitly in our e2e_test (simply cd .. and then use the workspace flag?)

elchead:
> If I understand this PR correctly, you update every use of file paths and prepend the workspace path to it. While this works and is stable, I wonder why we do not follow the convention of other tools and perform [`os.Chdir(workspace)`](https://pkg.go.dev/os#Chdir) and keep all paths relative to the CWD? I'm just trying to understand the tradeoffs and why this implementation was chosen.

I think this would simplify the changes a lot and remove unneeded parameters. I think it's better to choose this if there is no other reason for the current way. Even thought it would be some additional work :/

cli/internal/cmd/cmd.go
daniel-weisse: Moving the function makes sense.
Not sure about introducing a new type alias though: Since the value of the workspace is not consumed by functions written by us, but by the standard library, afero, and others, there would be a lot of type conversions.
Additionally I don't see a big benefit other than indicating what the variables is used for in the type name.

cli/internal/cmd/configfetchmeasurements_test.go
elchead: yeah not for this PR, but will add a backlog item to discuss.

daniel-weisse:
> While this works and is stable, I wonder why we do not follow the convention of other tools and perform [os.Chdir(workspace)](https://pkg.go.dev/os#Chdir) and keep all paths relative to the CWD?

Tried this, and found a good solution that works. 
Downside being that this won't correctly print filepaths relative to the execution directory. But it shouldn't be too much effort to fix that.

malt3 (approved):
Thanks! Works great and LGTM. 
Feel free to add or dismiss the short flag `-C`.

3u13r:
The one file under my ownership looks file. Unassigned review.

derpsteb (approved):
Did some basic manual testing and looked at codeowned files. 
:+1:

PR #2152, opened 2023-08-02
cli: node groups

malt3:
### Context

This is the second part of nodeGroup enablement. While previously, we modified the terraform templates to support nodeGroups, this PR exposes nodeGroups to the user via the constellation-conf.yaml.

As a side-effect, we deprecate the "--control-plane-nodes" and "--worker-nodes" flags from "constellation create".

### Proposed change(s)

- add nodeGroups to config
- update config version to v4
- deprecate old flags
- remove deprecated flags from docs

### Future changes (for other PRs)

- [ ] add section in the docs to guide users on how to use node groups
- [ ] config migration (v3 -> v4)

### Checklist

- [x] Link to Milestone
- [ ] [e2e test](https://github.com/edgelesssys/constellation/actions/runs/5760853599)


internal/config/migration/migration.go
daniel-weisse, line 1: The migrate command should migrate a config from the latest available release to a new format.
For example, v2.9.1 did not yet have the `CustomEndpoint` field in the config.
It might be worth to keep the migrate logic in a separate PR until we get closer to the next release and know what has to be included, and what can already be moved to the `v4` spec

malt3: > The migrate command should migrate a config from the latest available release to a new format.
For example, v2.9.1 did not yet have the CustomEndpoint field in the config.

Agreed. I believe this is the only field and it is optional so this wouldn't be an issue but I'll remove it for clarity.

> It might be worth to keep the migrate logic in a separate PR until we get closer to the next release and know what has to be included, and what can already be moved to the v4 spec

I can do that. I think force pushing on this PR is required to make this less confusing.

dev-docs/howto/nfs.md
```diff
@@ -7,7 +7,7 @@ This document describes how to deploy NFS in Constellation using Rook.
 The cluster needs at least 3 worker nodes, default machines are powerful enough.
 
 ```bash
-constellation create --name nfs -c 1 -w 3
+constellation create --name nfs
```
daniel-weisse, line 10:
```suggestion
constellation create
```
The name flag was also removed a while ago, no?

internal/config/config.go
```diff
@@ -377,6 +369,24 @@ func Default() *Config {
...
 			QEMUVTPM:           &QEMUVTPM{Measurements: measurements.DefaultsFor(cloudprovider.QEMU, variant.QEMUVTPM{})},
 		},
+		NodeGroups: map[string]NodeGroup{
+			"control_plane_default": {
+				Role:            "control-plane",
+				Zone:            "",
+				InstanceType:    "",
+				StateDiskSizeGB: 30,
+				StateDiskType:   "",
+				InitialCount:    3,
```
elchead, line 379: why 3 and only 1 worker? I think that 1 is a sensible default that a user would change if they need to?

daniel-weisse (approved):

elchead:
Some doc comments:

- [ ]  how to migrate the config to `NodeGroups`.
- [ ] describe what values can be set for each field of the `NodeGroup`, specifically `Zone` (link to CSP), `InstanceType` and `StateDiskType`
- [ ] explicitly state that you can configure multiple node groups of the same role to allow different node setups
- [ ] state that at least a `control_plane_default` and `worker_default` nodeGroup are required (no renaming)

internal/config/config.go
malt3: 3 is better since it is the minimum number that ensures high availability of the control-plane (including etcd).
My reasoning is that we want to have defaults for a sane cluster that is unlikely to need recovery.
Users will see the amount of nodes clearly displayed when running "constellation create" so they can still change this to `1` if desired.

internal/config/validation.go
```diff
@@ -129,6 +129,23 @@ func ValidateAWSZone(zone string) bool {
...
+
+// ValidateGCPZone validates that the zone is in the correct format.
+func ValidateGCPZone(zone string) bool {
+	gcpZoneRegex := regexp.MustCompile(`^[\w-]+-[a-z]$`)
+	return gcpZoneRegex.MatchString(zone)
+}
+
+// ValidateOpenStackRegion validates that the region is in the correct format.
+func ValidateOpenStackRegion(region string) bool {
+	return len(region) > 0
```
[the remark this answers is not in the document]
malt3: No. OpenStack is quite liberal in what a customer deployment use as a region naming scheme. [They even relax the allowed character set between versions](https://specs.openstack.org/openstack/nova-specs/specs/kilo/implemented/relax-resource-name-restrictions.html).

internal/config/config_test.go
```diff
@@ -889,7 +944,7 @@ func TestValidateProvider(t *testing.T) {
 }
 
 func TestConfigVersionCompatibility(t *testing.T) {
-	t.Skip() // TODO(daniel-weisse): re-enable and re-write for config v3
+	t.Skip() // TODO(daniel-weisse): re-enable and re-write for config v4
```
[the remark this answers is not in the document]
malt3: Done https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3326 :)

internal/config/migration/migration.go
```diff
@@ -9,305 +9,12 @@ package migration
...
-	}
-
-	// Write migrated config
-	if err := fileHandler.WriteYAML(path, cfgV3, file.OptOverwrite); err != nil {
-		return fmt.Errorf("writing %s: %w", path, err)
-	}
-
-	return nil
+func V2ToV3(_ string, _ file.Handler) error {
+	// TODO(malt3): add migration from v3 to v4
```
derpsteb, line 18: You are planning to add this in a separate PR?

malt3: Yes. It was actually part of this PR and was moved out to make this PR more readable.

internal/config/config.go
elchead: Curious to learn why the number is 3 :)
Can we maybe keep the node flags for us devs internally (so that we can quickly set up a test cluster with 1 node to save costs)?
(I know we can workaround with yq but seems like an unnecessary step)

I definetly think that sensible defaults for the other fields should be set. So that `config generate` + `create` works out of the box

internal/config/config_test.go
```diff
@@ -374,6 +393,24 @@ func TestValidate(t *testing.T) {
 				cnf.Attestation.AzureSEVSNP.Measurements = measurements.M{
 					0: measurements.WithAllBytes(0x00, measurements.Enforce, measurements.PCRMeasurementLength),
 				}
```
elchead, line 395: nit: use modifyConfigForAzureToPassValidate here?

internal/config/config.go
```diff
@@ -377,6 +369,24 @@ func Default() *Config {
 			GCPSEVES:           &GCPSEVES{Measurements: measurements.DefaultsFor(cloudprovider.GCP, variant.GCPSEVES{})},
 			QEMUVTPM:           &QEMUVTPM{Measurements: measurements.DefaultsFor(cloudprovider.QEMU, variant.QEMUVTPM{})},
 		},
+		NodeGroups: map[string]NodeGroup{
+			"control_plane_default": {
```
elchead, line 373: nit
```suggestion
		constants.DefaultControlPlaneGroupName : {
```

internal/config/config.go
```diff
@@ -377,6 +369,24 @@ func Default() *Config {
...
+				Role:            "control-plane",
+				Zone:            "",
+				InstanceType:    "",
+				StateDiskSizeGB: 30,
+				StateDiskType:   "",
+				InitialCount:    3,
+			},
+			"worker_default": {
```
elchead, line 381:
```suggestion
			constants.DefaultWorkerGroupName: {
```

internal/config/config.go
malt3: Here are the details of why three is a good default: https://etcd.io/docs/v3.5/faq/#deployment also https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/high-availability/#before-you-begin

> Can we maybe keep the node flags for us devs internally (so that we can quickly set up a test cluster with 1 node to save costs)?

I think as a developer you either have an existing config (with a customized number of initial nodes already set) or have to also set other config values (like debug, image). So I think we should not have a special flag and dogfood the sam UX that we want for our users. I also want to reduce the complexity of the cli.

docs/docs/workflows/create.md
```diff
@@ -33,7 +33,7 @@ Choose the initial size of your cluster.
 The following command creates a cluster with one control-plane and two worker nodes:
```
elchead, line 33: remove

malt3:
> For the regions / zones @thomasten said that we don't want to hardcode specific values in the validation, so that we can support new types without having to release a CLI. Is this different here?

@elchead what are you referring to? I think all of the regexes used for region / zone validation are already designed to allow new zones (as long as they follow the naming scheme used by the CSPs).
In case you are referring to the disk types: Those were already hardcoded in the config before. I just moved this.
I think there is no clear pattern for the disk type names so we should either keep the strict validation or remove it entirely.

elchead:
> > For the regions / zones @thomasten said that we don't want to hardcode specific values in the validation, so that we can support new types without having to release a CLI. Is this different here?
> 
> @elchead what are you referring to? I think all of the regexes used for region / zone validation are already designed to allow new zones (as long as they follow the naming scheme used by the CSPs). In case you are referring to the disk types: Those were already hardcoded in the config before. I just moved this. I think there is no clear pattern for the disk type names so we should either keep the strict validation or remove it entirely.

Yeah referring to disk + instance types. I prefer helping the user with the options but would just want @thomasten to confirm that this is what we want

cli/internal/terraform/variables.go
```diff
@@ -74,8 +74,8 @@ type AWSNodeGroup struct {
 	// StateDiskSizeGB is the size of the state disk to allocate to each node, in GB.
 	StateDiskSizeGB int `hcl:"disk_size" cty:"disk_size"`
 	// InitialCount is the initial number of nodes to create in the node group.
-	// During upgrades this value is not set.
-	InitialCount *int `hcl:"initial_count" cty:"initial_count"`
+	// During upgrades this value ignored.
+	InitialCount int `hcl:"initial_count" cty:"initial_count"`
```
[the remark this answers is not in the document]
malt3: I found it rather confusing to use since the same variable is passed around over multiple layers and is only considered a pointer type here.
The name makes it very obvious that this is only used for the *initial* count so I believe this is easier and safer (cannot be nil, no chance to deref a nil ptr).

internal/config/validation.go
```diff
@@ -254,8 +367,53 @@ func (c *Config) translateMoreThanOneAttestationError(ut ut.Translator, fe valid
 	return t
 }
 
+func registerTranslateDiskTypeError(ut ut.Translator) error {
```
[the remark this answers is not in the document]
malt3: I think this is fine because the error message already gives you a list of all allowed values.
Before, the URLs where part of the config yaml comments and we did not have an exhaustive list of allowed values in that comment.

malt3:
> Some doc comments:
> 
> * [ ]   how to migrate the config to `NodeGroups`.
> * [ ]  describe what values can be set for each field of the `NodeGroup`, specifically `Zone` (link to CSP), `InstanceType` and `StateDiskType`.
> * [ ]  explicitly state that you can configure multiple node groups of the same role to allow different node setups
> * [ ]  ~state that at least a `control_plane_default` and `worker_default` nodeGroup are required (no renaming)~ since we validate the config, it seems easier to maintain only the code

@elchead this will be addressed in a separate PR as pointed out in the PR description:

> add section in the docs to guide users on how to use node groups


I now also created AB#3326 to track this.

docs/docs/getting-started/first-steps.md
```diff
@@ -128,7 +128,7 @@ If you encounter any problem with the following steps, make sure to use the [lat
     :::
```
thomasten, line 128: fix "Create the cluster with one control-plane node and two worker nodes" above

docs/docs/getting-started/first-steps.md
```diff
@@ -128,7 +128,7 @@ If you encounter any problem with the following steps, make sure to use the [lat
...
 
     ```bash
-    constellation create --control-plane-nodes 1 --worker-nodes 2 -y
+    constellation create -y
     ```
 
     This should give the following output:
```
thomasten, line 134: replace `$ constellation create ...` with `$ constellation create -y` below

docs/docs/getting-started/first-steps-local.md
```diff
@@ -103,7 +103,7 @@ attaching persistent storage, or autoscaling aren't available.
 2. Now you can create your cluster and its nodes. `constellation create` uses the options set in `constellation-conf.yaml`.
 
   ```bash
-  constellation create --control-plane-nodes 1 --worker-nodes 1
+  constellation create
   ```
 
   This will create 2 VMs: one worker node, and one control plane node.
```
thomasten, line 109: fix

docs/docs/workflows/create.md
```diff
@@ -29,11 +29,10 @@ Before you create the cluster, make sure to have a [valid configuration file](./
 <tabs groupId="provider">
 <tabItem value="cli" label="CLI">
 
-Choose the initial size of your cluster.
-The following command creates a cluster with one control-plane and two worker nodes:
+Choose the initial size of your cluster by setting the `initialCount` field in the `constellation-config.yaml`.
```
thomasten, line 32: I think this doesn't belong here, but in the Configure workflow

docs/docs/workflows/create.md
```diff
@@ -29,11 +29,10 @@ Before you create the cluster, make sure to have a [valid configuration file](./
...
 
 ```bash
-constellation create --control-plane-nodes 1 --worker-nodes 2
+constellation create
 ```
 
 For details on the flags, consult the command help via `constellation create -h`.
```
thomasten, line 38:
```suggestion
```
remaining flags are not that interesting

docs/docs/workflows/troubleshooting.md
```diff
@@ -31,7 +31,7 @@ Afterward, set `ARM_SKIP_PROVIDER_REGISTRATION=true` as an environment variable
 For example:
 
 ```bash
-ARM_SKIP_PROVIDER_REGISTRATION=true constellation create --control-plane-nodes 1 --worker-nodes 2 -y
+ARM_SKIP_PROVIDER_REGISTRATION=true constellation create -y
```
thomasten, line 34:
```suggestion
ARM_SKIP_PROVIDER_REGISTRATION=true constellation create
```
nit: I'd prefer to omit the flag for this example

thomasten (changes requested):

derpsteb (approved):
Can't remove myself from PR :(
Didn't look in depth.

thomasten (approved):

PR #2153, opened 2023-08-02
cli: unify chart value setup

elchead:
### Context

With the bigger goal to unify the helm initialization and upgrade client, this PR moves all chart setup logic to the helm loader.

### Proposed change(s)

- move the chart extend value functions from the helm initializer to the helm loader
- remove helm chart marshaling since was only needed for grpc init request

### Related issue
- [AB#3319](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3319)

### Additional info
E2E GCP: https://github.com/edgelesssys/constellation/actions/runs/5739208007 ✅
E2E Azure: https://github.com/edgelesssys/constellation/actions/runs/5739831089 ✅
### Checklist

- [x] Link to Milestone


cli/internal/helm/overrides.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+	case cloudprovider.GCP:
+		serviceAccountKey, err := gcpshared.ServiceAccountKeyFromURI(serviceAccURI)
+		if err != nil {
+			return nil, fmt.Errorf("getting service account key: %w", err)
+		}
+		rawKey, err := json.Marshal(serviceAccountKey)
+		if err != nil {
+			return nil, fmt.Errorf("marshaling service account key: %w", err)
+		}
```
derpsteb, line 101: Could you move this marshalling up into `init.go`? This would remove the `gcpshared` import.

cli/internal/helm/overrides.go
```diff
@@ -0,0 +1,156 @@
...
+
+// extraCSIValues returns the values for the csi chart.
+func extraCSIValues(provider cloudprovider.Provider, serviceAccURI string) (map[string]any, error) {
+	var csiVals map[string]any
+	if provider == cloudprovider.OpenStack {
+		creds, err := openstack.AccountKeyFromURI(serviceAccURI)
+		if err != nil {
+			return nil, err
+		}
+		cinderIni := creds.CloudINI().CinderCSIConfiguration()
```
derpsteb, line 148: Same as with `gcpshared`.

cli/internal/helm/loader_test.go
```diff
@@ -20,27 +20,56 @@ import (
...
+			TenantID:            "TenantID",
+			Location:            "Location",
+			PreferredAuthMethod: azureshared.AuthMethodUserAssignedIdentity,
+			UamiResourceID:      "uid",
+		}
+		return creds.ToCloudServiceAccountURI()
+	default:
+		return ""
+	}
+}
```
derpsteb, line 63: If we moved marshalling up into `init.go`, this can be replaced by arbitrary `[]byte` values. Highlighting the fact that this package does not interact with the values it receives.

derpsteb:
This whole refactoring is such an improvement. Nice work!

cli/internal/helm/overrides.go
elchead: good suggestion. this is not marshalling though. Any idea for a general name of serviceAccURI then? marshalledServiceAccURI seems from wrong then

derpsteb (approved):
For visibilty: we decided to not move the csp*shared imports into `cli/internal/cmd` as there is a conceptual dependency on the CSP while loading charts. The resulting chart is different for each csp. 

LGTM.

PR #2155, opened 2023-08-03
terraform: fix uami parsing

3u13r:
### Proposed change(s)

- Count from the right side of the string to ignore optional prefix values that are still part of the users config

My config contains: 
`userAssignedIdentity: subscriptions/<id>/resourceGroups/<name>/providers/Microsoft.ManagedIdentity/userAssignedIdentities/<name>`

I think that the "correct way would be to use that e.g. the resource Group name is "prefixed" with `/resourceGroup/`. The regex would be a bit complex though. Therefore I went with the simpler solution.

### Checklist

- [x] Link to Milestone


elchead:
Do you know why your config is different? But counting from the right seems fine if there are not any other possible variations

elchead (approved):

PR #2158, opened 2023-08-03
cli: refactor terraform output parsing

3u13r:
### Proposed change(s)

- parse every output variable the same, this e.g. removes the nested parsing

Note: this changes the behavior a slight bit, since we now error in case `tfState.Values.Outputs["project"]` is not set. But I think this is the intended behavior.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/terraform/terraform.go
+++ b/cli/internal/terraform/terraform.go
@@ -213,45 +213,49 @@ func (c *Client) ShowCluster(ctx context.Context, provider cloudprovider.Provide
 		Secret:            secret,
 		UID:               uid,
 	}
-	// TODO add provider
+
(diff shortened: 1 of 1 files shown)


elchead:
Yes it's not intended. I actually tried to avoid this ugly parsing logic through `hcl` annotations in the struct, but the problem was that the parser would not enforce to match all fields so we would still have to check all fields if they are empty.  Even the json parser can only enforce to error when it finds unmatched fields but not if not all fields from the struct are found.

If you have a sneaky solution to that, we could make it nicer :)

elchead (approved):

daniel-weisse (approved):

PR #2165, opened 2023-08-04
cli: fix missing safety check in ShowIAM

daniel-weisse:
### Context
`*tfexec.Terraform.Show()` may return a `*tfjson.State` where `Values` is `nil`.
This can happen if the command is executed in a directory with no Terraform state, resulting in a panic in our Terraform wrapper, since we don't verify the returned state does not hold a `nil` pointer for `Values`.

### Proposed change(s)

- Add a safety check and unit tests for `ShowIAM`
- replace `someErr` in terraform unit tests with `assert.AnError`


elchead (approved):
sorry, the PR slipped through. Looks good

PR #2167, opened 2023-08-04
cli: declare mastersecret as immutable and print attestationCfg diff in warning

elchead:
### Context

In order to make an upgrade safe that applies our entire cluster configuration, we should take precaution to not update resources that can destroy the cluster.
We found `constellation-mastersecret` and the attestation config (inside `join-config`) to be such critical resources.
We decided to not make `join-config` immutable, because we actually allow updating it. Deleting the resource during an upgrade is troublesome when the upgrade fails.

### Proposed change(s)

- declare `constellation-mastersecret` immutable
- attestation config can be updated, but only through explicit user confirmation (already done). We know additionally print a diff to show the user what has changed.
- the attestationConfigDiff check now fetches the resource through the shared logic from `status` cmd to start decoupling from the upgrader which should soon be removed
 

### Related issue
- [AB#3321](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3321)

### Additional info
It would be nice if we could rollback a wrong attestation config, but it's currently not done since the attestation config update is outside of helm. With the planned `apply` functionality, the attestationconfig (`join-config`) should be updated as part of a helm upgrade and rollback when deployments fail.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/upgradeapply.go
+++ b/cli/internal/cmd/upgradeapply.go
@@ -74,7 +74,12 @@ func runUpgradeApply(cmd *cobra.Command, _ []string) error {
 	imagefetcher := imagefetcher.New()
 	configFetcher := attestationconfigapi.NewFetcher()
 
-	applyCmd := upgradeApplyCmd{upgrader: upgrader, log: log, imageFetcher: imagefetcher, configFetcher: configFetcher}
+	stableClient, err := kubernetes.NewStableClient()
+	if err != nil {
+		return fmt.Errorf("creating stable client: %w", err)
+	}
+
+	applyCmd := upgradeApplyCmd{upgrader: upgrader, log: log, imageFetcher: imagefetcher, configFetcher: configFetcher, stableClient: stableClient}
 	return applyCmd.upgradeApply(cmd, fileHandler)
 }
 
@@ -82,6 +87,7 @@ type upgradeApplyCmd struct {
 	upgrader      cloudUpgrader
 	imageFetcher  imageFetcher
 	configFetcher attestationconfigapi.Fetcher
+	stableClient  kubernetes.StableInterface
 	log           debugLog
 }
 
@@ -110,6 +116,31 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, fileHandler file.Hand
 				cmd.Println("Skipping upgrade.")
 				return nil
 			}
+		}
+	}
+	remoteAttestationCfg, err := getAttestationConfig(cmd.Context(), u.stableClient, conf.GetAttestationConfig().GetVariant())
+	if err != nil {
+		return fmt.Errorf("getting remote attestation config: %w", err)
+	}
+	attestationIsEqual, err := conf.GetAttestationConfig().EqualTo(remoteAttestationCfg)
+	if err != nil {
+		return fmt.Errorf("comparing attestation configs: %w", err)
+	}
+	if !attestationIsEqual {
+		cmd.Println("WARNING: The attestation config in the cluster is different from the local attestation config. Applying a wrong attestation config can lead to a loss of access to your cluster")
+		if !flags.yes {
+			yes, err := askToConfirm(cmd, "Do you really want to override the attestation config?")
+			if err != nil {
+				return fmt.Errorf("asking for confirmation: %w", err)
+			}
+			if !yes {
+				cmd.Println("Skipping upgrade.")
+				return nil
+			}
+		}
+		u.log.Debugf("Deleting join-config")
+		if err := u.stableClient.DeleteConfigMap(cmd.Context(), constants.JoinConfigMap); err != nil {
+			return fmt.Errorf("deleting join-config: %w", err)
 		}
 	}
 
--- a/cli/internal/cmd/upgradeapply.go
+++ b/cli/internal/cmd/upgradeapply.go
@@ -81,7 +81,7 @@ func runUpgradeApply(cmd *cobra.Command, _ []string) error {
 
 type stableClientFactory func(kubeconfigPath string) (getConfigMapper, error)
 
-// needed because StableClient returns the bigger kubernetes.StableInterface
+// needed because StableClient returns the bigger kubernetes.StableInterface.
 func stableClientFactoryImpl(kubeconfigPath string) (getConfigMapper, error) {
 	return kubernetes.NewStableClient(kubeconfigPath)
 }
@@ -135,7 +135,7 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, fileHandler file.Hand
 	conf.UpdateMAAURL(idFile.AttestationURL)
 
 	// If an image upgrade was just executed there won't be a diff. The function will return nil in that case.
-	stableClient, err := stableClientFactory(adminConfPath(flags.workspace))
+	stableClient, err := stableClientFactory(constants.AdminConfFilename)
 	if err != nil {
 		return fmt.Errorf("creating stable client: %w", err)
 	}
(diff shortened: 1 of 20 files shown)


cli/internal/cmd/upgradeapply.go
```diff
@@ -305,7 +330,14 @@ func (u *upgradeApplyCmd) upgradeAttestConfigIfDiff(cmd *cobra.Command, newConfi
...
 
 	if !flags.yes {
-		ok, err := askToConfirm(cmd, "You are about to change your cluster's attestation config. Are you sure you want to continue?")
+		cmd.Println("The configured attestation config is different from the attestation config in the cluster.")
+		diffStr, err := diffAttestationCfg(clusterAttestationConfig, newConfig)
+		if err != nil {
+			return fmt.Errorf("diffing attestation configs: %w", err)
+		}
+		cmd.Println("The following changes will be applied to the attestation config:")
+		cmd.Println(diffStr)
```
derpsteb, line 339: I guess this would be printed, even if there is no diff?

cli/internal/cmd/upgradeapply.go
```diff
@@ -291,7 +316,7 @@ type imageFetcher interface {
 // upgradeAttestConfigIfDiff checks if the locally configured measurements are different from the cluster's measurements.
 // If so the function will ask the user to confirm (if --yes is not set) and upgrade the measurements only.
 func (u *upgradeApplyCmd) upgradeAttestConfigIfDiff(cmd *cobra.Command, newConfig config.AttestationCfg, flags upgradeApplyFlags) error {
```
derpsteb, line 318: I have found it helpful in the past to add a unittest for functions like this. Functions that specify the output of the CLI. I found it helpful because it (a) forced me to specify the expected output, (b) made it easy to iterate all possible outputs and (c) made it easy to change the underlying string construction. Since there is a diff printed now, there is some complexity in the output generated. Potentially a unittest for `diffAttestationCfg` is easier/enough.  
Judge for yourself if you want to add such a test.

cli/internal/cmd/upgradeapply.go
elchead: no, because of https://github.com/edgelesssys/constellation/blob/9ae89133265b835c72b8d8fb62127df7a0d5f686/cli/internal/cmd/upgradeapply.go#L328 it returns early

cli/internal/cmd/upgradeapply.go
elchead: I think that specifically trivial output tests are rarely useful and cause more maintenance for writing the test. If the output was a json format that is part of our API contract that would be different of course.

derpsteb: Not doing trivial output tests has led me to fixing a bunch of bugs in the past, hence my recommendation ^^.
Again, judge for yourself.

derpsteb (approved):

cli/internal/cmd/upgradeapply.go
```diff
@@ -172,6 +183,20 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, fileHandler file.Hand
...
+	if err != nil {
+		return "", fmt.Errorf("marshalling remote attestation config: %w", err)
+	}
+	newYml, err := yaml.Marshal(newAttestationCfg)
+	if err != nil {
+		return "", fmt.Errorf("marshalling local attestation config: %w", err)
+	}
+	diff := string(diff.Diff("current", currentYml, "new", newYml))
+	return diff, nil
+}
```
malt3, line 198: Feel free to ignore, but we may want to only use one diffing library and we already use hexops/gotextdiff in other places.

Example:
https://github.com/edgelesssys/constellation/blob/cc106132520202a225821bfa8b7f2e33891b27a9/hack/bazel-deps-mirror/internal/bazelfiles/files.go#L122-L131

malt3 (approved):
Unsure about the diff implementation. Otherwise LGTM.

cli/internal/cmd/upgradeapply.go
elchead: We were discussing about it with Paul and he mentioned the library too, but decided against it cause it has no benefit, harder usability,  and the library was last updated 3 years ago. Since we already use `go-internal` anyways we decided for that one.

cli/internal/kubernetes/kubernetes.go
```diff
@@ -10,3 +10,77 @@ Package kubernetes provides functions to interact with a live cluster to the CLI
...
+	if err != nil {
+		return nil, err
+	}
+	return NewStableClientWithK8sClient(client), nil
+}
+
+// NewStableClientWithK8sClient returns a new StableInterface.
+func NewStableClientWithK8sClient(client kubernetes.Interface) StableInterface {
+	return &stableClient{client: client}
+}
```
daniel-weisse, line 58: Why do we have 2 functions for this now?
The first one is called in `runUpgradeApply`, the second one in `runStatus`.

IMO, we should not use the first function, as it assumes filepaths outside of the `cmd` package, something I tried to fix in https://github.com/edgelesssys/constellation/pull/2148

Ideally, we only have one function named `NewStableClient`, which takes either a `kubernetes.Interface`, or the path to a kubeconfig file as argument and returns a `stableClient`.

daniel-weisse (changes requested):

cli/internal/kubernetes/kubernetes.go
elchead: good point, fixed it.

cli/internal/cmd/upgradeapply.go
```diff
@@ -124,7 +138,11 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, fileHandler file.Hand
 	conf.UpdateMAAURL(idFile.AttestationURL)
 
 	// If an image upgrade was just executed there won't be a diff. The function will return nil in that case.
-	if err := u.upgradeAttestConfigIfDiff(cmd, conf.GetAttestationConfig(), flags); err != nil {
+	stableClient, err := stableClientFactory(adminConfPath(flags.workspace))
```
daniel-weisse, line 141:
```suggestion
	stableClient, err := stableClientFactory(constants.AdminConfFilename)
```
The workspace paths are just for printing correct filenames to the user

cli/internal/kubernetes/kubernetes.go
daniel-weisse, line 1: I think it would be the easiest if we just used `NewStableClient`, and passed a kubeconfig filepath as argument.
I currently don't see a good reason to have `NewClient` as an exported function, other to then pass the output of that function to `NewStableClient`.

This would also make things easier in e.g. `upgradeapply.go`: You could remove the `stableClientFactory` workaround, and simply use a function `func(string) (StableInterface, error)`

daniel-weisse (changes requested):

cli/internal/cmd/upgradeapply.go
```diff
@@ -73,9 +75,19 @@ func runUpgradeApply(cmd *cobra.Command, _ []string) error {
...
+
+type stableClientFactory func(kubeconfigPath string) (getConfigMapper, error)
+
+// needed because StableClient returns the bigger kubernetes.StableInterface.
+func stableClientFactoryImpl(kubeconfigPath string) (getConfigMapper, error) {
+	return kubernetes.NewStableClient(kubeconfigPath)
+}
+
+type getConfigMapper interface {
+	GetCurrentConfigMap(ctx context.Context, name string) (*corev1.ConfigMap, error)
```
daniel-weisse, line 90:
```suggestion
type configMapGetter interface {
	GetCurrentConfigMap(ctx context.Context, name string) (*corev1.ConfigMap, error)
```
As already mentioned [in another PR](https://github.com/edgelesssys/constellation/pull/2177#discussion_r1286740377), this should be a getter.

daniel-weisse (approved):

cli/internal/cmd/upgradeapply.go
elchead: it's also changed in the followup PR, so I let it be changed there 👍

PR #2168, opened 2023-08-04
ci: verify all pods in verify e2e 

katexochen:
### Proposed change(s)

- Currently, we are only verifying a random pod/node over the loadbalancer. Other nodes could have other measurements, so extending the test to verify all nodes.

### Checklist


katexochen:
@daniel-weisse looks like we need your fix first, run failed due to PCR mismatch.

katexochen:
Successful run: https://github.com/edgelesssys/constellation/actions/runs/5796556812/job/15710363569

PR #2169, opened 2023-08-04
cli: fix incorrect usage of masterSecret salt for clusterID generation

daniel-weisse:
### Context
Instead of measurement salt, the master secret salt was used in the helm charts

### Proposed change(s)

- Generate measurement salt on CLI side, provide it to bootstrapper on init, and set it in the helm chart


bootstrapper/initproto/init.proto
```diff
@@ -8,27 +8,31 @@ service API {
...
+  // ConformanceMode is a flag to indicate whether the cluster should be bootstrapped for Kubernetes conformance testing.
+  bool conformance_mode = 6;
+  // KubernetesComponents is a list of Kubernetes components to install.
+  repeated KubernetesComponent kubernetes_components = 7;
+  // InitSecret is a secret used to authenticate the initial bootstrapping node.
+  bytes init_secret = 8;
+  // ClusterName is the name of the cluster.
+  string cluster_name = 9;
+  // ApiserverCertSans is a list of Subject Alternative Names to add to the apiserver certificate.
+  repeated string apiserver_cert_sans = 10;
```
derpsteb, line 32: Idk why GH forces me to do a review rn. I just wanted to leave a comment...

I assume you are doing this to make incompatible init requests fail? A while back we decided that the numbers should not be reused. As far as I can tell, the [reasons for this](https://protobuf.dev/programming-guides/proto2/#consequences) are still valid. 
To achieve the same we could just increase the numbers to 19..29.

3u13r: I'd also like to clean this up eventually, but I also see that the style guide doesn't welcome such a change. 
A compromise could be that we remove the fields but don't re-assign their numbers? I guess we just need to keep track of the highest number we've assigned, i.e. we'd need to keep any fields after the last used one but can remove all the other fields.

3u13r (approved):
LGTM, please remember that the proto change necessitates new nightly images.

derpsteb (approved):
I still think we should not reuse field numbers and renumber the fields starting with 19. But that is only motivated from the language guide. Practially I don't expect problems, since the init command forces the image version to match the cli version. So feel free to merge.

Wasn't aware that I _also_ need to approve.

bootstrapper/initproto/init.proto
daniel-weisse: This was already merged, but I would like to quickly comment on this.

Our bootstrapper is not backward nor forward compatible, and also currently does not have to be.
Having our proto definitions be compatible in both ways doesn't help us, in fact at some occasions it produced hard to debug errors when initializing clusters with CLIs that were incompatible with older bootstrappers.
For that reason, and to finally have a cleaner proto definition I decided to remove the deprecated fields and added some comments to it.

PR #2172, opened 2023-08-04
cli: add maa token to the output of `verify` command

katexochen:
### Proposed change(s)

- Adding the parsed Microsoft Azure Attestation token to the output of verify command. This is especially interesting to validate that the IDKeyDigest from the SEV-SNP report with the MAA.
- This is the first step to transition the SNP reporter from the current custom runner based workflow to a simpler verify based workflow in the CI.

### Checklist

### DIFF
--- a/cli/internal/cmd/verify.go
+++ b/cli/internal/cmd/verify.go
@@ -15,7 +15,10 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"io"
 	"net"
+	"net/http"
+	"net/url"
 	"strconv"
 	"strings"
 
@@ -30,6 +33,7 @@ import (
 	"github.com/edgelesssys/constellation/v2/internal/file"
 	"github.com/edgelesssys/constellation/v2/internal/grpc/dialer"
 	"github.com/edgelesssys/constellation/v2/verify/verifyproto"
+	"github.com/golang-jwt/jwt/v5"
 	"github.com/google/go-sev-guest/abi"
 	"github.com/google/go-sev-guest/kds"
 	"github.com/spf13/afero"
@@ -127,7 +131,14 @@ func (c *verifyCmd) verify(cmd *cobra.Command, fileHandler file.Handler, verifyC
 	}
 
 	// certificates are only available for Azure
-	attDocOutput, err := formatter.format(rawAttestationDoc, conf.Provider.Azure == nil, flags.rawOutput, attConfig.GetMeasurements())
+	attDocOutput, err := formatter.format(
+		cmd.Context(),
+		rawAttestationDoc,
+		conf.Provider.Azure == nil,
+		flags.rawOutput,
+		attConfig.GetMeasurements(),
+		flags.maaURL,
+	)
 	if err != nil {
 		return fmt.Errorf("printing attestation document: %w", err)
 	}
@@ -240,7 +251,8 @@ func addPortIfMissing(endpoint string, defaultPort int) (string, error) {
 // an attestationDocFormatter formats the attestation document.
 type attestationDocFormatter interface {
 	// format returns the raw or formatted attestation doc depending on the rawOutput argument.
-	format(docString string, PCRsOnly bool, rawOutput bool, expectedPCRs measurements.M) (string, error)
+	format(ctx context.Context, docString string, PCRsOnly bool, rawOutput bool, expectedPCRs measurements.M,
+		attestationServiceURL string) (string, error)
 }
 
 type attestationDocFormatterImpl struct {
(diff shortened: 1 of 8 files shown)


thomasten:
I didn't look to deep in it yet, but this seems to re-implement things we do in go-azguestattestation with another library. Maybe we can just do this
https://github.com/edgelesssys/constellation/blob/ee0adfe8c787339417d35d9f92f52d70f15c0f80/internal/attestation/azure/snp/maa.go#L51-L55
and print the returned claims?

malt3 (approved):
I only reviewed the changes to CLI logic and did not review possible code overlap with go-azguestattestation

katexochen:
I've added validation for the JKU from the token against the URL from the ID/config file.

> I didn't look to deep in it yet, but this seems to re-implement things we do in go-azguestattestation with another library. Maybe we can just do this
> 
> https://github.com/edgelesssys/constellation/blob/ee0adfe8c787339417d35d9f92f52d70f15c0f80/internal/attestation/azure/snp/maa.go#L51-L55
> 
> and print the returned claims?

The library isn't that versatile. Especially claims are only parsed as map of string, but the JWT is a nested struct, so sub-structs would still need to be parsed afterwards.

We could use `maa.GetKeySet`, but not sure it's worth the import.

cli/internal/cmd/verify.go
```diff
@@ -447,6 +463,156 @@ func (f *attestationDocFormatterImpl) parseSNPReport(b *strings.Builder, reportB
...
+		}
+		kid, ok := token.Header["kid"].(string)
+		if !ok {
+			return nil, fmt.Errorf("invalid kid: %v", token.Header["kid"])
+		}
+		jku, ok := token.Header["jku"].(string)
+		if !ok {
+			return nil, fmt.Errorf("invalid jku: %v", token.Header["jku"])
+		}
+		if jku != webKeysURL+"/certs" {
```
thomasten, line 499: you may use url.JoinPath for this

cli/internal/cmd/verify.go
```diff
@@ -447,6 +463,156 @@ func (f *attestationDocFormatterImpl) parseSNPReport(b *strings.Builder, reportB
...
+			return nil, fmt.Errorf("jku from token (%s) does not match configured attestation service (%s)", jku, webKeysURL)
+		}
+
+		ctx, cancel := context.WithCancel(ctx)
+		defer cancel()
+		keySetBytes, err := httpGet(ctx, jku)
```
thomasten, line 505: What's the effect of deriving the new context here instead of using the existing one?

cli/internal/cmd/verify.go
```diff
@@ -447,6 +463,156 @@ func (f *attestationDocFormatterImpl) parseSNPReport(b *strings.Builder, reportB
...
+		if err := json.Unmarshal(keySetBytes, &rawKeySet); err != nil {
+			return nil, err
+		}
+
+		for _, key := range rawKeySet.Keys {
+			if key.Kid != kid {
+				continue
+			}
+			rawCert, _ := base64.StdEncoding.DecodeString(key.X5c[0])
+			cert, err := x509.ParseCertificate(rawCert)
```
thomasten, line 526:
```suggestion
				X5c [][]byte
				Kid string
			}
		}

		if err := json.Unmarshal(keySetBytes, &rawKeySet); err != nil {
			return nil, err
		}

		for _, key := range rawKeySet.Keys {
			if key.Kid != kid {
				continue
			}
			cert, err := x509.ParseCertificate(key.X5c[0])
```
I think you can slightly simplify it like this

katexochen: No, this doesn't work as the cert is base64 and ParseCertificate expects DER.

cli/internal/cmd/verify.go
thomasten: I was thinking of https://pkg.go.dev/net/url#JoinPath, but this is also ok

cli/internal/cmd/verify.go
thomasten: If you change the type of `X5c` to `[][]byte`, the base64 is decoded on unmarshaling.
I've done it like this here:
https://github.com/edgelesssys/go-azguestattestation/blob/a683be600fcfb6cca3e7c1e65216657a3a4d8c92/maa/validatetoken.go#L67-L106

But I'm also fine with keeping it as is.

thomasten (approved):

cli/internal/cmd/verify.go
katexochen: I've tried and it failed due to bad input. *confused*

PR #2174, opened 2023-08-07
cli: move `create` into own pkg

elchead:
### Proposed change(s)

- 

### Related issue
- [AB#3325](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3325)

### Checklist

- [ ] Link to Milestone


elchead:
After doing it for `create` , I'm not sure if we really gain so much here since the CLI commands actually share a fair part of helper logic. Especially testing helpers (should we have a testing subpkg?) But I definitely see that code is all over the place in this pkg and that it should be broken up into smaller pkgs.
What do you think?

derpsteb:
My understanding was that moving commands into subpkgs forces us to recognize this shared code and make conscious decisions about whether something should be shared or not. So the trouble you are running into is part of the task? Maybe we need to set our expectations a bit differntly: this could take some time :D.

I don't have a strong opinion either _for_ or _against_ this PR or the change in general. Slightly "pro change". 
 
Maybe we should split the task among everyone. And only move commands bit by bit, i.e. not multiple at the same time.

elchead:
> My understanding was that moving commands into subpkgs forces us to recognize this shared code and make conscious decisions about whether something should be shared or not. So the trouble you are running into is part of the task? Maybe we need to set our expectations a bit differntly: this could take some time :D.

I think there are many obvious parts that should be shared between the packages (how we parse flags, how we log, how we construct the workspace paths, askToConfirm). I think we are making our lives harder here and I want to be sure that we agree that it's worth this effort.

malt3:
After looking over the required changes, I think we should not merge this. I believe we should instead focus on moving code away from the cmd package.
If the individual commands are complex, moving logic into more appropriate packages results in cleaner code.
After we have done this simplification, moving cmds into individual packages may not be required at all.

elchead:
closing as we agree to restructure and not put the CMDs into separate pkgs. Further discussion in #2181

elchead:
closing as we agree to restructure and not put the CMDs into separate pkgs. Further discussion in #2181

PR #2181, opened 2023-08-08
cli: add design guidelines

daniel-weisse:
### Context
In preparation of refactoring I went over the CLI packages and wrote out some ideas and guidelines for the CLI.
The document aims to explain the design goals of packages in the CLI, not necessarily their function.
Any suggestions are welcome.

Especially from the codeowners of the packages I would like to hear:
* Does the suggested spec align with your view of the package
* Would you like to add any information on the package or change it in some way?

### Proposed change(s)
- Add CLI design guide
- Remove iamid package in favor of declaring the struct in the `cloudcmd` package.

### Checklist

- [ ] Link to Milestone


cli/README.md
```diff
@@ -0,0 +1,154 @@
...
+This package focuses on the interaction with the cloud provider.
+It separates the cloud provider specific code from the rest of the CLI, and
+provides a common interface for all cloud providers.
+Exported functions defined here should be cloud provider agnostic, meaning there should be no functions for specific CSPs.
+Rather the `cloudcmd` package should declare functions that take a `cloudprovier.Provider` as argument,
+perform CSP specific logic, and return a universally usable result.
+
+It is used by the [`cmd` package](#cmd) to handle creation of cloud resources and other CSP specific interactions.
+
+### [`clusterid`](./internal/)
```
derpsteb, line 37:
```suggestion
### [`clusterid`](./internal/clusterid)
```

cli/README.md
```diff
@@ -0,0 +1,154 @@
...
+To generate workspace correct filepaths for printing, use the functions from [`workspace.go`](./internal/cmd/workspace.go) // TODO: Discuss if we should move this to its own package
+
+### [`featureset`](./internal/featureset)
+
+`featureset` defines feature gate constants for the CLI and should not implement any logic.
+
+### [`helm`](./internal/helm)
+
+Helm is used to manage Kubernetes deployments of a Constellation cluster.
+The package implements functionality to install, update, and manage Helm charts.
```
derpsteb, line 62:
```suggestion
The package implements functionality to install, update, and manage Helm charts.
The package includes cloud-provider specific implementations, as charts need to be rendered differently for each CSP.
Exported functions should be CSP agnostic and take a cloudprovider argument, like described for pkg `cloudcmd`.
```

cli/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+The charts themselves are embedded in a built CLI binary, and values are dynamically updated depending on configuration.
+The charts can be found in [`cli/internal/helm/charts/`](./internal/helm/charts/).
```
derpsteb, line 65:
```suggestion
The charts can be found in [`cli/internal/helm/charts/`](./internal/helm/charts/).
Values should be added in the chart's `values.yaml` files, if they are static i.e. don't depend on user input.
```

cli/README.md
```diff
@@ -0,0 +1,154 @@
...
+The required files and Go code to start the container can be found in [`cli/internal/libvirt/`](./internal/libvirt/).
+
+The code in this package should be kept minimal, and likely won't need to be changed unless we do a major refactoring of our QEMU/libvirt installation.
+
+### [`terraform`](./internal/terraform)
+
+Terraform is used to create cloud resources and IAM resources required for Constellation clusters.
+The package implements functionality to create, update, delete, and view said resources.
+
+The Terraform files are embedded in a built CLI binary, and variables are dynamically update depending on configuration.
```
derpsteb, line 98:
```suggestion
The Terraform files are embedded in the CLI binary, and variables are dynamically updated depending on configuration.
```

cli/README.md
```diff
@@ -0,0 +1,154 @@
...
+The files can be found in [`cli/internal/terraform/terraform/`](./internal/terraform/terraform/).
+
+Functions of this package should be kept mostly cloud agnostic (There should be no "CreateAzureCluster" function),
```
derpsteb, line 101:
```suggestion
Functions of this package should be kept CSP agnostic (there should be no "CreateAzureCluster" function),
```

If you want to keep the "mostly", could you specify what that means?

derpsteb:
Maybe @katexochen should have a look as well if you didn't discuss with him already. Maybe he has some resources for pkg design/when (not) to add one.

katexochen:
Thanks for writing this down. lgtm in principal, but

- IMO the package specific descriptions belong in the pkg comment, not into a readme. Duplicating this information here in addition will likely only add to the problem, as we already fail to keep pkg comments up to date.
- The remaining parts aren't specific to the cli/cmd package, but general information on code design. IMO these belong into the dev-docs.

And some random thoughts while reading: Renaming helm->helmcmd and kubernetes->kubecmd might help to clarify the intent of these pkgs.

katexochen:
> Maybe he has some resources for pkg design/when (not) to add one.

No, sorry, none come to my mind right now.

cli/README.md
```diff
@@ -0,0 +1,154 @@
...
+It should be consulted by developers when making changes to the code of packages in [`cli/internal/`](./internal/).
+
+## Basic layout
+
+[`cli/`](./) consists of 2 packages, [`cli/cmd/`](./cmd/), and [`cli/internal/`](./internal/), and a [`main.go`](./main.go), which serves as the entrypoint for the CLI.
+No new packages or files should be added to this base directory.
+
+[`cli/cmd/`](./cmd/) defines the root command of the CLI. Any changes concerning the root command, or additions like persistent flags (flags that can be accessed by all subcommands) should be added here.
+
+TODO: Discuss: Root command could be added to `cli/internal/cmd/`
```
elchead, line 13: since root is not intended to be reused and is application code, it seems fine in its current place

cli/README.md
```diff
@@ -0,0 +1,154 @@
...
+Defines the commands of the CLI.
+Logic should be kept to input/output parsing whenever possible.
+Any more complex code should usually be implemented in one of the other CLI packages.
+
+The code here should be kept as cloud provider agnostic as possible.
+Any CSP specific tasks should be handled by [`cloudcmd`](#cloudcmd).
+
+All filepaths handled by the CLI code should originate from here.
+Common filepaths are defined as constants in [`internal/constants`](../internal/constants/).
+To generate workspace correct filepaths for printing, use the functions from [`workspace.go`](./internal/cmd/workspace.go) // TODO: Discuss if we should move this to its own package
```
elchead, line 53: own pkg makes sense to me

cli/README.md
elchead: I think `helm` can be CSP specific since otherwise the `cloudcmd` pkg would have to know about helm charts which is worse.

daniel-weisse:
Cleaned up and implemented suggestions:

* Moved the "Adding to a package/Adding new packages" to `dev-docs/conventions.md`
* Moved the package design goals from a separate README directly to the package documentation
* Moved the workspace path functions to their own (sub) package of `cli/internal/cmd/workspace`

cli/internal/cloudcmd/iam.go
```diff
@@ -201,23 +200,51 @@ func (c *IAMCreator) createAWS(ctx context.Context, cl tfIAMClient, opts *IAMCon
...
 		},
 	}, nil
 }
 
+// IAMOutput is the output of creating a new IAM profile.
+type IAMOutput struct {
+	// CloudProvider is the cloud provider of the cluster.
+	CloudProvider cloudprovider.Provider `json:"cloudprovider,omitempty"`
+
+	GCPOutput   GCPIAMOutput   `json:"gcpOutput,omitempty"`
```
elchead, line 225: should we make the fields optional through a ptr?

daniel-weisse: They weren't before. This struct also is never written to disk, so making fields optional doesn't change much, but requires us to make additional safety checks (the json tags also don't really matter here).

cli/internal/terraform/terraform.go
```diff
@@ -5,12 +5,15 @@ SPDX-License-Identifier: AGPL-3.0-only
 */
 
 /*
-Package terraform handles creation/destruction of a Constellation cluster using Terraform.
+Package terraform handles creation/destruction of a cloud and IAM resources required by Constellation using Terraform.
```
elchead, line 8:
```suggestion
Package terraform handles creation/destruction of cloud and IAM resources required by Constellation using Terraform.
```

elchead (approved):
thanks, these are good improvements and will be especially helpul for new onboardings :)

cli/internal/cmd/workspace/workspace.go
```diff
@@ -0,0 +1,65 @@
+/*
+Copyright (c) Edgeless Systems GmbH
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+/*
+Package workspace provides functions to generate paths relative to the workspace.
```
katexochen, line 8: I don't think I get the reason for this package yet. Can't we integrate this in a file handler or struct? Looks like something that should happen in a more automatic way. `PrefixFileHandler` or so...

katexochen (approved):
lgtm, thanks. :)

cli/internal/cmd/workspace/workspace.go
```diff
@@ -0,0 +1,65 @@
...
+
+Users may override the default workspace using the --workspace flag.
+The default workspace is the current working directory.
+
+The functions defined here return paths relative to the set workspace,
+and should be used when printing the path to the user.
+
+The MUST not be used when accessing files, as the workspace is changed
+using os.Chdir() before the command is executed.
```
3u13r, line 17: `return paths relative to the set workspace` this is not true, as far I understand it. They return the path to the actual pwd (before the -C execution). 

To avoid misuse, should we omit this functionality entirely and print the paths in relation to the set workdir? How do other tools behave here? I always understood `-C` as literally `cd` into the folder but if the program returns, I want to be back in the original path. I'd assume that e.g. `make` does not "normalize" the paths as we do here.

daniel-weisse: >I always understood -C as literally cd into the folder

Correct. One exit, you will be back where you started.
However, we have multiple cases where we print filepaths to the user for 
For example, after a successful init we print the path to the kubeconfig file:

```
Constellation ID file written to <workspace>/constellation-admin.conf
You can now connect to your cluster by executing:
    export KUBECONFIG="$PWD/<workspace>/constellation-admin.conf
```
If we omit this package, those paths will not work correctly.

(Looking at this again, this path may also not be printed correctly when one provides an absolute path. Should be easy to fix though.)

cli/internal/cmd/workspace/workspace.go
daniel-weisse: Do I understand you correctly that you would prefer some interface to handle the prefixing?
E.g. `PrefixHandler` would be set up at the start of a command using the workspace flag.
Then every time we want to print a filepath, instead of calling a distinct function, we instead call, e.g. `PrefixHandler.Path(constants.<SomeFilename>)`, which would return the correct path for the user?

cli/internal/cmd/workspace/workspace.go
3u13r: > However, we have multiple cases where we print filepaths to the user for

What are the other cases? Can we simply always print the absolute path?

daniel-weisse: > What are the other cases?

Usually any time we create or update files, e.g. during config generate, and upgrade apply/check. It is also required for printing error messages to the user.
We also print a lot of filenames for debug statements.

> Can we simply always print the absolute path?

In that case we would always have to call `filepath.Abs`, so there is no real advantage gained, other than not parsing the workspace flag, with the downside of paths looking less pretty.

I have also implemented the suggestion from https://github.com/edgelesssys/constellation/pull/2181#discussion_r1287386166, which should make usage of this package a lot easier.

malt3 (approved):

cli/internal/cmd/workspace/workspace.go
daniel-weisse: Implemented a struct to handle filepath prefixing, can you please review?

derpsteb (approved):
:muscle:

cli/internal/cmd/pathprefix/pathprefix.go
```diff
@@ -0,0 +1,38 @@
...
+/*
+Package pathprefix is used to print correct filepaths for a configured workspace.
+
+The default workspace is the current working directory.
+Users may override the default workspace using the --workspace flag.
+
+The functions defined here should be used when printing any filepath to the user,
+as they might otherwise be incorrect if the user has changed the workspace.
+
+The MUST not be used when accessing files, as the workspace is changed
```
katexochen, line 16:
```suggestion
The prefixer MUST not be used when accessing files, as the workspace is changed
```

katexochen (approved):

3u13r (approved):

PR #2184, opened 2023-08-09
e2e: remove existingConfig field

katexochen:
The existingConfig field is always set to true during create, as we use the IAM create step to generate the config in all cases. Accordingly, secret injection into config isn't needed anymore in create. This fixes a bug where other parameters like Kubernetes version and cluster name wouldn't be injected into the config due to existingConfig being true.

### Checklist

- [x] Link to Milestone


daniel-weisse (approved):
Code lgtm.
CI test on GCP with Kubernetes v1.27: https://github.com/edgelesssys/constellation/actions/runs/5807543805

PR #2186, opened 2023-08-09
ci: replace mastersecret flag in recover

derpsteb:
### Context

Since we added the workspace flag to the CLI we no longer support the dedicated mastersecret flag. Thus we need to adapt how the recover action gets the mastersecret.

### Proposed change(s)

- Add workspace output to constellation_create action
- Use workspace output in recover action

### Additional info
- [CI run](https://github.com/edgelesssys/constellation/actions/runs/5807255670)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/actions/constellation_create/action.yml
+++ b/.github/actions/constellation_create/action.yml
@@ -71,9 +71,9 @@ outputs:
   kubeconfig:
     description: "The kubeconfig for the cluster."
     value: ${{ steps.constellation-init.outputs.KUBECONFIG }}
-  masterSecret:
-    description: "The master-secret for the cluster."
-    value: ${{ steps.constellation-init.outputs.MASTERSECRET }}
+  workspace:
+    description: "The Constellation workspace directory."
+    value: ${{ steps.constellation-init.outputs.WORKSPACE }}
   osImageUsed:
     description: "The OS image used in the cluster."
     value: ${{ steps.setImage.outputs.image }}
@@ -211,8 +211,8 @@ runs:
       shell: bash
       run: |
         constellation init --force --debug
-        echo "KUBECONFIG=$(pwd)/constellation-admin.conf" >> $GITHUB_OUTPUT
-        echo "MASTERSECRET=$(pwd)/constellation-mastersecret.json" >> $GITHUB_OUTPUT
+        echo "KUBECONFIG=$(pwd)/constellation-admin.conf" | tee -a $GITHUB_OUTPUT
+        echo "WORKSPACE=$(pwd)" | tee -a $GITHUB_OUTPUT
 
     - name: Wait for nodes to join and become ready
       shell: bash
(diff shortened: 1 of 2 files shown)


.github/actions/constellation_create/action.yml
```diff
@@ -211,8 +211,8 @@ runs:
       shell: bash
       run: |
         constellation init --force --debug
-        echo "KUBECONFIG=$(pwd)/constellation-admin.conf" >> $GITHUB_OUTPUT
-        echo "MASTERSECRET=$(pwd)/constellation-mastersecret.json" >> $GITHUB_OUTPUT
+        echo "KUBECONFIG=$(pwd)/constellation-admin.conf" | tee -a $GITHUB_OUTPUT
+        echo "WORKSPACE=$(pwd)" | tee -a $GITHUB_OUTPUT
```
daniel-weisse, line 215:
```suggestion
```
The workspace during create should be the same during create and all other actions, I don't think we need to specifically set it just for recover

.github/actions/constellation_create/action.yml
```diff
@@ -211,8 +211,8 @@ runs:
       shell: bash
       run: |
         constellation init --force --debug
-        echo "KUBECONFIG=$(pwd)/constellation-admin.conf" >> $GITHUB_OUTPUT
-        echo "MASTERSECRET=$(pwd)/constellation-mastersecret.json" >> $GITHUB_OUTPUT
+        echo "KUBECONFIG=$(pwd)/constellation-admin.conf" | tee -a $GITHUB_OUTPUT
```
daniel-weisse, line 214: We should not print the Kubernetes config file during the e2e test

.github/actions/e2e_recover/action.yml
```diff
@@ -8,8 +8,8 @@ inputs:
   kubeconfig:
     description: "The kubeconfig for the cluster."
     required: true
-  masterSecret:
-    description: "The master-secret for the cluster."
+  workspace:
+    description: "The argument to use for constellation -C."
```
daniel-weisse, line 12:
```suggestion
```
This can be removed then

.github/actions/e2e_recover/action.yml
```diff
@@ -40,7 +40,7 @@ runs:
         start_time=$(date +%s)
         recovered=0
         while true; do
-            output=$(constellation recover --master-secret=${{ inputs.masterSecret }} --force)
+            output=$(constellation recover -C ${{ inputs.workspace }} --force)
```
daniel-weisse, line 43:
```suggestion
            output=$(constellation recover --force)
```
And this doesn't have to se the workspace

.github/actions/constellation_create/action.yml
derpsteb: this only prints the path.

daniel-weisse (approved):
lgtm if tests pass

PR #2180, opened 2023-08-08
cli: strict input validation on attestation version numbers

elchead:
### Context

For Azure-SEV SNP, we currently accept attestation version values such as:
2.6
09
As a user,  such input should fail config validation.

The version logically should only allow uint values such as: 1 or "1"

### Proposed change(s)

- don't allow float parsings and validate valid uint8 range

### Related issue
- [AB#3271](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3271)

### Checklist

- [ ] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)
- [ ] Add labels (e.g., for changelog category)
- [ ] Is PR title adequate for changelog?
- [ ] Link to Milestone

PR #2187, opened 2023-08-09
internal: fix unmarshalling attestation version numbers from JSON

daniel-weisse:
### Context
Unmarshalling a JSON representation of an AttestationConfig fails since https://github.com/edgelesssys/constellation/pull/2180.
This is because we now always try to unmarshal a given version as a Go string.

JSON has two distinct ways to represent numbers and strings: Numbers are always given without quotation marks, while strings have to use quotation marks.
This means we cannot simply unmarshal a value to string, like we do with YAML.
Additionally, unmarshalling to `any` causes Go to unmarshal numbers to float64.

### Proposed change(s)

- Try to unmarshal attestation version numbers to string, and then to int, when converting from JSON

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/internal/config/attestationversion.go
+++ b/internal/config/attestationversion.go
@@ -64,11 +64,11 @@ func (v *AttestationVersion) UnmarshalJSON(data []byte) (err error) {
 	// Therefore, try to unmarshal to string, and then to int, instead of using type assertions.
 	var unmarshalString string
 	if err := json.Unmarshal(data, &unmarshalString); err != nil {
-		var unmarshalInt int
+		var unmarshalInt int64
 		if err := json.Unmarshal(data, &unmarshalInt); err != nil {
 			return fmt.Errorf("unable to unmarshal to string or int: %w", err)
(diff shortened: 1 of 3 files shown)


internal/config/attestationversion.go
```diff
@@ -58,15 +58,24 @@ func (v AttestationVersion) MarshalJSON() ([]byte, error) {
...
+		unmarshalString = strconv.FormatInt(int64(unmarshalInt), 10)
 	}
-	return v.parseRawUnmarshal(rawUnmarshal)
+
+	return v.parseRawUnmarshal(unmarshalString)
 }
 
 func (v *AttestationVersion) parseRawUnmarshal(str string) error {
-	if strings.HasPrefix(str, "0") {
+	if strings.HasPrefix(str, "0") && len(str) != 1 {
```
elchead, line 78: good catch :)

derpsteb (approved):

PR #2188, opened 2023-08-09
e2e: detect changing idKeyDigests on azure

katexochen:
by setting the Azure SNP enforcement policy to equal in the weekly e2e.
The run should fail when there are unexpected ID Key digests used.

Run on the debug commit fails as expected, as policy is set to equal and the known digests were removed:
https://github.com/edgelesssys/constellation/actions/runs/5808575798/job/15745640059

### Checklist

- [x] Link to Milestone


malt3 (approved):

daniel-weisse (approved):
Remember to remove debug commit

derpsteb (approved):

PR #2201, opened 2023-08-09
cli: remove terraform node group migrations

daniel-weisse:
### Context
Manual terraform migrations were needed for v2.9.0.
When upgrading clusters from v2.9 to v2.10, they are no longer needed, and can therefore be removed from the CLI

### Proposed change(s)

- Remove manual Terraform migrations from the CLI
- Fix upgrade e2e test by making Kubernetes version argument truly optional

### Checklist

- [ ] Link to Milestone


cli/internal/terraform/terraform.go
```diff
@@ -597,7 +597,7 @@ func (c *Client) setLogLevel(logLevel LogLevel) error {
 }
 
 // StateMigration is a manual state migration that is not handled by Terraform due to missing features.
-// TODO(AB#3248): Remove this after we can assume that all existing clusters have been migrated.
+// TODO: Keep or remove? Currently not needed, but might be again in the future.
```
derpsteb, line 600: I am leaning towards "keep" tbh. And making the function no-ops. Otherwise we will have to figure out how to add migrations once again soon. Same with the calls in the cmd pkg. 
Unless you have a plan to add a more refined migration path during the refactoring. And you expect that to finish during 2.11-pre.

elchead: i have no strong opinion, but keeping the logic might make it easier for us. I can't foresee if we will need it again since it seemed like an edge case

elchead (approved):

malt3 (approved):

daniel-weisse:
Waiting for https://github.com/edgelesssys/constellation/pull/2205 to be merged so I can run e2e tests in a different region

derpsteb (approved):

PR #2218, opened 2023-08-11
e2e: upload TCB versions in verify test

katexochen:
### Proposed change(s)

- Upload TCB versions in verify.
The TCP versions are extracted from the MAA token, that itself is taken
from the verify command output. The configapi is adapted to directly
work on the MAA claims JSON.
- Also doing some refactoring on the config API tool
- Successful e2e run: https://github.com/edgelesssys/constellation/actions/runs/5833104416/job/15819807160

### Checklist


hack/configapi/cmd/delete.go
derpsteb, line 1: why remove the `delete` cmd?

katexochen: > It is relatively simple to delete these JSON structs in the S3 GUI, and
we don't need any automatic cleanup job like for images.

.github/actions/e2e_verify/action.yml
```diff
@@ -52,14 +59,35 @@ runs:
...
+            continue
+          fi
+
+          echo "Extracting TCB versions for API update"
+          startMAAToken="Microsoft Azure Attestation Token:"
+          endMAAToken="Verification OK"
+          sed -n "/${startMAAToken}/,/${endMAAToken}/ { /${startMAAToken}/d; /${endMAAToken}/d; p }" <<< "${verifyOut}" > "maa-claims-${node}.json"
+        done
+
+    - name: Upload extracted TCBs
```
elchead, line 83: this should only be run on azure right?

elchead:
> > It is relatively simple to delete these JSON structs in the S3 GUI, and
we don't need any automatic cleanup job like for images. 

I disagree. It's better to use an automatic tool than manually changing the GUI. The process requires not only deleting the json entry, but also the /list entry. Why not keep what we already have?

hack/configapi/cmd/delete.go
elchead: I disagree. It's better to use an automatic tool than manually changing the GUI. The process requires not only deleting the json entry, but also the /list entry. Why not keep what we already have?

katexochen: The code is never executed anywhere, and less code means less maintenance. But i you prefer to keep it, I can revert [777cc67].(https://github.com/edgelesssys/constellation/pull/2218/commits/777cc678a8e9d90e7b1d7eec3fc56edcf144681f)

elchead: Yes it's not meant to only be executed manually on demand, since it should ideally never be necessary to run it.  I advocate to keep it.

derpsteb (approved):
:+1:

katexochen:
@derpsteb should we scope the upload to runs on main only? I think for the TCB versions itself it doesn't matter, but maybe could be a good idea regarding WIP changes on configapi/workflow when executing on a branch?

elchead:
@katexochen good point! @derpsteb did we drop the safety check to only upload a version after we have seen them for a certain number of days?

katexochen:
@elchead it think we always uploaded directly, but only versions older than 2 weeks are used.

katexochen:
Run yesterday failed, as new TCB versions were detected and the CLI hadn't the needed permissions to upload. I've added the login step, but the re-run had the old firmware again. I won't try to force a run on the new firmware.

elchead:
so we wait for tomorrow? Should be safe to run no?

katexochen:
> so we wait for tomorrow? Should be safe to run no?

What should be safe? We don't need to wait for anything, IMO this is ready to merge.

PR #2222, opened 2023-08-14
cli: move helm and terraform out of kubernetes package

daniel-weisse:
### Context
The `kubernetes` package of the CLI currently has dependencies on packages it shouldn't have, as well as offering a public interface way larger in scope than necessary.
More specifically, the `cmd` package of the CLI for most of its use cases calls into the `kubernetes` to perform actions not directly associated with Kubernetes itself. This mostly affects the `terraform`, and `helm` package, both of which are called through the `kubernetes` package in many cases.

### Proposed change(s)

- Remove `terraform` and `helm` from the `kubernetes` package

### Additional info
There will be a follow up PR to further clean up the package. Mostly in regards to the usage of `StableClient` and `DynamicClient`, as well as renaming a couple things.


cli/internal/cmd/upgrade.go
```diff
@@ -23,3 +27,24 @@ func NewUpgradeCmd() *cobra.Command {
...
+type upgradeCmdKind int
+
+const (
+	// upgradeCmdKindCheck corresponds to the upgrade check command.
+	upgradeCmdKindCheck upgradeCmdKind = iota
+	// upgradeCmdKindApply corresponds to the upgrade apply command.
+	upgradeCmdKindApply
+)
+
+func generateUpgradeID(kind upgradeCmdKind) string {
```
elchead, line 41: nit: should iam upgrade use this too?

daniel-weisse: I applied this suggestion, but this also means IAM upgrades are placed in folders similar named to other upgrades.
Is this something we are OK with, or should be adjust this to stay as `iam-<uid>` ?

daniel-weisse: Added a new `upgradeCmdKind` for IAM

elchead (approved):

derpsteb (approved):
I would wait with merging until v2.10 is out.

PR #2235, opened 2023-08-15
cli: parse image and k8s versions as semver

daniel-weisse:
### Context
During `constellation upgrade check`, the Constellation image version, and Kubernetes version are validated to be valid semantic versions, yet we continue passing them around as regular strings.

### Proposed change(s)
- Enforce both versions to be semver by using the concrete data type

### DIFF
--- a/cli/internal/cmd/upgradecheck.go
+++ b/cli/internal/cmd/upgradecheck.go
@@ -286,10 +286,10 @@ func sortedMapKeys[T any](a map[string]T) []string {
 }
 
 // filterImageUpgrades filters out image versions that are not valid upgrades.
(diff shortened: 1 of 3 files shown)


derpsteb:
I guess this works since we don't pass image shortpaths here, only versions(?). LGTM.

Sanity check: https://github.com/edgelesssys/constellation/actions/runs/5935187416

derpsteb:
More sanity checks: https://github.com/edgelesssys/constellation/actions/runs/5935777757

derpsteb (approved):

daniel-weisse:
[e2e Kubernetes upgrade test](https://github.com/edgelesssys/constellation/actions/runs/5947819247/job/16130479728)

PR #2237, opened 2023-08-15
attestation: print ordered measurement verification warnings and errors

daniel-weisse:
### Context
Warnings and errors of measurement verification during an attestation are currently printed unordered.
Additionally, on error, only the first mismatching measurement is reported.

### Proposed change(s)

- Return warnings and errors in order of their indices
- Print all warnings before returning on error
- Return all mismatching measurements on error


malt3 (approved):
Nice job!

PR #2241, opened 2023-08-16
deps: limit Terraform version to FOSS releases

thomasten:
### Context
Starting with v1.6, Terraform will be licensed under BUSL. v1.5 will get critical security updates until end of 2023.

### Proposed change(s)

- Limit Terraform version to open-source releases until it's clear how we proceed.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/terraform/terraform.go
+++ b/cli/internal/terraform/terraform.go
@@ -39,7 +39,7 @@ import (
 )
 
 const (
-	tfVersion         = ">= 1.4.6"
+	tfVersion         = ">= 1.4.6, < 1.6.0"
 	terraformVarsFile = "terraform.tfvars"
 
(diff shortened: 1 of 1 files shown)


elchead:
have you tested that it works with the space? I only read "comma separated list"

thomasten:
> have you tested that it works with the space? I only read "comma separated list"

Yes, I have. Works.

3u13r:
I pushed a fix which enforces enforces the upper version constraint for the downloaded version. The constraints in "LatestRelease" are never used and the truly latest version is downloaded.
One can manually check this by uninstalling all local terraform versions and changing the upper bound to e.g. `<1.5.0`.

elchead (approved):

PR #2232, opened 2023-08-15
cli: refactor kubernetes package

daniel-weisse:
### Context
The CLI Kubernetes package needs some clean up.
The goal of this PR is:
* Make it clearer what the package does: It is supposed to be used for highlevel interactions with Kubernetes resources by the CLI
  * Low level stuff, like creating a resource or updating a resource (single function calls), should be kept in `internal/kubectl`
* Clean up the logic of some commands
* Remove unused functions

### Proposed change(s)

- Rename the `kubernetes` package to `kubecmd` to better highlight its intended use
- Rename `Upgrader` to `KubeCmd`
- Remove exported functions which take a client to interact with the kubernetes API as argument
  - Replaced with methods of the `kubecmd` client
- Replace usage of custom unstructured and structured Kubernetes API client with the client from `internal/kubernetes`
- Rename some variables and functions to better highlighted their functionality and intended use case

PR #2244, opened 2023-08-16
cli: helm install and upgrade unification

elchead:
### Proposed change(s)

-

### Checklist

- [ ] Link to Milestone


elchead:
testing E2E https://github.com/edgelesssys/constellation/actions/runs/5877599252

elchead:
E2E: https://github.com/edgelesssys/constellation/actions/runs/5893838079

cli/internal/helm/actionfactory.go
```diff
@@ -0,0 +1,205 @@
...
+
+type crdClient interface {
+	Initialize(kubeconfig []byte) error
+	ApplyCRD(ctx context.Context, rawCRD []byte) error
+	GetCRDs(ctx context.Context) ([]apiextensionsv1.CustomResourceDefinition, error)
+	GetCRs(ctx context.Context, gvr schema.GroupVersionResource) ([]unstructured.Unstructured, error)
+}
+
+// newActionFactory creates a new action factory for managing helm releases.
+func newActionFactory(kubeConfigPath string, log debugLog) (*actionFactory, error) {
```
elchead, line 49: to handle potential migration actions for charts we could add
`Register(releaseName,preHook,postHook)` to the factory or even just as a function with `[]applyActions` and set the respective hook on the action.

daniel-weisse (commented):
Only reviewed the changes to `cli/internal/cmd` package so far

cli/internal/cmd/upgradeapply.go
```diff
@@ -386,31 +387,72 @@ func (u *upgradeApplyCmd) handleServiceUpgrade(cmd *cobra.Command, conf *config.
...
-			cmd.PrintErrln("WARNING: Upgrading cert-manager will destroy all custom resources you have manually created that are based on the current version of cert-manager.")
-			ok, askErr := askToConfirm(cmd, "Do you want to upgrade cert-manager anyway?")
-			if askErr != nil {
-				return fmt.Errorf("asking for confirmation: %w", err)
+	options := helm.Options{
+		Force:        flags.force,
+		Conformance:  flags.conformance,
+		HelmWaitMode: flags.helmWaitMode,
+	}
+	helmApply := func(allowDestructive bool) error {
```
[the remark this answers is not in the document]
elchead: it would be a lot of parameter passing and the only changing argument is `allowDestructive`. I think its more readable. what do you think?

cli/internal/helm/action.go
```diff
@@ -0,0 +1,161 @@
...
+const (
+	// timeout is the maximum time given per helm action.
+	timeout = 10 * time.Minute
+)
+
+type applyAction interface {
+	Apply(context.Context) error
+	ReleaseName() string
+	IsAtomic() bool
+}
```
daniel-weisse, line 29: This interface (and the structs implementing it) are only used in the helm package, so they should be private.

cli/internal/helm/versionlister.go
```diff
@@ -0,0 +1,145 @@
...
+		return err != nil
+	}, func() error {
+		res, err = action.Run()
+		return err
+	})
+	return
+}
+
+// CurrentVersion returns the version of the currently installed helm release.
+func (c ReleaseVersionClient) CurrentVersion(release string) (semver.Semver, error) {
```
daniel-weisse, line 56:
```suggestion
// currentVersion returns the version of the currently installed helm release.
func (c ReleaseVersionClient) currentVersion(release string) (semver.Semver, error) {
```
Can be kept private as it is only used in the `helm` package

cli/internal/helm/versionlister.go
```diff
@@ -0,0 +1,145 @@
...
+		csiVersions[dep.Name], err = semver.New(dep.Version)
+		if err != nil {
+			return nil, fmt.Errorf("parsing CSI version %q: %w", dep.Name, err)
+		}
+	}
+	return csiVersions, nil
+}
+
+// Versions queries the cluster for running versions and returns a map of releaseName -> version.
+func (c ReleaseVersionClient) Versions() (ServiceVersions, error) {
```
daniel-weisse, line 108: This should be the first function (after struct definition and constructor) to follow our [source file element order](https://github.com/edgelesssys/wiki/blob/master/company/gostyle.md#order-of-elements-in-a-source-file)

cli/internal/helm/backup_test.go
```diff
@@ -6,191 +6,191 @@ SPDX-License-Identifier: AGPL-3.0-only
...
-	return c.crds, nil
-}
-
-func (c stubCrdClient) GetCRs(_ context.Context, _ schema.GroupVersionResource) ([]unstructured.Unstructured, error) {
-	if c.getCRsError != nil {
-		return nil, c.getCRsError
-	}
-	return c.crs, nil
-}
+// import (
```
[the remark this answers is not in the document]
daniel-weisse: #2232 is merged

cli/internal/helm/action.go
elchead: what do you think about extracting an extra `actions` subpkg as suggested in the description?

cli/internal/helm/loader.go
```diff
@@ -87,35 +93,39 @@ func NewLoader(csp cloudprovider.Provider, k8sVersion versions.ValidK8sVersion,
...
+// and then added as a new standalone chart (or as a dependency of another chart),
+// that the new release is installed after the existing one to avoid name conflicts.
+type ReleaseApplyOrder []Release
+
 // LoadReleases loads the embedded helm charts and returns them as a HelmReleases object.
-func (i *ChartLoader) LoadReleases(
-	config *config.Config, conformanceMode bool, helmWaitMode WaitMode, masterSecret uri.MasterSecret,
-	serviceAccURI string, idFile clusterid.File, output terraform.ApplyOutput,
-) (*Releases, error) {
+func (i *chartLoader) LoadReleases(conformanceMode bool, helmWaitMode WaitMode, masterSecret uri.MasterSecret,
```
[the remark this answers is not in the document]
elchead: since the chartLoader is already private there is no point in making `LoadReleases` private is there? I think it makes it more intuitive that this is the exposing function of the struct.

cli/internal/helm/action.go
daniel-weisse: Creating a subpackage sounds like a good idea.
However I'm unsure on how easy/difficult this would be to do while avoiding import cycles.
Maybe implement this as a follow up PR?

cli/internal/cmd/upgradeapply.go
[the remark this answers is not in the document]
elchead: looks good, applying it :)

cli/internal/helm/loader.go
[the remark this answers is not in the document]
daniel-weisse: If any function of the `helm` package returns the (private) `chartLoader`, the caller of that function will be able to call `LoadReleases` on the returned struct.
I feel like exported functions are less likely to be called from the package they are defined in, so this (imo) does not help much with showing a developer what method of `chartLoader` should be called from other parts of the `helm` package.

derpsteb:
For visibilty: re "We should discuss if we should not also manage CRD backups in helm.".

I am assuming we can not simplify the backup code if it is inside the helm pkg. If that assumption holds true, I vote for keeping the backup logic outside the helm pkg, as it is already done in this PR.
Reasons: the helm pkg is already quite big. The backup logic does not interact with the helm SDK. Backups seem like a feature of the Constellation CLI, less of `helm.Apply()`. I could see a `--no-upgrades` flag, which would be easier to implement if upgrades are created in the cmd pkg.

cli/internal/helm/action.go
elchead: created a new ticket

elchead:
running E2E upgrades and sonobuoy:
https://github.com/edgelesssys/constellation/actions/runs/5950370813

daniel-weisse (approved):
Code looks good to me.
We should run e2e test (including upgrades) (for all clouds) before merging.

derpsteb:
Really nice changes imo. Esp the upgrade unittest makes much more sense now.
I only have a few naming suggestions.

cli/internal/helm/helm.go
```diff
@@ -28,6 +28,104 @@ As such, the number of exported functions should be kept minimal.
...
+type Options struct {
+	Conformance      bool
+	HelmWaitMode     WaitMode
+	AllowDestructive bool
+	Force            bool
+}
+
+// ApplyCharts loads the charts and returns the executor to apply them.
+// TODO(elchead): remove validK8sVersion by putting ValidK8sVersion into config.Config, see AB#3374.
+func (h Client) ApplyCharts(conf *config.Config, validK8sversion versions.ValidK8sVersion, idFile clusterid.File, flags Options, tfOutput terraform.ApplyOutput, serviceAccURI string, masterSecret uri.MasterSecret) (Runner, bool, error) {
```
derpsteb, line 90:
```suggestion
func (h Client) PrepareApply(conf *config.Config, validK8sversion versions.ValidK8sVersion, idFile clusterid.File, flags Options, tfOutput terraform.ApplyOutput, serviceAccURI string, masterSecret uri.MasterSecret) (Runner, bool, error) {
```

Not sure what else to call it. I mainly want to move away from the current name, as it suggests that this function Applies something. However, it doesn't. It only talks to the cluster to fetch version information.

cli/internal/helm/helm.go
```diff
@@ -28,6 +28,104 @@ As such, the number of exported functions should be kept minimal.
...
+	helmLoader := newLoader(conf, idFile, validK8sVersion, h.cliVersion)
+	h.log.Debugf("Created new Helm loader")
+	return helmLoader.loadReleases(flags.Conformance, flags.HelmWaitMode, secret,
+		serviceAccURI, tfOutput)
+}
+
+// Runner runs the Helm actions.
+type Runner interface {
+	Run(ctx context.Context) error
+}
```
derpsteb, line 110:
```suggestion
type Applier interface {
	Apply(ctx context.Context) error
}
```

Together with the above comment, this would make it more clear where the action is executed.

cli/internal/helm/versionlister.go
```diff
@@ -0,0 +1,140 @@
...
+		return semver.Semver{}, fmt.Errorf("multiple releases found for %s", release)
+	}
+
+	if rel[0] == nil || rel[0].Chart == nil || rel[0].Chart.Metadata == nil {
+		return semver.Semver{}, fmt.Errorf("received invalid release %s", release)
+	}
+	return semver.New(rel[0].Chart.Metadata.Version)
+}
+
+func (c ReleaseVersionClient) csiVersions() (map[string]semver.Semver, error) {
```
derpsteb, line 110: A comment describing why this function is necessary (it looks very similar to `currentVersion`) would be nice.

cli/internal/helm/helm_test.go
```diff
@@ -107,3 +119,148 @@ func TestMergeMaps(t *testing.T) {
 		})
 	}
 }
+
+func TestHelmApply(t *testing.T) {
```
derpsteb, line 123: It's helpful to use outdated versions (e.g. 1.99.12) where possible. During release preparation it sometimes happens that you have to look for todos related to a specific version. In those cases, the unittests create unnecessary noise.

cli/internal/helm/helm_test.go
```diff
@@ -107,3 +119,148 @@ func TestMergeMaps(t *testing.T) {
...
+func helmListVersion(l *ReleaseVersionStub, releaseName string, installedVersion string) {
+	if installedVersion == "" {
+		l.On("currentVersion", releaseName).Return(semver.Semver{}, errReleaseNotFound)
+		return
+	}
+	v, _ := semver.New(installedVersion)
+	l.On("currentVersion", releaseName).Return(v, nil)
+}
+
+type ReleaseVersionStub struct {
```
derpsteb, line 259: Really nice use of a mock imo. Nice testcase in general.

cli/internal/helm/helm_test.go
```diff
@@ -107,3 +119,148 @@ func TestMergeMaps(t *testing.T) {
...
+		"constellation-services",
+		"constellation-operators",
+		"constellation-csi",
+	}
+	testCases := map[string]struct {
+		clusterMicroServiceVersion string
+		expectedActions            []string
+		clusterCertManagerVersion  *string
+		clusterAWSLBVersion        *string
+		expectNoUpgrade            bool
```
derpsteb, line 136:
```suggestion
		expectUpgrade            bool
```

Less double negation in our heads, more better :D

cli/internal/helm/actionfactory.go
```diff
@@ -0,0 +1,186 @@
...
+	for _, action := range actions {
+		if _, ok := action.(*upgradeAction); ok {
+			includesUpgrade = true
+			break
+		}
+	}
+	return actions, includesUpgrade, errors.Join(upgradeErrs...)
+}
+
+func (a actionFactory) appendNewAction(release Release, force, allowDestructive bool, actions *[]applyAction) error {
```
derpsteb, line 74: When I first read the function I expected a similar interface to Go's builtin `append` function. I.e. `([]applyAction, error)` as return value. I couldn't find a reason why such an interface wouldn't work. This obviously works as well, no need to change sth now.

cli/internal/helm/helm_test.go
elchead: I know ;) I only did it for convenience, otherwise we need to write expectUpgrade in every upgrade which is redundant, since `expectedActions`  makes that implicitly clear. Any suggestion?

3u13r:
> should a cilium upgrade also require the fix + restart logic?

I'm not 100% certain but it's quite unlikely. I'd just move ahead as if it is not needed. Otherwise I need to fix it later when upgrading.

cli/internal/helm/actionfactory.go
elchead: I think its just safer and more readible since you otherwise need to pass a pointer to the slice

cli/internal/helm/helm_test.go
derpsteb: I'd just add the `expectUpgrade: true,`.

3u13r:
The one change owned by me looks fine.

cli/internal/helm/actionfactory.go
derpsteb: This version requires you to pass a pointer. Doing sth like the following doesn't:
```
func appendSth(slice []int) []int {
	return append(slice, 1)
}
```

cli/internal/cmd/upgradeapply.go
```diff
@@ -381,12 +380,31 @@ func (u *upgradeApplyCmd) handleServiceUpgrade(
...
-		flags.conformance, flags.helmWaitMode, secret, serviceAccURI, validK8sVersion, tfOutput,
-	)
-	if errors.Is(err, helm.ErrConfirmationMissing) {
+	options := helm.Options{
+		Force:        flags.force,
+		Conformance:  flags.conformance,
+		HelmWaitMode: flags.helmWaitMode,
+	}
+
+	helmApply := func(allowDestructive bool) (helm.Applier, bool, error) {
```
derpsteb, line 389:
```suggestion
	prepareApply := func(allowDestructive bool) (helm.Applier, bool, error) {
```

cli/internal/helm/versionlister.go
```diff
@@ -0,0 +1,141 @@
...
+		return semver.Semver{}, fmt.Errorf("multiple releases found for %s", release)
+	}
+
+	if rel[0] == nil || rel[0].Chart == nil || rel[0].Chart.Metadata == nil {
+		return semver.Semver{}, fmt.Errorf("received invalid release %s", release)
+	}
+	return semver.New(rel[0].Chart.Metadata.Version)
+}
+
+// csi versions needs special handling because all versions of its subcharts should be gathered.
```
derpsteb, line 110:
```suggestion
// csi versions needs special handling because all versions of its subcharts should be gathered.
// If the CSI chart is not installed, no error is returned because the user can configure if the chart should be installed.
```

derpsteb (approved):

PR #2251, opened 2023-08-16
cli: remove helm management from join-config

daniel-weisse:
### Context

We allow to modify some Kubernetes resources through the Constellation config.
Some of these resources (at the moment only the `join-config` ConfigMap) are also part of our Helm charts.
The Helm upgrade logic does not allow upgrades of charts to the same version (without forcing an update using the `--force` flag.
This requires us to have additional logic outside of Helm to update resources deployed and managed by Helm.

For better logic separation and to avoid updating the same resource multiple times during a `constellation upgrade apply`, these resources should be managed completely separate from Helm.
In this case, the `join-config` should no longer be deployed by Helm, but by our CLI directly.

### Proposed change(s)

- Remove join-config Config Map from our Helm charts
- Manually create the join-config Config Map during `constellation init`
- Add migrations to remove Helm management from the join-config

### Checklist

- [ ] Link to Milestone


malt3 (approved):
I'm not very deep into the CLI code so another review is probably a good idea, but the high level changes LGTM.

daniel-weisse:
Added separate migration functions to `upgrade apply`, can you please review @derpsteb

elchead:
I'd add to the context:
Some K8s resources (at the moment only `join-config`) can be modified through the constellation config, but the helm upgrade logic does not allow same version upgrades. In order to allow changes of such config changes without version change, they should be managed outside of helm.

cli/internal/kubecmd/kubecmd.go
```diff
@@ -193,35 +194,35 @@ func (k *KubeCmd) GetClusterAttestationConfig(ctx context.Context, variant varia
 	return existingAttestationConfig, nil
 }
 
-// UpdateAttestationConfig updates the Constellation cluster's attestation config.
-// A backup of the previous config is created before updating.
-func (k *KubeCmd) UpdateAttestationConfig(ctx context.Context, newAttestConfig config.AttestationCfg) error {
-	// backup of previous measurements
-	joinConfig, err := k.kubectl.GetConfigMap(ctx, constants.ConstellationNamespace, constants.JoinConfigMap)
+// ApplyAttestationConfig creates or updates the Constellation cluster's attestation config.
```
elchead, line 197:
```suggestion
// ApplyAttestationConfig creates or updates the Constellation cluster's attestation config.
// A backup of the previous config is created with the suffix `_backup` in the config map data
```

cli/internal/kubecmd/kubecmd.go
```diff
@@ -193,35 +194,35 @@ func (k *KubeCmd) GetClusterAttestationConfig(ctx context.Context, variant varia
...
+		k.log.Debugf("ConfigMap %q does not exist in namespace %q, creating it now", constants.JoinConfigMap, constants.ConstellationNamespace)
+		if err := k.kubectl.CreateConfigMap(ctx, joinConfigMap(newConfigJSON, measurementSalt)); err != nil {
+			return fmt.Errorf("creating join-config ConfigMap: %w", err)
+		}
+		k.log.Debugf("Created %q ConfigMap in namespace %q", constants.JoinConfigMap, constants.ConstellationNamespace)
+		return nil
 	}
+
+	// create backup of previous config
+	joinConfig.Data[constants.AttestationConfigFilename+"_backup"] = joinConfig.Data[constants.AttestationConfigFilename]
```
elchead, line 219: Why are we changing the backup mechanism again? Then, we also need to upgrade the troubleshooting docs.

cli/internal/kubecmd/kubecmd.go
```diff
@@ -193,35 +194,35 @@ func (k *KubeCmd) GetClusterAttestationConfig(ctx context.Context, variant varia
 	return existingAttestationConfig, nil
 }
 
-// UpdateAttestationConfig updates the Constellation cluster's attestation config.
-// A backup of the previous config is created before updating.
-func (k *KubeCmd) UpdateAttestationConfig(ctx context.Context, newAttestConfig config.AttestationCfg) error {
-	// backup of previous measurements
-	joinConfig, err := k.kubectl.GetConfigMap(ctx, constants.ConstellationNamespace, constants.JoinConfigMap)
+// ApplyAttestationConfig creates or updates the Constellation cluster's attestation config.
+func (k *KubeCmd) ApplyAttestationConfig(ctx context.Context, newAttestConfig config.AttestationCfg, measurementSalt []byte) error {
```
elchead, line 198: Shouldn't this be called ApplyJoinConfig?

cli/internal/kubecmd/kubecmd.go
daniel-weisse: This reverts the backup mechanism back to what it was before v2.10.
I prefer this way because it is less effort for a user to revert an attestation config upgrade, and it is less error prone:
We only have to update a single ConfigMap, instead of creating/updating two.

cli/internal/kubecmd/kubecmd.go
```diff
@@ -193,35 +194,35 @@ func (k *KubeCmd) GetClusterAttestationConfig(ctx context.Context, variant varia
...
-	newConfigJSON, err := json.Marshal(newAttestConfig)
+	joinConfig, err := k.kubectl.GetConfigMap(ctx, constants.ConstellationNamespace, constants.JoinConfigMap)
 	if err != nil {
-		return fmt.Errorf("marshaling attestation config: %w", err)
+		if !k8serrors.IsNotFound(err) {
+			return fmt.Errorf("getting %s ConfigMap: %w", constants.JoinConfigMap, err)
+		}
+
+		k.log.Debugf("ConfigMap %q does not exist in namespace %q, creating it now", constants.JoinConfigMap, constants.ConstellationNamespace)
+		if err := k.kubectl.CreateConfigMap(ctx, joinConfigMap(newConfigJSON, measurementSalt)); err != nil {
```
elchead, line 211: why not use Apply instead of Create + Update?

daniel-weisse: Because we are doing two separate things here.
If a configmap exists, we simply want to update 2 keys in that configmap and update it.
If it does not, we create a new one.
The logic here does not fit well with calling `CoreV1().ConfigMaps().Apply()`

elchead: Why shouldn't this work?

cm, errNotExist := Get()
if errNotExist {
 cm = joinConfigMap(newConfigJSON, measurementSalt)
}
cm.Data =
cm.BinaryData = 
Apply(cm)

cli/internal/cmd/init.go
```diff
@@ -249,6 +257,14 @@ func (i *initCmd) initialize(cmd *cobra.Command, newDialer func(validator atls.V
 		return err
 	}
```
elchead, line 259: could we make the need for this more evident in the code documentation? I'd suggest a more general naming like .ApplyConfigurableNonHelmManagedResources with some doc comment. This way it's clear why we do this in the first place

daniel-weisse: IMO `ApplyConfigurableNonHelmManagedResources` is significantly less clear than `attestationConfigApplier`.
Especially since the client currently does nothing else than applying the attestation config to the cluster.

elchead (approved):

docs/docs/workflows/troubleshooting.md
```diff
@@ -55,14 +55,12 @@ When in doubt, check if the encountered [issue is known](https://github.com/edge
 
 :::
 
-
 :::tip
 
-During an upgrade with modified attestation config, a backup of the current configuration is stored in the `join-config-backup` config map in the `kube-system` namespace. To restore the old attestation config after a failed upgrade, you can copy the attestation config from this resource, put it in your configuration file and retry the upgrade.
+During an upgrade with modified attestation config, a backup of the current configuration is stored in the `join-config` config map in the `kube-system` namespace under the `attestationConfig_backup` key. To restore the old attestation config after a failed upgrade, replace the value of `attestationConfig` with the value from `attestationConfig_backup`. Alternatively you can copy the attestation config from `attestationConfig_backup`, put it in your configuration file and rerun `constellation upgrade apply`.
```
thomasten, line 60: Do you think there's any situation where the alternative action is preferable? If not, we should just scratch that sentence.

thomasten, line 60: How about adding the kubectl commands to replace attestationConfig with the backup?

elchead: I think letting the user change the config is safer because subsequent `upgrade apply` will try to apply the wrong config again.

elchead: @thomasten  I think letting the user change the config is safer because subsequent upgrade apply will try to apply the wrong config again.

daniel-weisse: This would require the user to convert the attestation config from JSON to YAML however.
If a user simply runs `constellation config fetch-measurements` + `constellation upgrade apply`, this problem should never occur in the first place

cli/internal/cmd/upgradeapply.go
```diff
@@ -413,6 +426,34 @@ func (u *upgradeApplyCmd) handleServiceUpgrade(cmd *cobra.Command, conf *config.
 	return err
 }
 
+// migrateFrom2_10 applies migrations necessary for upgrading from v2.10 to v2.11
+// TODO(v2.11): Remove this function after v2.11 is released.
+func migrateFrom2_10(ctx context.Context, kubeUpgrader kubernetesUpgrader) error {
+	// Sanity check to make sure we only run migrations on upgrades with CLI version 2.10 < v < 2.12
+	if !constants.BinaryVersion().MajorMinorEqual(semver.NewFromInt(2, 11, 0, "")) {
```
derpsteb, line 433: Wouldn't it be more robust to check the release version that is currently in the cluster?

daniel-weisse: This check was mostly meant as a sanity check so we don't run unnecessary migrations if we forget to remove code.
In case we want a more general approach to applying migration during `upgrade apply` with extra safety checks etc., I am open to discuss and implement them, but I would like to keep that separate from this PR.

docs/docs/workflows/troubleshooting.md
thomasten: Ok so if changing the config is our preferred way, we should mention that first. (And then we probably wouldn't want to provide the kubectl command for the other approach.)
What do you think @daniel-weisse ?

daniel-weisse:
[e2e Upgrade test](https://github.com/edgelesssys/constellation/actions/runs/5938352287)

docs/docs/workflows/troubleshooting.md
elchead: I agree that manual measurement management should not be a problem to the normal user. So I guess that either way is fine for troubleshooting

PR #2255, opened 2023-08-17
ci: fix Windows e2e test

msanft:
### Context

Windows e2e test was broken due to a renaming of the artifacts. Furthermore it was running on a separate schedule compared to the weekly e2e test, which is not very intuitive. 

### Proposed change(s)

- Use the correct artifact name.
- Call the workflow from `e2e-test-weekly`.
- Add a Teams notification if the workflow fails.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/e2e-windows.yml
+++ b/.github/workflows/e2e-windows.yml
@@ -1,9 +1,8 @@
 name: e2e test windows
 
 on:
+  workflow_call:
   workflow_dispatch:
-  schedule:
-    - cron: "0 3 * * 6" # At 03:00 on Saturday.
 
 jobs:
   build-cli:
@@ -21,13 +20,12 @@ jobs:
           targetOS: "windows"
           targetArch: "amd64"
           enterpriseCLI: true
-          outputPath: "build/constellation.exe"
 
       - name: Upload CLI artifact
         uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
         with:
-          path: "build/constellation.exe"
-          name: "constellation.exe"
+          path: "bazel-bin/cli/cli_enterprise_windows_amd64"
+          name: "constell-exe"
 
   e2e-test:
     name: E2E Test Windows
@@ -42,11 +40,12 @@ jobs:
       - name: Download CLI artifact
         uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
         with:
-          name: "constellation.exe"
+          name: "constell-exe"
 
       - name: Check CLI version
         shell: pwsh
         run: |
+          Move-Item -Path .\cli_enterprise_windows_amd64 -Destination .\constellation.exe
           .\constellation.exe version
 
       - name: Login to Azure (IAM service principal)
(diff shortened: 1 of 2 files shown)


msanft:
test run [here](https://github.com/edgelesssys/constellation/actions/runs/5888304238)

.github/workflows/e2e-windows.yml
```diff
@@ -137,3 +135,15 @@ jobs:
         shell: pwsh
         run: |
           .\constellation.exe iam destroy --debug -y
+
+      - name: Notify about failure
+        if: |
+          failure() && github.event_name == 'workflow_call'
```
katexochen, line 141: should look more like: https://github.com/edgelesssys/constellation/blob/main/.github/workflows/e2e-upgrade.yml#L298-L300

- check for main branch
- pass if event was scheduled to avoid notifications on dispatch triggered workflow caller

.github/workflows/e2e-windows.yml
```diff
@@ -137,3 +135,15 @@ jobs:
...
+        continue-on-error: true
+        uses: ./.github/actions/notify_failure
+        with:
+          projectWriteToken: ${{ secrets.PROJECT_WRITE_TOKEN }}
+          teamsWebhookUri: ${{ secrets.MS_TEAMS_WEBHOOK_URI }}
+          refStream: ${{ !github.event.pull_request.head.repo.fork && github.head_ref || '' }}
```
katexochen, line 147: The should be `/ref/main/stream/debug`

.github/workflows/e2e-windows.yml
```diff
@@ -137,3 +135,15 @@ jobs:
         shell: pwsh
         run: |
           .\constellation.exe iam destroy --debug -y
+
+      - name: Notify about failure
+        if: |
+          failure() && github.event_name == 'workflow_call'
+        continue-on-error: true
+        uses: ./.github/actions/notify_failure
```
katexochen, line 143: Kubernetes version is a required field.

katexochen, line 143: Please trigger a failure in a dispatch run while relaxing the if statement to check that the notification actually works.

katexochen (changes requested):

.github/workflows/e2e-windows.yml
msanft: Do you mean another field here? I don't think it's actually required

katexochen: true

msanft:
https://github.com/edgelesssys/constellation/actions/runs/5901266478/job/16007563219

msanft:
[e2e run](https://github.com/edgelesssys/constellation/actions/runs/5922697486)
[MS Teams notification for a failing run](https://teams.microsoft.com/l/message/19:e38928ad2c5d438d9c905be1c4de3419@thread.skype/1692362319266?tenantId=adb650a8-5da3-4b15-b4b0-3daf65ff7626&groupId=62348c88-6944-413d-9232-9d3adb03009e&parentMessageId=1692362319266&teamName=Edgeless&channelName=e2e-messages&createdTime=1692362319266&allowXTenantAccess=false)

.github/workflows/e2e-windows.yml
katexochen: The second point is still an issue.

katexochen (approved):
lgtm

PR #2256, opened 2023-08-17
cli: clean up terraform package

daniel-weisse:
### Context
Preliminary clean up of the terraform CLI package to prepare for future reworks.

### Proposed change(s)

- Reorder Functions in `cli/internal/terraform/terraform.go` in accordance with [our guidelines](https://github.com/edgelesssys/wiki/blob/master/company/gostyle.md#order-of-elements-in-a-source-file)
- Rename `CreateCluster` and `ApplyIAMConfig` to `ApplyCluster` and `ApplyIAM`
- reduce some code duplication

<!-- (uncomment if applicable)
### Additional info
@malt3 I removed some commented out code in the manual state migration function. Not sure if we want to keep this in the future, or if this part was already obsolete for a while. Please review

### DIFF
--- a/cli/internal/terraform/terraform.go
+++ b/cli/internal/terraform/terraform.go
@@ -458,6 +458,7 @@ func (c *Client) apply(ctx context.Context, logLevel LogLevel) error {
 }
 
 // applyManualStateMigrations applies manual state migrations that are not handled by Terraform due to missing features.
+// This functions expects to be run on an initialized Terraform workspace.
 // Each migration is expected to be idempotent.
 // This is a temporary solution until we can remove the need for manual state migrations.
 func (c *Client) applyManualStateMigrations(ctx context.Context) error {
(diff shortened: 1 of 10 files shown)


cli/internal/terraform/terraform.go
```diff
@@ -560,11 +467,6 @@ func (c *Client) applyManualStateMigrations(ctx context.Context) error {
 		}
 	}
 	return nil
-	// expects to be run on initialized workspace
```
elchead, line 563: since the restriction of running init first still remains, we should bake this requirement into the code signature (e.g. accept an init result) or at least keep a warning at the top?

malt3 (approved):

cli/internal/terraform/terraform.go
daniel-weisse: A migration in the current code may do anything, not limited to interacting with a Terraform file, so I am strongly against adding any kind of hard requirements to executing this function.
In the worst case, this should simply thrown an error.

I think the comment was explicitly for the manual state migration that was commented out here, maybe @malt3 can comment on this?

elchead (approved):

PR #2259, opened 2023-08-17
bazel: fix output paths of container

katexochen:
This fixes the output paths of builds within the container by mounting directories to paths that exist on the host. We also explicitly set the output path in a .bazelrc to the user specific path. The rc file is mounted into the container and overrides the host rc. Also adding automatic stop in case start is called and a containers is already running.
Sym links like bazel-out and paths bazel outputs should generally work with this change.

### Checklist

### DIFF
--- a/bazel/container/container.sh
+++ b/bazel/container/container.sh
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
 function setup {
+  (stopBazelServer && sleep 1) || true
+
   # Ensure that the cache directories exist, so they are not created by docker with root permissions.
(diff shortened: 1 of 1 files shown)


bazel/container/container.sh
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
 function setup {
+  (stopBazelServer && sleep 1) || true
+
```
malt3, line 5: I'm not sure why this is needed

malt3 (approved):
Nit. Otherwise LGTM.

PR #2269, opened 2023-08-21
attestation: use `go-sev-guest` library

msanft:
### Context

We want to use the [`go-sev-guest` library](https://github.com/google/go-sev-guest) to have an easily maintainable interface to the verification / validation of SEV-SNP attestation reports. (For now on Azure only due to deviations in the attestation processes of the CSPs) 

### Proposed change(s)

- Provide a 1:1-translation of our current custom attestation strategy to a `go-sev-guest` policy. However, there are checks which a) we currently perform but do not seem to be expressable in a `go-sev-guest` policy or b) are automatically performed by `go-sev-guest` but not currently done by us. Below lists all occurances of those cases. -> See the sections below for how our previous checks are covered by which checks from `go-sev-guest` 
- **@Reviewers**: In addition to carefully reviewing the code, please pay special attention to the checks listed below and if they make sense and are restrictive enough.

### Checks we currently do but can't be expressed in `go-sev-guest`
- CommittedTCB >= TCB specified in Config: https://github.com/edgelesssys/constellation/blob/590931f3ac98db668b77e24c048eb3e1f9d4a16d/internal/attestation/azure/snp/validator.go#L151-L153
> covered by "LaunchTCB >= Minimum LaunchTCB". This implies that the config now defines the minimum launchTCB. This probably makes more sense. Since this was never documented I think we can silently change this. The new verification is more permissive with SVNs that are newer the the configured, which is good (more robust). ~ @derpsteb  
- LaunchTCB != CommittedTCB: https://github.com/edgelesssys/constellation/blob/590931f3ac98db668b77e24c048eb3e1f9d4a16d/internal/attestation/azure/snp/validator.go#L154-L156
> fine to remove. This is unnecessarily restrictive. Can't remember why it's there. ~ @derpsteb 
- CurrentTCB >= CommittedTCB: https://github.com/edgelesssys/constellation/blob/590931f3ac98db668b77e24c048eb3e1f9d4a16d/internal/attestation/azure/snp/validator.go#L157-L159
> we should set PermitProvisionalFirmware to true so this check is kept. I currently don't see a reason why we would want to prevent this. It should improve robustness during firmware upgrades imo. ~ @derpsteb 
- MAA checks depending on EnforcementPolicy set in the Constellation config:
https://github.com/edgelesssys/constellation/blob/590931f3ac98db668b77e24c048eb3e1f9d4a16d/internal/attestation/azure/snp/validator.go#L190-L216
Might be achievable by ignoring the go-sev-guest IdKey hash check if EnforcementPolicy is MaaFallback or WarnOnly and performing a custom check after. (See current idea of an implementation in the PR)
> we need to keep this. ~ @derpsteb 
-> @msanft : implement custom check after `go-sev-guest` validation
- TCB in VCEK == CommittedTCB 
https://github.com/edgelesssys/constellation/blob/5272e7c86fd4c7dba435cf3c2b9e1d549b3c5a17/internal/attestation/azure/snp/validator.go#L221-L222
This might be replaced by the CurrentTCB >= TCB specified in VCEK check mentioned below.
> Re 5.: I agree that this should be replaced with "CurrentTCB >= TCB specified in VCEK". The current check is overly restrictive. ~ @derpsteb 

### Checks we didn't do before but are automatically done by `go-sev-guest`
- Depending on an allow flag, either CurrentTCB >= CommittedTCB **OR** CurrentTCB != CommittedTCB: https://github.com/google/go-sev-guest/blob/d8660ce1f6ee9699f62ff463e6787e32c318ae1c/validate/validate.go#L394-L399
- LaunchTCB >= Minimum LaunchTCB (provided by us): https://github.com/google/go-sev-guest/blob/d8660ce1f6ee9699f62ff463e6787e32c318ae1c/validate/validate.go#L401
(May be easy for us by just giving in the Minimum TCB specified in our Config)
- ReportedTCB != TCB specified in VCEK: https://github.com/google/go-sev-guest/blob/d8660ce1f6ee9699f62ff463e6787e32c318ae1c/validate/validate.go#L408
- CurrentTCB >= TCB specified in VCEK: https://github.com/google/go-sev-guest/blob/d8660ce1f6ee9699f62ff463e6787e32c318ae1c/validate/validate.go#L409
- ReportedTCB >= Minimum TCB (provided by us): https://github.com/google/go-sev-guest/blob/d8660ce1f6ee9699f62ff463e6787e32c318ae1c/validate/validate.go#L410

### Checklist

- [x] Link to Milestone


internal/attestation/azure/snp/validator.go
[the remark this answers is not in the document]
msanft: > Ideally, we can drop all of our own SNP report validation (including VCEK verification) except for MAA and maybe IDKey.
Yes, I think this is the main point of adopting the library now.

> Regarding IDKey, the implementation may be simpler if we never use the library for it and just keep our current check. 
I think so too. Double-checking with the library has no obvious advantages to me while making the code harder to read.

derpsteb:
I didn't check the code yet as the pipeline is still failing and this is still a draft. 

Regarding the TCB checks.
From "Checks we currently do but can't be expressed in go-sev-guest":
re 1.: fine to remove. covered by "LaunchTCB >= Minimum LaunchTCB". This implies that the config now defines the minimum launchTCB. This probably makes more sense. Since this was never documented I think we can silently change this. The new verification is more permissive with SVNs that are newer the the configured, which is good (more robust).

re 2.: fine to remove. This is unnecessarily restrictive. Can't remember why it's there.

re 3.: we should set `PermitProvisionalFirmware` to true so this check is kept. I currently don't see a reason why we would want to prevent this. It should improve robustness during firmware upgrades imo.

re 4.: we need to keep this.

Regarding the checks that we gain: seems like a good addition to me. I always read the AMD docs in a way that suggests that the PSP firmware enforces these conditions and that a attestation verifier wouldn't have to check them. But this isn't stated anywhere. Better to have them. :+1:

derpsteb:
Re 5.: I agree that this should be replaced with "CurrentTCB >= TCB specified in VCEK". The current check is overly restrictive.

msanft:
latest update: Both unit test `TestTrustedKeyFromSNP` and manual test with an actual cluster report the following error:
```
verifying SNP attestation: report signature verification error: x509: ECDSA verification failure
```
of which I don't know why they report it yet. If anyone has ideas to debug this, please let me know.

thomasten:
To me it seems that the report and the vcek just don't match. Please try using the vcek and chain that the issuer is getting from Azure THIM instead of downloading it from AMD.
If this turns out to work, let's discuss pros and cons of both ways.

msanft:
needs https://github.com/google/go-sev-guest/pull/73

msanft:
One other bug I need to track down: 

A test with an actual cluster returns:

```
Error: init call: rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: validating attestation public key: verifying SNP attestation: report signature verification error: x509: ECDSA verification failure"
```

If I print the report that is being verified in the above trace it works when checking it "manually" in my test environment with the following snippet:

```go
report, err := abi.ReportToProto(testdata.AttestationReport)
if err != nil {
	panic(err)
}

fmt.Println("Report HWID:", hex.EncodeToString(report.GetChipId()))
fmt.Println("TCB level:", kds.DecomposeTCBVersion(kds.TCBVersion(report.GetCurrentTcb())))

attestation, err := verify.GetAttestationFromReport(report, &verify.Options{})
if err != nil {
	panic(err)
}

err = verify.SnpAttestation(attestation, &verify.Options{})
if err != nil {
	panic(err)
}
```

I assume that either somewhere in the build process, the pseudo-version of `go-sev-guest` (including our patch) isn't being respected, or I make some silly mistake at another point

msanft:
^ Possibly caused by an indirect dependency on v0.6.1 of `go-sev-guest` in our `hack` folder

internal/attestation/azure/snp/validator.go
```diff
@@ -52,235 +55,157 @@ func NewValidator(cfg *config.AzureSEVSNP, log attestation.Logger) *Validator {
...
-	}
-
+	// Verify the attestation report's certificates.
+	if err := verify.SnpAttestation(att, &verify.Options{}); err != nil {
+		return nil, fmt.Errorf("verifying SNP attestation: %w", err)
+	}
+
+	// If the enforcement policy is set to MAAFallback or WarnOnly, the check of the IDKey hashes should not be enforced.
+	requireIDBlock := !(v.config.FirmwareSignerConfig.EnforcementPolicy == idkeydigest.MAAFallback ||
+		v.config.FirmwareSignerConfig.EnforcementPolicy == idkeydigest.WarnOnly)
```
daniel-weisse, line 92:
```suggestion
	requireIDBlock := v.config.FirmwareSignerConfig.EnforcementPolicy == idkeydigest.Enforce
```
?

internal/attestation/azure/snp/validator.go
```diff
@@ -52,235 +55,157 @@ func NewValidator(cfg *config.AzureSEVSNP, log attestation.Logger) *Validator {
...
+		TrustedIDKeyHashes: v.config.FirmwareSignerConfig.AcceptedKeyDigests,
+		// The IDKey hashes should not be checked if the enforcement policy is set to MAAFallback or WarnOnly to prevent
+		// an error from being returned because of the TrustedIDKeyHashes validation. In this case, we should perform a
+		// custom check of the MAA-specific values later. Right now, this is a double check, since a custom MAA check
+		// is performed either way.
+		RequireIDBlock: requireIDBlock,
+	}); err != nil {
+		return nil, fmt.Errorf("validating SNP attestation: %w", err)
+	}
+	if requireIDBlock {
```
daniel-weisse, line 129:
```suggestion
	if !requireIDBlock {
```
This check should run when `requireIDBlock` is set to false, correct?

internal/attestation/azure/snp/validator.go
```diff
@@ -52,235 +55,157 @@ func NewValidator(cfg *config.AzureSEVSNP, log attestation.Logger) *Validator {
...
+		// Custom WarnOnly / MAAFallback check of the IDKeyDigests.
+		v.checkIDKeyDigest(ctx, att, instanceInfo.MAAToken, extraData)
```
daniel-weisse, line 131:
```suggestion
		// Custom WarnOnly / MAAFallback check of the IDKeyDigests.
		if err := v.checkIDKeyDigest(ctx, att, instanceInfo.MAAToken, extraData); err != nil {
			return err
		}
```
Return value not checked

internal/attestation/azure/snp/validator.go
```diff
@@ -52,235 +55,157 @@ func NewValidator(cfg *config.AzureSEVSNP, log attestation.Logger) *Validator {
...
 		}
 	}
 
 	if !hasExpectedIDKeyDigest {
+		// IDKeyDigests that were not expected are present, check the enforcement policy and verify against
+		// the MAA if necessary.
 		switch v.config.FirmwareSignerConfig.EnforcementPolicy {
 		case idkeydigest.MAAFallback:
 			v.log.Infof(
 				"configured idkeydigests %x don't contain reported idkeydigest %x, falling back to MAA validation",
```
daniel-weisse, line 166:
```suggestion
				"Configured idkeydigests %x don't contain reported idkeydigest %x, falling back to MAA validation",
```
Log messages should start in upper case

internal/attestation/azure/snp/validator.go
```diff
@@ -52,235 +55,157 @@ func NewValidator(cfg *config.AzureSEVSNP, log attestation.Logger) *Validator {
...
 				v.config.FirmwareSignerConfig.AcceptedKeyDigests,
-				report.IDKeyDigest[:],
+				report.Report.IdKeyDigest,
 			)
 			return v.maa.validateToken(ctx, v.config.FirmwareSignerConfig.MAAURL, maaToken, extraData)
 		case idkeydigest.WarnOnly:
 			v.log.Warnf(
 				"configured idkeydigests %x don't contain reported idkeydigest %x",
```
daniel-weisse, line 173:
```suggestion
				"Configured idkeydigests %x don't contain reported idkeydigest %x",
```

msanft:
> ^ Possibly caused by an indirect dependency on v0.6.1 of `go-sev-guest` in our `hack` folder

Correcting this fixed it.

daniel-weisse (approved):

internal/attestation/azure/snp/validator.go
```diff
@@ -52,235 +55,154 @@ func NewValidator(cfg *config.AzureSEVSNP, log attestation.Logger) *Validator {
...
-
+	// Verify the attestation report's certificates.
+	if err := verify.SnpAttestation(att, &verify.Options{}); err != nil {
+		return nil, fmt.Errorf("verifying SNP attestation: %w", err)
+	}
+
+	// Checks if the attestation report matches the given constraints.
+	// Some constraints are implicitly checked by validate.SnpAttestation:
+	// - the report is not expired
+	if err := validate.SnpAttestation(att, &validate.Options{
```
malt3, line 93: We should check that `VMPL`  is 0.
This is a check that was missing in our old implementation.

internal/attestation/azure/snp/validator.go
```diff
@@ -52,235 +55,154 @@ func NewValidator(cfg *config.AzureSEVSNP, log attestation.Logger) *Validator {
...
+		GuestPolicy: abi.SnpPolicy{
+			Debug: false, // Debug means the VM can be decrypted by the host for debugging purposes and thus is not allowed.
+			SMT:   true,  // Allow Simultaneous Multi-Threading (SMT). Normally, we would want to disable SMT
```
derpsteb, line 96: Did you check this? Because our kernel cmdline on AWS includes `mitigation=nosmt`.

internal/attestation/azure/snp/validator_test.go
```diff
@@ -333,59 +370,265 @@ func TestValidateAk(t *testing.T) {
 	}
 }
 
-func int32ToByte(val uint32) []byte {
-	r := make([]byte, 4)
-	binary.PutUvarint(r, uint64(val))
-	return r
-}
+// TestGetTrustedKey tests the verification and validation of attestation report.
+func TestTrustedKeyFromSNP(t *testing.T) {
```
derpsteb, line 374: It might be worth it to add a custom assert function to each testcase that checks the returned error. This will make the test a bit more brittle because we will have to do string-based testing. But it will prevent scenarios where tests that should fail, fail for the wrong reasons. E.g. because the report was malformed, but we wanted to test the signature verification.

internal/attestation/azure/snp/validator.go
msanft: This implementation is Azure-only. On Azure, I didn't find such an option in their docs

internal/attestation/azure/snp/validator_test.go
msanft: Added some basic assertions. For all values that alter the report and thus also result in an invalid signature, which I didn't notice before. For this, the current (placeholder) assertion is obviously not good. I will probably need to include a valid VCEK / report pair in the testdata package for each test case I guess.

internal/attestation/azure/snp/validator.go
derpsteb: Our images all specify `nosmt` on the kernel cmdline. Except for AWS. But I rechecked and Azure VMs report SMT to be enabled when querying `/sys/devices/system/cpu/smt/control`. I guess this is fine.

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,368 @@ type Validator struct {
...
+	attestationVerifier  attestationVerifier
+	attestationValidator attestationValidator
 
 	config *config.AzureSEVSNP
 
 	log attestation.Logger
 }
 
+type attestationVerifier interface {
+	SnpAttestation(attestation *spb.Attestation, options *verify.Options) error
```
3u13r, line 55:
```suggestion
	SNPAttestation(attestation *spb.Attestation, options *verify.Options) error
```

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,368 @@ type Validator struct {
...
+}
+
+type attestationValidator interface {
+	SnpAttestation(attestation *spb.Attestation, options *validate.Options) error
```
3u13r, line 59:
```suggestion
	SNPAttestation(attestation *spb.Attestation, options *validate.Options) error
```

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,368 @@ type Validator struct {
...
+}
+
+type attestationVerifierImpl struct{}
+
+// SnpAttestation verifies the VCEK certificate as well as the certificate chain of the attestation report.
+func (attestationVerifierImpl) SnpAttestation(attestation *spb.Attestation, options *verify.Options) error {
```
thomasten, line 65: it also verifies the report's signature

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,368 @@ type Validator struct {
...
+	// and inherently trusted. This check should also be performed by go-sev-guest, but as we have to specify both ASK and ARK for
+	// the trusted roots, this needs to be done here as well.
+	rootPool := x509.NewCertPool()
+	rootPool.AddCert(&trustedArk)
+	if _, err := ask.Verify(x509.VerifyOptions{
+		Roots: rootPool,
+	}); err != nil {
+		return nil, fmt.Errorf("verifying ASK certificate: %w", err)
+	}
+	if err := v.attestationVerifier.SnpAttestation(att, &verify.Options{
```
thomasten, line 131: To me it seems that the library does it correctly here
https://github.com/google/go-sev-guest/blob/81d183f8e73dee86e453bcfacb454af3d3880abb/verify/verify.go#L376C1-L376C66
because of Options created as
https://github.com/google/go-sev-guest/blob/81d183f8e73dee86e453bcfacb454af3d3880abb/verify/trust/trust.go#L235-L239
Can you add a test with (self-generated) ASK not matching the ARK? It should fail with the verification here. And if you then remove the verification, it should fail in the library.

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,368 @@ type Validator struct {
...
+			BlSpl:    v.config.BootloaderVersion.Value, // Bootloader
+			TeeSpl:   v.config.TEEVersion.Value,        // TEE (Secure OS)
+			SnpSpl:   v.config.SNPVersion.Value,        // SNP
+			UcodeSpl: v.config.MicrocodeVersion.Value,  // Microcode
+		},
+		// Check that CurrentTCB >= CommittedTCB.
+		PermitProvisionalFirmware: true,
+		// Check if the IDKey hashes in the report are in the list of accepted hashes.
+		TrustedIDKeyHashes: v.config.FirmwareSignerConfig.AcceptedKeyDigests,
+		// The IDKey hashes should not be checked if the enforcement policy is set to MAAFallback or WarnOnly to prevent
```
thomasten, line 175:
```suggestion
		// Check if the IDKey hash in the report is in the list of accepted hashes.
		TrustedIDKeyHashes: v.config.FirmwareSignerConfig.AcceptedKeyDigests,
		// The IDKey hash should not be checked if the enforcement policy is set to MAAFallback or WarnOnly to prevent
```

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,368 @@ type Validator struct {
...
+		if bytes.Equal(digest, report.Report.IdKeyDigest) {
 			hasExpectedIDKeyDigest = true
 			break
 		}
 	}
 
 	if !hasExpectedIDKeyDigest {
+		// IDKeyDigests that were not expected are present, check the enforcement policy and verify against
+		// the MAA if necessary.
 		switch v.config.FirmwareSignerConfig.EnforcementPolicy {
```
thomasten, line 218: fix comment

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,368 @@ type Validator struct {
...
+	// VCEK is the PEM-encoded VCEK certificate for the attestation report.
+	VCEK []byte
+	// CertChain is the PEM-encoded certificate chain for the attestation report.
+	CertChain []byte
+	// AttestationReport is the attestation report from the vTPM (NVRAM) of the CVM.
+	AttestationReport []byte
+	// RuntimeData is the Azure runtime data from the vTPM (NVRAM) of the CVM.
+	RuntimeData []byte
+	// MAAToken is the token of the MAA for the attestation report, used as a fallback
+	// if the IDKeyDigests cannot be verified.
```
thomasten, line 257:
```suggestion
	// if the IDKeyDigest cannot be verified.
```

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,368 @@ type Validator struct {
...
+	vcek, vcekParsingErr := a.parseVCEK()
+	if err != nil {
+		logger.Warnf("Error parsing VCEK: %v", vcekParsingErr)
+	}
+	if vcek != nil {
+		att.CertificateChain.VcekCert = vcek.Raw
+	}
+	// Otherwise, retrieve it from AMD KDS.
+	if att.CertificateChain.VcekCert == nil {
+		logger.Infof("VCEK certificate not present, falling back to retrieving it from AMD KDS")
```
thomasten, line 290: can you explain why we can't use the "fill in missing certs" mechanism of the library?

3u13r (changes requested):

internal/attestation/azure/snp/validator.go
msanft: The [method they expose in their library](https://github.com/google/go-sev-guest/blob/bab68c55ee83798dc313c70cf9d98b61055e6cb9/verify/verify.go#L630-L642) unfortunately always fills in all the certificates, no matter which are already present. There is an [unexposed method](https://github.com/google/go-sev-guest/blob/bab68c55ee83798dc313c70cf9d98b61055e6cb9/verify/verify.go#L577-L628) that would do the job though. I will try to make this a public method upstream

internal/attestation/azure/snp/validator.go
msanft: Re "Adding test":
yes I will add this

Why we need this check at all:
The library unfortunately doesn't allow to "only" specify the ARK (root cert) for establishing trust in the chain. It always requires ASK and ARK to be present, since AMD also publishes them together publicly [here](https://github.com/google/go-sev-guest/blob/bab68c55ee83798dc313c70cf9d98b61055e6cb9/verify/trust/trust.go#L37-L39). In the Constellation config, the user is only able to specify ARK. Therefore, to be able to use this ARK at all, we have to verify that it signs the ASK (which the library does as well, as you said) to be able to not use the embedded certificates from the library and let the user explicitly specify which root cert he wants to trust. We have multiple possible solutions to this:
- Do it as we do, verify that the ARK from the Constellation config signs the received ASK (from THIM or KDS) first, then pass this ARK-ASK-pair (ARK from Constellation config, ASK from THIM or KDS) into the library which checks them against the ARK-ASK-pair (both from THIM or KDS) 
- Put the whole CertChain in the Constellation config, this would probably be most consistent with what AMD "expects you to do", since they publish the whole certchain publicly as well, and not just the ARK, as mentioned above.
- Remove the ARK field from the Constellation config and embed the root certs ourselves in the Constellation CLI. We will have to issue a security fix either way in case of a disaster, therefore I don't see any particular reason to have the field in the config except for transparency reasons.
- **Preferred solution:** Get something upstreamed in the library that will allow you to only specify the ARK for the trusted roots. /etc/profiles/per-user/moritzs/bin/go test -timeout 30s -run ^TestTrustedKeyFromSNP$ github.com/edgelesssys/constellation/v2/internal/attestation/azure/snp would be the relevant code part presumably.

msanft: cc @derpsteb @katexochen

thomasten: Ok with me to pass ARK and ASK. It's just about whether it's required to verify the ASK against the ARK if the library does it as well.

internal/attestation/azure/snp/validator.go
msanft: Opened an issue for this: 
https://github.com/google/go-sev-guest/issues/77

internal/attestation/azure/snp/validator.go
msanft: I assume passing the ASK directly would cause the trust to only be established from there (i.e. ARK wouldn't be respected).  I will inspect the relevant code parts and add the test you've mentioned to clarify this

msanft: Indeed, the library does not seem to error on an invalid ASK if we remove this check and just pass it into the trusted roots.

internal/attestation/azure/snp/validator_test.go
```diff
@@ -333,59 +502,500 @@ func TestValidateAk(t *testing.T) {
...
+			assertion: func(assert *assert.Assertions, err error) {
+				assert.ErrorContains(err, "report VMPL 1 is not 0")
+			},
+		},
+		"invalid ASK": {
+			report:               defaultReport,
+			runtimeData:          defaultRuntimeData,
+			acceptedIDKeyDigests: defaultIDKeyDigest,
+			enforcementPolicy:    idkeydigest.Equal,
+			verifier:             skipVerifier,
```
thomasten, line 900:
```suggestion
			verifier:             defaultVerifier,
```

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,372 @@ type Validator struct {
...
+	// We verify the ASK certificate against the ARK certificate, since the ARK certificate is the root certificate
+	// and inherently trusted. This check should also be performed by go-sev-guest, but as we have to specify both ASK and ARK for
+	// the trusted roots, this needs to be done here as well.
+	rootPool := x509.NewCertPool()
+	rootPool.AddCert(&trustedArk)
+	if _, err := ask.Verify(x509.VerifyOptions{
+		Roots: rootPool,
+	}); err != nil {
+		return nil, fmt.Errorf("verifying ASK certificate: %w", err)
+	}
```
thomasten, line 130:
```suggestion
```

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,372 @@ type Validator struct {
...
+// according to the enforcement policy. If the enforcement policy is set to MAAFallback, the maaToken is validated against the MAA.
+// If the enforcement policy is set to WarnOnly, a warning is logged. If the enforcement policy is set to neither WarnOnly or MAAFallback, an
+// error is returned.
+func (v *Validator) checkIDKeyDigest(ctx context.Context, report *spb.Attestation, maaToken string, extraData []byte) error {
 	hasExpectedIDKeyDigest := false
 	for _, digest := range v.config.FirmwareSignerConfig.AcceptedKeyDigests {
-		if bytes.Equal(digest, report.IDKeyDigest[:]) {
+		if bytes.Equal(digest, report.Report.IdKeyDigest) {
 			hasExpectedIDKeyDigest = true
 			break
```
thomasten, line 211: you can `return nil` here, remove the `hasExpectedIDKeyDigest` variable, and simplify the control flow a bit

internal/attestation/azure/snp/validator.go
msanft: Thanks for pointing this out. I didn't catch the skipped verifier in the test. With the default verifier, this check is done in the library as expected.

internal/attestation/azure/snp/validator.go
thomasten: It's also invoked here
https://github.com/google/go-sev-guest/blob/bab68c55ee83798dc313c70cf9d98b61055e6cb9/verify/verify.go#L548-L557
Wouldn't that suffice for us?

msanft: Actually, we require the ASK to be present before calling the `SnpAttestation` method, due to the trustedRoots thing

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,359 @@ type Validator struct {
...
-	reverseEndian(sigR)
-	reverseEndian(sigS)
-
-	rParam := new(big.Int).SetBytes(sigR)
-	sParam := new(big.Int).SetBytes(sigS)
-	sequence := ecdsaSig{rParam, sParam}
-	sigEncoded, err := asn1.Marshal(sequence)
+	// If the VCEK certificate is present, parse it and format it.
+	vcek, vcekParsingErr := a.parseVCEK()
 	if err != nil {
```
thomasten, line 269:
```suggestion
	if vcekParsingErr != nil {
```

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,359 @@ type Validator struct {
...
-				if err != nil {
-					return fmt.Errorf("unmarshalling microcode version: %w", err)
-				}
-				if certVersion != int(report.CommittedTCB.Microcode) {
-					return fmt.Errorf("bootloader version %d from report does not match VCEK version %d", int(report.CommittedTCB.Microcode), certVersion)
-				}
-			}
-		}
+// parseVCEK parses the VCEK certificate from the instanceInfo into an x509-formatted certificate.
+func (a *azureInstanceInfo) parseVCEK() (vcek *x509.Certificate, retErr error) {
```
thomasten, line 373: I don't see a reason for named return values

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,359 @@ type Validator struct {
...
+	newlinesTrimmed := strings.TrimSpace(string(a.VCEK))
```
thomasten, line 374:
```suggestion
	newlinesTrimmed := strings.TrimSpace(string(a.VCEK))
	if newlinesTrimmed == "" {
		return nil, nil
	}
```
Judging from the caller's usage of this func, do you want this? So that the caller doesn't log an error for an empty vcek?

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,359 @@ type Validator struct {
...
+// If less than 2 certificates are present, only the present certificate is returned.
+// If more than 2 certificates are present, an error is returned.
+func (a *azureInstanceInfo) parseCertChain() (ask, ark *x509.Certificate, retErr error) {
+	newlinesTrimmed := strings.TrimSpace(string(a.CertChain))
+
+	i := 1
+	var rest []byte
+	var block *pem.Block
+	for block, rest = pem.Decode([]byte(newlinesTrimmed)); block != nil; block, rest = pem.Decode(rest) {
+		if i > 3 {
```
thomasten, line 329:
```suggestion
		if i > 2 {
```

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,359 @@ type Validator struct {
...
 	}
 
-	return nil
+	switch {
+	case i == 1:
+		retErr = fmt.Errorf("no PEM blocks found")
+		return
+	case len(rest) != 0:
+		retErr = fmt.Errorf("remaining PEM block is not a valid certificate: %s", rest)
+		return
```
thomasten, line 366:
```suggestion
	case i == 1:
		retErr = fmt.Errorf("no PEM blocks found")
	case len(rest) != 0:
		retErr = fmt.Errorf("remaining PEM block is not a valid certificate: %s", rest)
```

internal/attestation/azure/snp/validator.go
msanft: I think this should be correct here? We expect 2 PEM blocks to be present, and `i` is initialized with `1`. So we would only want to error on the third certificate, i.e. from `i >= 4`

thomasten: 1st iteration (1st cert): i=1, i>2==false
2nd iteration (2nd cert): i=2, i>2==false
3rd iteration (3rd cert): i=3, i>2==true

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,361 @@ type Validator struct {
...
 	}
-
-	report, err := newSNPReportFromBytes(instanceInfo.AttestationReport)
+	att, err := instanceInfo.attestationWithCerts(v.log, v.getter)
 	if err != nil {
 		return nil, fmt.Errorf("parsing attestation report: %w", err)
 	}
 
-	vcek, err := v.validateVCEK(instanceInfo.Vcek, instanceInfo.CertChain)
+	// Verify the attestation report's certificates.
```
thomasten, line 115:
```suggestion
```

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,361 @@ type Validator struct {
...
+		if err != nil {
+			return nil, fmt.Errorf("retrieving VCEK certificate from AMD KDS: %w", err)
+		}
+		att.CertificateChain.VcekCert = vcek
+	}
+
+	// If the certificate chain is present, parse it and format it.
+	ask, ark, certChainParsingErr := a.parseCertChain()
+	if certChainParsingErr != nil {
+		logger.Warnf("Error parsing certificate chain: %v", certChainParsingErr)
```
thomasten, line 289:
```suggestion
	ask, ark, err := a.parseCertChain()
	if err != nil {
		logger.Warnf("Error parsing certificate chain: %v", err)
```
nit: for consistency

go.mod
```diff
@@ -33,6 +33,7 @@ replace (
 
 replace (
 	github.com/edgelesssys/constellation/v2/operators/constellation-node-operator/v2/api => ./operators/constellation-node-operator/api
+	github.com/google/go-sev-guest v0.7.1 => github.com/msanft/go-sev-guest v1.1.0
```
thomasten, line 36: instead of forking, I'd prefer to `require` the latest (or whatever desired) commit from main

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,358 @@ type Validator struct {
...
 	}
-
-	report, err := newSNPReportFromBytes(instanceInfo.AttestationReport)
+	att, err := instanceInfo.attestationWithCerts(v.log, v.getter)
 	if err != nil {
 		return nil, fmt.Errorf("parsing attestation report: %w", err)
 	}
 
-	vcek, err := v.validateVCEK(instanceInfo.Vcek, instanceInfo.CertChain)
+	// Verify the attestation report's certificates.
```
thomasten, line 115: 3f729a2 unintentionally reverts some things

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,358 @@ type Validator struct {
...
-	hasExpectedIDKeyDigest := false
-	for _, digest := range v.config.FirmwareSignerConfig.AcceptedKeyDigests {
-		if bytes.Equal(digest, report.IDKeyDigest[:]) {
-			hasExpectedIDKeyDigest = true
-			break
+// parseCertChain parses the certificate chain from the instanceInfo into x509-formatted ASK and ARK certificates.
+// If less than 2 certificates are present, only the present certificate is returned.
+// If more than 2 certificates are present, an error is returned.
+func (a *azureInstanceInfo) parseCertChain() (ask, ark *x509.Certificate, retErr error) {
+	rest := []byte(strings.TrimSpace(string(a.CertChain)))
```
thomasten, line 322:
```suggestion
	rest := bytes.TrimSpace(a.CertChain)
```

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,358 @@ type Validator struct {
...
-				if certVersion != int(report.CommittedTCB.Microcode) {
-					return fmt.Errorf("bootloader version %d from report does not match VCEK version %d", int(report.CommittedTCB.Microcode), certVersion)
-				}
-			}
-		}
+// parseVCEK parses the VCEK certificate from the instanceInfo into an x509-formatted certificate.
+// If the VCEK certificate is not present, nil is returned.
+func (a *azureInstanceInfo) parseVCEK() (*x509.Certificate, error) {
+	newlinesTrimmed := strings.TrimSpace(string(a.VCEK))
+	if newlinesTrimmed == "" {
```
thomasten, line 374:
```suggestion
	newlinesTrimmed := bytes.TrimSpace(a.VCEK)
	if len(newlinesTrimmed) == 0 {
```

internal/attestation/azure/snp/validator.go
```diff
@@ -36,249 +41,358 @@ type Validator struct {
...
+		// VCEK is not present.
+		return nil, nil
 	}
 
-	return nil
-}
+	block, rest := pem.Decode([]byte(newlinesTrimmed))
```
thomasten, line 379:
```suggestion
	block, rest := pem.Decode(newlinesTrimmed)
```

hack/go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/edgelesssys/constellation/v2/hack
 go 1.20
 
 replace (
+	github.com/google/go-sev-guest v0.7.1 => github.com/msanft/go-sev-guest v1.1.0
```
thomasten, line 6:
```suggestion
```

go.mod
```diff
@@ -77,7 +77,7 @@ require (
 	github.com/go-playground/universal-translator v0.18.1
 	github.com/go-playground/validator/v10 v10.14.1
 	github.com/golang-jwt/jwt/v5 v5.0.0
-	github.com/google/go-sev-guest v0.6.1
+	github.com/google/go-sev-guest v0.7.2-0.20230904235445-81d183f8e73d
```
thomasten, line 80:
```suggestion
	github.com/google/go-sev-guest v0.8.0
```

thomasten (changes requested):

cli/internal/cmd/verify.go
```diff
@@ -459,12 +459,12 @@ func (f *attestationDocFormatterImpl) parseSNPReport(b *strings.Builder, reportB
 	writeIndentfln(b, 2, "Platform Info:")
 	writeIndentfln(b, 3, "Symmetric Multithreading enabled (SMT): %t", platformInfo.SMTEnabled)
 	writeIndentfln(b, 3, "Transparent secure memory encryption (TSME): %t", platformInfo.TSMEEnabled)
-	writeIndentfln(b, 2, "Author Key ID: %x", report.AuthorKeyEn)
+	writeIndentfln(b, 2, "Author Key ID: %x", report.SignerInfo)
```
thomasten, line 462: This line seems to have already been wrong before.
Now, if you want to print the SignerInfo, you should use abi.ParseSignerInfo to get the values.

cli/internal/cmd/verify.go
```diff
@@ -459,12 +459,12 @@ func (f *attestationDocFormatterImpl) parseSNPReport(b *strings.Builder, reportB
...
 	writeIndentfln(b, 2, "Report Data: %x", report.ReportData)
 	writeIndentfln(b, 2, "Measurement: %x", report.Measurement)
 	writeIndentfln(b, 2, "Host Data: %x", report.HostData)
 	writeIndentfln(b, 2, "ID Key Digest: %x", report.IdKeyDigest)
-	writeIndentfln(b, 2, "Author Key Digest: %x", report.AuthorKeyDigest)
+	writeIndentfln(b, 2, "Author Key Digest: %t", ((report.SignerInfo & 1) != 0))
```
thomasten, line 467: This change seems unnecessary and wrong. AuthorKeyDigest is still there.

thomasten (changes requested):

cli/internal/cmd/verify.go
msanft: Yes, this is accidental

derpsteb (approved):
LGTM. I didn't do another deep dive as I am unde the impression Thomas and Paul already did that. I think it's great that you extended the unittests.

internal/attestation/azure/snp/testdata/testdata.go
```diff
@@ -0,0 +1,35 @@
+/*
+Copyright (c) Edgeless Systems GmbH
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+// Package testdata contains testing data for an attestation process.
+package testdata
```
3u13r, line 8: This is not guarded by tags. Does that mean that we also embed those files in release binaries or is go smart enough here to see that they are unused?

msanft: How would I tag the file so that it get's included when running unit tests?

thomasten: As long as this package isn't imported by any files other than *_test.go, it won't be included in release binaries.

3u13r (approved):
LGTM

daniel-weisse (approved):

katexochen (approved):

msanft:
[One last E2E run](https://github.com/edgelesssys/constellation/actions/runs/6261046453), merging afterwards if successful

PR #2277, opened 2023-08-23
ci: fix ccm build when no new version are found

katexochen:
Previous output of findvers.sh would be [""] in case no version were found, now the output is []. Also, GitHub cannot handle empty arrays in the matrix field, so we add an if and check if the array is empty.

### Checklist


katexochen:
:red_circle: https://github.com/edgelesssys/constellation/actions/runs/5946882885
:green_circle: https://github.com/edgelesssys/constellation/actions/runs/5949813508

malt3 (approved):

PR #2278, opened 2023-08-24
ci: idempotent release pipeline

elchead:
### Context

The release pipeline needed manual cleanup steps to be able to retrigger it. This PR makes the pipeline retriggerable without manual steps.

### Proposed change(s)

- force push the release tag to overwrite the existing one
- don't fail when there is no update commit for the container images 

- missing AWS permissions for image deletion are in a PR of the respective repo.

- skip the find temporary branch step in the `on-release` workflow, which failed in all past releases.

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/.github/workflows/on-release.yml
+++ b/.github/workflows/on-release.yml
@@ -27,6 +27,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          fetch-depth: 0 # fetch all history
 
       - name: Determine branch names
         run: |
@@ -35,20 +37,8 @@ jobs:
           PART_MINOR=${WITHOUT_V#*.}
           MAJOR_MINOR=${PART_MAJOR}.${PART_MINOR}
           RELEASE_BRANCH="release/v${MAJOR_MINOR}"
-          WORKING_BRANCH_PREFIX="origin/tmp/${FULL_VERSION}"
+          WORKING_BRANCH="tmp/${FULL_VERSION}"
           echo "RELEASE_BRANCH=${RELEASE_BRANCH}" | tee -a "$GITHUB_ENV"
-          echo "WORKING_BRANCH_PREFIX=${WORKING_BRANCH_PREFIX}" | tee -a "$GITHUB_ENV"
-
-      - name: Find temporary branch
-        id: find-temporary-branch
-        run: |
-          working_branch_count=$(git branch --list -r "${WORKING_BRANCH_PREFIX}*" | wc -l)
-          if [[ "${working_branch_count}" -ne 1 ]]; then
-            echo "More than one (or zero) temporary branches found. Cannot finish transaction."
-            git branch --list "${WORKING_BRANCH_PREFIX}*"
-            exit 1
-          fi
-          WORKING_BRANCH=$(git branch --list "${WORKING_BRANCH_PREFIX}*" | head -n 1 | sed -e 's/^[[:space:]]*//')
           echo "WORKING_BRANCH=${WORKING_BRANCH}" | tee -a "$GITHUB_ENV"
 
       - name: Check if we are strictly ahead of the release branch (if it exists)
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,6 +13,9 @@ on:
         required: true
         default: "minor"
 
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
 jobs:
   verify-inputs:
     name: Verify inputs
@@ -44,8 +47,8 @@ jobs:
 
       - name: Verify temporary branch
         run: |
-          if [[ ! "${WORKING_BRANCH}" =~ ^tmp/v[0-9]+\.[0-9]+\.[0-9]+(/.+)? ]]; then
-            echo "Workflow can only be triggered from a temporary branch in the form of tmp/vX.Y.Z/..."
+          if [[ ! "${WORKING_BRANCH}" =~ ^tmp/v[0-9]+\.[0-9]+\.[0-9] ]]; then
+            echo "Workflow can only be triggered from a temporary branch in the form of tmp/vX.Y.Z"
             exit 1
           fi
 
@@ -176,8 +179,12 @@ jobs:
         run: |
           git config --global user.name "edgelessci"
           git config --global user.email "edgelessci@users.noreply.github.com"
-          git commit -m "deps: update images to ${VERSION}"
-          git push
+          if git diff-index --quiet HEAD --; then
+            echo "No changes to commit"
+          else
+            git commit -m "deps: update images to ${VERSION}"
+            git push
+          fi
 
   os-image:
     name: Build OS image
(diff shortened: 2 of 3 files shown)


dev-docs/workflows/release.md
```diff
@@ -78,58 +78,3 @@ Then wait until the node / Kubernetes upgrades are finished by periodically chec
 3.  If the release is a minor version release, bump the pre-release version in the `version.txt` file.
 4. Update the `fromVersion` in `e2e-test-release.yml` and `e2e-test-weekly.yaml` to the newly released version. To check the current values, run: `grep "fromVersion: \[.*\]" -R .github`.
 5. Reset `upgradeRequiresIAMMigration`  in `iamupgradeapply.go`.
-
```
derpsteb, line 81: I would prefer to keep this documentation. While we shouldn't have to rely on those steps, it might still be handy to have them.

.github/workflows/on-release.yml
```diff
@@ -27,30 +27,18 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
-
+        with:
+          fetch-depth: 0 # fetch all history
```
katexochen, line 31:
```suggestion
          fetch-depth: 0 # fetch all history
          
```

.github/workflows/on-release.yml
```diff
@@ -27,30 +27,18 @@ jobs:
...
-
-      - name: Find temporary branch
-        id: find-temporary-branch
-        run: |
-          working_branch_count=$(git branch --list -r "${WORKING_BRANCH_PREFIX}*" | wc -l)
-          if [[ "${working_branch_count}" -ne 1 ]]; then
-            echo "More than one (or zero) temporary branches found. Cannot finish transaction."
-            git branch --list "${WORKING_BRANCH_PREFIX}*"
-            exit 1
-          fi
```
katexochen, line 50: This isn't idempotent. We should cleanup the old branch before creating the new one instead.

We should also add a concurrency group to cancel ongoing runs of the release before we delete the old branch.

elchead: IMHO this seems a bit overengineered. We would never run 2 concurrent releases would we? We can only retrigger after the existing one is finished.

elchead: Did you write this step? Why is it relevant if there is tmp branch from another release? The release should only be concerned with its own release, or am I wrong?

.github/workflows/on-release.yml
```diff
@@ -62,7 +50,6 @@ jobs:
             echo "The current branch is not strictly ahead of the release branch. Cannot finish transaction without touching release branch history."
             exit 1
           fi
-
       - name: Create or update release branch
         run: git push origin "${WORKING_BRANCH}":"${RELEASE_BRANCH}"
```
katexochen, line 54: WORKING_BRANCH isn't set anymore, is it?

.github/workflows/release.yml
```diff
@@ -251,7 +255,7 @@ jobs:
           git config --global user.name "edgelessci"
           git config --global user.email "edgelessci@users.noreply.github.com"
           git tag -a "${VERSION}" -m "Release ${VERSION}"
-          git push origin "refs/tags/${VERSION}"
+          git push --force origin "refs/tags/${VERSION}" # force push to overwrite existing tag
```
katexochen, line 258: This force push should never be needed, please remove.

.github/workflows/on-release.yml
katexochen: > IMHO this seems a bit overengineered. We would never run 2 concurrent releases would we? We can only retrigger after the existing one is finished.

I don't think it is. This are literally 3 lines.

malt3: The problem is that here we need to find the correct tmp branch. So if there are multiple, it is possible we choose an old stale tmp branch. This code could be simplified by saying the tmp branch name must be `tmp/va.b.c` and not allow a suffix. Allowing multiple branches here is a recipe for disaster.

.github/workflows/on-release.yml
```diff
@@ -27,30 +27,18 @@ jobs:
...
+          fetch-depth: 0 # fetch all history
       - name: Determine branch names
         run: |
           WITHOUT_V=${FULL_VERSION#v}
           PART_MAJOR=${WITHOUT_V%%.*}
           PART_MINOR=${WITHOUT_V#*.}
           MAJOR_MINOR=${PART_MAJOR}.${PART_MINOR}
           RELEASE_BRANCH="release/v${MAJOR_MINOR}"
-          WORKING_BRANCH_PREFIX="origin/tmp/${FULL_VERSION}"
+          WORKING_BRANCH_PREFIX="/tmp/${FULL_VERSION}"
```
malt3, line 39:
```suggestion
          WORKING_BRANCH_PREFIX="tmp/${FULL_VERSION}"
```

.github/workflows/release.yml
malt3: Let's move the tagging to the on-release workflow in a separate PR. We can allow the force for now.

.github/workflows/on-release.yml
elchead: > > IMHO this seems a bit overengineered. We would never run 2 concurrent releases would we? We can only retrigger after the existing one is finished.
> 
> I don't think it is. This are literally 3 lines.

did it, can you check @katexochen ?

.github/workflows/on-release.yml
elchead: done

.github/workflows/on-release.yml
```diff
@@ -35,22 +37,9 @@ jobs:
...
-        run: |
-          working_branch_count=$(git branch --list -r "${WORKING_BRANCH_PREFIX}*" | wc -l)
-          if [[ "${working_branch_count}" -ne 1 ]]; then
-            echo "More than one (or zero) temporary branches found. Cannot finish transaction."
-            git branch --list "${WORKING_BRANCH_PREFIX}*"
-            exit 1
-          fi
-          WORKING_BRANCH=$(git branch --list "${WORKING_BRANCH_PREFIX}*" | head -n 1 | sed -e 's/^[[:space:]]*//')
           echo "WORKING_BRANCH=${WORKING_BRANCH}" | tee -a "$GITHUB_ENV"
-
```
katexochen, line 53: newline

dev-docs/workflows/release.md
```diff
@@ -79,11 +79,8 @@ Then wait until the node / Kubernetes upgrades are finished by periodically chec
 4. Update the `fromVersion` in `e2e-test-release.yml` and `e2e-test-weekly.yaml` to the newly released version. To check the current values, run: `grep "fromVersion: \[.*\]" -R .github`.
 5. Reset `upgradeRequiresIAMMigration`  in `iamupgradeapply.go`.
 
-## Pipeline cleanup
-
-When testing changes to the pipeline or when the pipeline fails during a release it might become necessary to clean up the images created by the pipeline.
-These are the necessary steps.
-
+## Troubleshooting: Pipeline cleanup
+No manual steps should be necessary anymore but in case you encounter issues, these are instructions to do some cleanup steps manually:
```
katexochen, line 83:
```suggestion
No manual steps should be necessary anymore but in case you encounter issues, create a ticket to fix it. These are instructions to do some cleanup steps manually:
```

elchead:
> code looks good, please provide a successful run from the latest commit. 👍

cherry picked the commit and re-triggering the pipeline now:
https://github.com/edgelesssys/constellation/actions/runs/5953130091

elchead:
> > code looks good, please provide a successful run from the latest commit. 👍
> 
> cherry picked the commit and re-triggering the pipeline now: https://github.com/edgelesssys/constellation/actions/runs/5953130091

@katexochen the pipeline went through until the tests

malt3 (approved):

katexochen (approved):

PR #2281, opened 2023-08-24
rfc: Constellation state file

daniel-weisse:
### Context
We plan to replace our current cluster ID file (`constellation-id.json`) with a new "state file", which also holds values we currently have to read from Terraform output files during `init`.
This file should also make it clear when what sections of the file are expected to be available for reading.
This is an RFC for the file spec.

### Additional info
- Naming (and the proposed file spec) should not be considered final. I would like reviewers to provide any suggestions to improve the format. Especially the `runtimeResources` key is more of a quick placeholder than anything we probably want in the final implementation


rfc/state-file.md
```diff
@@ -0,0 +1,94 @@
...
+The file should be considered "read-only" by users, if Constellation's cloud resource were created using the Constellation CLI.
+If the cloud resources were created manually by a user, they will need to manually fill in some of the state file's fields.
+
+This will allow us to more easily decouple resource creation/management from Constellation.
+Meaning it should be much easier for Constellation users to have different parties create and update cloud resources and IAM using their own tooling (e.g. Terraform),
+and only use the Constellation CLI for Kubernetes cluster management (Kubernetes versions, service versions, image upgrades etc.).
+
+## Format
+
+The state file will be in either YAML or JSON format, and contain a version number to keep track of necessary migrations or breaking formats.
```
elchead, line 30: I advocate for YAML:
- more readable, can add comments, we already use yaml for the config file

rfc/state-file.md
```diff
@@ -0,0 +1,94 @@
...
+* Runtime resources
+
+    cluster ID, measurement salt etc.
+    This field will be automatically populated during `constellation init`
+
+YAML file example:
+
+```yaml
+version: "v1"
+cloudResources:
```
elchead, line 56: one thought: if only `cloudResources` are ever thought to be touched by the user shouldn't we let the user create a minimal file and have a separate file for `runtimeResources`?

rfc/state-file.md
```diff
@@ -0,0 +1,94 @@
...
+        IPCidrPod: "192.0.2.0/24"
+runtimeResources:
+    clusterID: "00112233445566778899AABBCCDDEEFF"
+    ownerID: "00112233445566778899AABBCCDDEEFF"
+    measurementSalt: "c2VjcmV0Cg=="
+```
+
+## Updates to the state file
+
+Values depending on Terraform, that change during a Constellation upgrade, should be automatically updated by the CLI if the cloud resources are not manually managed by the user.
```
elchead, line 82: Another reason that file separation seems cleaner to me:
- with manual resources the cloudResources file is entirely managed by the user
- the other file is only managed by the CLI

elchead:
we could name the `runtimeResources` `constellationResources` to indicate that they are managed by the CLI?

daniel-weisse:
I would prefer to not manage multiple files, but if the majority agrees we can split the state file into CSP resources and CLI resources

rfc/state-file.md
```diff
@@ -0,0 +1,94 @@
...
+This will allow us to more easily decouple resource creation/management from Constellation.
+Meaning it should be much easier for Constellation users to have different parties create and update cloud resources and IAM using their own tooling (e.g. Terraform),
+and only use the Constellation CLI for Kubernetes cluster management (Kubernetes versions, service versions, image upgrades etc.).
+
+## Format
+
+The state file will be in either YAML or JSON format, and contain a version number to keep track of necessary migrations or breaking formats.
+The file should be split into multiple sections for each relevant resource.
+
+Information that is only used to upgrade Terraform resources (e.g. IAM) should not be part of this file.
```
msanft, line 33: Why is that?

msanft (commented):
Format looks good to me, as you said, we should still think about naming here though

rfc/state-file.md
```diff
@@ -0,0 +1,94 @@
...
+During a cluster's lifecycle, Constellation needs to keep track of multiple different values and state information of the cluster,
+like loadbalancer IPs or UID of the cluster.
+
+If this information is configurable by a user, and required to set up the cluster, the user saves it in Constellation's config file.
+Information that is not configurable, and that is generated by the cluster during the creation or init phase is either saved in Terraform output files,
+or in the cluster ID file (`constellation-id.json`).
+
+This creates problems since the relevant information may be split over multiple files, and it is not always clear at what point certain information is available in the cluster ID file.
+Additionally, commands like `upgrade` and `init` currently depend on Terraform files being present in the working directory,
+which may not be the case if the cloud resources were manually created.
```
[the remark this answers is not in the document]
derpsteb: Agreed that it would be nice if we can remove the need for tf state in the workspace with this file. Ideally we would have a command in the CLI to parse TF state into Constellation state, imo.
Sth like `constellation config parse-state <tfstate-path>`.

rfc/state-file.md
```diff
@@ -0,0 +1,94 @@
...
+Values depending purely on Kubernetes/node state should be automatically updated by the CLI.
+
+Breaking format changes should be automatically handled for `runtimeResources`.
+They should be automatically handled for `cloudResources` if the resources are managed by the Constellation CLI,
+otherwise we should provide documentation and/or tooling for migrating old file formats.
+
+## Migrating from terraform outputs + id file
+
+The Constellation version implementing the state file should provide an upgrade path for moving the Terraform outputs and cluster id file content into the new state file.
+This should happen at the beginning of `constellation upgrade apply`.
```
derpsteb, line 94: I feel like we could make this a separate command from the beginning. That would be a step towards user-manager TF. And I feel like we have to implement the functionality anyways. Forces us to design the state migration code as a separated entity.

rfc/state-file.md
derpsteb, line 1: I think it would be nice to write down the design goals for the file in a separate section `# Design goals`.
This would make it easier to assess the rest of the RFC.

From what I am seeing so far that would be: 
- have a single source of truth the Constellation CLI relies on to read/store information for lifecycle management.
- remove the dependency on a tfstate file during runtime. Whatever infra management tool is used to create the infra - the Constellation CLI consumes the infra state from the statefile.

rfc/state-file.md
derpsteb: I would like to not have multiple files. My understanding is that part of the problem we want to address with the statefile is that we have multiple files.

I agree with the point on safety and feel like we should make it obvious to the user what can be changed and what not.

rfc/state-file.md
```diff
@@ -0,0 +1,94 @@
...
+and only use the Constellation CLI for Kubernetes cluster management (Kubernetes versions, service versions, image upgrades etc.).
+
+## Format
+
+The state file will be in either YAML or JSON format, and contain a version number to keep track of necessary migrations or breaking formats.
+The file should be split into multiple sections for each relevant resource.
+
+Information that is only used to upgrade Terraform resources (e.g. IAM) should not be part of this file.
+
+The following is a list of proposed sections:
```
derpsteb, line 35: Rename to "Writeable" and "Readonly"? It's a bit clumsy, but should make things obvious. And I feel like the current categories aren't that precisely defined, yet.

rfc/state-file.md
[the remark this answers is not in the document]
daniel-weisse: Agreed, but implementation details and integration of parsing of existing terraform state is out of scope for this rfc

rfc/state-file.md
daniel-weisse: The information we keep track of in the state file should be agnostic to how that information was created.
If we require some information to upgrade Terraform state, it should be kept in the Constellation config file instead.

rfc/state-file.md
```diff
@@ -0,0 +1,94 @@
...
+
+* Cloud resources
+
+    Information about resources we currently create using Terraform.
+    Essentially, we want to record the information we currently write to the cluster ID file during `create`,
+    as well as the Terraform output during `create`.
+
+    This field is expected to be manually set if cloud resources were manually deployed by a user.
+    If cloud resource were created with the Constellation CLI (using `constellation create`),
```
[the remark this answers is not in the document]
daniel-weisse: I feel like adding a `Generated Resources: DO NOT EDIT` to the things a user shouldn't touch manually should be enough.

rfc/state-file.md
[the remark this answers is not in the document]
daniel-weisse: While the idea sounds nice, I don't think that is easily possible without us potentially accidentally corrupting/destroying resources of the user.
The only way I could see this work was if the user used the Terraform files from our CLI to generated the resources, at which point they might as well have used the CLI directly.

rfc/state-file.md
```diff
@@ -0,0 +1,94 @@
...
+## Updates to the state file
+
+Values depending on Terraform, that change during a Constellation upgrade, should be automatically updated by the CLI if the cloud resources are not manually managed by the user.
+If they are manually managed, the user has to update the values on their own.
+
+Values depending purely on Kubernetes/node state should be automatically updated by the CLI.
+
+Breaking format changes should be automatically handled for `runtimeResources`.
+They should be automatically handled for `cloudResources` if the resources are managed by the Constellation CLI,
+otherwise we should provide documentation and/or tooling for migrating old file formats.
```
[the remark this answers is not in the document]
daniel-weisse: That is the goal described here.
The "manual" migration is only for e.g. Terraform created by the user through a third party tool.

rfc/state-file.md
daniel-weisse: A couple things about this:
First, we need to handle the id file separately anyways, which probably shouldn't happen in a CLI command dedicated to parsing terraform state.

I'm also not yet 100% sold on the idea of implementing a command to parse Terraform state files for the information needed for the Constellation state file.
While it could be a nice convenience for people who do use Terraform, ideally the state file would allow users to use any method for creating their cloud and IAM resources.
Adding a command specifically for Terraform would be a nice convenience, but IMO this shouldnt need to be a high priority.

rfc/state-file.md
daniel-weisse: Writeable and Readonly are IMO not a good fit, since this would only describe how a user can interact with them, but not what these resources actually are.
The name of these keys should describe accurately what these resources are and ideally also at what point in Constellation's lifecycle they are available.
Part of the problem with the current id file is that we write multiple different values to the same struct, but at different times.
This makes it unclear during which command what information is expected to be available from the id file.
The state file should fix this issue, and I dont think Writeable and Readonly would do that.

rfc/state-file.md
daniel-weisse: I feel like this is already taken care of by the `# Proposal` section, no?

malt3 (approved):
This sounds like the right solution to me.

derpsteb (approved):

rfc/state-file.md
```diff
@@ -0,0 +1,101 @@
...
+Breaking format changes should be automatically handled for `runtimeResources`.
+Breaking format changes should be automatically handled for `cloudResources` if the resources are managed by the Constellation CLI.
+Otherwise, we should provide documentation and/or tooling for migrating old file formats.
+
+## Migrating from terraform outputs + id file
+
+The Constellation version implementing the state file should provide an upgrade path for moving the Terraform outputs and cluster id file content into the new state file.
+This should happen at the beginning of `constellation upgrade apply`.
+
+We might want to implement a CLI command to automatically create the state file from and existing Terraform state.
```
3u13r, line 100:
```suggestion
We might want to implement a CLI command to automatically create the state file from an existing Terraform state.
```

3u13r (approved):
LGTM

elchead:
> we could name the `runtimeResources` `constellationResources` to indicate that they are managed by the CLI?

Is there any update on the naming or we keep `runtimeResources`?

rfc/state-file.md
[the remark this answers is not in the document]
elchead: can u add this to the rfc?

elchead (approved):

daniel-weisse:
> Is there any update on the naming or we keep `runtimeResources`?

What about `infrastructure` instead of `cloudResources`?
I'm still not sure about what to do with `runtimeResources`. `constellationResources` is a bit to vague IMO. Especially since technically everything in this file can be considered a "Constellation resources". I also thought about naming it `clusterResources`, but that also doesn't do the best job of highlighting that these values are only available once the cluster is up and running.

daniel-weisse:
Added some preliminary doc comments to the file example, and renamed `runtimeResources` to `clusterValues`

malt3:
> What about `infrastructure` instead of `cloudResources`? 

`infrastructure` sounds great 👍

rfc/state-file.md
```diff
@@ -0,0 +1,113 @@
...
+Where possible, links to documentation for specific keys should be provided to further explain their usage.
+
+Each section and key should have a comment dedicated to explaining what it does, where to get it from, and if it is to be considered read-only,
+or if a user has to supplement a value themselves (if they manage their own infrastructure).
+
+Information that is only used to upgrade Terraform resources (e.g. IAM) should not be part of this file.
+
+The following is a list of proposed sections:
+
+* Cloud resources
```
thomasten, line 43:
```suggestion
* Infrastructure
```

rfc/state-file.md
```diff
@@ -0,0 +1,113 @@
...
+
+    This field is expected to be manually set if cloud resources were manually deployed by a user.
+    If cloud resource were created with the Constellation CLI (using `constellation create`),
+    it will be populated automatically.
+
+* Runtime resources
+
+    Cluster ID, measurement salt etc.
+    This field will be automatically populated during `constellation init`.
+    This entry should be marked with `DO NOT EDIT` to make it clear these resources are expected to not be touched by a user manually.
```
thomasten, line 57:
```suggestion
* Cluster values

    Cluster ID, measurement salt etc.
    This field will be automatically populated during `constellation init`.
    This entry should be marked with `DO NOT EDIT` to make it clear these values are expected to not be touched by a user manually.
```

rfc/state-file.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Updates to the state file
+
+Values depending on Terraform, that change during a Constellation upgrade, should be automatically updated by the CLI if the cloud resources are not manually managed by the user.
+If they are manually managed, the user has to update the values on their own.
+
+Values depending purely on Kubernetes/node state should be automatically updated by the CLI.
+
+Breaking format changes should be automatically handled for `runtimeResources`.
+Breaking format changes should be automatically handled for `cloudResources` if the resources are managed by the Constellation CLI.
```
thomasten, line 104:
```suggestion
Breaking format changes should be automatically handled for `clusterValues`.
Breaking format changes should be automatically handled for `infrastructure` if the resources are managed by the Constellation CLI.
```

thomasten (approved):

PR #2282, opened 2023-08-24
cli: cleanup terraform files when create fails

elchead:
### Context

If `constellation create` fails due to terraform validation (the validation specified in variables.tf) triggering an abort, a constellation-terraform folder is left behind. No resources should be created at this point and thus no state folder should be left behind.

Example:
set cluster name with capital letter in gcp:

```

Error: "name" ("adrianGCP-5cfcb64c") doesn't match regexp "^(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)$"
```

### Proposed change(s)

- defer the cleanup of tf files

### Related issue
- [AB#3282](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3282)

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/cli/internal/cloudcmd/rollback.go
+++ b/cli/internal/cloudcmd/rollback.go
@@ -18,7 +18,7 @@ import (
 
 // rollbacker does a rollback.
 type rollbacker interface {
-	rollback(ctx context.Context, logLevel terraform.LogLevel) error
+	rollback(ctx context.Context, w io.Writer, logLevel terraform.LogLevel) error
 }
 
 // rollbackOnError calls rollback on the rollbacker if the handed error is not nil,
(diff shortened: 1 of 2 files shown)


daniel-weisse:
I don't think we should implement this, since this would delete all Terraform files even when the failure cause was a transient error (e.g. network related).
Properly cleaning up the resource could then be rather difficult.

I would rather print a hint for the user to remove the empty directory (if it exists), or add extra logic to only remove the directory if its actually empty.

elchead:
> I don't think we should implement this, since this would delete all Terraform files even when the failure cause was a transient error (e.g. network related). Properly cleaning up the resource could then be rather difficult.
> 
> I would rather print a hint for the user to remove the empty directory (if it exists), or add extra logic to only remove the directory if its actually empty.

There is no easy way to check if there is state. 
I will just print a user hint

PR #2287, opened 2023-08-26
cli: store kubernetes version as strong type in config

elchead:
### Context

Storing the K8s version as validated type saves a lot of parameter passing and centralizes the validation logic

### Proposed change(s)

- store K8s version as weakly validated type (ignoring patch version) in the config
- ~allow to specify K8s version without patch (e.g. 1.27) as valid config value and resolve the patch version~
- allow to specify a patch version in `config generate` (e.g. 1.27.4)
- add more unit test cases

### Related issue
- [AB#3374](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3374)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/configgenerate.go
+++ b/cli/internal/cmd/configgenerate.go
@@ -133,7 +133,7 @@ func parseGenerateFlags(cmd *cobra.Command) (generateFlags, error) {
 	if err != nil {
 		return generateFlags{}, fmt.Errorf("parsing Kubernetes flag: %w", err)
 	}
-	resolvedVersion, err := versions.NewValidK8sVersion(k8sVersion, true) // allow versions without specified patch
+	resolvedVersion, err := versions.NewValidK8sVersion(k8sVersion, true)
 	if err != nil {
 		return generateFlags{}, fmt.Errorf("resolving Kubernetes version from flag: %w", err)
 	}
--- a/cli/internal/cmd/configgenerate.go
+++ b/cli/internal/cmd/configgenerate.go
@@ -14,13 +14,12 @@ import (
 	"github.com/edgelesssys/constellation/v2/internal/attestation/variant"
 	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
 	"github.com/edgelesssys/constellation/v2/internal/config"
-	"golang.org/x/mod/semver"
-
 	"github.com/edgelesssys/constellation/v2/internal/constants"
 	"github.com/edgelesssys/constellation/v2/internal/file"
 	"github.com/edgelesssys/constellation/v2/internal/versions"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
+	"golang.org/x/mod/semver"
 )
 
 func newConfigGenerateCmd() *cobra.Command {
--- a/cli/internal/cmd/configgenerate.go
+++ b/cli/internal/cmd/configgenerate.go
@@ -13,14 +13,13 @@ import (
 	"github.com/edgelesssys/constellation/v2/cli/internal/cmd/pathprefix"
 	"github.com/edgelesssys/constellation/v2/internal/attestation/variant"
 	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
-	"github.com/edgelesssys/constellation/v2/internal/compatibility"
 	"github.com/edgelesssys/constellation/v2/internal/config"
+
 	"github.com/edgelesssys/constellation/v2/internal/constants"
 	"github.com/edgelesssys/constellation/v2/internal/file"
 	"github.com/edgelesssys/constellation/v2/internal/versions"
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
-	"golang.org/x/mod/semver"
 )
 
 func newConfigGenerateCmd() *cobra.Command {
@@ -35,7 +34,7 @@ func newConfigGenerateCmd() *cobra.Command {
 		ValidArgsFunction: generateCompletion,
 		RunE:              runConfigGenerate,
 	}
-	cmd.Flags().StringP("kubernetes", "k", semver.MajorMinor(config.Default().KubernetesVersion), "Kubernetes version to use in format MAJOR.MINOR")
+	cmd.Flags().StringP("kubernetes", "k", string(config.Default().KubernetesVersion), "Kubernetes version to use in format MAJOR.MINOR")
 	cmd.Flags().StringP("attestation", "a", "", fmt.Sprintf("attestation variant to use %s. If not specified, the default for the cloud provider is used", printFormattedSlice(variant.GetAvailableAttestationVariants())))
 
 	return cmd
@@ -43,7 +42,7 @@ func newConfigGenerateCmd() *cobra.Command {
 
 type generateFlags struct {
 	pf                 pathprefix.PathPrefixer
-	k8sVersion         string
+	k8sVersion         versions.ValidK8sVersion
 	attestationVariant variant.Variant
 }
 
@@ -124,19 +123,6 @@ func createConfig(provider cloudprovider.Provider) *config.Config {
 	return res
 }
 
-// supportedVersions prints the supported version without v prefix and without patch version.
-// Should only be used when accepting Kubernetes versions from --kubernetes.
-func supportedVersions() string {
-	builder := strings.Builder{}
-	for i, version := range versions.SupportedK8sVersions() {
-		if i > 0 {
-			builder.WriteString(" ")
-		}
-		builder.WriteString(strings.TrimPrefix(semver.MajorMinor(version), "v"))
-	}
-	return builder.String()
-}
-
 func parseGenerateFlags(cmd *cobra.Command) (generateFlags, error) {
 	workDir, err := cmd.Flags().GetString("workspace")
 	if err != nil {
@@ -144,11 +130,11 @@ func parseGenerateFlags(cmd *cobra.Command) (generateFlags, error) {
 	}
 	k8sVersion, err := cmd.Flags().GetString("kubernetes")
 	if err != nil {
-		return generateFlags{}, fmt.Errorf("parsing kuberentes flag: %w", err)
+		return generateFlags{}, fmt.Errorf("parsing Kubernetes flag: %w", err)
 	}
-	resolvedVersion, err := resolveK8sVersion(k8sVersion)
+	resolvedVersion, err := versions.NewValidK8sVersion(k8sVersion, true) // allow versions without specified patch
 	if err != nil {
-		return generateFlags{}, fmt.Errorf("resolving kuberentes version from flag: %w", err)
+		return generateFlags{}, fmt.Errorf("resolving Kubernetes version from flag: %w", err)
 	}
 
 	attestationString, err := cmd.Flags().GetString("attestation")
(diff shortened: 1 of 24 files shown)


cli/internal/kubecmd/kubecmd.go
```diff
@@ -124,17 +124,13 @@ func (k *KubeCmd) UpgradeNodeVersion(ctx context.Context, conf *config.Config, f
 	nodeVersion.Spec.ImageReference = imageReference
 	nodeVersion.Spec.ImageVersion = imageVersion.Version()
 
+	// TODO(elchead): why?
```
daniel-weisse, line 127: Because we don't to force a user to upgrade their Kubernetes version with each release.
If we publish a new Constellation version which ships with Kubernetes `v1.27.4`, and a user has their cluster running on Kubernetes `v1.27.2`, they should be able to run `constellation upgrade apply` without necessarily increasing their Kubernetes patch version,

docs/docs/reference/cli.md
```diff
@@ -79,7 +79,7 @@ constellation config generate {aws|azure|gcp|openstack|qemu|stackit} [flags]
 ```
   -a, --attestation string   attestation variant to use {aws-sev-snp|aws-nitro-tpm|azure-sev-snp|azure-trustedlaunch|gcp-sev-es|qemu-vtpm}. If not specified, the default for the cloud provider is used
   -h, --help                 help for generate
-  -k, --kubernetes string    Kubernetes version to use in format MAJOR.MINOR (default "v1.27")
+  -k, --kubernetes string    Kubernetes version to use in format MAJOR.MINOR (default "v1.27.4")
```
daniel-weisse, line 82: You would need to check in with other devs here, but I think the idea was for users to not have to specify a patch version when using this flag

derpsteb: That is indeed the point of the flag.

elchead: fixed

elchead:
Discuss: supporting versions without patch as K8s version might be worth putting in the release notes

cli/internal/cmd/upgradeapply.go
```diff
@@ -149,7 +149,7 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, upgradeDir string, fl
 			}
 		}
 	}
-	validK8sVersion, err := validK8sVersion(cmd, conf.KubernetesVersion, flags.yes)
+	conf.KubernetesVersion, err = validK8sVersion(cmd, string(conf.KubernetesVersion), flags.yes)
```
derpsteb, line 152: This should be persisted, shouldn't it?

cli/internal/cmd/init.go
```diff
@@ -153,6 +152,11 @@ func (i *initCmd) initialize(
 	if err != nil {
 		return err
 	}
+	conf.KubernetesVersion, err = versions.NewValidK8sVersion(string(conf.KubernetesVersion), true)
+	// cfg validation does not check k8s patch version since upgrade may accept an outdated patch version.
```
derpsteb, line 156:
```suggestion
	// cfg validation does not check k8s patch version since upgrade may accept an outdated patch version.
	conf.KubernetesVersion, err = versions.NewValidK8sVersion(string(conf.KubernetesVersion), true)
```

cli/internal/cmd/init.go
```diff
@@ -170,10 +174,7 @@ func (i *initCmd) initialize(
 
 	// config validation does not check k8s patch version since upgrade may accept an outdated patch version.
 	// init only supported up-to-date versions.
```
derpsteb, line 176:
```suggestion
```

cli/internal/cmd/upgradeapply.go
elchead: why should it be persisted? If a user specifies 'v1.27' I'm not sure if we should add the patch version. He might want to always have the latest patch version. Otherwise I would expect that he sets the complete version (e.g. through config generate)

cli/internal/cmd/upgradecheck.go
```diff
@@ -576,7 +576,7 @@ func (v *versionUpgrade) writeConfig(conf *config.Config, fileHandler file.Handl
 		conf.MicroserviceVersion = v.newServices
 	}
 	if len(v.newKubernetes) > 0 {
-		conf.KubernetesVersion = v.newKubernetes[0]
+		conf.KubernetesVersion = versions.ValidK8sVersion(v.newKubernetes[0])
```
derpsteb, line 579:
```suggestion
		conf.KubernetesVersion = versions.NewValidK8sVersion(v.newKubernetes[0], true)
```
I know, this should be safe. But better safe than sorry. And if others see this use of the type they may start imitating it.

elchead:
ran 
E2E upgrade: https://github.com/edgelesssys/constellation/actions/runs/5998073530 
E2E sonobuoy: https://github.com/edgelesssys/constellation/actions/runs/5999162031

internal/versions/versions.go
```diff
@@ -36,29 +36,102 @@ func SupportedK8sVersions() []string {
...
 type ValidK8sVersion string
 
 // NewValidK8sVersion validates the given string and produces a new ValidK8sVersion object.
+// It accepts a major minor version (e.g. 1.26) and transforms it into a supported patch version (e.g. 1.26.7).
+// It also accepts a full version (e.g. 1.26.7) and validates it.
 // Returns an empty string if the given version is invalid.
 // strict controls whether the patch version is checked or not.
-// If strict is false, the patch version is ignored and the returned
-// ValidK8sVersion is a supported patch version for the given major.minor version.
+// If strict is false, the patch version is ignored and the returned.
```
daniel-weisse, line 55:
```suggestion
// If strict is false, the patch version is ignored and valid version with patch is returned.
```
Is this correct?

internal/versions/versions.go
```diff
@@ -36,29 +36,102 @@ func SupportedK8sVersions() []string {
...
+	return nil
+}
+
+// resolveK8sPatchVersion takes the user input from --kubernetes and transforms a MAJOR.MINOR definition into a supported
+// MAJOR.MINOR.PATCH release.
+func resolveK8sPatchVersion(k8sVersion string) (string, error) {
+	if !semver.IsValid(k8sVersion) {
+		return "", fmt.Errorf("kubernetes flag does not specify a valid semantic version: %s", k8sVersion)
+	}
+	fmt.Println("COMPARE", semver.MajorMinor(k8sVersion), k8sVersion)
```
daniel-weisse, line 94:
```suggestion
```

internal/versions/versions.go
```diff
@@ -36,29 +36,102 @@ func SupportedK8sVersions() []string {
...
+	if err != nil {
+		return fmt.Errorf("unsupported Kubernetes version %s, supported versions are %s", version, strings.Join(SupportedK8sVersions(), ", "))
+	}
+	*v = valid
+	return nil
+}
+
+// resolveK8sPatchVersion takes the user input from --kubernetes and transforms a MAJOR.MINOR definition into a supported
+// MAJOR.MINOR.PATCH release.
+func resolveK8sPatchVersion(k8sVersion string) (string, error) {
```
daniel-weisse, line 90: A function in `internal/` should not know about flags from `cli/`.
Either adjust this function to remove references to said flags, or move this function back into `cli/`

elchead:
@derpsteb I reverted the support for no patch specification in the config due to complicated case handling. I think it's not worth to support it, but have saved the stash if we decide to do so in the future

daniel-weisse (approved):

cli/internal/cmd/init_test.go
```diff
@@ -125,15 +125,34 @@ func TestInitialize(t *testing.T) {
...
 		"k8s version without v works": {
 			provider:      cloudprovider.Azure,
 			idFile:        &clusterid.File{IP: "192.0.2.1"},
 			initServerAPI: &stubInitServer{res: &initproto.InitResponse{Kind: &initproto.InitResponse_InitSuccess{InitSuccess: testInitResp}}},
-			configMutator: func(c *config.Config) { c.KubernetesVersion = strings.TrimPrefix(string(versions.Default), "v") },
+			configMutator: func(c *config.Config) {
+				res, err := versions.NewValidK8sVersion(strings.TrimPrefix(string(versions.Default), "v"), true)
+				if err != nil {
+					panic("invalid k8s version")
+				}
```
derpsteb, line 140:
```suggestion
				require.NoError(t, err)
```

internal/versions/versions.go
```diff
@@ -36,29 +36,104 @@ func SupportedK8sVersions() []string {
...
 	}
 	if !supported {
-		return "", fmt.Errorf("invalid Kubernetes version: %s; supported versions are %v", k8sVersion, SupportedK8sVersions())
+		return "", fmt.Errorf("invalid Kubernetes version: %s; supported versions are %v", prefixedVersion, SupportedK8sVersions())
+	}
+	return ValidK8sVersion(prefixedVersion), nil
+}
+
+// hasPatchVersion returns if the given version has specified a patch version.
+func hasPatchVersion(version string) bool {
```
derpsteb, line 70: This should be further down in the file. [wiki](https://github.com/edgelesssys/wiki/blob/master/company/gostyle.md#order-of-elements-in-a-source-file)

derpsteb (approved):

internal/versions/versions.go
```diff
@@ -36,29 +36,99 @@ func SupportedK8sVersions() []string {
 	return validVersionsSorted
 }
 
+// SupportedValidK8sVersions returns a typed list of supported Kubernetes versions.
+func SupportedValidK8sVersions() (res []ValidK8sVersion) {
+	for _, v := range SupportedK8sVersions() {
+		res = append(res, ValidK8sVersion(v))
+	}
+	return
+}
```
3u13r, line 45: I don't think that we should implement a function here which is only used in test and which simply converts the type. This is quite confusing as `SupportedK8sVersions()` already outputs only supported versions.  
Another way would be to replace the `SupportedK8sVersions()` function with `SupportedValidK8sVersions()`

3u13r (changes requested):

internal/versions/versions.go
derpsteb: What about making this a private function in *_test.go?

elchead: @derpsteb  i did that now :)

elchead: @3u13r can u have another look?

3u13r (dismissed):

cli/internal/cmd/init.go
```diff
@@ -168,12 +172,7 @@ func (i *initCmd) initialize(
...
-	// config validation does not check k8s patch version since upgrade may accept an outdated patch version.
-	// init only supported up-to-date versions.
-	k8sVersion, err := versions.NewValidK8sVersion(compatibility.EnsurePrefixV(conf.KubernetesVersion), true)
-	if err != nil {
-		return err
-	}
+	k8sVersion := conf.KubernetesVersion
 	i.log.Debugf("Validated k8s version as %s", k8sVersion)
 	if versions.IsPreviewK8sVersion(k8sVersion) {
 		cmd.PrintErrf("Warning: Constellation with Kubernetes %v is still in preview. Use only for evaluation purposes.\n", k8sVersion)
```
3u13r, line 178: Having a look at this again, I think we should go with option 2 https://github.com/edgelesssys/constellation/pull/2287#discussion_r1318484238.
I don't think it is readable to have 2 places where we refer to `conf.KubernetesVersion` and validate it in the first part while printing that we've validated it in the second part.

3u13r (changes requested):

cli/internal/cmd/init.go
elchead: @3u13r fixed it

elchead:
@3u13r can u give it another look pls when u find a moment?

cli/internal/cmd/configgenerate.go
```diff
@@ -13,14 +13,14 @@ import (
 	"github.com/edgelesssys/constellation/v2/cli/internal/cmd/pathprefix"
 	"github.com/edgelesssys/constellation/v2/internal/attestation/variant"
 	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
-	"github.com/edgelesssys/constellation/v2/internal/compatibility"
 	"github.com/edgelesssys/constellation/v2/internal/config"
+	"golang.org/x/mod/semver"
+
```
3u13r, line 18: Please remove the space so that `"golang.org/x/mod/semver"` is placed at the old location at the bottom of the list by the formatter.

3u13r (approved):
One small nit, otherwise LGTM

PR #2305, opened 2023-09-04
cli: save Helm charts to disk before running upgrades

daniel-weisse:
### Context
`constellation upgrade apply` upgrades and/or installs chart to the cluster.
What charts are used for this, and how the values look can be rather hard to determine for a user, making potential troubleshooting difficult.
Additionally, the values used to apply these upgrades are not shown or saved anywhere.

### Proposed change(s)
- Save the charts, values and override values used during `constellation upgrade apply` to disk before running said operation
  - Adds a new `SaveCharts` method to the Helm interface
  - Charts are saved to `constellation-upgrade/upgrade-<timestamp>-<uid>`
  - Each action defines a `SaveChart` method that is called by the interface
    - install actions save their charts to `<dir>/helm-install`
    - upgrade actions save their charts to `<dir>/helm-upgrade`
  - override values (values generated by our CLI and are merged with the charts values) are saved as `overrides.yaml` in the chart directory

### Additional info
- I copied and adjusted the [`SaveDir`](https://pkg.go.dev/helm.sh/helm/v3/pkg/chartutil#SaveDir) function from helm's  `chartutil` package because the original function saved dependencies as tar files instead of writing chart files directly. 
- [AB#3409](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3409)

### Checklist

- [ ] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs): 
  - See [Azure boards ticket](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3413)
- [x] Link to Milestone

### DIFF
--- a/cli/internal/helm/action.go
+++ b/cli/internal/helm/action.go
@@ -9,9 +9,11 @@ package helm
 import (
 	"context"
 	"fmt"
+	"path/filepath"
 	"time"
 
 	"github.com/edgelesssys/constellation/v2/internal/constants"
+	"github.com/edgelesssys/constellation/v2/internal/file"
 
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/cli"
@@ -24,6 +26,7 @@ const (
 
 type applyAction interface {
 	Apply(context.Context) error
+	SaveChart(chartsDir string, fileHandler file.Handler) error
 	ReleaseName() string
 	IsAtomic() bool
 }
@@ -94,6 +97,11 @@ func (a *installAction) Apply(ctx context.Context) error {
 	return nil
 }
 
+// SaveChart saves the chart to the given directory under the `install/<chart-name>` subdirectory.
+func (a *installAction) SaveChart(chartsDir string, fileHandler file.Handler) error {
+	return saveChart(a.release, chartsDir, fileHandler)
+}
+
 func (a *installAction) apply(ctx context.Context) error {
 	_, err := a.helmAction.RunWithContext(ctx, a.release.Chart, a.release.Values)
 	return err
@@ -145,6 +153,11 @@ func (a *upgradeAction) Apply(ctx context.Context) error {
 	return nil
 }
 
+// SaveChart saves the chart to the given directory under the `upgrade/<chart-name>` subdirectory.
+func (a *upgradeAction) SaveChart(chartsDir string, fileHandler file.Handler) error {
+	return saveChart(a.release, chartsDir, fileHandler)
+}
+
 func (a *upgradeAction) apply(ctx context.Context) error {
 	_, err := a.helmAction.RunWithContext(ctx, a.release.ReleaseName, a.release.Chart, a.release.Values)
 	return err
@@ -158,4 +171,15 @@ func (a *upgradeAction) ReleaseName() string {
 // IsAtomic returns true if the action is atomic.
 func (a *upgradeAction) IsAtomic() bool {
 	return a.helmAction.Atomic
+}
+
+func saveChart(release Release, chartsDir string, fileHandler file.Handler) error {
+	if err := saveChartToDisk(release.Chart, chartsDir, fileHandler); err != nil {
+		return fmt.Errorf("saving chart %s to %q: %w", release.ReleaseName, chartsDir, err)
+	}
+	if err := fileHandler.WriteYAML(filepath.Join(chartsDir, release.Chart.Metadata.Name, "overrides.yaml"), release.Values); err != nil {
+		return fmt.Errorf("saving override values for chart %s to %q: %w", release.ReleaseName, filepath.Join(chartsDir, release.Chart.Metadata.Name), err)
+	}
+
+	return nil
 }
--- a/cli/internal/cmd/init_test.go
+++ b/cli/internal/cmd/init_test.go
@@ -232,6 +232,10 @@ func (s stubRunner) Apply(_ context.Context) error {
 	return nil
 }
 
+func (s stubRunner) SaveCharts(_ string, _ file.Handler) error {
+	return nil
+}
+
 func TestGetLogs(t *testing.T) {
 	someErr := errors.New("failed")
 
--- a/cli/internal/cmd/upgradeapply.go
+++ b/cli/internal/cmd/upgradeapply.go
@@ -422,6 +422,13 @@ func (u *upgradeApplyCmd) handleServiceUpgrade(
 		}
 	}
 
+	// Save the Helm charts for the upgrade to disk
+	chartDir := filepath.Join(upgradeDir, "helm-charts")
+	if err := executor.SaveCharts(chartDir, u.fileHandler); err != nil {
+		return fmt.Errorf("saving Helm charts to disk: %w", err)
+	}
+	u.log.Debugf("Helm charts saved to %s", chartDir)
+
 	if includesUpgrades {
 		u.log.Debugf("Creating backup of CRDs and CRs")
 		crds, err := u.kubeUpgrader.BackupCRDs(cmd.Context(), upgradeDir)
--- a/cli/internal/helm/BUILD.bazel
+++ b/cli/internal/helm/BUILD.bazel
@@ -6,6 +6,7 @@ go_library(
     srcs = [
         "action.go",
         "actionfactory.go",
+        "chartutil.go",
         "ciliumhelper.go",
         "helm.go",
         "loader.go",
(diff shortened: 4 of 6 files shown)


cli/internal/helm/action.go
```diff
@@ -9,21 +9,28 @@ package helm
...
 	// timeout is the maximum time given per helm action.
-	timeout = 10 * time.Minute
+	timeout         = 10 * time.Minute
+	installChartDir = "helm-install"
+	upgradeChartDir = "helm-upgrade"
 )
 
 type applyAction interface {
 	Apply(context.Context) error
+	SaveChart(upgradeDir string, fileHandler file.Handler) error
```
derpsteb, line 33:
```suggestion
	SaveChart(fileHandler file.Handler) error
```

What was the original motivation for differentiating between `helm-install` and `helm-upgrade`? 
The content of both folders should be the same, irrespective of `init` or `upgrade apply`. Saving the files to different folders introduces a (admittedly small) difference between `init` and `upgrade`. I think this isn't necessary?

cli/internal/helm/action.go
```diff
@@ -159,3 +176,84 @@ func (a *upgradeAction) ReleaseName() string {
...
+}
+
+// saveChartToDisk saves a chart as files in a directory.
+//
+// This takes the chart name, and creates a new subdirectory inside of the given dest
+// directory, writing the chart's contents to that subdirectory.
+// Dependencies are written using the same format, instead of writing them as tar files
+//
+// View the SaveDir implementation in chartutil as reference: https://github.com/helm/helm/blob/3a31588ad33fe3b89af5a2a54ee1d25bfe6eaa5e/pkg/chartutil/save.go#L40
+func saveChartToDisk(c *chart.Chart, dest string, fileHandler file.Handler) error {
```
derpsteb, line 198: What speaks against using the SDK's `Save` or `SaveDir` functions? I understand it's an extra step for the user to unpack the charts. But they only need to do that to modify the charts. The helm cli supports loading packaged charts. 

Would be less maintenance burden for us if we did use the SDK directly.

cli/internal/helm/action.go
daniel-weisse: They are not saved to different directories for `init` or `upgrade`, but depending on if it is a new chart that has not been installed yet, or if it is a chart that will be upgrade.
A `constellation upgrade apply` can write both directories.
My reasoning for this was to make it more clear for end users what charts are updated by an upgrade and what charts have to be newly installed.

cli/internal/helm/action.go
daniel-weisse: I would personally prefer writing the files directly, instead of having tar files of the charts. (The code for `SaveDir` also hasn't been touched in the last 4 years, so I think it is unlikely there will be any relevant changes in the future)

cli/internal/helm/action.go
derpsteb: We can tell users to just use `helm upgrade --install`? We will have to resolve the difference between `init` and `upgrade` very soon anyways, so I would prefer to do it now.

cli/internal/helm/action.go
derpsteb: Even if it didn't change I would prefer to use the SDK's interface. Let's let @malt3 decide.

cli/internal/helm/action.go
daniel-weisse: > We can tell users to just use `helm upgrade --install`?

That would be possible.

> We will have to resolve the difference between `init` and `upgrade`

Again, this has nothing to do with init and upgrade, this is purely to differentiate if a chart will be installed as a new chart (because it does not yet exist in the cluster) or if it has to be upgraded. The `helm-install` directory may also exist on `constellation upgrade apply` if we added new charts to the release.
My reasoning for keeping them in separate folder was to highlight what is installed new and what is upgraded, which could make trouble shooting easier.

cli/internal/helm/action.go
elchead: I also agree with @derpsteb. What is the disadvantage of storing tars? Afais, the files will only be checked in rare cases anyways?

elchead:
This could be a good chance to add a unit test with some default setups for the CSPs and to check that the chart is rendered as expected.
-> covers the feature
-> covers the helm rendering

Happy to help out

elchead:
I think this feature should be released with https://github.com/edgelesssys/constellation/pull/2310 to be able to actually skip the apply.

elchead (changes requested):
lgtm besides open points

cli/internal/helm/action.go
daniel-weisse: My reasoning for saving the files instead of tar was that it is way easier to look at the chart that way.
Since @malt3 was the one who brought this feature up, I would like to hear his opinion on this.
If we are OK with using tars, I would opt to just use [`chartutil.Save`](https://pkg.go.dev/helm.sh/helm/v3/pkg/chartutil#Save) which saves everything as tar, not just the dependencies.

daniel-weisse:
> This could be a good chance to add a unit test with some default setups for the CSPs and to check that the chart is rendered as expected.

I'm not sure how well this would fit a unit test.
The helm code interacts with the file system directly using `os.Create` etc., which makes unit tests suboptimal.
Additionally, I am not really sure what this test should cover, as saving the chart to disk is mostly handled by an external dependency (even more so if we go with just saving charts as tar files).

> -> covers the helm rendering

Saving the charts to disk does not actually perform any rendering of the charts. We do not template the charts, we simply save the chart we previously loaded from the embedded file system.
The only exceptions are updating the version numbers in `Chart.yaml` for charts we packaged, and generating an additional `overrides.yaml` with the values we loaded dynamically depending on the users config file.
At this point we could maybe check the values we write to disk as `overrides.yaml`, but again, this would simply check the output of a `fileHandler.WriteYaml` call on values we have previously loaded in other functions.

> Can we document this behavior

Please view the linked ticket in the PR description

> Can we document this at the same time? (The user can review the helm chart files and skip the helm chart application through `--skip-phases=helm`)

This is not what this feature implementation was made for. It is intended to allow for better troubleshooting in case of failing to upgrade helm charts during `constellation upgrade apply`.
An additional "review the proposed changes step" is something we have already talked about at some occasions, and will likely be implemented in the future.

cli/internal/helm/action.go
malt3: One reason I want to store the helm charts to disk is that I want to be able to modify the helm values (if necessary) because they are no longer usable. The example we saw in a user deployment was a reference to an old Azure service account that had to be replaced due to a customer-internal IAM change.
So making these easily editable (and storing yaml files directly) is preferable.

daniel-weisse: The values are stored separately tho (at least our overrides), they are not saved by the helm `chartutil` code.

malt3: Oh that makes sense. In that case my preference is still slightly towards being able to read/edit the helm charts easily since in very rare cases, a customer may actually need to edit the charts.

malt3 (approved):
LGTM and testing was successful.

elchead:
> > Can we document this at the same time? (The user can review the helm chart files and skip the helm chart application through `--skip-phases=helm`)
> 
> This is not what this feature implementation was made for. It is intended to allow for better troubleshooting in case of failing to upgrade helm charts during `constellation upgrade apply`. An additional "review the proposed changes step" is something we have already talked about at some occasions, and will likely be implemented in the future.

Would it be valuable to add this in AB#3413 though since we now support this review?

elchead:
> > This could be a good chance to add a unit test with some default setups for the CSPs and to check that the chart is rendered as expected.
> 
> I'm not sure how well this would fit a unit test. The helm code interacts with the file system directly using `os.Create` etc., which makes unit tests suboptimal. Additionally, I am not really sure what this test should cover, as saving the chart to disk is mostly handled by an external dependency (even more so if we go with just saving charts as tar files).
> 
> > -> covers the helm rendering
> 
> Saving the charts to disk does not actually perform any rendering of the charts. We do not template the charts, we simply save the chart we previously loaded from the embedded file system. The only exceptions are updating the version numbers in `Chart.yaml` for charts we packaged, and generating an additional `overrides.yaml` with the values we loaded dynamically depending on the users config file. At this point we could maybe check the values we write to disk as `overrides.yaml`, but again, this would simply check the output of a `fileHandler.WriteYaml` call on values we have previously loaded in other functions.

Fair point, only real rendering (through dry-run) could probably provide any value and also OS makes the testing less ideal.
Let's defer it

elchead (approved):

PR #2307, opened 2023-09-05
ci: use virtee project for sev-snp-measure-go

derpsteb:
### Context

We have added our sev-snp-measure port to the virtee org. To keep it up to date we should use it.

### Proposed change(s)

- Switch from our sev-snp-measure-go repo to the virtee one. Pipeline [run](https://github.com/edgelesssys/constellation/actions/runs/6086300397) :green_circle: 
### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/aws-snp-launchmeasurement.yml
+++ b/.github/workflows/aws-snp-launchmeasurement.yml
@@ -52,8 +52,8 @@ jobs:
 
     - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
       with:
-        repository: edgelesssys/sev-snp-measure-go.git
-        ref: main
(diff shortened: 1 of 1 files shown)


.github/workflows/aws-snp-launchmeasurement.yml
```diff
@@ -52,7 +52,7 @@ jobs:
 
     - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
       with:
-        repository: edgelesssys/sev-snp-measure-go.git
+        repository: virtee/sev-snp-measure-go.git
         ref: main
```
thomasten, line 56: should we pin this to a commit hash given that we aren't the only ones anymore that can commit to the repo?

thomasten (approved):

PR #2312, opened 2023-09-06
csi: let constructor take care of setting up cryptsetup

daniel-weisse:
### Context
The common constructor for the `deviceMapper` interface used by the CSI package is defined in our internal package only.
This causes problems for external dependents of the CSI package.
Additionally, there is no good reason to let the caller handle setting up cryptsetup.

### Proposed change(s)
- Let the constructor (`cryptmapper.New()`) handle setting up cryptsetup, instead of expecting an interface which implements the required functionality.


malt3 (approved):
LGTM

PR #2310, opened 2023-09-06
cli: new flag to skip phases of upgrade

elchead:
### Context

Currently it is not possible to select / deselect specific upgrades (terraform, helm, k8s, node image).

### Proposed change(s)

- add a new `--skip-phases` flag that accepts a list of phases to skip during the upgrade

### Related issue
- [AB#3407](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3407)

### Checklist

- [x] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)
- [x] Add labels (e.g., for changelog category)
- [x] Is PR title adequate for changelog?
- [x] Link to Milestone

PR #2313, opened 2023-09-06
doc: `--skip-flag` in the upgrade workflow

elchead:
### Proposed change(s)

-

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/docs/docs/workflows/upgrade.md
+++ b/docs/docs/workflows/upgrade.md
@@ -62,6 +62,13 @@ When applying an upgrade, backup files of Constellation-managed Custom Resource 
 You can use the Terraform state backup to restore previous resources in case an upgrade misconfigured or erroneously deleted a resource.
 You can use the Custom Resource (Definition) backup files to restore Custom Resources and Definitions manually (e.g., via `kubectl apply`) if the automatic migration of those resources fails.
 
+::note
+
+For advanced users: the upgrade consists of several phases that can be individually skipped through the `--skip-phases` flag.
+The phases are `infrastracture` for the cloud resource management through Terraform, `helm` for the chart management of the microservices, `node` for OS image upgrades, and `k8s` for Kubernetes version upgrades.
+
+:::
+
 ## Check the status
 
 Upgrades are asynchronous operations.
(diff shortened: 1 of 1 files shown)


docs/docs/workflows/upgrade.md
```diff
@@ -62,6 +62,13 @@ When applying an upgrade, backup files of Constellation-managed Custom Resource
 You can use the Terraform state backup to restore previous resources in case an upgrade misconfigured or erroneously deleted a resource.
 You can use the Custom Resource (Definition) backup files to restore Custom Resources and Definitions manually (e.g., via `kubectl apply`) if the automatic migration of those resources fails.
 
+::note
+
+For advanced users: the upgrade consists of several phases that can be individually skipped through the `--skip-phases` flag.
+The phases are `infrastracture` for the cloud resource management through Terraform, `helm` for the chart management of the microservices, `node` for OS image upgrades, and `k8s` for Kubernetes version upgrades.
```
thomasten, line 68:
```suggestion
The phases are `infrastracture` for the cloud resource management through Terraform, `helm` for the chart management of the microservices, `image` for OS image upgrades, and `k8s` for Kubernetes version upgrades.
```

thomasten, line 68: If we expect users to mostly do this after consultation with us, I think this is okay. But if this is meant for self-service, a more detailed explanation of the phases and when it could make sense to skip them would be needed.

3u13r:
Obviously it would have been best if this change was already included in https://github.com/edgelesssys/constellation/pull/2310. I'd hope that this is approved before Tuesday, otherwise we should backport this (which is easy for docs of course).

elchead:
> Obviously it would have been best if this change was already included in #2310. I'd hope that this is approved before Tuesday, otherwise we should backport this (which is easy for docs of course).

Daniel advised me to split it up. This should be merged by Tuesday, indeed.

thomasten (approved):
Maybe Daniel thought it would be more work do document this. In general and according to the PR checklist and our review guidelines, docs can be in the same PR.
We can merge this as is. It's sufficient for releasing this feature. If we come to the conclusion that more documentation is needed, we can extend it later.

PR #2317, opened 2023-09-07
cli: perform upgrades in-place in Terraform workspace

msanft:
### Context

Previously, on an upgrade including Terraform changes, Constellation would create an external folder `./constellation-upgrade/abc/terraform` with a copy of the pre-upgrade Terraform state and the post-upgrade wanted Terraform configuration, and do a `terraform plan` and `terraform apply` there. Only if the upgrade would complete successfully, the folder would be copied from `./constellation-upgrade/abc/terraform` into `./constellation-terraform`. If an upgrade failed, the intermediate state would be lost and users weren't able to use `terraform import` on the intermediate state.

### Proposed change(s)

- On a Terraform (IAM or cluster) upgrade, first back up the corresponding folder (containing the pre-upgrade state) into `./constellation-upgrade/abc/terraform-backup` or `./constellation-upgrade/abc/terraform-iam-backup`
- Then, copy the new embedded Terraform configuration (what the user wants to have post-upgrade) into the `constellation-terraform` or `constellation-iam-terraform` folder.
- Then, show the diff of said folder and apply it if the user confirms.

### Checklist

- [x] Link to Milestone


cli/internal/cmd/iamupgradeapply.go
```diff
@@ -91,16 +88,15 @@ func runIAMUpgradeApply(cmd *cobra.Command, _ []string) error {
 	}
 
 	i := iamUpgradeApplyCmd{
-		fileHandler:   fileHandler,
-		configFetcher: configFetcher,
-		log:           log,
```
daniel-weisse, line 96: `configFetcher` should stay a variable of `iamUpgradeApplyCmd`, otherwise we cannot stub it for unit tests.

cli/internal/cmd/upgradeapply.go
```diff
@@ -271,9 +271,10 @@ func (u *upgradeApplyCmd) migrateTerraform(cmd *cobra.Command, conf *config.Conf
 			}
 			if !ok {
 				cmd.Println("Aborting upgrade.")
-				// Remove the upgrade directory
-				if err := u.fileHandler.RemoveAll(upgradeDir); err != nil {
-					return res, fmt.Errorf("cleaning up upgrade directory %s: %w", upgradeDir, err)
+				// User doesn't expect to see any changes in his workspace after aborting an "upgrade apply",
+				// therefore, roll back to the backed up state.
+				if err := u.clusterUpgrader.RollbackClusterWorkspace(); err != nil {
+					u.log.Debugf("Failed to rollback Terraform workspace: %s", err)
```
daniel-weisse, line 277: We should return an error if the rollback fails, or at least use `cmd.PrintErr`, so the error is always logged, not only with debug output enabled.
In case restoring the backup fails, we should also tell the user where the backup is located so they may restore it manually.

cli/internal/cmd/upgradecheck.go
```diff
@@ -235,9 +235,10 @@ func (u *upgradeCheckCmd) upgradeCheck(cmd *cobra.Command, fetcher attestationco
 		return fmt.Errorf("planning terraform migrations: %w", err)
 	}
 	defer func() {
-		// Remove the upgrade directory
-		if err := u.fileHandler.RemoveAll(upgradeDir); err != nil {
-			u.log.Debugf("Failed to clean up Terraform migrations: %s", err)
+		// User doesn't expect to see any changes in his workspace after an "upgrade plan",
+		// therefore, roll back to the backed up state.
+		if err := u.terraformChecker.RollbackClusterWorkspace(); err != nil {
+			u.log.Debugf("Failed to rollback Terraform workspace: %s", err)
```
daniel-weisse, line 241: Same here.
Since we now write directly into the terraform workspace, we need to tell the user directly if restoring the backup failed.
Just printing it in the debug log is not enough.

cli/internal/cloudcmd/tfupgrade.go
```diff
@@ -52,20 +51,20 @@ func planUpgrade(
 	return hasDiff, nil
 }
 
-// moveUpgradeToCurrent replaces the an existing Terraform workspace with a workspace holding migrated Terraform resources.
-func moveUpgradeToCurrent(fileHandler file.Handler, existingWorkspace, upgradeWorkingDir string) error {
-	if err := fileHandler.RemoveAll(existingWorkspace); err != nil {
-		return fmt.Errorf("removing old terraform directory: %w", err)
+// rollbackToBackup replaces the existing Terraform workspace with the backup.
+func rollbackToBackup(fileHandler file.Handler, workingDir, backupDir string) error {
```
daniel-weisse, line 55:
```suggestion
// restoreBackup replaces the existing Terraform workspace with the backup.
func restoreBackup(fileHandler file.Handler, workingDir, backupDir string) error {
```
naming nit: We don't actually "rollback" any changes in case things go wrong, and this names reads a bit easier imo

cli/internal/terraform/loader_test.go
```diff
@@ -186,21 +186,16 @@ func TestPrepareUpgradeWorkspace(t *testing.T) {
...
 				require.Error(err)
 			} else {
 				require.NoError(err)
 			}
-			checkFiles(t, file, func(err error) { assert.NoError(err) }, tc.newWorkingDir, tc.expectedFiles)
-			checkFiles(t, file, func(err error) { assert.NoError(err) },
-				tc.backupDir,
-				tc.oldWorkspaceFiles,
-			)
+			checkFiles(t, file, func(err error) { assert.NoError(err) }, tc.workingDir, tc.expectedFiles)
```
elchead, line 198: could we check that the embedded file content is written and that the backupdir contains the old content?

cli/internal/cmd/upgradecheck.go
```diff
@@ -235,9 +235,14 @@ func (u *upgradeCheckCmd) upgradeCheck(cmd *cobra.Command, fetcher attestationco
...
-		// Remove the upgrade directory
-		if err := u.fileHandler.RemoveAll(upgradeDir); err != nil {
-			u.log.Debugf("Failed to clean up Terraform migrations: %s", err)
+		// User doesn't expect to see any changes in his workspace after an "upgrade plan",
+		// therefore, roll back to the backed up state.
+		if err := u.terraformChecker.RestoreClusterWorkspace(); err != nil {
+			cmd.PrintErrf(
+				"restoring Terraform workspace: %s, restore the Terraform workspace manually from %s ",
+				err,
+				filepath.Join(constants.UpgradeDir, "<upgrade-id>", constants.TerraformUpgradeBackupDir),
```
elchead, line 244: I think the user should be given the upgrade id

cli/internal/cmd/upgradeapply_test.go
```diff
@@ -83,6 +83,15 @@ func TestUpgradeApply(t *testing.T) {
 			wantErr:           true,
 			stdin:             "no\n",
 		},
+		"abort, rollback terraform err": {
```
elchead, line 86: nit: restore seems like a more intuitive name since we use that name in the code. let's just use one of either words

cli/internal/cmd/upgradeapply_test.go
```diff
@@ -83,6 +83,15 @@ func TestUpgradeApply(t *testing.T) {
...
+			kubeUpgrader: &stubKubernetesUpgrader{
+				currentConfig: config.DefaultForAzureSEVSNP(),
```
elchead, line 88: can we check that TF is restored when the upgrade is aborted (use the mock for that?) and also not called when we don't abort?

cli/internal/cmd/iamupgradeapply_test.go
```diff
@@ -0,0 +1,181 @@
...
+			fh: setupFs(true, true),
+			iamUpgrader: &stubIamUpgrader{
+				rollbackErr: assert.AnError,
+			},
+			configFetcher: &stubConfigFetcher{
+				fetchLatestErr: assert.AnError,
+			},
+			yesFlag: true,
+			wantErr: true,
+		},
```
elchead, line 122: thanks for adding the tests. could we add the restore called check here too?

cli/internal/cmd/upgradeapply_test.go
msanft: I don't have a strong opinion on mocks, but I think we should discuss if we want to actively use mock testing in the team and either do it in every package when someone edits it

cli/internal/cmd/iamupgradeapply_test.go
msanft: s.a.

cli/internal/cmd/upgradeapply_test.go
elchead: I agree that mocking is a last resort. Of course it would be more resilient to only mock the filesystem and actually check that the files are restored

cli/internal/cmd/upgradecheck.go
```diff
@@ -241,7 +243,7 @@ func (u *upgradeCheckCmd) upgradeCheck(cmd *cobra.Command, fetcher attestationco
 			cmd.PrintErrf(
 				"restoring Terraform workspace: %s, restore the Terraform workspace manually from %s ",
 				err,
-				filepath.Join(constants.UpgradeDir, "<upgrade-id>", constants.TerraformUpgradeBackupDir),
+				filepath.Join(constants.UpgradeDir, u.upgradeID, constants.TerraformUpgradeBackupDir),
```
elchead, line 246: I think we wanted to get rid of the upgradeID and only expose the path (Daniel refactored that).
Can we move the upgradeDir arg in again and use that instead?

cli/internal/cmd/upgradeapply_test.go
msanft: I do see the point in using mocks for verifying that the expected code path is taken here. However, I don't think that this fits very well with our current testing strategy here, as it would require us to encode logic in our tests, which is often recognized as an anti-pattern. (see [here](https://abseil.io/resources/swe-book/html/ch12.html), 12-14 to 12-16) I do see the values coming from these tests, but I think if we decide to adopt this, we should do so as a whole. I think we should avoid mocking only in certain tests, to keep some consistency. I might have a wrong understanding of how to use these mocks too. If you see a way to use it without putting too much logic into the test, feel free to let me know.

cli/internal/terraform/loader_test.go
```diff
@@ -186,31 +197,44 @@ func TestPrepareUpgradeWorkspace(t *testing.T) {
...
+			err := prepareUpgradeWorkspace(path, file, tc.workingDir, tc.backupDir)
 
 			if tc.wantErr {
 				require.Error(err)
 			} else {
 				require.NoError(err)
+				checkFiles(
+					t, file,
+					func(err error) { assert.NoError(err) },
+					func(content []byte) { assert.NotEqual(defaultFileContent, content) },
```
elchead, line 211: nit: can we rename to 'oldStateContent' or something similar to indicate that this is the current TF file state that should be updated?

elchead (approved):
lgtm with the suggested changes

PR #2319, opened 2023-09-07
cli: check chart versions against target version in users config before upgrading

daniel-weisse:
### Context
Our CLI embeds the Helm charts used for the Kubernetes cluster.
For charts we pull in from external providers, the version of these charts is dependent on the version we set for that release.
For charts we package ourselves, the version of the chart is dependent on the version of the CLI binary.
For these charts, on `constellation upgrade apply`, we should only apply an upgrade if the version specifies in the user's config is an upgrade to the version running in their cluster. And only to the version set in the config.
For this the Helm code is currently missing some checks.

### Proposed change(s)

- Check if config version matches CLI version when installing new charts
- Check if version specified in config is a valid upgrade to the chart running in the Kubernetes cluster
- If a chart is a valid upgrade, check that the version specified in the user's config matches the version of charts embedded in the CLI
- Add unit tests for the `appendNewAction` method

### Additional info
- From what I can tell, this bug was not present in the v2.10 releases

### DIFF
--- a/cli/internal/helm/actionfactory.go
+++ b/cli/internal/helm/actionfactory.go
@@ -29,7 +29,6 @@ type actionFactory struct {
 	versionLister releaseVersionLister
 	cfg           *action.Configuration
 	kubeClient    crdClient
-	cliVersion    semver.Semver
 	log           debugLog
 }
 
@@ -38,9 +37,8 @@ type crdClient interface {
 }
 
 // newActionFactory creates a new action factory for managing helm releases.
-func newActionFactory(kubeClient crdClient, lister releaseVersionLister, actionConfig *action.Configuration, cliVersion semver.Semver, log debugLog) *actionFactory {
+func newActionFactory(kubeClient crdClient, lister releaseVersionLister, actionConfig *action.Configuration, log debugLog) *actionFactory {
 	return &actionFactory{
-		cliVersion:    cliVersion,
 		versionLister: lister,
 		cfg:           actionConfig,
 		kubeClient:    kubeClient,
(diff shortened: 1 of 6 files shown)


derpsteb:
LGTM. We definitely need those checks. I will let Adrian give the final approval as he is working on that code much more actively.

cli/internal/helm/actionfactory_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+	"github.com/edgelesssys/constellation/v2/internal/compatibility"
+	"github.com/edgelesssys/constellation/v2/internal/logger"
+	"github.com/edgelesssys/constellation/v2/internal/semver"
+	"github.com/stretchr/testify/assert"
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart"
+)
+
+func TestAppendNewAction(t *testing.T) {
```
derpsteb, line 21: :heart:

cli/internal/helm/actionfactory.go
```diff
@@ -71,13 +69,21 @@ func (a actionFactory) GetActions(releases []Release, force, allowDestructive bo
...
-func (a actionFactory) appendNewAction(release Release, force, allowDestructive bool, actions *[]applyAction) error {
+func (a actionFactory) appendNewAction(release Release, configTargetVersion semver.Semver, force, allowDestructive bool, actions *[]applyAction) error {
 	newVersion, err := semver.New(release.Chart.Metadata.Version)
 	if err != nil {
 		return fmt.Errorf("parsing chart version: %w", err)
 	}
 	currentVersion, err := a.versionLister.currentVersion(release.ReleaseName)
 	if errors.Is(err, errReleaseNotFound) {
+		// Don't install a new release if the user's config specifies a different version than the CLI offers.
+		if !force && isCLIVersionedRelease(release.ReleaseName) && configTargetVersion.Compare(newVersion) != 0 {
```
elchead, line 80: can we move the version validation to before getting the current version?

cli/internal/helm/actionfactory.go
```diff
@@ -71,13 +69,21 @@ func (a actionFactory) GetActions(releases []Release, force, allowDestructive bo
...
+			return fmt.Errorf(
+				"unable to install release %s at %s: this CLI only supports microservice version %s for upgrading",
```
elchead, line 82: I think we should throw an InvalidUpgradeError in all error cases because we don't want to fail the entire upgrade but just skip right? Our policy was to skip on errors and not to fail entirely? Could we add an unit test that checks the error behavior for an invalid microservice version?

daniel-weisse: We [previously returned a hard error if this was the case](https://github.com/edgelesssys/constellation/blob/094e0744d9aebd29469699d88ddb396b9246b54d/cli/internal/helm/upgrade.go#L106-L108).
But I'm open to changing this. Any opinions on this @derpsteb

derpsteb: Before the helm refactoring we only showed the user that this upgrade won't be executed for $ERROR reasons and continued with other upgrades. I prefer that version over exiting after an error. 

As long as we don't mess up our version compatibility it should always be fine to miss one microservice upgrade while images are already upgraded.^^

daniel-weisse: Ok, then lets just skip the upgrade with `InvalidUpgradeError` if the CLI does not support the version.

elchead (approved):
should we run a no-op + upgrade test E2E before merging?

PR #2321, opened 2023-09-08
cli: write infrastructure to new state file

elchead:
### Context

This is the first implementation part of the new constellation state (#2281).

### Proposed change(s)
create
- aggregate infrastructure state from TF output
- write the infrastructure output to  `constellation-state.yaml`  with version v1 after terraform apply

upgrade
- extend TF client with ShowInfrastrucutre() to output the infrastructure part of the state
- use ShowInfrastructure() and overwrite the infrastructure output to  `constellation-state.yaml`  with version v1 after terraform apply

### Related issue
- [AB#3422](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3422)

### Checklist

- [x] Link to Milestone


cli/internal/state/state.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// Infrastructure describe the state related to the cloud resources of the cluster.
+type Infrastructure struct {
+	UID               string   `yaml:"uid"`
+	PublicIP          string   `yaml:"publicIP"`
+	InitSecret        string   `yaml:"initSecret"`
+	APIServerCertSANs []string `yaml:"apiServerCertSANs"`
+	Azure             *Azure   `yaml:"azure"`
+	GCP               *GCP     `yaml:"gcp"`
```
derpsteb, line 23: It seems more idiomatic to me to make this an interface and add marshalling methods. Would prevent setting both keys, makes nil checks obsolete, makes the intent of the property clearer. Was there a specific reason you chose this approach, apart from the fact that this is how we do it in the config?

derpsteb:
LGTM. Would just like to discuss that one point.

cli/internal/state/state.go
elchead: you mean to make this field contain a CSP interface similar to the attestation config?
In the end the properties of the CSPs are very different so I don't see how to generalize the fields in an interface.
What is unclear about the intent of the property?

Do you have a suggestion?

derpsteb: True. It isn't clear how to write that interface.

derpsteb (approved):

derpsteb:
Before merging this: what is the plan to ensure that the migration from the idfile to the statefile happens within 1 release cycle? How do we migrate?

elchead:
> Before merging this: what is the plan to ensure that the migration from the idfile to the statefile happens within 1 release cycle? How do we migrate?

The idFile is only removed in the last step (see user stories in Azure). So the code is not broken at any point.
When we prioritize this story, it should be very feasible to merge this in the next release.
Otherwise, we would have both the idFile and state file during the intermediate release.

derpsteb:
Good to merge from my side.

cli/internal/cmd/create.go
```diff
@@ -162,21 +165,39 @@ func (c *createCmd) create(cmd *cobra.Command, creator cloudCreator, fileHandler
...
 		return translateCreateErrors(cmd, c.pf, err)
 	}
 	c.log.Debugf("Successfully created the cloud resources for the cluster")
 
+	idFile := convertToIDFile(infraState, provider)
 	if err := fileHandler.WriteJSON(constants.ClusterIDsFilename, idFile, file.OptNone); err != nil {
 		return err
 	}
+	state := state.State{
+		Version:        "v1",
```
daniel-weisse, line 180: This should be a constant.
See the version number of the config file for reference

cli/internal/cmd/upgradeapply.go
```diff
@@ -189,18 +190,25 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, upgradeDir string, fl
...
 		}
 	} else {
-		tfOutput, err = u.migrateTerraform(cmd, conf, upgradeDir, flags)
+		infraState, err = u.migrateTerraform(cmd, conf, upgradeDir, flags)
 		if err != nil {
 			return fmt.Errorf("performing Terraform migrations: %w", err)
 		}
 	}
+	state := state.State{
+		Version:        "v1",
```
daniel-weisse, line 206: Same as before

cli/internal/cmd/upgradeapply.go
```diff
@@ -189,18 +190,25 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, upgradeDir string, fl
...
+		Infrastructure: infraState,
+	}
+	if err := u.fileHandler.WriteYAML(constants.StateFilename, state, file.OptOverwrite); err != nil {
+		return fmt.Errorf("writing state file: %w", err)
+	}
```
daniel-weisse, line 211: Consider moving the updating of an existing state file to `updateClusterIDFile` (rename the function eventually),
and have an explicit migration function if the state file does not exist when running `upgrade apply`.
IMO does not need to happen in this PR, but should be done before this is in any release.
If you decide to implement in a different PR, please add a comment to this section.

cli/internal/state/state.go
```diff
@@ -0,0 +1,41 @@
...
+type State struct {
+	Version        string         `yaml:"version"`
+	Infrastructure Infrastructure `yaml:"infrastructure"`
+}
+
+// Infrastructure describe the state related to the cloud resources of the cluster.
+type Infrastructure struct {
+	UID               string   `yaml:"uid"`
+	PublicIP          string   `yaml:"publicIP"`
+	InitSecret        string   `yaml:"initSecret"`
```
daniel-weisse, line 20: I'm unsure if `InitSecret` contains potentially unprintable characters.
It was of type `[]byte` in the id file, maybe we should keep it as such here?
@3u13r can you give an opinion on this?

cli/internal/state/state.go
daniel-weisse, line 1: Can we add a Codeowner for this package?

daniel-weisse, line 1: Doc comments in this file need some work, but should probably be handled in [AB#3426](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3426)

cli/internal/terraform/terraform.go
```diff
@@ -170,6 +171,44 @@ func (c *Client) ShowIAM(ctx context.Context, provider cloudprovider.Provider) (
 	}
 }
 
+// ShowInfrastructure reads the state of Constellation cluster resources from Terraform.
+func (c *Client) ShowInfrastructure(ctx context.Context, provider cloudprovider.Provider) (state.Infrastructure, error) {
```
daniel-weisse, line 175: Is there still a reason to keep the old `ShowCluster` (and also the `CreateCluster`) function which returns `ApplyOutput`, instead of letting the terraform package just return `state.Infrastructure`?

daniel-weisse (changes requested):

cli/internal/terraform/terraform.go
daniel-weisse: Nvm, there is already a PR for that

cli/internal/state/state.go
elchead: Sure, who should be it? I would opt in since i don't have so many yet. Or do you want it?

cli/internal/state/state.go
3u13r: Going for `string` is fine here. The `initSecret` is generated in terraform (e.g. https://github.com/edgelesssys/constellation/blob/main/cli/internal/terraform/terraform/gcp/main.tf#L69) and in terraform `random_password` is equivalent to `random_string` which only consists of printable characters.

cli/internal/state/state.go
daniel-weisse: Sure, go ahead.

daniel-weisse (approved):

3u13r (dismissed):

3u13r (approved):
LGTM

PR #2322, opened 2023-09-08
cli: retry helm apply on any error

daniel-weisse:
### Context
We have retry logic for installing/upgrading helm charts.
However, we have a rather restrictive check for what errors are retried in case installing/upgrading fails.

### Proposed change(s)

- Instead of checking for specific errors to retry, retry on any error
  - If we ever encounter some error cases we definitely don't want to retry, we should just add exceptions for these errors, instead of keeping a list of errors we want to retry.

### Related issue
- [AB#3432](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3432)

### Checklist
- [x] Link to Milestone


elchead (approved):
LGTM

derpsteb (approved):

PR #2336, opened 2023-09-14
joinservice: cache certificates for Azure SEV-SNP attestation

msanft:
### Context

Caching ASK intermediate certificates is a prerequisite for VLEK-based signer attestation as performed on AWS SEV-SNP, but can also be used on Azure. 

### Proposed change(s)

- On cluster initialization, request the Milan certificate chain from the AMD KDS in the join service, and save it in a configmap.
- When creating the validator for a joinservice pod, pass the cached certificate to the validator if it exists.
- In the validator, use the cached certificate. On Azure, this is a special case as there is a 3-step precedence in which certificates should be taken:
  - If the ARK and ASK from the THIM response are intact, use them.
  - If the THIM response does not contain the cert chain, use the ARK from the Constellation config and the **cached ASK**.
  - If neither of these cases applies, retrieve ARK and ASK from AMD KDS.
https://github.com/edgelesssys/constellation/blob/d3c976977d7f882fc3b80a0f9f41146a7679ddb9/internal/attestation/azure/snp/validator.go#L267-L271

### Additional info
- [E2E Test](https://github.com/edgelesssys/constellation/actions/runs/6300085599)
- [More recent E2E Test](https://github.com/edgelesssys/constellation/actions/runs/6323307816)
- [Even more recent E2E Test](https://github.com/edgelesssys/constellation/actions/runs/6347643459)

### Checklist

- [x] Link to Milestone


thomasten:
This introduces attestation-variant-specific code in otherwise attestation-variant-agnostic code. This can get messy if we add more variants in the future. We should try to reduce this where possible.
Thing that come into my mind:
* Adding a parameter for attestation-variant-specific data (like `cfg`, but for non-user-defined data) to the `Validator` func instead of adding a very specific `ValidatorWithASKCertCache` func.
* Do we need a separate config map for the cached cert? If yes, should it be named more generically for all kinds of (future) attestation-variant-specific data?
* Better separate the attestation-variant-specific code in joinservice's main.go. (At least move it to an own func.)

Not sure if all of this makes sense or if there are better ways. Please think about it.

msanft:
> This introduces attestation-variant-specific code in otherwise attestation-variant-agnostic code. This can get messy if we add more variants in the future. We should try to reduce this where possible.

I definitely see the point here and agree.

Regarding having a `ValidatorWithASKCertCache`, I discussed this with Leo and we agreed that changing current validator function signatures would come at the cost of a lot of refactoring to do. However, I also think that this might be worth it, in case other attestation-variant-specific features are introduced in the fututre. Potentially we should consider a builder pattern or something similar here. I also think that wrapping the csp/attestation-specific things behind a returned interface is not a great abstraction necessarily. Regarding the other 2 points, i will think about it.

daniel-weisse (commented):
What about adding the ASK as an optional value to the Azure attestation config?
While this would still require attestation variant specific code in the join-service, it would keep the function signature intact, and also allows us to easily disable the caching behavior for the CLI (by not setting the optional ASK field).
We already do something similar for the MAA URL on Azure.

joinservice/internal/certcache/certcache.go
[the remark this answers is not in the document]
msanft: See [AB#3451](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3451)

joinservice/internal/kubernetes/kubernetes.go
```diff
@@ -65,15 +67,27 @@ func (c *Client) GetComponents(ctx context.Context, configMapName string) (compo
...
 }
 
-func (c *Client) getConfigMapData(ctx context.Context, name, key string) (string, error) {
+// GetConfigMapData returns the data for the given key in the configmap with the given name.
+func (c *Client) GetConfigMapData(ctx context.Context, name, key string) (string, error) {
 	cm, err := c.client.CoreV1().ConfigMaps("kube-system").Get(ctx, name, metav1.GetOptions{})
 	if err != nil {
+		if errors.IsNotFound(err) {
+			return "", &ConfigMapNotExistError{}
+		}
```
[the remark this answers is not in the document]
msanft: This wouldn't allow me to wrap the error

internal/config/config.go
```diff
@@ -1119,6 +1119,9 @@ type AzureSEVSNP struct {
 	// description: |
 	//   AMD Root Key certificate used to verify the SEV-SNP certificate chain.
 	AMDRootKey Certificate `json:"amdRootKey" yaml:"amdRootKey"`
+	// description: |
+	//   AMD Signing Key certificate used to verify the SEV-SNP VCEK / VLEK certificate.
+	AMDSigningKey Certificate `json:"amdSigningKey" yaml:"amdSigningKey" validate:"len=0"`
```
daniel-weisse, line 1124:
```suggestion
	AMDSigningKey Certificate `json:"amdSigningKey,omitempty" yaml:"amdSigningKey,omitempty" validate:"len=0"`
```
Should be hidden when not in use

joinservice/internal/watcher/validator.go
```diff
@@ -28,21 +29,23 @@ type Updatable struct {
...
 		variant:     variant,
 	}
+}
 
-	if err := u.Update(); err != nil {
-		return nil, err
-	}
-	return u, nil
+// WithCachedCerts sets the available cached certificates in the updatable validator.
+func (u *Updatable) WithCachedCerts(cachedCerts *certcache.CachedCerts) *Updatable {
```
daniel-weisse, line 46: What is the purpose of keeping this function?
We don't expect to ever call `NewValidator` in the join-service without it, so why not add `cachedCerts` as an argument to the constructor directly?

joinservice/internal/watcher/validator.go
```diff
@@ -76,11 +79,30 @@ func (u *Updatable) Update() error {
...
+			switch c := cfg.(type) {
+			case *config.AzureSEVSNP:
+				c.AMDSigningKey = config.Certificate(*ask)
+				return c
+			}
+			// TODO(derpsteb): Add AWS SEV-SNP
+		}
+	}
+	return cfg
+}
```
daniel-weisse, line 108:
```suggestion
// addCachedCerts adds the certificates cached by the validator to the config, if applicable.
func (u *Updatable) addCachedCerts(cfg config.AttestationCfg) (config.AttestationCfg, error) {
	switch c := cfg.(type) {
	case *config.AzureSEVSNP:
		ask, err := u.getCachedCerts()
		if err != nil {
			return nil, err
		}
		c.AMDSigningKey = config.Certificate(ask)
		return c, nil
	}
	// TODO(derpsteb): Add AWS SEV-SNP

	return cfg, nil
}

func (u *Updatable) getCachedCerts() (x509.Certificate, error) {
	if u.cachedCerts == nil {
		return x509.Certificate{}, fmt.Errorf("no cached certs available")
	}
	ask, ark := u.cachedCerts.SevSnpCerts()
	if ask == nil {
		return x509.Certificate{}, fmt.Errorf("no ASK available")
	}
	if ark == nil {
		return x509.Certificate{}, fmt.Errorf("no ARK available")
	}
	return *ask, nil
}
```
I think frontloading the attestation variant check makes this function a bit easier to read and maintain. It also handles errors better (cached certs not set, or certificates missing)

joinservice/internal/certcache/certcache.go
```diff
@@ -0,0 +1,165 @@
...
+		return nil, nil, fmt.Errorf("failed to encode ark: %w", err)
+	}
+
+	c.log.Debugf("Creating certificate chain cache configmap")
+	if err := c.kubeClient.CreateConfigMap(ctx, constants.SevSnpCertCacheConfigMapName, map[string]string{
+		constants.CertCacheAskKey: askWriter.String(),
+		constants.CertCacheArkKey: arkWriter.String(),
+	}); err != nil {
+		return nil, nil, fmt.Errorf("failed to create certificate chain cache configmap: %w", err)
+	}
```
daniel-weisse, line 114:
```suggestion
	if err := c.kubeClient.CreateConfigMap(ctx, constants.SevSnpCertCacheConfigMapName, map[string]string{
		constants.CertCacheAskKey: askWriter.String(),
		constants.CertCacheArkKey: arkWriter.String(),
	}); err != nil {
		// If ConfiMap already exists, another JoinService instance created the certcache while this operation was running
		// Calling this function again should now load the cached certificates
		if k8serrors.IsAlreadyExists(err) {
			return c.getCertChainCache(ctx context.Context)
		}
		return nil, nil, fmt.Errorf("failed to create certificate chain cache configmap: %w", err)
	}
```
There is a potential data race here if two join-service pods start at the same time while no certificate cache exists.
I doubt this will be a problem in practice when creating new clusters, but could cause minor errors when upgrading from a version that did not implement this caching behavior.
We can prevent this by calling `getCertChainCache` again in case the ConfigMap already exists.

joinservice/internal/kubernetes/kubernetes.go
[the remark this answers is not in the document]
daniel-weisse: But you are not wrapping the error, you are simply replacing it with another?

joinservice/internal/certcache/certcache.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// getCertChainCache returns the cached ASK and ARK certificate.
+func (c *Client) getCertChainCache(ctx context.Context) (ask, ark *x509.Certificate, err error) {
+	c.log.Debugf("Retrieving certificate chain from cache")
+	askRaw, err := c.kubeClient.GetConfigMapData(ctx, constants.SevSnpCertCacheConfigMapName, constants.CertCacheAskKey)
+	if err != nil {
+		if _, ok := err.(kubernetes.ConfigMapNotExistError); ok {
+			return nil, nil, nil
+		}
```
daniel-weisse, line 126: Instead of returning a `nil` error, and then checking on the return values in the caller, this function should let the caller handle checking for specific errors and then decide on what actions to take.

joinservice/internal/certcache/certcache.go
```diff
@@ -0,0 +1,165 @@
...
+func (c *Client) createCertChainCache(ctx context.Context, signingType abi.ReportSigner) (ask, ark *x509.Certificate, err error) {
+	c.log.Debugf("Creating certificate chain cache")
+	ask, ark, err = c.getCertChainCache(ctx)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get certificate chain cache: %w", err)
+	}
+	if ask != nil && ark != nil {
+		c.log.Debugf("Certificate chain cache already exists")
+		return ask, ark, nil
+	}
```
daniel-weisse, line 91: As said in my previous comment, we should check for the error of `getCertChainCache` to decide whether the cert cache ConfigMap already exists.
However, we might want to implement additional logic to update an existing cert cache ConfigMap in case the ASK or ARK are missing from it.
In this case we would fetch the missing certificate and then call an Update function instead of a Create function.

joinservice/internal/kubernetes/kubernetes.go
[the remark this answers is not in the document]
msanft: I am wrapping any error which is not a non-existent config map. Also, bubbling up the error would require us to import `"k8s.io/apimachinery/pkg/api/errors"` in the upper-level `certcache` package, which - as of now - only has to import our kubernetes client

joinservice/internal/watcher/validator.go
msanft: Yes, this became obsolete with the latest update

joinservice/internal/kubernetes/kubernetes.go
```diff
@@ -181,3 +181,17 @@ func (c *Client) CreateConfigMap(ctx context.Context, name string, data map[stri
...
 }
+
+// UpdateConfigMap updates the configmap with the provided name by writing the provided key and value.
+func (c *Client) UpdateConfigMap(ctx context.Context, name, key, value string) error {
+	cm, err := c.client.CoreV1().ConfigMaps("kube-system").Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get configmap: %w", err)
+	}
+	cm.Data[key] = value
+	_, err = c.client.CoreV1().ConfigMaps("kube-system").Update(ctx, cm, metav1.UpdateOptions{})
```
daniel-weisse, line 192:
```suggestion
	cm, err := c.client.CoreV1().ConfigMaps(constants.ConstellationNamespace).Get(ctx, name, metav1.GetOptions{})
	if err != nil {
		return fmt.Errorf("failed to get configmap: %w", err)
	}
	cm.Data[key] = value
	_, err = c.client.CoreV1().ConfigMaps(constants.ConstellationNamespace).Update(ctx, cm, metav1.UpdateOptions{})
```

internal/attestation/azure/snp/validator.go
```diff
@@ -45,6 +45,9 @@ type Validator struct {
 	attestationVerifier  attestationVerifier
 	attestationValidator attestationValidator
 
+	// ASK certificate cached by the join service.
+	cachedASKCert *x509.Certificate
```
daniel-weisse, line 49: Shouldn't be needed anymore, since its part of the `config` now

internal/attestation/azure/snp/validator.go
```diff
@@ -98,22 +101,38 @@ func NewValidator(cfg *config.AzureSEVSNP, log attestation.Logger) *Validator {
 	return v
 }
 
+// WithCachedASKCert sets the cached ASK certificate.
+func (v *Validator) WithCachedASKCert(cert *x509.Certificate) *Validator {
+	v.cachedASKCert = cert
+	return v
+}
```
daniel-weisse, line 108:
```suggestion
```
Same here

daniel-weisse (commented):
Implementation looks good to me now

daniel-weisse (approved):
Please run e2e tests before merging.
Otherwise looks good to me

thomasten (changes requested):

malt3:
Looks good from a high level. I think Daniel and Thomas are the best reviewers when it comes to the attestation details.

joinservice/internal/kubernetes/kubernetes.go
```diff
@@ -159,3 +161,37 @@ func k8sCompliantHostname(in string) (string, error) {
...
+	_, err := c.client.CoreV1().ConfigMaps(constants.ConstellationNamespace).Create(ctx, cm, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to create configmap: %w", err)
+	}
+	return nil
+}
+
+// UpdateConfigMap updates the configmap with the provided name by writing the provided key and value.
+func (c *Client) UpdateConfigMap(ctx context.Context, name, key, value string) error {
+	cm, err := c.client.CoreV1().ConfigMaps(constants.ConstellationNamespace).Get(ctx, name, metav1.GetOptions{})
```
3u13r, line 187: Depending on how frequently this resource is updated by other components, we might want to wrap this in a retry like in https://github.com/edgelesssys/constellation/blob/feat/joinservice/manage-ask/internal/kubernetes/kubectl/kubectl.go#L142.

joinservice/cmd/main.go
```diff
@@ -47,6 +49,9 @@ func main() {
 	verbosity := flag.Int("v", 0, logger.CmdLineVerbosityDescription)
 	flag.Parse()
 
+	ctx, cancel := context.WithTimeout(context.Background(), vpcIPTimeout)
+	defer cancel()
```
3u13r, line 53: Moving this ctx up and using it for another function call goes aginst the original intention of this ctx. Either we should simply create a new `constext.Background` from which we later create a context with timeout OR we change the name of the timeout const.

3u13r (changes requested):

joinservice/internal/kubernetes/kubernetes.go
msanft: Yes, we should do that in the refactoring of the joinservice by directly using the kubectl utility.

3u13r (approved):
LGTM

internal/attestation/azure/snp/validator.go
```diff
@@ -101,19 +102,29 @@ func NewValidator(cfg *config.AzureSEVSNP, log attestation.Logger) *Validator {
...
 		return nil, fmt.Errorf("unmarshalling instanceInfo: %w", err)
 	}
-	att, err := instanceInfo.attestationWithCerts(v.log, v.getter)
+
+	att, err := instanceInfo.attestationWithCerts(v.log, v.getter, cachedCerts)
 	if err != nil {
 		return nil, fmt.Errorf("parsing attestation report: %w", err)
 	}
 
 	// Verify the attestation report's certificates.
```
thomasten, line 125:
```suggestion
```

joinservice/cmd/main.go
```diff
@@ -47,6 +49,9 @@ func main() {
 	verbosity := flag.Int("v", 0, logger.CmdLineVerbosityDescription)
 	flag.Parse()
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
thomasten, line 53: Is there a semantic difference to `ctx := context.Background()`?

joinservice/cmd/main.go
```diff
@@ -56,21 +61,34 @@ func main() {
...
+	}
+
+	attVariant, err := variant.FromString(*attestationVariant)
 	if err != nil {
 		log.With(zap.Error(err)).Fatalf("Failed to parse attestation variant")
 	}
-	validator, err := watcher.NewValidator(log.Named("validator"), variant, handler)
+
+	certCacheClient := certcache.NewClient(log.Named("certcache"), kubeClient, attVariant)
+	cachedCerts, err := certCacheClient.CreateCertChainCache(ctx)
```
thomasten, line 75: we shouldn't introduce a dependency to AMD KDS for non-SNP clusters

thomasten (changes requested):

joinservice/cmd/main.go
msanft: The Attestation-Variant-Specific logic is hidde behind `CreateCertChainCache`. I thought we agreed to move it out of the `main.go` file.

> Better separate the attestation-variant-specific code in joinservice's main.go. (At least move it to an own func.)

thomasten: sorry, I somehow misread the code in `CreateCertChainCache` (although it's simple :D)

joinservice/internal/certcache/amdkds/amdkds.go
```diff
@@ -0,0 +1,38 @@
...
+	return &KDSClient{
+		getter: getter,
+	}
+}
+
+// CertChain queries the AMD KDS for the certificate chain for given signing type (VCEK / VLEK).
+func (c *KDSClient) CertChain(signingType abi.ReportSigner) (ask, ark *x509.Certificate, err error) {
+	askark, err := trust.GetProductChain("Milan", signingType, c.getter)
+	if err != nil {
+		return nil, nil, fmt.Errorf("retrieving certificate chain: %v", err)
```
thomasten, line 34:
```suggestion
		return nil, nil, fmt.Errorf("retrieving certificate chain: %w", err)
```

joinservice/internal/certcache/certcache.go
```diff
@@ -0,0 +1,201 @@
...
+// CreateCertChainCache creates a certificate chain cache for the given attestation variant
+// and returns the cached certificates, if applicable.
+// If the certificate chain cache already exists, nothing is done.
+func (c *Client) CreateCertChainCache(ctx context.Context) (*CachedCerts, error) {
+	switch c.attVariant {
+	case variant.AzureSEVSNP{}:
+		c.log.Debugf("Creating azure SEV-SNP certificate chain cache")
+		ask, ark, err := c.createCertChainCache(ctx, abi.VcekReportSigner)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create azure SEV-SNP certificate chain cache: %w", err)
```
thomasten, line 55:
```suggestion
		c.log.Debugf("Creating Azure SEV-SNP certificate chain cache")
		ask, ark, err := c.createCertChainCache(ctx, abi.VcekReportSigner)
		if err != nil {
			return nil, fmt.Errorf("creating Azure SEV-SNP certificate chain cache: %w", err)
```

joinservice/internal/certcache/certcache.go
```diff
@@ -0,0 +1,201 @@
...
+func (c *Client) createCertChainCache(ctx context.Context, signingType abi.ReportSigner) (ask, ark *x509.Certificate, err error) {
+	c.log.Debugf("Creating certificate chain cache")
+	var shouldCreateConfigMap bool
+
+	cacheAsk, cacheArk, err := c.getCertChainCache(ctx)
+	if k8serrors.IsNotFound(err) {
+		c.log.Debugf("Certificate chain cache does not exist")
+		shouldCreateConfigMap = true
+	} else if err != nil {
+		return nil, nil, fmt.Errorf("failed to get certificate chain cache: %w", err)
```
thomasten, line 92:
```suggestion
		return nil, nil, fmt.Errorf("getting certificate chain cache: %w", err)
```

joinservice/internal/certcache/certcache.go
```diff
@@ -0,0 +1,201 @@
...
+		ask = cacheAsk
+	}
+	if cacheArk != nil {
+		ark = cacheArk
+	}
+
+	c.log.Debugf("Retrieving certificate chain from KDS")
+	kdsAsk, kdsArk, err := c.kdsClient.CertChain(signingType)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to retrieve certificate chain from KDS: %w", err)
```
thomasten, line 108: same here and at multiple other places

joinservice/internal/certcache/certcache.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	if kdsAsk != nil {
+		ask = kdsAsk
+	}
+	if kdsArk != nil {
+		ark = kdsArk
+	}
```
thomasten, line 115: It's not clear if the preference to take certs from KDS than from Cache is wanted. Please clarify, at least with a comment, or with a test case.

joinservice/internal/certcache/certcache.go
```diff
@@ -0,0 +1,201 @@
...
+
+	askWriter := &strings.Builder{}
+	if err := pem.Encode(askWriter, &pem.Block{Type: "CERTIFICATE", Bytes: ask.Raw}); err != nil {
+		return nil, nil, fmt.Errorf("failed to encode ask: %w", err)
+	}
+	arkWriter := &strings.Builder{}
+	if err := pem.Encode(arkWriter, &pem.Block{Type: "CERTIFICATE", Bytes: ark.Raw}); err != nil {
+		return nil, nil, fmt.Errorf("failed to encode ark: %w", err)
+	}
```
thomasten, line 124: I'd suggest adding a `func PEMEncode(cert *x509.Certificate) (string, error)` to our `internal/crypto` package.

joinservice/internal/certcache/certcache.go
```diff
@@ -0,0 +1,201 @@
...
+func (c *Client) getCertChainCache(ctx context.Context) (ask, ark *x509.Certificate, err error) {
+	c.log.Debugf("Retrieving certificate chain from cache")
+	askRaw, err := c.kubeClient.GetConfigMapData(ctx, constants.SevSnpCertCacheConfigMapName, constants.CertCacheAskKey)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get ASK: %w", err)
+	}
+	if askRaw != "" {
+		c.log.Debugf("ASK cache hit")
+		askBlock, _ := pem.Decode([]byte(askRaw))
+		ask, err = x509.ParseCertificate(askBlock.Bytes)
```
thomasten, line 171:
```suggestion
		ask, err = crypto.PemToX509Cert([]byte(askRaw))
```
pem.Decode returns nil on error, which needs to be catched. And we have a util func for this.

joinservice/internal/certcache/certcache.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	arkRaw, err := c.kubeClient.GetConfigMapData(ctx, constants.SevSnpCertCacheConfigMapName, constants.CertCacheArkKey)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get ARK: %w", err)
+	}
+	if arkRaw != "" {
+		c.log.Debugf("ARK cache hit")
+		arkBlock, _ := pem.Decode([]byte(arkRaw))
+		ark, err = x509.ParseCertificate(arkBlock.Bytes)
```
thomasten, line 184:
```suggestion
		ark, err = crypto.PemToX509Cert([]byte(arkRaw))
```

joinservice/internal/certcache/certcache_test.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	if c.arkResponse != nil {
+		ark = mustParsePEM(c.arkResponse)
+	}
+	return ask, ark, c.certChainErr
+}
+
+func mustParsePEM(pemBytes []byte) *x509.Certificate {
+	pemBlock, _ := pem.Decode(pemBytes)
+	cert, err := x509.ParseCertificate(pemBlock.Bytes)
```
thomasten, line 151:
```suggestion
	cert, err := crypto.PemToX509Cert(pemBytes)
```

joinservice/internal/certcache/certcache_test.go
```diff
@@ -0,0 +1,231 @@
...
+			ctx := context.Background()
+
+			c := NewClient(logger.NewTest(t), tc.kubeClient, variant.Dummy{})
+
+			_, _, err := c.getCertChainCache(ctx)
+			if tc.wantErr {
+				assert.Error(err)
+			} else {
+				assert.NoError(err)
+			}
```
thomasten, line 202: also test functionality of this func by doing something with the return values

joinservice/internal/watcher/validator.go
```diff
@@ -76,11 +81,47 @@ func (u *Updatable) Update() error {
...
+	if u.cachedCerts == nil {
+		return x509.Certificate{}, fmt.Errorf("no cached certs available")
+	}
+	ask, ark := u.cachedCerts.SevSnpCerts()
+	if ask == nil {
+		return x509.Certificate{}, fmt.Errorf("no ASK available")
+	}
+	if ark == nil {
+		return x509.Certificate{}, fmt.Errorf("no ARK available")
+	}
```
thomasten, line 125:
```suggestion
	ask, _ := u.cachedCerts.SevSnpCerts()
	if ask == nil {
		return x509.Certificate{}, fmt.Errorf("no ASK available")
	}
```

thomasten (changes requested):

derpsteb:
Removing myself as this seems to have had enough eyes.

joinservice/internal/watcher/validator.go
msanft: Do we only want to cache the ASK certificate completely for now? This would let us remove a lot of the logic inside this function, but also closes the door on keeping the ARK here instead of the Constellation config.

joinservice/internal/certcache/certcache.go
thomasten: Sort of. If only the ASK or only the ARK is cached, both will be replaced with the certs from the KDS.

joinservice/internal/watcher/validator.go
thomasten: Note that this code is from the watcher. If it only needs the ASK (for now), I wouldn't let it fail here.

Regarding "a lot of the logic inside this function", I guess you mean inside certcache? I also thought about this, but wasn't sure. That's why I only commented on the client of that cache (i.e., the watcher validator).

msanft: Yes, my reply was in regards to the certcache, sorry. I think it would let us spare some code now, but since the general attestation design (i.e. ARK -> ASK certchain) should be somewhat stable, I don't think it would require a lot of maintenance if its tested well enough. But I don't have a strong opinion on removing it either.

thomasten (approved):

internal/crypto/crypto.go
```diff
@@ -76,7 +76,7 @@ func PemToX509Cert(raw []byte) (*x509.Certificate, error) {
 	return cert, nil
 }
 
-// X509CertToPEM takes an x.509 certificate and returns it as a PEM-encoded certificate.
+// X509CertToPem takes an x.509 certificate and returns it as a PEM-encoded certificate.
 func X509CertToPem(cert *x509.Certificate) ([]byte, error) {
```
daniel-weisse, line 80:
```suggestion
// X509CertToPEM takes an x.509 certificate and returns it as a PEM-encoded certificate.
func X509CertToPEM(cert *x509.Certificate) ([]byte, error) {
```
PEM should be the correct form here

PR #2338, opened 2023-09-14
ci: remove verify test for macos during release

3u13r:
### Proposed change(s)

- macos runners use a different `sed` therefore the verify test breaks currently on macos: https://github.com/edgelesssys/constellation/actions/runs/6184742123/job/16788978904

Since we only need one test for macos anyway, removing the verify test for now. Still, we should fix the test for macos at some point.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/e2e-test-release.yml
+++ b/.github/workflows/e2e-test-release.yml
@@ -154,10 +154,12 @@ jobs:
           #
           # Tests on macOS runner
           #
-          - test: "verify"
-            provider: "azure"
-            kubernetes-version: "v1.28"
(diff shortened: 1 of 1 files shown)


daniel-weisse (approved):
Maybe add a comment why we dont run the verify test currently?
E.g.
```yml
          # Skipping verify test on MacOS since the runner uses a different version of sed
          # TODO: Update verify test to work on MacOS runners
          # - test: "verify"
          #  provider: "azure"
          #  kubernetes-version: "v1.28"
          #  runner: "macos-12"
```

PR #2346, opened 2023-09-19
debugd: send requests over lb

3u13r:
### Proposed change(s)

- Remove unnecessary `CanSend()` function, as `receiveFinished` is already checked during `transfer.SendFiles()`
- Adds loadbalancer IP to the ips to try to get the bootstrapper from

The second change is necessary as in the future, all nodes in the cluster are not allowed to send out unencrypted traffic to the node CIDR. 

### Checklist

- [x] Link to Milestone


malt3 (approved):
Good job!

PR #2348, opened 2023-09-19
join: join over lb if available

3u13r:
### Context

This is needed since the upcoming node-to-node strict mode does not allow traffic which is not WireGuard encrypted by Cilium (with the exception of etcd). Therefore nodes cannot join via directly talking to another node.

### Proposed change(s)

- add join port to all LBs
- Let rejoinclient and joinclient join over LB (if available, otherwise fallback to node VPC IPs)

The fallback is needed to still support Miniconstellation as it does not have an LB.

e2e upgrade tests: 
azure: https://github.com/edgelesssys/constellation/actions/runs/6237449241
gcp: https://github.com/edgelesssys/constellation/actions/runs/6238824048
aws: https://github.com/edgelesssys/constellation/actions/runs/6238826460

### Checklist

- [x] Link to Milestone

### DIFF
--- a/bootstrapper/internal/joinclient/joinclient.go
+++ b/bootstrapper/internal/joinclient/joinclient.go
@@ -188,8 +188,14 @@ func (c *JoinClient) Stop() {
 func (c *JoinClient) tryJoinWithAvailableServices() error {
 	ips, err := c.getControlPlaneIPs()
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get control plane IPs: %w", err)
 	}
+
+	ip, _, err := c.metadataAPI.GetLoadBalancerEndpoint(context.TODO())
+	if err != nil {
+		return fmt.Errorf("failed to get load balancer endpoint: %w", err)
+	}
+	ips = append(ips, ip)
 
 	if len(ips) == 0 {
 		return errors.New("no control plane IPs found")
@@ -425,6 +431,8 @@ type MetadataAPI interface {
 	List(ctx context.Context) ([]metadata.InstanceMetadata, error)
 	// Self retrieves the current instance.
 	Self(ctx context.Context) (metadata.InstanceMetadata, error)
+	// GetLoadBalancerEndpoint retrieves the load balancer endpoint.
+	GetLoadBalancerEndpoint(ctx context.Context) (host, port string, err error)
 }
 
 type encryptedDisk interface {
--- a/bootstrapper/internal/joinclient/joinclient.go
+++ b/bootstrapper/internal/joinclient/joinclient.go
@@ -186,23 +186,29 @@ func (c *JoinClient) Stop() {
 }
 
 func (c *JoinClient) tryJoinWithAvailableServices() error {
-	ips, err := c.getControlPlaneIPs()
-	if err != nil {
-		return fmt.Errorf("failed to get control plane IPs: %w", err)
-	}
+	ctx, cancel := c.timeoutCtx()
+	defer cancel()
 
-	ip, _, err := c.metadataAPI.GetLoadBalancerEndpoint(context.TODO())
+	var endpoints []string
+
+	ip, _, err := c.metadataAPI.GetLoadBalancerEndpoint(ctx)
 	if err != nil {
 		return fmt.Errorf("failed to get load balancer endpoint: %w", err)
 	}
-	ips = append(ips, ip)
+	endpoints = append(endpoints, net.JoinHostPort(ip, strconv.Itoa(constants.JoinServiceNodePort)))
 
-	if len(ips) == 0 {
+	ips, err := c.getControlPlaneIPs(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get control plane IPs: %w", err)
+	}
+	endpoints = append(endpoints, ips...)
+
+	if len(endpoints) == 0 {
 		return errors.New("no control plane IPs found")
 	}
 
-	for _, ip := range ips {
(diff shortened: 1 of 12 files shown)


elchead:
could you add the context why we need this?

bootstrapper/internal/joinclient/joinclient.go
```diff
@@ -188,8 +188,14 @@ func (c *JoinClient) Stop() {
...
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get control plane IPs: %w", err)
+	}
+
+	ip, _, err := c.metadataAPI.GetLoadBalancerEndpoint(context.TODO())
+	if err != nil {
+		return fmt.Errorf("failed to get load balancer endpoint: %w", err)
 	}
+	ips = append(ips, ip)
```
elchead, line 198: doesn't adding it here make the LB the fallback option?

elchead:
can u also provide a run for miniconstellation?

cli/internal/terraform/terraform/aws/main.tf
```diff
@@ -96,6 +98,7 @@ resource "aws_lb" "front_end" {
   internal           = false
   load_balancer_type = "network"
   tags               = local.tags
+  security_groups    = [aws_security_group.security_group.id]
```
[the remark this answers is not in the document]
3u13r: fyi this link does not work on Firefox.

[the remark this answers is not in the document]
elchead: yeah I know it's not a browser standard but the link should still work: I refer to security_groups :)

bootstrapper/internal/joinclient/joinclient.go
```diff
@@ -188,8 +188,14 @@ func (c *JoinClient) Stop() {
 func (c *JoinClient) tryJoinWithAvailableServices() error {
 	ips, err := c.getControlPlaneIPs()
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to get control plane IPs: %w", err)
+	}
+
+	ip, _, err := c.metadataAPI.GetLoadBalancerEndpoint(context.TODO())
```
daniel-weisse, line 194: Consider moving the `ctx, cancel := c.timeoutCtx()` part out of `*JoinClient.getControlPlaneIPs`, and move it into `tryJoinWithAvailableServices`.
This would allow using the same context to fetch the control plane IPs, as well as the Load Balancer IPs.

disk-mapper/internal/rejoinclient/rejoinclient.go
```diff
@@ -130,19 +133,37 @@ func (c *RejoinClient) requestRejoinTicket(endpoint string) (*joinproto.IssueRej
...
+				joinEndpoints = append(joinEndpoints, net.JoinHostPort(instance.VPCIP, strconv.Itoa(constants.JoinServiceNodePort)))
+			}
+		}
 	}
+
+	lbEndpoint, _, err := c.metadataAPI.GetLoadBalancerEndpoint(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("retrieving load balancer endpoint from cloud provider: %w", err)
+	}
+	joinEndpoints = append(joinEndpoints, net.JoinHostPort(lbEndpoint, strconv.Itoa(constants.JoinServiceNodePort)))
```
daniel-weisse, line 160: Same as in `joinclient.go`, this adds the LB as the last entry, making it the fallback.

internal/cloud/metadata/metadata.go
```diff
@@ -43,21 +39,3 @@ type InstanceLister interface {
 	// List retrieves all instances belonging to the current constellation.
 	List(ctx context.Context) ([]InstanceMetadata, error)
 }
-
-// JoinServiceEndpoints returns the list of endpoints for the join service, which are running on the control plane nodes.
-func JoinServiceEndpoints(ctx context.Context, lister InstanceLister) ([]string, error) {
```
daniel-weisse, line 48: Since `joinclient.go` and `rejoinclient.go` have to perform the same action to fetch loadbalancer endpoint, and control plane IPs, what about adapting this function to include fetching of LB endpoint and using it for both?
With the current implementation, the code is slightly different for `joinclient.go` and `rejoinclient.go`.

Function could look like the following:
```Go
type endpointLister interface {
    List(ctx context.Context) ([]InstanceMetadata, error)
    GetLoadBalancerEndpoint(ctx context.Context) (host, port string, err error)
}

func JoinServiceEndpoints(ctx context.Context, lister endpointLister) ([]string, error) {
    lbIP, _, err := lister.GetLoadBalancerEndpoint(ctx)
    if err != nil {
        // Maybe do additional error handling here in case of no loadbalancer (ignore error?)
        return nil, fmt.Errorf("retrieving load balancer endpoint from cloud provider: %w", err)
    }
    // Old JoinServiceEndpoints code
    joinEndpoints := []string{}
    ...

    // Return all endpoints, with LB as the first endpoint
    return append(net.JoinHostPort(lbEndpoint, strconv.Itoa(constants.JoinServiceNodePort)), joinEndpoints), nil
}
```

3u13r: Hm, I didn't like the place of this function as part of the metadata package as it is too specific to me. Also, without this function the metadata package looks cleaner. 
I see you point of introducing a potential difference in both implementation though, but I haven't found the right spot.

daniel-weisse (approved):

elchead (approved):
started a mini test and lgtm when green

PR #2357, opened 2023-09-25
cli: read Azure SEV-SNP version values from launch TCB in `constellation verify` output

elchead:
UPDATE waiting for #2391 

### Context

As described in #2343, the current version fetching is not reliable. Since it is not transparent how the Azure MAA values are deduced, we rather refer to the Launch TCB values from `constellation verify` 

### Proposed change(s)

- read the verify output in the E2E verify test to read the version values from the LaunchTCB

### Related issue
- [AB#3447](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3447)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/api/attestationconfigapi/cli/parse_report.go
+++ b/internal/api/attestationconfigapi/cli/parse_report.go
@@ -1,0 +1,76 @@
+/*
+Copyright (c) Edgeless Systems GmbH
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+
+	"github.com/edgelesssys/constellation/v2/internal/api/attestationconfigapi"
+)
+
+// ParseSNPReport parses the SNP report and returns the version information.
+func ParseSNPReport(reader io.Reader) (attestationconfigapi.AzureSEVSNPVersion, error) {
+	mp := parseLaunchTCBSection(reader)
+	return parseVersion(mp)
+}
+
+func parseLaunchTCBSection(reader io.Reader) map[string]string {
+	scanner := bufio.NewScanner(reader)
+	parsedValues := make(map[string]string)
+	inLaunchTCBSection := false
+
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if strings.Contains(line, "Launch TCB:") {
+			inLaunchTCBSection = true
+			continue // skip the current line as it is the section header
+		}
+
+		// Stop scanning if we have reached the end of the Launch TCB section
+		if inLaunchTCBSection && strings.Contains(line, "Signature (DER)") {
+			break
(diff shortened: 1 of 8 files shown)


internal/api/attestationconfigapi/cli/parse_report.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+
+	"github.com/edgelesssys/constellation/v2/internal/api/attestationconfigapi"
+)
+
+// ParseSNPReport parses the SNP report and returns the version information.
+func ParseSNPReport(reader io.Reader) (attestationconfigapi.AzureSEVSNPVersion, error) {
```
3u13r, line 20: We have no stability guarantees of Constellation's cli output format. Therefore, this way of parsing is very brittle.

My suggestion is that we should write the full json (before parsing) to disk on `constellation verify` and then use the exact same parsing here.

elchead: Is this a sensible default to write it to file at all times? or should we add a new flag to show the output as json? I can do that but was not sure this was worth the implementation effort since we also control the constellation verify output

thomasten:
Note that we changed this feature to also consider the Committed and Reported TCB. Please check the AB ticket again.

internal/api/attestationconfigapi/cli/parse_report.go
elchead: When we use --raw to parse the content the instance info containing the report is base64 encoded this makes the file parsing yet more obscure in the e2e test setup. Not sure we should go this route. @derpsteb what do you think?
I don't see the verify output as very unstable.
The current parsing would only fail if we rename the affected fields in the report or move the signature field in the output. But the latter could be fixed too.

3u13r:
I discussed with @katexochen and we currently see the following way forward:
1. Refactor the parsing of `cmd/verify.go` EITHER refactor out the parsing to a location that is used by the verify command and your tool OR implement a custom mashaller and unmarshaller which recursively parses the `attestationDoc` to a struct containing the parsed `InstanceInfo` (not as base64 but as e.g. `azureInstanceInfo` struct). We prefer the second option.
2. If the second option was chosen: Use the refactored parsing to enable `--json` or `--output json` for the `constellation verify` command. 

Both in one separate PR.

Then we can go ahead and implement this story. For the second option based on the json output which can simply be marshalled and unmarshalled again or fir the first option based on the raw output.

derpsteb:
I agree that having a `--json`/`--output json` flag on `constellation verify` would be ideal. What I would definitely aim to do is that `verify` and this cli share the parsing logic. Probably by the way that Leo and Paul suggest.

elchead:
closing in favor of new implementation in #2429

PR #2358, opened 2023-09-25
ci: add more filters to e2e failure OpenSearch links

daniel-weisse:
### Context
Our daily and weekly e2e tests generate logs and upload them to OpenSearch.
To make it easier for devs to connect logs with specific tests, on test failure, we automatically generate a URL filtering the logs to ones from the failed workflow.
Since one workflow may run multiple tests (in the case of the weekly e2e test 30+), this filtering is not sufficient anymore.
This PR aims to change that

### Proposed change(s)
* Add `ref-stream` and `kubernetes-version` fields to e2e log messages
* Update the upgrade e2e test to use `upgrade` as its test payload instead of `nop`
  * `upgrade` is an alias for `nop`. This should make it easier to filter the relevant logs
* Add `kubernetes-version`, `ref-stream`, and `e2e-test-payload` filters to the URL for failed workflow runs
  *  For some tests `kubernetes-version` and `ref-stream` are left empty in the URL. In those cases they are also left empty in the log messages, e.g. for e2e upgrade tests

### Additional info
[Example failure card](https://github.com/orgs/edgelesssys/projects/3/views/1?pane=issue&itemId=39707193)

### Checklist

- [ ] Link to Milestone


msanft (commented):
To have consistent fields in debud and K8s deployments, we should change the fields in the debugd ([here](https://github.com/edgelesssys/constellation/blob/main/debugd/internal/debugd/logcollector/fields.go) and [here](https://github.com/edgelesssys/constellation/blob/8f5a2867b47be0e396b1b71ccd70a92c1c1ed9c2/.github/actions/constellation_create/action.yml#L120-L143)) accordingly.

msanft (approved):

malt3 (approved):

PR #2385, opened 2023-09-27
s3proxy: add initial implementation

derpsteb:
### Context

**EDIT:** This PR now includes keyservice integration and proper envelope encryption (#2398).

This PR adds a new service to Constellation called s3proxy. It can be used to transparently encrypt and decrypt data when using AWS S3. Data that is created while using s3proxy is tagged with `constellation-encryption` and the body is encrypted using AES-256-GCM. Data that is retrieved while using s3proxy is decrypted if has an encrypted data encryption key in it's metadata. 

This PR is missing:
- e2e tests based on minio/mint: I would like to discuss again how we want to put this into the Constellation repo.
- deployment automation as a bazel target: plan is to have a target that builds the image, pushes it and deploys s3proxy with the image into a cluster that was setup separately. This seems like the easiest development setup. 
- ~integration with Constellation keyservice to derive keys: immediate next step.~

I decided to leave these things out to get some eyes on the implementation as it is now working with the e2e tests and Filestash.

There are no unittests so far as I didn't know what the correct behavior was until now. The implementation is validated through the minio/mint e2e tests and Filestash so far. Ideally I would like to revisit unittests after we have the MVP in the documentation and rely on e2e tests until then. 

### Proposed change(s)
- Add initial implementation of s3proxy

### Additional information
- ~Tagging as no-changelog as at least the keyservice integration is required before users should be pointed to this. That separate PR can link to this one to have the full story.~ Tagging as feature as this now includes the full MVP.
- Added Thomas as reviewer, but feedback from anyone else is welcome.

### Checklist

- [x] Link to Milestone


s3proxy/internal/crypto/crypto.go
```diff
@@ -0,0 +1,79 @@
+/*
+Copyright (c) Edgeless Systems GmbH
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+/*
+Package crypto provides encryption and decryption functions for the s3proxy.
```
3u13r, line 8: While this package's job seems simple, I'd suggest to use Tink's AEAD encryption: https://pkg.go.dev/github.com/google/tink/go/aead#example-package. This makes it even harder to make crypto mistakes.

s3proxy/internal/crypto/crypto.go
```diff
@@ -0,0 +1,79 @@
...
+// ciphertext is formatted as 12 byte nonce + ciphertext.
+func Decrypt(ciphertext []byte, key []byte) ([]byte, error) {
+	// Enforce AES-256
+	if len(key) != 32 {
+		return nil, aes.KeySizeError(len(key))
+	}
+
+	// Extract the nonce from the ciphertext.
+	nonce := ciphertext[:12]
+	ciphertext = ciphertext[12:]
```
3u13r, line 67: I think we should check the length of the ciphertext before those operations.

s3proxy/internal/router/object.go
```diff
@@ -0,0 +1,185 @@
...
+
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/edgelesssys/constellation/v2/internal/logger"
+	"github.com/edgelesssys/constellation/v2/s3proxy/internal/crypto"
+)
+
+const (
+	// testingKey is a temporary encryption key used for testing.
+	// TODO (derpsteb): This key needs to be fetched from Constellation's keyservice.
+	testingKey = "01234567890123456789012345678901"
```
3u13r, line 27: I think we should address this right now. I'm very uneasy with variables that start with `testing` who not reside in a `_test.go` file. Especially if they are crypto keys. Especially if they are used right now.
Maybe since we aren't in a hurry with this feature and this is quite separate, let's wait for approval here and then create a PR with the KMS implementation to this branch?

s3proxy/internal/router/router.go
```diff
@@ -0,0 +1,459 @@
...
+	var bucket string
+
+	client, err := s3.NewClient(r.region)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	path := req.URL.Path
+	switch {
```
3u13r, line 69: I'd refactor this switch case:
1. factor out the function calls like `containsBucket(req.Host)` to before the switch/case and instead assign those to bool variables which have a clear name. This should improve readability.
2. refactor each case "body" to its own function like `handle...()`.

s3proxy/internal/crypto/crypto.go
3u13r: If we decide to still implement this ourselves, then we should at least add tests for this.

s3proxy/deploy/create_cert.sh
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env bash
+# service is the name of the s3proxy service in kubernetes.
```
3u13r, line 2:
```suggestion
# $service is the name of the s3proxy service in kubernetes.
```
I think this makes it faster to parse out the variable. I don't know of any style guides for bash commenting tough.

3u13r (changes requested):

s3proxy/deploy/create_cert.sh
daniel-weisse, line 1: Why not use cert-manager to generate this certificate?

s3proxy/internal/crypto/crypto.go
derpsteb: Addressed here: https://github.com/edgelesssys/constellation/pull/2398

s3proxy/internal/router/object.go
derpsteb: Addressed here: https://github.com/edgelesssys/constellation/pull/2398

s3proxy/internal/crypto/crypto.go
```diff
@@ -0,0 +1,79 @@
...
+
+// Encrypt takes a 32 byte key and encrypts a plaintext using AES-256-GCM.
+// Output format is 12 byte nonce + ciphertext.
+func Encrypt(plaintext, key []byte) ([]byte, error) {
+	// Enforce AES-256
+	if len(key) != 32 {
+		return nil, aes.KeySizeError(len(key))
+	}
+
+	// None should not be reused more often that 2^32 times:
```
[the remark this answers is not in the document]
derpsteb: Removed in: https://github.com/edgelesssys/constellation/pull/2398

s3proxy/internal/crypto/crypto.go
derpsteb: Removed in: https://github.com/edgelesssys/constellation/pull/2398

s3proxy/deploy/create_cert.sh
derpsteb: Replaced by cert-manager solution.

s3proxy/internal/crypto/crypto.go
```diff
@@ -0,0 +1,79 @@
...
+	// https://pkg.go.dev/crypto/cipher#NewGCM
+	// Assuming n encryption operations per second, the key has to be rotated every:
+	// n=1: 2^32 / (60*60*24*365*10) = 135 years.
+	// n=10: 2^32 / (60*60*24*365*10) = 13.5 years.
+	// n=100: 2^32 / (60*60*24*365*10) = 1.3 years.
+	// n=1000: 2^32 / (60*60*24*365*10) = 50 days.
+	nonce := make([]byte, 12)
```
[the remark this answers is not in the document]
derpsteb: Removed in: https://github.com/edgelesssys/constellation/pull/2398

s3proxy/internal/router/router.go
derpsteb: Code should a lot more concise now. Sorry for the convoluted initial version :see_no_evil:

s3proxy/cmd/main.go
```diff
@@ -0,0 +1,119 @@
...
+	server := http.Server{
+		Addr:    fmt.Sprintf("%s:%d", flags.ip, flags.port),
+		Handler: http.HandlerFunc(router.Serve),
+		// Disable HTTP/2. Serving HTTP/2 will cause some clients to use HTTP/2.
+		// It seems like AWS S3 does not support HTTP/2.
+		// Having HTTP/2 enabled will at least cause the aws-sdk-go V1 copy-object operation to fail.
+		TLSNextProto: map[string]func(*http.Server, *tls.Conn, http.Handler){},
+	}
+
+	if flags.port == 443 {
```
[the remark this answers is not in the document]
derpsteb: Replaced port flag with `--no-tls` flag.

s3proxy/internal/router/router.go
```diff
@@ -0,0 +1,416 @@
...
+	if !bytes.Equal(actual[:], expected) {
+		return fmt.Errorf("content-md5 mismatch, header is %x, body is %x", expected, actual)
+	}
+
+	return nil
+}
+
+// match reports whether path matches pattern, and if it matches,
+// assigns any capture groups to the *string or *int vars.
+func match(path, pattern string, vars ...interface{}) bool {
```
daniel-weisse, line 359:
```suggestion
func match(path, pattern string, vars ...any) bool {
```
nit

s3proxy/internal/router/router.go
```diff
@@ -0,0 +1,416 @@
...
+		case *string:
+			*p = match
+		case *int:
+			n, err := strconv.Atoi(match)
+			if err != nil {
+				return false
+			}
+			*p = n
+		default:
+			panic("vars must be *string or *int")
```
daniel-weisse, line 376: This function should not panic.
Return errors instead

s3proxy/internal/router/router.go
```diff
@@ -0,0 +1,416 @@
...
+	relock.Lock()
+	defer relock.Unlock()
+
+	regex := regexen[pattern]
+	if regex == nil {
+		regex = regexp.MustCompile("^" + pattern + "$")
+		regexen[pattern] = regex
+	}
+	return regex
+}
```
daniel-weisse, line 392: This function is only used for 2 "hardcoded" patterns.
Additionally, this function may panic at runtime if someone decides to use this function with new patterns.
Instead of using this functions, compile the expressions as package scoped vars, and pass them to the `match` function.

E.g.:
```Go
var (
    // adjust naming
    matchKey = regexp.MustCompile("/(.+)")
    matchBucketAndKey = regexp.MustCompile("/([^/?]+)/(.+)")
)

func (r Router) Serve(w http.ResponseWriter, req *http.Request) {
    // ...
    if containsBucket(req.Host) {
        // BUCKET.s3.REGION.amazonaws.com
       parts := strings.Split(req.Host, ".")
       bucket = parts[0]

       matchingPath = match(req.URL.Path, matchKey, &key)
   } else {
       matchingPath = match(req.URL.Path, matchBucketAndKey, &bucket, &key)
    }
    // ...
}

func match((path, rxp *regexp.Regexp, vars ...any) bool {
    matches := rxp.FindStringSubmatch(path)
    // ...
}
```

daniel-weisse (changes requested):

s3proxy/deploy/deployment-s3proxy.yaml
```diff
@@ -0,0 +1,95 @@
...
+    template:
+        metadata:
+          labels:
+            app: s3proxy
+        spec:
+          imagePullSecrets:
+          - name: regcred
+          containers:
+          - name: s3proxy
+            image: ghcr.io/derpsteb/constellation/s3proxy@sha256:0463a52da3c69056f8c3d2509a7c552a49fd8dc14ff2f00ac029548b7b3683c4
```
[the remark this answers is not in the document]
derpsteb: Not sure how to get an image from the constellation registry without merging.

s3proxy/internal/router/router.go
derpsteb: Simplified this so the panic/error is not necessary.

s3proxy/cmd/main.go
```diff
@@ -54,20 +54,25 @@ func main() {
 }
 
 func runServer(flags cmdFlags, log *logger.Logger) error {
-	log.Infof("listening", "ip", flags.ip, "port", flags.port, "region", flags.region)
+	log.Infof("listening", "ip", flags.ip, "port", flags.noTLS, "region", flags.region)
```
daniel-weisse, line 57:
```suggestion
	log.With(zap.String("ip", flags.ip), zap.Bool("tlsEnabled", !flags.noTLS), zap.String("region", flags.Region)).Infof("listening")
```

daniel-weisse (approved):

3u13r (changes requested):

s3proxy/cmd/main.go
```diff
@@ -0,0 +1,125 @@
...
+
+func runServer(flags cmdFlags, log *logger.Logger) error {
+	log.With(zap.String("ip", flags.ip), zap.Bool("port", flags.noTLS), zap.String("region", flags.region)).Infof("listening")
+
+	router := router.New(flags.region, log)
+
+	port := defaultPort
+	if flags.noTLS {
+		port = 80
+	}
```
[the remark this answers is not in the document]
derpsteb: s3proxy now uses port 4433. Non-configurable for now. The deployment's service remaps that to 443.

3u13r (approved):
LGTM, should we add e2e test tickets to the backlog?

derpsteb:
Yes. Put into AB#3479 and AB#3480.

PR #2387, opened 2023-09-27
cli: write MAAUrl to idfile again

3u13r:
### Proposed change(s)

- fixes: https://github.com/edgelesssys/constellation/pull/2321
- The MAAUrl is read on e.g. `constellation verify` on Azure

### Checklist

- [x] Link to Milestone


malt3 (approved):

msanft:
We should read from the state file in `constellation verify` instead.

3u13r:
> As a long-term solution, we should read from the state file in constellation verify instead. 

Yes, but the authors original intention seems to be to already create the state file but still keed the id file (correctly) around, hence this way of fixing this.

PR #2388, opened 2023-09-27
Support internal load balancers

3u13r:
### Proposed change(s)

- Each CSP terraform has a `internal_loadbalancer` flag which default can be switched to `true`. Then the constellation infra is set up using internal LB(s) only. 
- Remove the lb svc for the verify deployment

I manually verified this on every of the 3 CSPs.

### Checklist

- [x] Link to Milestone


bootstrapper/cmd/bootstrapper/main.go
```diff
@@ -125,6 +125,8 @@ func main() {
 		if err != nil {
 			log.With(zap.Error(err)).Fatalf("Failed to set up cloud logger")
 		}
+		go metadata.PrepareAzureControlPlaneNode(ctx, log)
```
3u13r, line 128: I thought about wrapping is in channels to be able to return an error in this function but that also didn't look very nice. Maybe the place is also wrong. Maybe we should add `PrepareControlPlaneModes` as a general function to the CSP interface. This could serve as a hook point for CSP specific hacks.

.github/actions/constellation_create/action.yml
```diff
@@ -115,6 +118,12 @@ runs:
       run: |
         yq eval -i '(.debugCluster) = true' constellation-conf.yaml
 
+    - name: Enable debugCluster flag
```
daniel-weisse, line 121:
```suggestion
    - name: Enable internalLoadBalancer flag
```

.github/workflows/e2e-test-manual-internal.yml
daniel-weisse, line 1: Why have this as a separate test instead of making it an option for the e2e-test-manual workflow?

bootstrapper/cmd/bootstrapper/main.go
daniel-weisse, line 128: Why is this done as a Go routine?
What would prevent us from simply executing it and waiting for an error?
The way this works right now could cause it to fail when the bootstrapper is done quickly and shuts down before the routine finishes

internal/cloud/azure/azure.go
```diff
@@ -308,6 +325,44 @@ func (c *Cloud) getVMInterfaces(ctx context.Context, vm armcompute.VirtualMachin
...
+
+	lb, err := c.getLoadBalancer(ctx, resourceGroup, uid)
+	if err != nil {
+		return "", fmt.Errorf("retrieving load balancer: %w", err)
+	}
+	if lb == nil || lb.Properties == nil {
+		return "", errors.New("could not dereference load balancer IP configuration")
+	}
+
+	var privIP []string
```
daniel-weisse, line 347:
```suggestion
	var privIP string
```
Why not make this just a string, since we only ever expect a single IP address and stop iterating over the configuration once we find the first?

internal/cloud/azure/azure.go
```diff
@@ -388,6 +446,89 @@ func (c *Cloud) getLoadBalancerDNSName(ctx context.Context) (string, error) {
...
+// and the 5-tuple is (VM IP, <some port>, LB IP, 6443, TCP).
+// Now the load balancer does not re-write the source IP address only the destination (DNAT).
+// Therefore the 5-tuple is (VM IP, <some port>, VM IP, 6443, TCP).
+// Now the VM responds to the SYN packet with a SYN-ACK packet, but the outgoing
+// connection waits on a response from the load balancer and not the VM therefore
+// dropping the packet.
+//
+// OpenShift also uses the same mechanism to redirect traffic to the API server:
+// https://github.com/openshift/machine-config-operator/blob/e453bd20bac0e48afa74e9a27665abaf454d93cd/templates/master/00-master/azure/files/opt-libexec-openshift-azure-routes-sh.yaml
+func (c *Cloud) PrepareAzureControlPlaneNode(ctx context.Context, log *logger.Logger) {
```
daniel-weisse, line 464:
```suggestion
func (c *Cloud) PrepareControlPlaneNode(ctx context.Context, log *logger.Logger) {
```
This is a method on the `Cloud` struct in the `azure` package, so dropping the `Azure` from the method name seems reasonable to me.

daniel-weisse, line 464: This goes together with my earlier comment:
Why let this function throw fatal errors instead of returning it to the caller?

internal/cloud/azure/azure.go
```diff
@@ -388,6 +446,89 @@ func (c *Cloud) getLoadBalancerDNSName(ctx context.Context) (string, error) {
...
+	chainName := "azure-lb-nat"
+
+	selfMetadata, err := c.Self(ctx)
+	if err != nil {
+		log.Fatalf("failed to get self metadata: %v", err)
+	}
```
daniel-weisse, line 470:
```suggestion
	if err != nil {
		log.With(zap.Error(err).Fatalf("failed to get self metadata")
	}
```
Also applies to most other log statements in this function

internal/cloud/azure/azure.go
```diff
@@ -388,6 +446,89 @@ func (c *Cloud) getLoadBalancerDNSName(ctx context.Context) (string, error) {
...
+	req, err := http.NewRequestWithContext(ctx, "GET", "https://localhost:6443/readyz", nil)
+	if err != nil {
+		log.Fatalf("failed to create request: %v", err)
+	}
+
+	for {
+		time.Sleep(5 * time.Second)
+		resp, err := httpClient.Do(req)
+		if err != nil {
+			fmt.Printf("failed to request endpoint: %v\n", err)
```
daniel-weisse, line 500:
```suggestion
			log.With(zap.Error(err)).Errorf("failed to request endpoint")
```

internal/cloud/azure/azure.go
```diff
@@ -388,6 +446,89 @@ func (c *Cloud) getLoadBalancerDNSName(ctx context.Context) (string, error) {
...
+	}
+
+	if selfMetadata.Role != role.ControlPlane {
+		log.Infof("not a control plane node, skipping iptables setup")
+		return
+	}
+
+	loadbalancerIP, err := c.getLoadBalancerPrivateIP(ctx)
+	if err != nil {
+		log.Infof("failed to get load balancer private IP: %v", err)
```
daniel-weisse, line 479:
```suggestion
		log.Errorf("failed to get load balancer private IP: %v", err)
```

internal/cloud/azure/azure.go
```diff
@@ -388,6 +446,89 @@ func (c *Cloud) getLoadBalancerDNSName(ctx context.Context) (string, error) {
...
+		}
+		defer resp.Body.Close()
+
+		if resp.StatusCode == http.StatusOK {
+			break
+		}
+	}
+
+	// Create the iptables chain
+	out, err := exec.CommandContext(ctx, "iptables", "-w", "-t", "nat", "-N", chainName).CombinedOutput()
```
daniel-weisse, line 511: Have you considered using [`iptables` from CoreOS](https://pkg.go.dev/github.com/coreos/go-iptables/iptables)?
In the end this is still just a wrapper for the iptables binary, but gives a nicer interface.

internal/config/config.go
```diff
@@ -830,6 +833,12 @@ func (c *Config) Validate(force bool) error {
 		}
 	}
 
+	if c.InternalLoadBalancer {
+		if c.GetProvider() != cloudprovider.AWS && c.GetProvider() != cloudprovider.GCP {
+			return &ValidationError{validationErrMsgs: []string{"internalLoadBalancer is only supported for AWS and GCP"}}
+		}
```
daniel-weisse, line 839: Doesn't the `PrepareAzureControlPlane` function allow internal Loadbalancers on Azure?

.github/workflows/e2e-test-manual-internal.yml
3u13r: Because the e2e-test-manual has already 10 dispatch events and I was uncomfortable merging even more inputs.

internal/config/config.go
3u13r: I forgot to mention the state of Azure. I now added a note in the PR description. In short: we need this function but also need to remove konnectivity. 

If you want to, we could remove (comment out?) the function until it is actually used.

internal/cloud/azure/azure.go
3u13r: Yes, I had a look at wrapping the commands but decided against it since it seemed like that we'd only need a very small part of the package. But I have no strong opinions here. I'll implement implement it with the package and see how it looks.

elchead:
I looked at the Terraform changes which lgtm.

> I discussed with @malt3 a possible testing which should be compatible with most (all?) of our current CI tests.
> 
> Essentially:
> 
> 1. create a jump host VM with public IP which runs socat.
> 2. run  `constellation create`
> 3. hack the PublicIP into the constellation-id.json as `customEndpoint` so that the kubeAPI adds it as SAN.
> 4. change the ClusterEndpoint to the PublicIP
> 5. run `constellation init`
> 6. replace the endpoint in the kubeconfig with the PublicIP
> 
> Now we should be able to operate as normal from our local system / the CI.
> 
> Thanks @malt3 for the discussion!

Thanks for the test coverage! Should we run them as part of the weekly (at least for GCP)?

elchead (approved):

internal/cloud/azure/azure.go
3u13r: I went with using `Warnf` here, since this is expected to fail for public LB architectures. 
But I'm open to going with `Errorf` anyway.

internal/cloud/azure/azure.go
3u13r: Looks great, using it.

bootstrapper/cmd/bootstrapper/main.go
3u13r: Rewrote it as blocking function.

3u13r:
The commit [add in-cluster endpoint to terraform output](https://github.com/edgelesssys/constellation/pull/2388/commits/8d5ea24474e5e6d86801e9b92cfe64f108cc44d6) is necessary to add the in-cluster endpoint to the Cilium helm values.

daniel-weisse (approved):

malt3 (approved):

derpsteb (approved):
Only looked at Helm pkg. :+1:

PR #2392, opened 2023-09-28
cli: omitempty infrastructure fields

3u13r:
### Proposed change(s)

- Omit some empty fields in `constellation-state.yaml`

It's expected that at least one or even both fields are empty. Therefore we should not print those to file.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/state/state.go
+++ b/cli/internal/state/state.go
@@ -32,8 +32,8 @@ type Infrastructure struct {
 	ClusterEndpoint   string   `yaml:"clusterEndpoint"`
 	InitSecret        string   `yaml:"initSecret"`
 	APIServerCertSANs []string `yaml:"apiServerCertSANs"`
-	Azure             *Azure   `yaml:"azure"`
-	GCP               *GCP     `yaml:"gcp"`
+	Azure             *Azure   `yaml:"azure,omitempty"`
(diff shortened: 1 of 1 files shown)


malt3 (approved):

daniel-weisse:
We should think about this change again.
With self managed infrastructure, users are expected to fill in the (optional) fields that are now hidden by default.
This seems like bad UX to me.
An alternative could be adding the CSP specific fields as comments, and then letting the user choose which parts to use.
This could require some extra maintenance effort tho, instead of just letting the user delete the csp specific fields they dont need.

elchead:
> We should think about this change again. With self managed infrastructure, users are expected to fill in the (optional) fields that are now hidden by default. This seems like bad UX to me. An alternative could be adding the CSP specific fields as comments, and then letting the user choose which parts to use. This could require some extra maintenance effort tho, instead of just letting the user delete the csp specific fields they dont need.

I agree. This should be part of the doc story to provide a UX friendly template to fill in the values.

3u13r:
I don't think that how we write out the file has anything to do with UX. The user must manually generate the file since it is not generated by `constellation create`, right? So we need to document the template somewhere else anyway or provide a interactive script or something.

PR #2393, opened 2023-09-28
init: overwrite kubeconfig address

3u13r:
### Context

Since the bootstrapper discovers the load balancer itself and does not receives the endpoint used from the init request, it is possible that the bootstrapper "sees the world differently than the CLI". 

This is especially true for privateIP deployments. Since the lb discovers the internal load balancer but the CLI side might want to query another endpoint which redirects to the internal load balancer. Therefore the client side should always use the cluster endpoint from the `constellation-id.json` or `constellation-state.yaml` file as those are the truths for the CLI. 

The kubeconfig is generated by the cluster. Therefore it points to the endpoint used by the cluster. When we receive the kubeconfig we therefore replace it's endpoint with the endpoint from the `constellation-id.json`. 

### Proposed change(s)

- We automatically overwrite the server address in the kubeconfig with the cluster endpoint from the `constellation-id.json` (later: constellation-state.yaml`) 

This is the minimal change I need to continue with https://github.com/edgelesssys/constellation/pull/2388. For publicIP deployments we don't change anything as we replace the value with itself.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/init.go
+++ b/cli/internal/cmd/init.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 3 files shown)


cli/internal/cmd/init.go
```diff
@@ -472,7 +472,23 @@ func (i *initCmd) writeOutput(
...
+	kubeconfig, err := clientcmd.Load(initResp.GetKubeconfig())
+	if err != nil {
+		return fmt.Errorf("loading kubeconfig: %w", err)
+	}
+	if len(kubeconfig.Clusters) != 1 {
+		return fmt.Errorf("expected exactly one cluster in kubeconfig, got %d", len(kubeconfig.Clusters))
+	}
+	for _, cluster := range kubeconfig.Clusters {
+		cluster.Server = "https://" + net.JoinHostPort(idFile.IP, strconv.Itoa(constants.KubernetesPort))
+	}
```
msanft, line 485:
```suggestion
	kubeconfig.Clusters[0].Server = fmt.Sprintf("https://%s", net.JoinHostPort(idFile.IP, strconv.Itoa(constants.KubernetesPort)))
```
nit

cli/internal/cmd/init_test.go
```diff
@@ -339,12 +351,34 @@ func TestWriteOutput(t *testing.T) {
 	assert := assert.New(t)
 	require := require.New(t)
 
+	clusterEndpoint := "cluster-endpoint"
+
+	expectedKubeconfig := k8sclientapi.Config{
+		Clusters: map[string]*k8sclientapi.Cluster{
+			"cluster": {
+				Server: "https://" + clusterEndpoint + ":6443",
```
msanft, line 359:
```suggestion
				Server: fmt.Sprintf("https://%s:6443", clusterEndpoint),
```
nit

msanft (approved):
I can't say much regarding the technical details of this, but the explanation you've given makes sense to me. Implementation LGTM!

daniel-weisse (approved):

malt3 (approved):

PR #2395, opened 2023-09-29
cli: use state file on init and upgrade

msanft:
### Context

We are in the transition from the ID-File (`constellation-id.json`) to the state file (`constellation-state.yaml`). This PR aims to implement part of this transition by using the state-file instead of the ID-file in the init and upgrade operations of a cluster. This PR does not intend to remove the ID-file completely, which should be done in a follow-up PR aligned with [AB#3425](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3425).

### Proposed change(s)

- Implement basic read / write / merge handlers for the state file.
- Use the state file in `constellation init`, where the [`ClusterValues`](https://github.com/edgelesssys/constellation/blob/main/rfc/state-file.md) output is now being populated (in addition to populating the old ID-file)
- Use the state file in `constellation upgrade apply`, where the old, pre-upgrade `constellation-id.json` is being read. It's output, merged with the output of the upgrade itself is written to the state file.

### Additional info
- [E2E Test](https://github.com/edgelesssys/constellation/actions/runs/6352260025)
- [E2E Test Upgrade](https://github.com/edgelesssys/constellation/actions/runs/6376515073)
- [AB#3424](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3424)
- This PR shows a bug in MiniConstellation, where the cluster is not able to initialize ("connection refused" when talking to the bootstrapper node), which is presumably caused by the bootstrapper not starting.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/init.go
+++ b/cli/internal/cmd/init.go
@@ -206,10 +206,6 @@ func (i *initCmd) initialize(
 	}
 	idFile.MeasurementSalt = measurementSalt
 
-	stateFile.SetClusterValues(state.ClusterValues{
-		MeasurementSalt: measurementSalt,
-	})
-
 	clusterName := clusterid.GetClusterName(conf, idFile)
 	i.log.Debugf("Setting cluster name to %s", clusterName)
 
@@ -250,7 +246,7 @@ func (i *initCmd) initialize(
 	idFile.CloudProvider = provider
 
 	bufferedOutput := &bytes.Buffer{}
-	if err := i.writeOutput(idFile, stateFile, resp, flags.mergeConfigs, bufferedOutput); err != nil {
+	if err := i.writeOutput(idFile, stateFile, resp, flags.mergeConfigs, bufferedOutput, measurementSalt); err != nil {
 		return err
 	}
 
@@ -450,8 +446,14 @@ func (d *initDoer) handleGRPCStateChanges(ctx context.Context, wg *sync.WaitGrou
 	})
 }
 
+// writeOutput writes the output of a cluster initialization to the
+// state- / id- / kubeconfig-file and saves it to disk.
 func (i *initCmd) writeOutput(
-	idFile clusterid.File, stateFile *state.State, initResp *initproto.InitSuccessResponse, mergeConfig bool, wr io.Writer,
+	idFile clusterid.File,
+	stateFile *state.State,
+	initResp *initproto.InitSuccessResponse,
+	mergeConfig bool, wr io.Writer,
+	measurementSalt []byte,
 ) error {
 	fmt.Fprint(wr, "Your Constellation cluster was successfully initialized.\n\n")
 
@@ -504,8 +506,11 @@ func (i *initCmd) writeOutput(
 	idFile.OwnerID = ownerID
 	idFile.ClusterID = clusterID
 
-	stateFile.ClusterValues.OwnerID = ownerID
-	stateFile.ClusterValues.ClusterID = clusterID
+	stateFile.SetClusterValues(state.ClusterValues{
+		MeasurementSalt: measurementSalt,
+		OwnerID:         ownerID,
+		ClusterID:       clusterID,
+	})
 
 	if err := stateFile.WriteToFile(i.fileHandler, constants.StateFilename); err != nil {
 		return fmt.Errorf("writing state file: %w", err)
--- a/cli/internal/cmd/init.go
+++ b/cli/internal/cmd/init.go
@@ -9,7 +9,6 @@ package cmd
 import (
 	"bytes"
 	"context"
-	"encoding/base64"
 	"encoding/hex"
 	"errors"
 	"fmt"
@@ -157,7 +156,7 @@ func (i *initCmd) initialize(
 		cmd.PrintErrln("WARNING: Attestation temporarily relies on AWS nitroTPM. See https://docs.edgeless.systems/constellation/workflows/config#choosing-a-vm-type for more information.")
 	}
 
-	// TODO(msanft): Remove IDFile as per AB#3354
+	// TODO(msanft): Remove IDFile as per AB#3425
 	i.log.Debugf("Checking cluster ID file")
 	var idFile clusterid.File
 	if err := i.fileHandler.ReadJSON(constants.ClusterIDsFilename, &idFile); err != nil {
@@ -208,7 +207,7 @@ func (i *initCmd) initialize(
 	idFile.MeasurementSalt = measurementSalt
 
 	stateFile.SetClusterValues(state.ClusterValues{
-		MeasurementSalt: base64.StdEncoding.EncodeToString(measurementSalt),
+		MeasurementSalt: measurementSalt,
 	})
 
 	clusterName := clusterid.GetClusterName(conf, idFile)
@@ -246,7 +245,7 @@ func (i *initCmd) initialize(
 	}
 	i.log.Debugf("Initialization request succeeded")
 
-	// TODO(msanft): Remove IDFile as per AB#3354
+	// TODO(msanft): Remove IDFile as per AB#3425
 	i.log.Debugf("Writing Constellation ID file")
 	idFile.CloudProvider = provider
 
--- a/cli/internal/cmd/init.go
+++ b/cli/internal/cmd/init.go
@@ -9,6 +9,7 @@ package cmd
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
 	"encoding/hex"
 	"errors"
 	"fmt"
@@ -41,8 +42,6 @@ import (
 	"github.com/edgelesssys/constellation/v2/cli/internal/helm"
 	"github.com/edgelesssys/constellation/v2/cli/internal/kubecmd"
 	"github.com/edgelesssys/constellation/v2/cli/internal/state"
-	"github.com/edgelesssys/constellation/v2/cli/internal/terraform"
-	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
 	"github.com/edgelesssys/constellation/v2/internal/config"
 	"github.com/edgelesssys/constellation/v2/internal/constants"
 	"github.com/edgelesssys/constellation/v2/internal/crypto"
@@ -73,24 +72,19 @@ func NewInitCmd() *cobra.Command {
 }
 
 type initCmd struct {
-	log           debugLog
-	merger        configMerger
-	spinner       spinnerInterf
-	fileHandler   file.Handler
-	clusterShower infrastructureShower
-	pf            pathprefix.PathPrefixer
+	log         debugLog
+	merger      configMerger
+	spinner     spinnerInterf
+	fileHandler file.Handler
+	pf          pathprefix.PathPrefixer
 }
 
-func newInitCmd(
-	clusterShower infrastructureShower, fileHandler file.Handler,
-	spinner spinnerInterf, merger configMerger, log debugLog,
-) *initCmd {
+func newInitCmd(fileHandler file.Handler, spinner spinnerInterf, merger configMerger, log debugLog) *initCmd {
 	return &initCmd{
-		log:           log,
-		merger:        merger,
-		spinner:       spinner,
-		fileHandler:   fileHandler,
-		clusterShower: clusterShower,
+		log:         log,
+		merger:      merger,
+		spinner:     spinner,
+		fileHandler: fileHandler,
 	}
 }
 
@@ -116,12 +110,7 @@ func runInitialize(cmd *cobra.Command, _ []string) error {
 	defer cancel()
 	cmd.SetContext(ctx)
 
-	tfClient, err := terraform.New(ctx, constants.TerraformWorkingDir)
-	if err != nil {
-		return fmt.Errorf("creating Terraform client: %w", err)
-	}
-
-	i := newInitCmd(tfClient, fileHandler, spinner, &kubeconfigMerger{log: log}, log)
+	i := newInitCmd(fileHandler, spinner, &kubeconfigMerger{log: log}, log)
 	fetcher := attestationconfigapi.NewFetcher()
 	newAttestationApplier := func(w io.Writer, kubeConfig string, log debugLog) (attestationConfigApplier, error) {
 		return kubecmd.New(w, kubeConfig, fileHandler, log)
@@ -168,10 +157,16 @@ func (i *initCmd) initialize(
 		cmd.PrintErrln("WARNING: Attestation temporarily relies on AWS nitroTPM. See https://docs.edgeless.systems/constellation/workflows/config#choosing-a-vm-type for more information.")
 	}
 
+	// TODO(msanft): Remove IDFile as per AB#3354
 	i.log.Debugf("Checking cluster ID file")
 	var idFile clusterid.File
 	if err := i.fileHandler.ReadJSON(constants.ClusterIDsFilename, &idFile); err != nil {
 		return fmt.Errorf("reading cluster ID file: %w", err)
+	}
+
+	stateFile, err := state.ReadFromFile(i.fileHandler, constants.StateFilename)
+	if err != nil {
+		return fmt.Errorf("reading state file: %w", err)
 	}
 
 	i.log.Debugf("Validated k8s version as %s", k8sVersion)
@@ -212,6 +207,10 @@ func (i *initCmd) initialize(
 	}
 	idFile.MeasurementSalt = measurementSalt
 
+	stateFile.SetClusterValues(state.ClusterValues{
+		MeasurementSalt: base64.StdEncoding.EncodeToString(measurementSalt),
+	})
+
 	clusterName := clusterid.GetClusterName(conf, idFile)
 	i.log.Debugf("Setting cluster name to %s", clusterName)
 
@@ -247,11 +246,12 @@ func (i *initCmd) initialize(
 	}
 	i.log.Debugf("Initialization request succeeded")
 
+	// TODO(msanft): Remove IDFile as per AB#3354
 	i.log.Debugf("Writing Constellation ID file")
 	idFile.CloudProvider = provider
 
 	bufferedOutput := &bytes.Buffer{}
-	if err := i.writeOutput(idFile, resp, flags.mergeConfigs, bufferedOutput); err != nil {
+	if err := i.writeOutput(idFile, stateFile, resp, flags.mergeConfigs, bufferedOutput); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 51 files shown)


cli/internal/cmd/upgradeapply.go
```diff
@@ -172,11 +176,18 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, upgradeDir string, fl
 		return err
 	}
 
+	// TODO(msanft): Remove reading from idFile once v2.12.0 is released and read from state file directly.
+	// For now, this is only here to ensure upgradability from an id-file to a state file version.
 	var idFile clusterid.File
 	if err := u.fileHandler.ReadJSON(constants.ClusterIDsFilename, &idFile); err != nil {
 		return fmt.Errorf("reading cluster ID file: %w", err)
 	}
```
daniel-weisse, line 184: We should add some additional logic here in case a state file already exists.
E.g. try to read from state file first, and in case of file-not-found fall back to reading ID file

cli/internal/cmd/upgradeapply.go
```diff
@@ -186,37 +197,52 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, upgradeDir string, fl
...
+
+	// Write the post-migration state to disk
+	if err := stateFile.WriteToFile(u.fileHandler, constants.StateFilename); err != nil {
 		return fmt.Errorf("writing state file: %w", err)
 	}
+
+	cmd.Printf("Infrastructure migrations applied successfully and output written to: %s and %s\n"+
+		"A backup of the pre-upgrade state has been written to: %s\n",
+		flags.pf.PrefixPrintablePath(constants.StateFilename),
+		flags.pf.PrefixPrintablePath(constants.ClusterIDsFilename),
```
daniel-weisse, line 236:
```suggestion
	cmd.Printf("Infrastructure migrations applied successfully and output written to: %s\n"+
		"A backup of the pre-upgrade state has been written to: %s\n",
		flags.pf.PrefixPrintablePath(constants.StateFilename),
```
nit: remove mention of ID file

cli/internal/cmd/upgradeapply.go
```diff
@@ -186,37 +197,52 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, upgradeDir string, fl
...
+	// If infrastructure phase is skipped, we expect the new infrastructure
+	// to be in the Terraform configuration already. Otherwise, perform
+	// the Terraform migrations.
+	var postMigrationInfraState state.Infrastructure
 	if flags.skipPhases.contains(skipInfrastructurePhase) {
-		infraState, err = u.clusterShower.ShowInfrastructure(cmd.Context(), conf.GetProvider())
+		postMigrationInfraState, err = u.clusterShower.ShowInfrastructure(cmd.Context(), conf.GetProvider())
 		if err != nil {
-			return fmt.Errorf("getting infra state: %w", err)
+			return fmt.Errorf("getting Terraform state: %w", err)
```
daniel-weisse, line 211: Can you add a note about replacing this with reading the state file after the next minor update?

cli/internal/cmd/upgradeapply.go
```diff
@@ -186,37 +197,52 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, upgradeDir string, fl
...
-	state := state.NewState(infraState)
-	// TODO(elchead): AB#3424 move this to updateClusterIDFile and correctly handle existing state when writing state
-	if err := u.fileHandler.WriteYAML(constants.StateFilename, state, file.OptOverwrite); err != nil {
+
+	// Write the post-migration state to disk
+	if err := stateFile.WriteToFile(u.fileHandler, constants.StateFilename); err != nil {
 		return fmt.Errorf("writing state file: %w", err)
 	}
+
+	cmd.Printf("Infrastructure migrations applied successfully and output written to: %s and %s\n"+
```
daniel-weisse, line 233: Moving this outside of the `migrateTerraform` function will cause it to get printed every time `upgrade apply` is run. Even when the Terraform is skipped, or there are no updates to be done.

elchead:
i added the link AB link :)

cli/internal/state/state.go
```diff
@@ -7,23 +7,126 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+	}
+}
+
+// NewFromIDFile creates a new cluster state file from the given ID file.
+func NewFromIDFile(idFile clusterid.File) *State {
+	s := New()
+
+	if idFile.ClusterID != "" {
+		s.ClusterValues.ClusterID = idFile.ClusterID
+	}
```
daniel-weisse, line 53: Since this function creates a new `State`, and `idFile` is not a pointer, you can make most of the assignments here unconditionally.

cli/internal/state/state.go
```diff
@@ -7,23 +7,126 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+	if err := mergo.Merge(s, other, mergo.WithOverride); err != nil {
+		return nil, fmt.Errorf("merging state file: %w", err)
+	}
+	return s, nil
+}
+
+// ClusterName returns the name of the cluster.
+func (s *State) ClusterName(cfg *config.Config) string {
+	return cfg.Name + "-" + s.Infrastructure.UID
+}
```
daniel-weisse, line 120: We could just make this a field in the state file to avoid having to keep this function

cli/internal/state/state.go
```diff
@@ -7,23 +7,126 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+
 const (
 	// Version1 is the first version of the state file.
 	Version1 = "v1"
 )
 
+// ReadFromFile reads the state file at the given path and returns the state.
+func ReadFromFile(fileHandler file.Handler, path string) (*State, error) {
+	state := &State{}
+	if err := fileHandler.ReadYAML(path, &state); err != nil {
```
elchead, line 27: just curious why you didn't choose
```
var state State
ReadYAML(path,&state)
return &state
```
why is a double pointer necessary in your implementation?

msanft: This is just a typo actually. Thanks for pointing it out!

go.mod
```diff
@@ -133,6 +133,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require dario.cat/mergo v1.0.0
```
daniel-weisse, line 136:
```suggestion
	sigs.k8s.io/yaml v1.3.0
	dario.cat/mergo v1.0.0
)
```

elchead:
Originally, the scope of this ticket was not to use the state in upgrade and apply yet. It seems that now the idFile is only relevant to migrate during an upgrade?
It seems that from the original AB#3425 only:
- upgrade: remove idFile after writing values to clusterValues

is still open or is there anything else left to do to remove it? If that is the case we could also do it in this ticket I think

cli/internal/cmd/init.go
```diff
@@ -511,6 +504,13 @@ func (i *initCmd) writeOutput(
 	idFile.OwnerID = ownerID
 	idFile.ClusterID = clusterID
 
+	stateFile.ClusterValues.OwnerID = ownerID
+	stateFile.ClusterValues.ClusterID = clusterID
```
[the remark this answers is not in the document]
elchead: I agree. ClusterValues should only be atomically set. Such that you always know that you have a complete state when you read from it.
I also don't like the empty New() for that reasons.
What do you think of the state struct approach that I posted some time ago (https://teams.microsoft.com/l/message/19:87121fb7568f4364bfaf6fc2f15977a5@thread.skype/1686490443211?tenantId=adb650a8-5da3-4b15-b4b0-3daf65ff7626&groupId=62348c88-6944-413d-9232-9d3adb03009e&parentMessageId=1686490443211&teamName=Edgeless&channelName=Today%20I%20Learned&createdTime=1686490443211&allowXTenantAccess=false)

This way we could distinguish between a state with only InfrastructureState (after create and during init) and a state with both (during upgrade)

cli/internal/cmd/upgradeapply.go
msanft: True. I think we can spare the "migrations applied successfully" print in case of a manual infrastructure migration and leave the implementation of this simple. What do you think?

cli/internal/cmd/init_test.go
```diff
@@ -103,31 +104,36 @@ func TestInitialize(t *testing.T) {
 		"initialize some gcp instances": {
 			provider:      cloudprovider.GCP,
 			idFile:        &clusterid.File{IP: "192.0.2.1"},
+			stateFile:     &state.State{Version: state.Version1, Infrastructure: state.Infrastructure{ClusterEndpoint: "192.0.2.1"}},
```
elchead, line 107: nit: should we use the constructor and make the fields private if we want to go with the struct approach for stronger type safety? This would also allow to do some validation if we want to later on

elchead:
please also run the miniconstellation test in the end

msanft:
> Originally, the scope of this ticket was not to use the state in upgrade and apply yet. It seems that now the idFile is only relevant to migrate during an upgrade? It seems that from the original [AB#3425](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3425) only:
> 
> * upgrade: remove idFile after writing values to clusterValues
> 
> is still open or is there anything else left to do to remove it? If that is the case we could also do it in this ticket I think

Not really imo. At this point (in this PR), we still:
- Write to `idFile` on create and init.
- Read the state from ` idFile` during init.

If we choose to implement the fallback from state-file to id-file on upgrade, as Daniel explained above, in this PR, this would be part of `- read state file instead of idFile during init + upgrade`. Other than that, I don't really see anything of this PR involved in AB#3425

cli/internal/cmd/upgradeapply.go
daniel-weisse: You could just move this back into `migrateTerraform`

cli/internal/cmd/init_test.go
msanft: We won't be able to marshal it to YAML then without the hacks Otto mentioned. I don't know if we want to commit to this yet. I think using constructors in non-test code is a great idea. But here, IMO, it makes it less readable, as you know what's inside the struct directly.

cli/internal/cmd/upgradeapply.go
msanft: "note" as in note to the user, or note to the developers? The latter doesn't really make sense to me, as this specific code part should be independent from id- or state-file as it reads from Terraform state.

cli/internal/cmd/upgradeapply.go
msanft: This is the obvious way, but it would come at the consequence of no notice to the user that the state-file has been updated with the new infrastructure values in case of a manual Terraform migration (i.e. skipTerraformPhase)

daniel-weisse: In case of manual Terraform migration, the state file changes would have been made by the user, so there shouldnt be a need to inform them about the changes

cli/internal/state/state.go
msanft: Can I just edit this in the RFC?

cli/internal/cmd/init_test.go
elchead: oh you are right! Yeah lets keep as is then

elchead:
> > Originally, the scope of this ticket was not to use the state in upgrade and apply yet. It seems that now the idFile is only relevant to migrate during an upgrade? It seems that from the original [AB#3425](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3425) only:
> > 
> > * upgrade: remove idFile after writing values to clusterValues
> > 
> > is still open or is there anything else left to do to remove it? If that is the case we could also do it in this ticket I think
> 
> Not really imo. At this point (in this PR), we still:
> 
> * Write to `idFile` on create and init.
> * Read the state from ` idFile` during init.
> 
> If we choose to implement the fallback from state-file to id-file on upgrade, as Daniel explained above, in this PR, this would be part of `- read state file instead of idFile during init + upgrade`. Other than that, I don't really see anything of this PR involved in [AB#3425](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3425) The [original ticket](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3425) mentions to read from `idFile` and write to the state-file in upgrade apply, which, to me, is what I did in this PR.

By "read state" i originally meant that the idFile is still used in all function signatures to keep the PRs smaller, but I see that it was unclear and its fine as is now :)
My point is that since we use the new state in the upgrade now, we could already remove the idFile at this stage.
Then only this is left:
- don't write idFile during init + create anymore
which seems very small too.

msanft:
> I originally meant that the idFile is still used in all function signatures to keep the PRs smaller (in [AB#3424](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3424) (this PR)), but I see that it was unclear and its fine as is now :) My point is that since we use the new state in the upgrade now, we could already remove the idFile at this stage. Then only this is left:
> 
> * don't write idFile during init + create anymore
>   which seems very small too.

Sorry for the misunderstanding here. My intention was to still keep the file intact for now to not break existing development clusters, while relying on the inputs from the state file where possible. I do really like the point of splitting this up into many smaller PRs though!

I already have a separate branch where I'm working on the idFile-Removal. My suggestion would be to open a separate PR with these changes, based on this PR here. This way, reviewers have an easier time checking the changes. I would edit the PR description to reflect exactly which changes I made within which PR then. What do you think?

cli/internal/cmd/upgradeapply.go
```diff
@@ -301,52 +328,39 @@ func (u *upgradeApplyCmd) migrateTerraform(cmd *cobra.Command, conf *config.Conf
...
-	infraState, err := u.clusterShower.ShowInfrastructure(cmd.Context(), conf.GetProvider())
-	if err != nil {
-		return infraState, fmt.Errorf("getting Terraform output: %w", err)
-	}
-	state := state.NewState(infraState)
-	// TODO(elchead): AB#3424 move this to updateClusterIDFile and correctly handle existing state when writing state
-	if err := u.fileHandler.WriteYAML(constants.StateFilename, state, file.OptOverwrite); err != nil {
-		return infraState, fmt.Errorf("writing state file: %w", err)
-	}
-	return infraState, nil
```
daniel-weisse, line 349: Is it safe to remove this?
Afaik we still need to check if a state file is missing the infrastructure state and otherwise add it.
Though this could also be done outside of this function, at the start of `upgrade apply` as part of the migration to v2.12

cli/internal/cmd/upgradeapply.go
```diff
@@ -292,7 +319,7 @@ func (u *upgradeApplyCmd) migrateTerraform(cmd *cobra.Command, conf *config.Conf
 
 	hasDiff, err := u.clusterUpgrader.PlanClusterUpgrade(cmd.Context(), cmd.OutOrStdout(), vars, conf.GetProvider())
 	if err != nil {
-		return res, fmt.Errorf("planning terraform migrations: %w", err)
+		return state.Infrastructure{}, fmt.Errorf("planning terraform migrations: %w", err)
 	}
 
 	if hasDiff {
```
daniel-weisse, line 325: Could probably remove a lot of nesting by returning early if `!hasDiff`

cli/internal/cmd/upgradeapply.go
msanft: How would one get into a case where a state file exists, but no infrastructure is there?

- For new clusters, the state file should be generated.
- For pre-state-file clusters, the migration should happen from id-file to state-file.

daniel-weisse: But the `NewFromIDFile` does not set all relevant infrastructure values for the state file, does it?
We still need to supplement some from the Terraform output, which currently does not happen if that phase is skipped or there are not Terrraform changes to be made.

msanft: I may misunderstand, but merging postMigrationInfraState into the state-file should do this IMO, as it is either populated with the values from the Terraform configuration in case of a manual migration, or the outputs of the Terraform migration in case of the "automatic" migration. Or do I miss something here?
https://github.com/edgelesssys/constellation/blob/a631138f2819bdea812fd62a83fae1afe35de19b/cli/internal/cmd/upgradeapply.go#L211-L238

daniel-weisse: If `migrateTerraform` detects no diff, the function returns an empty Infrastructure struct, meaning this part wouldn't add any missing values to the state file.

cli/internal/state/state_test.go
```diff
@@ -0,0 +1,392 @@
...
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+
+			err := tc.state.WriteToFile(tc.fh, constants.StateFilename)
+
+			if tc.wantErr {
+				assert.Error(err)
+			} else {
+				assert.NoError(err)
+				assert.Equal(mustMarshalYaml(t, tc.state), mustReadFromFile(t, tc.fh))
```
elchead, line 98: nit: why not
_, err := Read(..)
assert.NoError(err)

cli/internal/cmd/upgradeapply.go
```diff
@@ -186,37 +204,47 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, upgradeDir string, fl
...
+	if err := stateFile.WriteToFile(u.fileHandler, constants.StateFilename); err != nil {
 		return fmt.Errorf("writing state file: %w", err)
 	}
+
 	// extend the clusterConfig cert SANs with any of the supported endpoints:
 	// - (legacy) public IP
 	// - fallback endpoint
 	// - custom (user-provided) endpoint
-	sans := append([]string{idFile.IP, conf.CustomEndpoint}, idFile.APIServerCertSANs...)
+	// At this point, state file and id-file should have been merged, so we can use the state file.
```
elchead, line 246: add comment to remove in next PR

elchead (approved):
lgtm

msanft:
[E2E Upgrade](https://github.com/edgelesssys/constellation/actions/runs/6403445784)
[E2E Azure](https://github.com/edgelesssys/constellation/actions/runs/6403451044)
[E2E MiniConstellation](https://github.com/edgelesssys/constellation/actions/runs/6403457210)

daniel-weisse (approved):

derpsteb (approved):
Changes to `cli/internal/helm` LGTM :+1:

docs/docs/architecture/orchestration.md
```diff
@@ -28,7 +28,7 @@ Altogether, the following files are generated during the creation of a Constella
 
 After the creation of your cluster, the CLI will provide you with a Kubernetes `kubeconfig` file.
 This file grants you access to your Kubernetes cluster and configures the [kubectl](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/) tool.
-In addition, the cluster's [identifier](orchestration.md#post-installation-configuration) is returned and stored in a file called `constellation-id.json`
+In addition, the cluster's [identifier](orchestration.md#post-installation-configuration) is returned and stored in a file called `constellation-state.yaml`
```
thomasten, line 31: Please check if the rest of this section is up to date. At least "an ID file" needs to be updated in the bulleted list above.

thomasten (approved):
only reviewed the docs, which LGTM

malt3 (approved):

PR #2397, opened 2023-10-02
github: add AB ticket link to PR template

msanft:
### Context

Mentioning Azure DevOps tickets in GitHub PRs links the ticket to the PR in Azure DevOps. This helps at backlog refinements and general mapping of tickets to PRs.

### Proposed change(s)

- Add a link to an AB ticket in the GitHub pull request template.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -20,6 +20,7 @@ Feel free to edit, complete or extend this list while the PR is open.
 
 <!-- (uncomment if applicable)
 ### Additional info
+- [AB#1234](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/1234)
 - Any additional information or context
 -->
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -18,11 +18,10 @@ Feel free to edit, complete or extend this list while the PR is open.
 - link to the issue
 -->
 
-<!-- (uncomment if applicable)
 ### Additional info
+<!-- Remove items that do not apply -->
 - [AB#1234](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/1234)
 - Any additional information or context
--->
 
 ### Checklist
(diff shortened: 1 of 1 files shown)


elchead (changes requested):
I like to put in the reminder, but I think that the link is not very helpful.
You can simply type AB#$NBR to add the link.
My suggestion would be to make additional info not optional and add AB# in the template

msanft:
> You can simply type AB#$NBR to add the link. 

What do you mean by that?

> My suggestion would be to make additional info not optional and add AB# in the template

Agreed.

elchead:
You can ofc also provide the link but I think the template should not "force" using a link (i.e. let's just put AB# and let the creator decide how he links)

msanft:
> You can ofc also provide the link but I think the template should not "force" using a link (i.e. let's just put AB# and let the creator decide how he links)

I don't really see the point for only putting the number. When you paste the number, you could also just paste it again in the link. What's your point against putting the link exactly?

elchead:
> > You can ofc also provide the link but I think the template should not "force" using a link (i.e. let's just put AB# and let the creator decide how he links)
> 
> I don't really see the point for only putting the number. When you paste the number, you could also just paste it again in the link. What's your point against putting the link exactly?

I will just always have to delete it since I think its easier to memorize a number than copying the link and also adjusting the link name. But it's ok.

elchead (approved):

msanft:
> I will just always have to delete it since I think its easier to memorize a number than copying the link and also adjusting the link name. But it's ok.

Just to clarify the misunderstanding (at least I think it is one) here; My idea was that one only memorizes the number and replaces both `1234` with the memorized number (the one in the link name as well as the one in the link). This way, both name and link should be correct

elchead:
FYI putting AB#number already creates a link by the github bot ;) @msanft

PR #2398, opened 2023-10-02
s3proxy: add keyservice integration

derpsteb:
### Context
#2385 introduces a minimal version of the s3proxy. This PR adds integration with the C11n keyservice on top of #2385.

### Proposed change(s)

- Use tink for crypto operations. Previously used stdlib's `crypto`. This currently points to a fork. A PR is open, but Bazel kept failing to find the upstream reference in the PR. So I added the replace for now.
- Fetch a key encryption key (KEK) from the C11n keyservice. This kek will encrypt all data encryption keys (DEK). For reference: we didn't think of this approach ourself. It is called envelope encryption and seems to a widely accepted pattern.
- Generate a new DEK for each `PutObject` request. Use the DEK to encrypt the object. Store the encrypted DEK as metadata for the object. 
- Use tink's keywrapping to encrypt DEKs.

### Additional info
- Tested manually through filestash, and mint.

### Checklist

- [x] Link to Milestone


daniel-weisse (approved):

s3proxy/internal/router/router.go
```diff
@@ -41,12 +55,24 @@ var (
 // Router implements the interception logic for the s3proxy.
 type Router struct {
 	region string
+	kek    []byte
```
3u13r, line 58: Could we maybe simply use `[32]byte` as a type? 
The downside is future crypto/key agility.

s3proxy/internal/kms/kms.go
```diff
@@ -0,0 +1,76 @@
...
+		endpoint: endpoint,
+		grpc:     client{},
+	}
+}
+
+// GetDataKey returns a data encryption key for the given UUID.
+func (c Client) GetDataKey(ctx context.Context, keyID string, length int) ([]byte, error) {
+	log := c.log.With("keyID", keyID, "endpoint", c.endpoint)
+	// the KMS does not use aTLS since traffic is only routed through the Constellation cluster
+	// cluster internal connections are considered trustworthy
```
3u13r, line 43: I think we should also document this in a README.md or something, should someone external decide to use this outside of Constellation (I know this is somewhat hard given the KMS dependency).

3u13r (changes requested):

s3proxy/internal/router/router.go
derpsteb: If we want to change from that keysize we are probably also changing the crypto primitive. So a software update will be necessary anyways.

Changed.

s3proxy/internal/kms/kms.go
derpsteb: Will be part of the docs. Also added to the readme.

3u13r (approved):
LGTM

PR #2402, opened 2023-10-04
cli: remove id-file

msanft:
### Context

As #2395 introduces the switch from the id-file to the more extensive state-file, the ID-file should be removed from the existing commands. (i.e. `constellation init` and `constellation upgrade apply`)

### Proposed change(s)

- Write to the state-file during `constellation create`
- Read / Write from / to the state-file during `constellation init`
- Move the ID-file to a backup on `constellation upgrade apply` (i.e. when upgrading from a id-file version to a state-file version)
- Read from state-file instead of ID-file in `constellation verify`
- Read from state-file instead of ID-file in `constellation mini`
- Read from state-file instead of ID-file in `constellation recover`
- Read from state-file instead of ID-file in `constellation terminate`
- Read from state-file instead of ID-file in `constellation iam destroy`
- Read from state-file instead of ID-file in `cdbg deploy`
- Remove references to the ID-file from the docs, without fully documenting the self-managed infrastructure workflow yet.

### Additional info

- [AB#3425](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3425)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/init.go
+++ b/cli/internal/cmd/init.go
@@ -36,7 +36,6 @@ import (
 
 	"github.com/edgelesssys/constellation/v2/bootstrapper/initproto"
 	"github.com/edgelesssys/constellation/v2/cli/internal/cloudcmd"
-	"github.com/edgelesssys/constellation/v2/cli/internal/clusterid"
 	"github.com/edgelesssys/constellation/v2/cli/internal/cmd/pathprefix"
 	"github.com/edgelesssys/constellation/v2/cli/internal/helm"
 	"github.com/edgelesssys/constellation/v2/cli/internal/kubecmd"
@@ -156,13 +155,6 @@ func (i *initCmd) initialize(
 		cmd.PrintErrln("WARNING: Attestation temporarily relies on AWS nitroTPM. See https://docs.edgeless.systems/constellation/workflows/config#choosing-a-vm-type for more information.")
 	}
 
-	// TODO(msanft): Remove IDFile as per AB#3425
-	i.log.Debugf("Checking cluster ID file")
-	var idFile clusterid.File
-	if err := i.fileHandler.ReadJSON(constants.ClusterIDsFilename, &idFile); err != nil {
-		return fmt.Errorf("reading cluster ID file: %w", err)
-	}
-
 	stateFile, err := state.ReadFromFile(i.fileHandler, constants.StateFilename)
 	if err != nil {
 		return fmt.Errorf("reading state file: %w", err)
@@ -181,7 +173,10 @@ func (i *initCmd) initialize(
 	}
 	i.log.Debugf("Checked license")
 
-	conf.UpdateMAAURL(idFile.AttestationURL)
+	if stateFile.Infrastructure.Azure != nil {
+		conf.UpdateMAAURL(stateFile.Infrastructure.Azure.AttestationURL)
+	}
+
 	i.log.Debugf("Creating aTLS Validator for %s", conf.GetAttestationConfig().GetVariant())
 	validator, err := cloudcmd.NewValidator(cmd, conf.GetAttestationConfig(), i.log)
 	if err != nil {
@@ -199,14 +194,15 @@ func (i *initCmd) initialize(
 	if err != nil {
 		return fmt.Errorf("generating master secret: %w", err)
 	}
-	i.log.Debugf("Generated measurement salt")
+
+	i.log.Debugf("Generating measurement salt")
 	measurementSalt, err := crypto.GenerateRandomBytes(crypto.RNGLengthDefault)
 	if err != nil {
 		return fmt.Errorf("generating measurement salt: %w", err)
 	}
-	idFile.MeasurementSalt = measurementSalt
+	stateFile.ClusterValues.MeasurementSalt = measurementSalt
 
-	clusterName := clusterid.GetClusterName(conf, idFile)
+	clusterName := stateFile.ClusterName(conf)
 	i.log.Debugf("Setting cluster name to %s", clusterName)
 
 	cmd.PrintErrln("Note: If you just created the cluster, it can take a few minutes to connect.")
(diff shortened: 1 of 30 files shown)


msanft:
`check` fail is due to actionlint and can be ignored

cli/internal/state/state.go
```diff
@@ -56,7 +56,7 @@ func NewFromIDFile(idFile clusterid.File, cfg *config.Config) *State {
 			UID:               idFile.UID,
 			ClusterEndpoint:   idFile.IP,
 			APIServerCertSANs: idFile.APIServerCertSANs,
-			InitSecret:        string(idFile.InitSecret),
```
elchead, line 59: just curious: what was the reason to make it a byte slice?

cli/internal/cmd/terminate_test.go
daniel-weisse, line 1: Can you add a TODO to `terminate.go` for removing the removal of ID file after the v1.12 release?

cli/internal/state/state.go
msanft: Initially I thought a string fits best as this essentially is just a random string that we receive from the applied Terraform config. However, at almost all places we use this, it needs to be a byte slice. And as it contains only printable characters either way, I figured to just save it as a byte slice directly to save on unnecessary conversions.

cli/internal/cmd/verify.go
```diff
@@ -186,24 +186,28 @@ func (c *verifyCmd) parseVerifyFlags(cmd *cobra.Command, fileHandler file.Handle
...
 		if emptyIDs {
-			cmd.Printf("Using ID from %q. Specify --cluster-id to override this.\n", pf.PrefixPrintablePath(constants.ClusterIDsFilename))
-			ownerID = idFile.OwnerID
-			clusterID = idFile.ClusterID
+			cmd.Printf("Using ID from %q. Specify --cluster-id to override this.\n", pf.PrefixPrintablePath(constants.StateFilename))
+			ownerID = stateFile.ClusterValues.OwnerID
+			clusterID = stateFile.ClusterValues.ClusterID
+		}
+		if stateFile.Infrastructure.Azure != nil {
+			attestationURL = stateFile.Infrastructure.Azure.AttestationURL
```
daniel-weisse, line 210: Performing this check inside the `if emptyEndpoint || emptyIDs` block, will cause `attestationURL` to always be an empty string if a user specifies IDs and endpoint manually

cli/internal/cmd/init.go
```diff
@@ -205,15 +194,15 @@ func (i *initCmd) initialize(
...
 	}
-	i.log.Debugf("Generated measurement salt")
+
+	i.log.Debugf("Generating measurement salt")
 	measurementSalt, err := crypto.GenerateRandomBytes(crypto.RNGLengthDefault)
 	if err != nil {
 		return fmt.Errorf("generating measurement salt: %w", err)
 	}
-	idFile.MeasurementSalt = measurementSalt
+	stateFile.ClusterValues.MeasurementSalt = measurementSalt
```
elchead, line 203: we write the cluster values in writeOutput
```suggestion
```

cli/internal/cmd/verify.go
msanft: Since we don't want a failure if the arguments are specified but no state-file is present, how about something like this?

```go
var attestationURL string
	stateFile, err := state.ReadFromFile(fileHandler, constants.StateFilename)
	if err == nil {
		if stateFile.Infrastructure.Azure != nil {
			attestationURL = stateFile.Infrastructure.Azure.AttestationURL
		}
	}
```

`if err == nil``` smells like an antipattern though

cli/internal/cmd/init_test.go
```diff
@@ -156,30 +149,27 @@ func TestInitialize(t *testing.T) {
 			masterSecretShouldExist: true,
 			wantErr:                 true,
 		},
-		"empty id file": {
+		"empty state file": {
 			provider:      cloudprovider.GCP,
-			idFile:        &clusterid.File{},
 			stateFile:     &state.State{},
```
elchead, line 154: nit: use state.New()

msanft: IMO you wouldn't want to use constructors in the tests here, as they could potentially hide which values the struct is being populated with. In this case, it's not a problem, as right now it only fills in the version. But I think we should generally avoid using constructors in tests

cli/internal/cmd/verify.go
daniel-weisse: You could just keep the same logic as before, but replace ID file with state file.

cli/internal/cmd/init_test.go
elchead: mmh but then add the version. The test might in theory error only because the version is missing.

cli/internal/cmd/upgradeapply.go
```diff
@@ -239,6 +239,13 @@ func (u *upgradeApplyCmd) upgradeApply(cmd *cobra.Command, upgradeDir string, fl
 		return fmt.Errorf("writing state file: %w", err)
 	}
 
+	// TODO(msanft): Remove this after v2.12.0 is released, as we do not support
```
elchead, line 242: Why don't we just remove the file after succesful migration to the state file?

msanft: If something goes wrong in the process (god knows what), IMO it's better to have a backup file than just remove it

elchead:
We should update the references to the id file in the docs. Should this be done in a separate PR? i think it fits well in this PR

msanft:
> We should update the references to the id file in the docs. Should this be done in a separate PR? i think it fits well in this PR

Looking at where we write about this currently in the docs, I think we should fully support the self-managed infrastructure flow with the state file first and then write the docs all in one. What do you think?

cli/internal/cmd/verify.go
```diff
@@ -186,27 +186,35 @@ func (c *verifyCmd) parseVerifyFlags(cmd *cobra.Command, fileHandler file.Handle
...
 	c.log.Debugf("Flag 'raw' set to %t", force)
 
-	var idFile clusterid.File
-	if err := fileHandler.ReadJSON(constants.ClusterIDsFilename, &idFile); err != nil && !errors.Is(err, afero.ErrFileNotFound) {
-		return verifyFlags{}, fmt.Errorf("reading cluster ID file: %w", err)
+	// Get empty values from state file
+	stateFile, err := state.ReadFromFile(fileHandler, constants.StateFilename)
+	isFileNotFound := errors.Is(err, afero.ErrFileNotFound)
+	if isFileNotFound {
+		stateFile = state.New() // error compat
```
daniel-weisse, line 193:
```suggestion
	if isFileNotFound {
		cmd.Errorf("Failed to load state file: %s", err)
		stateFile = state.New() // error compat
```
Should we log an error here just to make sure users notice potential errors correctly?

daniel-weisse (approved):

PR #2408, opened 2023-10-04
csi: fix concurrent use of `cryptmapper` package

daniel-weisse:
### Context
The `csi/cryptmapper` package currently has some problems with concurrency.
More specifically, if one application has multiple threads making use of the functions offered by the package, calling `Open` or `Close` can return errors as the device in use was not yet freed by another thread.

### Proposed change(s)
- Work on a fresh [`cryptsetup` instance](https://github.com/edgelesssys/constellation/blob/7e899d09c401f96cee8e3192aeb9486c2d8e4d35/internal/cryptsetup/cryptsetup.go#L52) in every function
- Allow idempotent calling of `Open`
  - If a mapped device with the name already exists (and the integrity device exists) simply return without an error

### Additional info
- Required for a new release of the CSI drivers
- Does not affect the currently released driver versions

### Checklist
- [x] Link to Milestone

### DIFF
--- a/csi/cryptmapper/cryptmapper.go
+++ b/csi/cryptmapper/cryptmapper.go
@@ -12,6 +12,7 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"os"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


malt3 (approved):
Nice work

PR #2418, opened 2023-10-06
docs: add note about current AWS CVM issues

thomasten:
What do you think of this kind of note? Is it at the right place?

### DIFF
--- a/docs/docs/getting-started/first-steps.md
+++ b/docs/docs/getting-started/first-steps.md
@@ -170,6 +170,20 @@ If you encounter any problem with the following steps, make sure to use the [lat
 
     :::
 
+    :::note
+
+    **AWS**: Currently, confidential VMs may temporarily fail to boot.
+    AWS is actively working on a solution.
+    If creating or initializing a cluster fails, terminate the cluster and try again later.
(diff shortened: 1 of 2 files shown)


m1ghtym0 (approved):

docs/docs/getting-started/first-steps.md
```diff
@@ -170,6 +170,16 @@ If you encounter any problem with the following steps, make sure to use the [lat
 
     :::
 
+    :::note
+
+    **AWS**: Currently, confidential VMs may temporarily fail to boot. If creating or initializing a cluster fails, terminate the cluster and try again later. Alternatively, you can use non-confidential Nitro VMs for evaluation purposes. To do so, generate a new configuration with the `aws-nitro-tpm` attestation variant:
```
derpsteb, line 175:
```suggestion
    **AWS**: Currently, confidential VMs may temporarily fail to boot. AWS is actively working on a solution. If creating or initializing a cluster fails, terminate the cluster and try again later. Alternatively, you can use non-confidential Nitro VMs for evaluation purposes. To do so, generate a new configuration with the `aws-nitro-tpm` attestation variant:
```
Mention that this is out of our control? Could also be sth like `We are actively working with AWS to remediate the issue.`

PR #2417, opened 2023-10-06
s3proxy: add new page to documentation

derpsteb:
### Context
s3proxy is now merged to main. This PR adds user-facing docs for deployment, background information and the example application Filestash.

### Proposed change(s)

- Add user-facing docs for s3proxy. Includes link to instructions for example application.

### Checklist

- [x] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)
- [x] Add labels (e.g., for changelog category)
- [x] Is PR title adequate for changelog?
- [x] Link to Milestone

PR #2420, opened 2023-10-09
s3proxy: add allow-multipart flag

derpsteb:
### Context

Currently the s3proxy MVP does not implement encryption for multipart uploads. This may lead to inadvertent data leakage, if users are not aware of this. Thus this PR changes the default behavior to a more secure one: block multipart uploads. The old behavior can still be triggered by supplying the new flag `--allow-multipart`.

### Proposed change(s)

- Move handler functions into separate file
- Change default behavior of s3proxy to block multipart upload requests
- Add `--allow-multipart` flag to trigger the previous default behavior

### Additional info

- Manually tested through mint e2e tests + filestash.

### Checklist

- [x] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs) #2417 
- [x] Link to Milestone

### DIFF
--- a/s3proxy/cmd/main.go
+++ b/s3proxy/cmd/main.go
@@ -100,7 +100,7 @@ func parseFlags() (cmdFlags, error) {
 	region := flag.String("region", defaultRegion, "AWS region in which target bucket is located")
 	certLocation := flag.String("cert", defaultCertLocation, "location of TLS certificate")
 	kmsEndpoint := flag.String("kms", "key-service.kube-system:9000", "endpoint of the KMS service to get key encryption keys from")
-	forwardMultipartReqs := flag.Bool("allow-multipart", false, "forward multipart requests to the target bucket; beware: this may leak data to AWS")
+	forwardMultipartReqs := flag.Bool("allow-multipart", false, "forward multipart requests to the target bucket; beware: this may store unencrypted data on AWS. See the documentation for more information")
 	level := flag.Int("level", defaultLogLevel, "log level")
 
 	flag.Parse()
--- a/s3proxy/cmd/main.go
+++ b/s3proxy/cmd/main.go
@@ -49,6 +49,10 @@ func main() {
 
 	logger := logger.New(logger.JSONLog, logger.VerbosityFromInt(flags.logLevel))
 
+	if flags.forwardMultipartReqs {
+		logger.Warnf("configured to forward multipart uploads, this may leak data to AWS")
+	}
+
 	if err := runServer(flags, logger); err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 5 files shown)


s3proxy/cmd/main.go
```diff
@@ -96,6 +100,7 @@ func parseFlags() (cmdFlags, error) {
 	region := flag.String("region", defaultRegion, "AWS region in which target bucket is located")
 	certLocation := flag.String("cert", defaultCertLocation, "location of TLS certificate")
 	kmsEndpoint := flag.String("kms", "key-service.kube-system:9000", "endpoint of the KMS service to get key encryption keys from")
+	forwardMultipartReqs := flag.Bool("allow-multipart", false, "forward multipart requests to the target bucket; beware: this may leak data to AWS")
```
thomasten, line 103:
```suggestion
	forwardMultipartReqs := flag.Bool("allow-multipart", false, "forward multipart requests to the target bucket; beware: this may store unencrypted data on AWS")
```
Should we be more explicit here?
Maybe also "see the docs for more info"?

s3proxy/internal/router/handler.go
```diff
@@ -0,0 +1,191 @@
...
+		if _, err := w.Write(body); err != nil {
+			log.With(zap.Error(err)).Errorf("Write")
+			http.Error(w, fmt.Sprintf("writing body: %s", err.Error()), http.StatusInternalServerError)
+			return
+		}
+	}
+}
+
+// handleCreateMultipartUpload logs the request and blocks with an error message.
+func handleCreateMultipartUpload(log *logger.Logger) http.HandlerFunc {
```
thomasten, line 154: Is my understanding correct that it would suffice to only handle this request? Could reduce code a bit, but I don't have a strong opinion here. Maybe better be safe than sorry if we're unsure.

derpsteb: I have the same understanding. Adding stub handlers for all endpoints related to the function seemed easier to debug to me. A user who didn't fully read the docs might wonder why only the upload fails, otherwise. But I see it's a weak argument. 
More "keep the code" arguments: establish a pattern for how to add more endpoints; stubs for eventual multipart upload implementation. 

Happy to remove it if you think we should keep code minimal.

thomasten: Makes sense. Let's keep it.

s3proxy/cmd/main.go
derpsteb: Added both. Not sure about formatting.
I feel like your suggestion is as explicit as we can get. Apart from capital letters. But I feel like we are mentioning the aspect every step of the way now (docs, help text, logs).

PR #2429, opened 2023-10-11
api: for Azure attestationconfigapi use TCB values from SNP report instead of MAA token

elchead:
### Context

Since the MAA version values are intransparent and were found to not be reliable for TCB version inference, we are now using the values from the SNP report instead.

### Proposed change(s)

- store the json output of verify on Azure and pass to the attestationconfigapi upload tool
- check that Launch TCB, Committed TCB, and Reported TCB are equal
- fix go-sev-guest library version to newer hash on main, because current release has debug prints

### Additional info

- [AB#3447](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3447)

### Checklist

- [ ] Link to Milestone


elchead:
Run: https://github.com/edgelesssys/constellation/actions/runs/6479025795/job/17591789611

elchead:
The verify test currently only allows uploads to the API on main. Is this restriction necessary? It would be easier to test without it and also I don't see any no harm

internal/api/attestationconfigapi/cli/main.go
```diff
@@ -148,8 +152,32 @@ func runCmd(cmd *cobra.Command, _ []string) (retErr error) {
 	return nil
 }
 
+func allEqual[T comparable](args ...T) bool {
```
derpsteb, line 155: Please add a unittest with at least the following cases:
1. one input arg
2. three input args that are equal
3. three input args where second and third element are different

Args should be snpReports.

.github/actions/e2e_verify/action.yml
```diff
@@ -66,36 +66,32 @@ runs:
...
-          echo "Extracting TCB versions for API update"
-          startMAAToken="Microsoft Azure Attestation Token:"
-          endMAAToken="Verification OK"
-          sed -n "/${startMAAToken}/,/${endMAAToken}/ { /${startMAAToken}/d; /${endMAAToken}/d; p }" <<< "${verifyOut}" > "maa-claims-${node}.json"
+          kill $forwarderPID
         done
 
     - name: Login to AWS
-      if: github.ref_name == 'main' && inputs.cloudProvider == 'azure'
+      if: inputs.cloudProvider == 'azure'
```
derpsteb, line 80: I would like to keep the check: unreviewed code shouldn't push to our production API.

Since this only affects the upload step the action can still be run on the feature branch. This will catch syntax errors. The CLI itself can be tested with the e2e-attestationconfigapi test. Incorrect invocations will not be caught, which I agree is an inconvenience. But in this case it's better to live with this small inconvenience over pushing malformed data to the API imo.

elchead: I see the potential danger, but I still believe that we should disable this check temporarily after review and test before merging into main.

elchead:
e2e succeeded https://github.com/edgelesssys/constellation/actions/runs/6533033933/job/17737387605.

.github/actions/e2e_verify/action.yml
elchead: I removed the check again

go.mod
```diff
@@ -137,7 +138,10 @@ require (
...
 )
 
-require github.com/google/go-tdx-guest v0.2.2 // indirect
+require (
+	github.com/google/go-attestation v0.5.0 // indirect
+	github.com/google/go-tdx-guest v0.2.2 // indirect
+)
 
 require (
 	cloud.google.com/go v0.110.2 // indirect
```
daniel-weisse, line 147:
```suggestion
require (
	github.com/google/go-attestation v0.5.0 // indirect
	github.com/google/go-tdx-guest v0.2.2 // indirect
	cloud.google.com/go v0.110.2 // indirect
```
nit: group dependencies

daniel-weisse (commented):
Only reviewed the Go code.
Looks mostly good to me

daniel-weisse (approved):

PR #2451, opened 2023-10-12
ci: remove python from codeql

malt3:
### Context

The python linter checks always fail. We have almost no python code to begin with, so this should not be problem.
We also never saw any valid issues found by CodeQL.

### Proposed change(s)

- ci: remove python from codeql

### Checklist

- [x] Link to Milestone


3u13r (approved):
LGTM

PR #2453, opened 2023-10-16
cli: add field docs to the state file

msanft:
### Context

The state file was missing in-line documentation, aligned with what the config file has.

### Proposed change(s)

- Add in-line documentation to the fields in the state file.
- Mark the fields that (always) need to be filled manually in case of self-managed infrastructure.

### Additional info

- [AB#3426](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3426)

### Checklist

- [x] Link to Milestone


cli/internal/state/state.go
```diff
@@ -30,9 +35,99 @@ func ReadFromFile(fileHandler file.Handler, path string) (*State, error) {
...
+	//   State of the cluster's cloud resources. These values are retrieved during
+	//   cluster creation. In the case of self-managed infrastructure, the marked
+	//   fields in this struct should be filled by the user as per
+	//   https://docs.edgeless.systems/constellation/workflows/create.
 	Infrastructure Infrastructure `yaml:"infrastructure"`
-	ClusterValues  ClusterValues  `yaml:"clusterValues"`
+	// description: |
+	//   DO NOT EDIT. State of the Constellation Kubernetes cluster.
+	//   These values are set during cluster initialization and should not be changed.
+	ClusterValues ClusterValues `yaml:"clusterValues"`
```
elchead, line 53: idea:

cli/internal/state/state.go
```diff
@@ -30,9 +35,99 @@ func ReadFromFile(fileHandler file.Handler, path string) (*State, error) {
...
+	// description: |
+	//   Salt used to generate the ClusterID on the bootstrapping node.
+	MeasurementSalt []byte `yaml:"measurementSalt"`
+}
+
+// Infrastructure describe the state related to the cloud resources of the cluster.
+type Infrastructure struct {
+	// description: |
+	//   Unique identifier the cluster's cloud resources are tagged with.
+	UID string `yaml:"uid"`
```
elchead, line 73: should we indicate optional values with a pointer?

daniel-weisse (approved):
We should rethink custom marshaling of `[]byte` for the state file.
Currently, `MeasurementSalt` and `InitSecret` get rendered as a list of byte values, e.g.

```yaml
MeasurementSalt:
    - 12
    - 53
    - 12
    - 53
    // ...
```

Should probably be done in a separate PR however.

Otherwise looks good to me.

cli/internal/state/state.go
msanft: This makes it quite tedious to work with in code, as you need non-nil checks everywhere. I see a value in using a pointer to the nested structs, but I wouldn't use them for all fields

daniel-weisse (approved):

PR #2454, opened 2023-10-16
cli: revert libvirt terraform provider upgrade

daniel-weisse:
### Context
https://github.com/edgelesssys/constellation/pull/2434 upgraded the libvirt terraform provider to v7.4.
That version seems to have broken support for connection URIs, making Terraform seem to completely ignore the paramter.
This means that even if the libivrt URI is set to something like `qemu+tcp://localhost:16599/system`, it will default to the system host libvirt.

We need to revert that commit for our Docker-libvirt setup to still be functional.

### Proposed change(s)

- Revert libvirt terraform provider to latest working release (v7.1)

### Additional info

- [AB#3488](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3488)
- This is was not part of the v2.12 release, so we don't need any patch releases

### DIFF
--- a/cli/internal/terraform/terraform/qemu/.terraform.lock.hcl
+++ b/cli/internal/terraform/terraform/qemu/.terraform.lock.hcl
@@ -2,37 +2,28 @@
 # Manual edits may be lost in future updates.
 
 provider "registry.terraform.io/dmacvicar/libvirt" {
-  version     = "0.7.4"
-  constraints = "0.7.4"
+  version     = "0.7.1"
+  constraints = "0.7.1"
   hashes = [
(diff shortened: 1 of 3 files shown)


elchead (approved):
lgtm. how did you see that? would an e2e test before merging uncovered that?

daniel-weisse:
> how did you see that?

miniConstellation e2e test have been failing since this was merged

> would an e2e test before merging uncovered that?

Yes, but I don't think running the e2e tests specifically for the one PR is necessary, since our daily/weekly tests catch this issue. I would also not expect every dependency update to bring in issues that inhibit usage of said dependency (which is the case here).

PR #2460, opened 2023-10-16
cli: use custom byte-slice marshalling for state file

msanft:
### Context

Fields of type `[]byte` were marshalled as YAML arrays in the state file, resulting in very large files to scroll through when expanded.

### Proposed change(s)

- Use a custom type `hexBytes` to (un)marshal to/from YAML that encodes/decodes values to hex strings.

### Additional info

- [AB#3493](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3493)

### Checklist

- [x] Link to Milestone


cli/internal/state/state.go
```diff
@@ -164,3 +165,27 @@ func (s *State) Merge(other *State) (*State, error) {
...
+
+// hexBytes is a byte slice that is marshalled to and from a hex string.
+type hexBytes []byte
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface.
+func (h *hexBytes) UnmarshalYAML(unmarshal func(any) error) error {
+	var hexString string
+	if err := unmarshal(&hexString); err != nil {
+		return err
+	}
```
daniel-weisse, line 177:
```suggestion
// UnmarshalYAML implements the yaml.Unmarshaler interface.
func (h *hexBytes) UnmarshalYAML(unmarshal func(any) error) error {
	var hexString string
	if err := unmarshal(&hexString); err != nil {
		var hexBytes []byte
		if err := unmarshal(&hexBytes); err != nil {
			return err
		}
		hexString = hex.EncodeToString(hexBytes)
	}
```
Since we already released a version of Constellation that supported a statefile, we need compatibility code here to make sure we can unmarshal from the yaml array

daniel-weisse (approved):

elchead (approved):

cli/internal/state/state_test.go
```diff
@@ -326,3 +327,80 @@ func TestMerge(t *testing.T) {
 		})
 	}
 }
+
+func TestMarshalHexBytes(t *testing.T) {
```
elchead, line 331: nit: I would test marshal and unmarshal together to verify that they are "compatible"

elchead (approved):

malt3:
Just a nit: I think the label `feature` is not fitting here. At least the PR title has no meaning for end users.

msanft:
> Just a nit: I think the label `feature` is not fitting here. At least the PR title has no meaning for end users.

Should we go for "no changelog" instead? Or what do you have in mind?

PR #2458, opened 2023-10-16
docs: add minimal docs for self-managed infrastructure

msanft:
### Context

Self-managed infrastructure needs "MVP" documentation to be recognized as "supported". This PR documents the process of self-managed infrastructure very roughly but good enough for the general case. Specific cases (e.g. MAA-usage) differ in parts of what's described here, but are too specific to put into the main documentation.

### Proposed change(s)

- Add "MVP" documentation of self-managed infrastructure in Constellation.

### Additional info

- AB#3428

### Checklist

- [x] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)
- [x] Add labels (e.g., for changelog category)
- [x] Is PR title adequate for changelog?
- [x] Link to Milestone

PR #2472, opened 2023-10-17
ci: add e2e test for self-managed infrastructure

msanft:
### Context

There should be an e2e-test of the self-managed infrastructure workflow we also describe in our [docs](https://docs.edgeless.systems/constellation/workflows/create). This is related to #2458.

### Proposed change(s)

- Add a test payload `self-managed infra` to test self-managed infrastructure creation as a substitution to `constellation create`

### Additional info

- [AB#3429](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3429)
- [E2E AWS](https://github.com/edgelesssys/constellation/actions/runs/6636476213)
- [E2E Azure](https://github.com/edgelesssys/constellation/actions/runs/6657452461)
- [E2E GCP](https://github.com/edgelesssys/constellation/actions/runs/6636477659)

### Checklist

- [x] Link to Milestone


.github/actions/constellation_create/action.yml
```diff
@@ -124,17 +124,28 @@ runs:
...
+    - name: Show Cluster Configuration
       shell: bash
       run: |
         echo "Creating cluster using config:"
         cat constellation-conf.yaml
         sudo sh -c 'echo "127.0.0.1 license.confidential.cloud" >> /etc/hosts' || true
+
+    # Uses --force flag since the CLI currently does not have a pre-release version and is always on the latest released version.
+    # However, many of our pipelines work on prerelease images. Thus the used images are newer than the CLI's version.
+    # This makes the version validation in the CLI fail.
```
daniel-weisse, line 136: Does this still hold true?
We don't need to test in this PR, but I'm wondering if we can drop the force flag for the CI by now

.github/workflows/e2e-test-manual.yml
```diff
@@ -35,6 +35,7 @@ on:
           - "verify"
           - "recover"
           - "malicious join"
+          - "self-managed infra"
```
daniel-weisse, line 38: I think it would be beneficial to not have this as a test payload, but as a separate test input (New workflow if needed because of input limitations).
Otherwise we are limited to a single test payload for self-managed infrastructure tests.

.github/actions/self_managed_create/action.yml
```diff
@@ -0,0 +1,91 @@
...
+      shell: bash
+      working-directory: ${{ github.workspace }}/e2e-infra
+      run : |
+        echo "image_ref=$(bazel run //hack/image-fetch:image-fetch)" >> $GITHUB_OUTPUT
+
+    - name: Write Terraform variables
+      shell: bash
+      working-directory: ${{ github.workspace }}/e2e-infra
+      run : |
+        echo "name = \"$(yq '.name' constellation-conf.yaml)\"" >> terraform.tfvars
```
3u13r, line 30: Thinking about using a `-bash` file for this so the we can link it (maybe even embed it) in the docs more easily and have a single source of truth. Do you think that this is possible?

hack/image-fetch/main.go
3u13r, line 1: Should we maybe use the bash command mentioned in the docs instead, so that this more closely resembles the user behavior?

3u13r (changes requested):

.github/actions/self_managed_create/action.yml
msanft: We have many places in our docs where embedding files would be really useful. I don't know if / how this could work, but it would be good. In this specific case though, its not really possible as the `{{ inputs.foo }}` references should not be in the docs, env vars would be a possible workaround

hack/image-fetch/main.go
msanft: I agree that we should mimick the expected user behaviour as close as possible here, but its really hard since we would have to write complicated bash to pick the right CSP / Attestation variant here. The Imagefetcher should fetch from the same S3, so the data should be the same at least. I dont know if the complexity is worth the tradeoff to inaccuracy here

3u13r (approved):
LGTM

.github/workflows/e2e-test-manual.yml
```diff
@@ -260,12 +260,14 @@ jobs:
...
           cosignPrivateKey: ${{ secrets.COSIGN_PRIVATE_KEY }}
           fetchMeasurements: ${{ contains(needs.find-latest-image.outputs.image, '/stream/stable/') }}
+          selfManagedInfra: "true"
 
       - name: Always terminate cluster
         if: always()
         uses: ./.github/actions/constellation_destroy
         with:
           kubeconfig: ${{ steps.e2e_test.outputs.kubeconfig }}
+          selfManagedInfra: "true"
```
daniel-weisse, line 270: Needs to be reverted/moved to its own workflow before merging

daniel-weisse (approved):

msanft:
We need to patch the MAA manually before init on Azure, which we currently only do in the CLI.
```
Oct 26 15:04:52 fedora bootstrapper[8984]: {"level":"WARN","ts":"2023-10-26T15:0
4:52Z","logger":"bootstrapper","caller":"vtpm/attestation.go:109","msg":"Failed 
to issue attestation statement: fetching instance info: creating MAA token: gett
ing encrypted token: MAA returned 400 Bad Request: {\"error\":{\"code\":\"Policy
EvaluationError\",\"message\":\"Policy Evaluation Error has occurred\",\"innerer
ror\":{\"code\":\"PolicyValidationFailure\",\"details\":[{\"code\":\"PolicyFailu
reLine\",\"target\":\"Line\",\"value\":\"[type==\\\"secureboot\\\", value==false
] => deny();\"},{\"code\":\"PolicyFailureClaim\",\"target\":\"secureboot\",\"val
ue\":\"false\"}],\"message\":\"A Deny claim was issued, authorization failed.\"}
}}"}
```

PR #2480, opened 2023-10-19
validation: add generic validation framework

msanft:
### Context

We want to have a generic framework for document validation in Constellation, for use on the config and state file.

### Proposed change(s)

- Add a framework for config validation. A usage example can be found in [`validation_test.go`](https://github.com/edgelesssys/constellation/blob/0ee82ee88c6b0d1ebb52915881c502be7034e45c/internal/validation/validation_test.go)  and in the Go doc comments of the exported functions. If more thorough documentation on this is required, I can also prepare a page in the dev-docs.
- This PR uses quite hacky reflection to implement JSONPath traces of fields that fail validation. This approach is quite prone to errors, but is verified to work for all types tested in the unit tests. In case any type that you might want to validate is not covered by this, please let me know.

### Additional info

- [AB#3496](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3496)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/validation/constraints.go
+++ b/internal/validation/constraints.go
@@ -21,13 +21,15 @@ type Constraint struct {
 /*
 WithFieldTrace adds a well-formatted trace to the field to the error message
 shown when the constraint is not satisfied. Both "doc" and "field" must be pointers:
-	- "doc" must be a pointer to the top level document
-	- "field" must be a pointer to the field to be validated
+  - "doc" must be a pointer to the top level document
+  - "field" must be a pointer to the field to be validated
 
 Example for a non-pointer field:
+
 	Equal(d.IntField, 42).WithFieldTrace(d, &d.IntField)
 
 Example for a pointer field:
+
 	NotEmpty(d.StrPtrField).WithFieldTrace(d, d.StrPtrField)
 
 Due to Go's addressability limititations regarding maps, if a map field is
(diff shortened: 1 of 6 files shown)


internal/validation/errors.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+// appendMapKey appends the key of a map to the path.
+func (p pathBuilder) appendMapKey(k string) pathBuilder {
+	p.buf = append(p.buf, fmt.Sprintf("[\"%s\"]", k))
+	return p
+}
+
+// String returns the path.
+func (p pathBuilder) String() string {
```
daniel-weisse, line 260:
```suggestion
// string returns the path.
func (p pathBuilder) string() string {
```
nit: doesn't need to be exported

internal/validation/errors_test.go
```diff
@@ -0,0 +1,470 @@
...
+		OnlyYamlKey:   "abc",
+	}
+
+	doc, field := references(t, st, &st.OnlyYamlKey, "")
+	err := newError(doc, field, assert.AnError)
+	require.Error(t, err)
+	require.Contains(t, err.Error(), fmt.Sprintf("validating ErrorTestDoc.onlyYamlKey: %s", assert.AnError))
+}
+
+type ErrorTestDoc struct {
```
daniel-weisse, line 418: nit: these testing types don't need to be exported, only their field names have to be

internal/validation/constraints.go
```diff
@@ -0,0 +1,153 @@
...
+import (
+	"fmt"
+	"reflect"
+	"regexp"
+)
+
+// Constraint is a constraint on a document or a field of a document.
+type Constraint struct {
+	// Satisfied returns true if the constraint is satisfied.
+	Satisfied func() (valid bool, err error)
```
daniel-weisse, line 18: Would there be any downsides to having `Satisifed` return just an error, instead of bool + error?

daniel-weisse (commented):
Nice.
Just a couple smaller things

PR #2462, opened 2023-10-16
Enable Cilium node-to-node strict encryption

3u13r:
### Proposed change(s)

- Deploy Cilium 1.15-pre with our custom node-to-node strict mode (see the last 2 commits on: https://github.com/edgelesssys/cilium/releases/tag/v1.15.0-pre.2-edg.1)
- Remove konnectivity
- Remove everything around `FixCilium()`. There is no need to restart the agent anymore.

### Additional info

- [AB#3316](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3316)

Minicon: https://github.com/edgelesssys/constellation/actions/runs/6864239262 (This fails due to the main nightly image still setting up konnectivity, when using the image from this branch the test works: `ref/feat-cilium-strict-node-to-node/stream/nightly/v2.13.0-pre.0.20231114141207-0d4d3c3d8386`)
e2e, azure, lb: https://github.com/edgelesssys/constellation/actions/runs/6864247842
e2e, gcp, verify: https://github.com/edgelesssys/constellation/actions/runs/6864253985
e2e, aws, autoscaling: https://github.com/edgelesssys/constellation/actions/runs/6864259968
e2e, upgrade, gcp 3:2, from v2.12.0: https://github.com/edgelesssys/constellation/actions/runs/6877667384

Again:
e2e, azure, lb: https://github.com/edgelesssys/constellation/actions/runs/6878967924
e2e, gcp, verify: https://github.com/edgelesssys/constellation/actions/runs/6878976138
e2e, aws, autoscaling: https://github.com/edgelesssys/constellation/actions/runs/6878980995
e2e, upgrade, gcp 3:2, from v2.12.0: https://github.com/edgelesssys/constellation/actions/runs/6878954935

### Checklist

- [x] Add labels (e.g., for changelog category)
- [x] Is PR title adequate for changelog?
- [x] Link to Milestone

PR #2481, opened 2023-10-19
terraform: always output node cidr

3u13r:
### Context

We need to expose the node CIDR of the Constellation so that we can use the value later (in https://github.com/edgelesssys/constellation/pull/2462) as a strict CIDR in the Cilium helm values.

### Proposed change(s)

- Always output the node CIDR and write into our state file

### Additional info

Upgrade test gcp: https://github.com/edgelesssys/constellation/actions/runs/6576667618

### Checklist

- [x] Link to Milestone


elchead (approved):
code lgtm. can you please align with @msanft to potentially rebase on the self managed infra PR such that this is documented too?

derpsteb (approved):

PR #2449, opened 2023-10-12
cli: common backend for `init` and `upgrade apply` commands

daniel-weisse:
### Context
Our CLI commands `constellation init` and `constellation upgrade apply` share a lot of code.
Additionally, we would like to eventually deprecate both commands in favor of a single `constellation apply` command.
To reduce code duplication and allow for a single command, this PR aims to provide a single backend for both commands.
This backend should take care of initializing the cluster and/or applying upgrades.

### Proposed change(s)

- Add a common `apply` backend that handles the logic for both `constellation init` and `constellation upgrade apply`
- Fix an issue with the `ExtendClusterConfigCertSANs` that would add an empty string to the clusters apiserver SANs field

### Additional info

- [AB#3481](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3481)

### TODO
- [x] Run more tests
  - [x] [MiniConstellation](https://github.com/edgelesssys/constellation/actions/runs/6531299634)
  - [x] [AWS e2e](https://github.com/edgelesssys/constellation/actions/runs/6531794719)
  - [x] [AWS upgrade](https://github.com/edgelesssys/constellation/actions/runs/6534470192)
  - [x] [Azure e2e](https://github.com/edgelesssys/constellation/actions/runs/6531790913)
  - [x] [Azure upgrade](https://github.com/edgelesssys/constellation/actions/runs/6534466647)
  - [x] [GCP e2e](https://github.com/edgelesssys/constellation/actions/runs/6543160645)
  - [x] [GCP upgrade](https://github.com/edgelesssys/constellation/actions/runs/6506950819)
- [x] Move existing unit tests to common apply backend
- [x] Remove old code
- [x] (Maybe) Move apply code into multiple files and/or create interfaces to run apply phases, e.g. initPhase, infraPhase etc.
  - Split into multiple files for now (terraform, init, helm)

### Checklist

- [x] Add labels (e.g., for changelog category)
- [x] Is PR title adequate for changelog?
- [x] Link to Milestone

PR #2484, opened 2023-10-20
cli: add `constellation apply` command to replace `init` and `upgrade apply`

daniel-weisse:
### Context

https://github.com/edgelesssys/constellation/pull/2449 replaces the backend code for `init` and `upgrade apply` with one common implementation.
This PR adds a new command to replace both `init` and `upgrade apply` with a single command.

### Proposed change(s)

- Add `constellation apply` command
- Mark `constellation init` and `constellation upgrade apply` as deprecated
- Use `constellation apply` in CI

### Additional info

- [AB#3494](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3494)
- [GCP e2e test](https://github.com/edgelesssys/constellation/actions/runs/6584992230)
- [e2e upgrade test](https://github.com/edgelesssys/constellation/actions/runs/6585561665)

### Checklist

  - See [AB#3499](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3499)
- [x] Link to Milestone

### DIFF
--- a/.github/actions/constellation_create/action.yml
+++ b/.github/actions/constellation_create/action.yml
@@ -147,7 +147,11 @@ runs:
       id: constellation-init
       shell: bash
       run: |
-        constellation init --debug
+        cmd=apply
+        if constellation --help | grep -q init; then
+          cmd=init
+        fi
+        constellation $cmd --force --debug
         echo "KUBECONFIG=$(pwd)/constellation-admin.conf" | tee -a $GITHUB_OUTPUT
 
     - name: Wait for nodes to join and become ready
--- a/.github/actions/constellation_create/action.yml
+++ b/.github/actions/constellation_create/action.yml
@@ -147,11 +147,12 @@ runs:
       id: constellation-init
       shell: bash
       run: |
+        # TODO(v2.14): Remove workaround for CLIs not supporting apply command
         cmd=apply
         if constellation --help | grep -q init; then
           cmd=init
         fi
-        constellation $cmd --force --debug
+        constellation $cmd --debug
         echo "KUBECONFIG=$(pwd)/constellation-admin.conf" | tee -a $GITHUB_OUTPUT
 
     - name: Wait for nodes to join and become ready
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -49,10 +49,9 @@ func NewRootCmd() *cobra.Command {
 
 	must(rootCmd.MarkPersistentFlagDirname("workspace"))
 
-	rootCmd.AddCommand(cmd.NewApplyCmd())
 	rootCmd.AddCommand(cmd.NewConfigCmd())
 	rootCmd.AddCommand(cmd.NewCreateCmd())
-	rootCmd.AddCommand(cmd.NewInitCmd())
+	rootCmd.AddCommand(cmd.NewApplyCmd())
 	rootCmd.AddCommand(cmd.NewMiniCmd())
 	rootCmd.AddCommand(cmd.NewStatusCmd())
 	rootCmd.AddCommand(cmd.NewVerifyCmd())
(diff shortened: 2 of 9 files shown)


cli/internal/cmd/apply.go
```diff
@@ -40,6 +40,32 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
+// NewApplyCmd creates the apply command.
+func NewApplyCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "apply",
+		Short: "Apply a configuration to a Constellation cluster",
+		Long:  "Apply an upgrade to a Constellation cluster by applying the chosen configuration.",
```
3u13r, line 48: I think that `Apply an upgrade` might be a bit misleading. Maybe somehting like `Apply the current configuration. This can be used to initialize or upgrade the Constellation.`

3u13r (changes requested):

elchead:
should we also apply the apply pattern to `iam` to be consistent again?

.github/actions/constellation_create/action.yml
```diff
@@ -141,7 +141,11 @@ runs:
       id: constellation-init
       shell: bash
       run: |
-        constellation init --force --debug
+        cmd=apply
+        if constellation --help | grep -q init; then
+          cmd=init
+        fi
+        constellation $cmd --force --debug
```
elchead, line 148: add a TODO to remove in 2.14?

cli/internal/cmd/apply.go
```diff
@@ -40,6 +40,32 @@ import (
...
+		Args:  cobra.NoArgs,
+		RunE:  runApply,
+	}
+
+	cmd.Flags().Bool("conformance", false, "enable conformance mode")
+	cmd.Flags().Bool("skip-helm-wait", false, "install helm charts without waiting for deployments to be ready")
+	cmd.Flags().Bool("merge-kubeconfig", false, "merge Constellation kubeconfig file with default kubeconfig file in $HOME/.kube/config")
```
elchead, line 55: should we mark flags that are only relevant for init?

cli/cmd/root.go
```diff
@@ -49,6 +49,7 @@ func NewRootCmd() *cobra.Command {
 
 	must(rootCmd.MarkPersistentFlagDirname("workspace"))
 
+	rootCmd.AddCommand(cmd.NewApplyCmd())
 	rootCmd.AddCommand(cmd.NewConfigCmd())
 	rootCmd.AddCommand(cmd.NewCreateCmd())
 	rootCmd.AddCommand(cmd.NewInitCmd())
```
thomasten, line 55: I'd move apply below create and move deprecated init to the end

cli/internal/cmd/apply.go
```diff
@@ -40,6 +40,82 @@ import (
...
+func (s *skipPhases) add(phases ...skipPhase) {
+	if *s == nil {
+		*s = make(skipPhases)
+	}
+	for _, phase := range phases {
+		(*s)[skipPhase(strings.ToLower(string(phase)))] = struct{}{}
+	}
+}
+
+func printAllPhases() string {
```
thomasten, line 82: Maybe move this func right below the consts so that if someone changes this in the future, it's more likely that they change both places. Alternatively, you could add comments to keep these in sync.

thomasten, line 82: nit/optional: to me, the name implies that the func prints to stdout or the like. Maybe there's a better name.

cli/internal/cmd/apply.go
daniel-weisse: I'm not sure how I feel about this, because the goal is to remove the distinction between `upgrade apply` and `init`.
Ideally, we would instead update the `apply` command to offer the functionality of the `init` specific flags also when running it to just upgrade the cluster.

derpsteb: I also think that eventually all flags that were originally init-only should become modifiable through upgrade. Unless there is a specific reason why this is dangerous. But in the meantime it seems like a nice idea to me to have a note like `applies to phase X`.

derpsteb (approved):

cli/internal/cmd/apply.go
elchead: i went over the flags again and don't see any critical problem by not having the hint. is fine by me.

elchead (approved):

thomasten (approved):

3u13r (approved):
LGTM

PR #2489, opened 2023-10-20
cli: remove unnecessary check from QEMU rollbacker

daniel-weisse:
### Context
Terraform does not care if `destroy` is called on an empty directory.
In that case it simply returns `nil`, so we can skip the extra check during `cloudcmd.Create`.

### Proposed change(s)

- Remove unnecessary rollback destroy guards form QEMU rollbacker

### Additional info

- Setup work for [AB#3288](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3288)


elchead (approved):
lgtm. verified that `terraform destroy` in an empty dir returns 0.

PR #2501, opened 2023-10-23
cli: refactor terraform code to be update/create agnostic

daniel-weisse:
### Context
Our Terraform code currently only differs between upgrades and new creations when setting up the workspace.
To allow creating or upgrading a workspace using the same code, this PR aims to fix this difference by moving relevant code out of the Terraform package.

### Proposed change(s)

- Move backup creation of Terraform workspace from `terraform` package to `cloudcmd`
- Always allow overwriting files when setting up the Terraform workspace
  - In case of upgrades, this is the behavior we want regardless
  - In case of new creations, we ensure correct behavior by first checking that no Terraform state exists in the workspace during `constellation create`
- Use a constant to set the path of embedded Terraform files

### Additional info

- Preparation for [AB#3383](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3383)

### DIFF
--- a/cli/internal/cloudcmd/tfupgrade.go
+++ b/cli/internal/cloudcmd/tfupgrade.go
@@ -21,14 +21,14 @@ import (
 func planUpgrade(
 	ctx context.Context, tfClient tfUpgradePlanner, fileHandler file.Handler,
 	outWriter io.Writer, logLevel terraform.LogLevel, vars terraform.Variables,
-	templateDir, backupDir string,
+	templateDir, existingWorkspace, backupDir string,
 ) (bool, error) {
 	if err := ensureFileNotExist(fileHandler, backupDir); err != nil {
 		return false, fmt.Errorf("backup directory %s already exists: %w", backupDir, err)
 	}
 
 	// Backup old workspace
-	if err := fileHandler.CopyDir(templateDir, backupDir); err != nil {
+	if err := fileHandler.CopyDir(existingWorkspace, backupDir); err != nil {
 		return false, fmt.Errorf("backing up old workspace: %w", err)
 	}
 
--- a/cli/internal/cloudcmd/tfupgrade.go
+++ b/cli/internal/cloudcmd/tfupgrade.go
@@ -27,13 +27,13 @@ func planUpgrade(
 		return false, fmt.Errorf("backup directory %s already exists: %w", backupDir, err)
 	}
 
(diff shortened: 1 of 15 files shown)


cli/internal/terraform/terraform.go
```diff
@@ -481,26 +467,19 @@ func (c *Client) applyManualStateMigrations(ctx context.Context) error {
 }
 
 // writeVars tries to write the Terraform variables file or, if it exists, checks if it is the same as we are expecting.
```
elchead, line 469:
```suggestion
// writeVars writes / overwrites the Terraform variables file.
```

cli/internal/cloudcmd/tfupgrade_test.go
```diff
@@ -19,58 +19,61 @@ import (
 )
 
 func TestPlanUpgrade(t *testing.T) {
+	defaultFs := func(require *require.Assertions) file.Handler {
+		fs := file.NewHandler(afero.NewMemMapFs())
+		require.NoError(fs.MkdirAll("test"))
```
elchead, line 24: nit: extract "test" as variable templateDir or simply name it so, to help indicate that this is the folder where the TF templates are expected. Then rename `defaultFs` to `fsWithTemplateDir`

cli/internal/cloudcmd/tfupgrade.go
```diff
@@ -27,13 +27,13 @@ func planUpgrade(
...
 
-	// Backup the old Terraform workspace and move the embedded Terraform files into the workspace.
-	err := tfClient.PrepareUpgradeWorkspace(
-		templateDir,
-		backupDir,
-		vars,
-	)
-	if err != nil {
+	// Backup old workspace
+	if err := fileHandler.CopyDir(templateDir, backupDir); err != nil {
```
elchead, line 31: we need to copy the workspace dir (template dir are the newly embedded files) to the backup dir.

elchead: please add a test case that checks that the files from the working dir are copied to the backup and that new embedded files are copied to the working dir

elchead:
good simplification again 👍 after our discussion toward the terraform provider i was wondering how much we should invest into unifying the terraform logic when we soon want to get rid of it.

elchead (approved):

PR #2503, opened 2023-10-23
terraform: Terraform module for AWS

elchead:
### Context

This is the first implementation to manage constellation through Terraform using a module. It internally still relies on the CLI to call `constellation apply`.
However, the CLI is automatically fetched (and yq), so it's hidden from the user.

### Proposed change(s)

- create a terraform-module for aws (shipped as zipped artifact)
- add an e2e test for the setup and run during weekly tests

### Additional info

- [AB#3502](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3502)
- Successful [E2E run](https://github.com/edgelesssys/constellation/actions/runs/6704150039/job/18216009331) older [here](https://github.com/edgelesssys/constellation/actions/runs/6692137659) 
- Terraform changes will be rebased once reviewed. Since the files moved to a new location any Terraform change will cause conflicts right now.
- [x] should we fetch `yq` like the constellation CLI in the background? It was found that `yq` has incompatible changes in different minor releases. E.g. one version required `-` to be present to process `stdout` input and another didn't. I'm not aware of any incompatibility issues in the current code, but I think it's safer to ensure one yq version.
- [x] consider publication to module registry? https://developer.hashicorp.com/terraform/registry/modules/publish~ -> not until asked for
- [x] How should we solve the reference to the constellation CLI until the release?
- [x] Is it ok to use `yq`? and potentially install as dependency too?
You need to perform a bazel devbuild inside `aws-constellation` such that the binary is available.

#### Usage
1. Download the `terraform-module` artifact.
2. `cd terraform-module/aws-constellation`
3. Create a `terraform.tfvars` file with the input (see E2E test)
4. (Optional): If you want to use a custom constellation CLI, you can place the binary inside the `terraform-module/aws-constellation`. Otherwise the latest CLI release is fetched during the execution.
5. Run terraform apply:
```
terraform apply -var-file=terraform.tfvars  -auto-approve
...
module.constellation.terraform_data.apply: Still creating... [15m40s elapsed]
module.constellation.terraform_data.apply (local-exec): Your Constellation cluster was successfully initialized.

module.constellation.terraform_data.apply (local-exec): Constellation cluster identifier  ff9a90dff66b2d4fe6043593bdef6abe28b26cd9d73b30febf709c1820f818e6
module.constellation.terraform_data.apply (local-exec): Kubernetes configuration          constellation-admin.conf

module.constellation.terraform_data.apply (local-exec): You can now connect to your cluster by executing:
module.constellation.terraform_data.apply (local-exec): 	export KUBECONFIG="/Users/adria/work/constellation/cli/internal/terraform/terraform/constellation-admin.conf"

module.constellation.terraform_data.apply: Creation complete after 15m49s [id=7451e54c-6620-1f58-29b6-f88472b1109f]

Apply complete! Resources: 3 added, 0 changed, 1 destroyed.
21.35s user 4.51s system 2% cpu 15:55.62s total
````

### Checklist

- [x] Link to Milestone

### CODE REVIEW

**Review by 3u13r — CHANGES_REQUESTED** [review]

File: cli/internal/terraform/terraform/constellation-cluster/main.tf [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+
+  depends_on = [
+    terraform_data.csp_config
+  ]
+}
+
+
+resource "terraform_data" "apply" {
+  provisioner "local-exec" {
+    command = "./constellation init --force" # TODO use apply --yes --skip-phases infrastructure
```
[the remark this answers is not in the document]
elchead: no reason. was just handy during testing. This should be replaced by apply as annotated before merging at any rate.

cli/internal/terraform/terraform/constellation-cluster/install-constellation.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+if ! command -v constellation &> /dev/null; then
+  OS=$(uname -s)
+  ARCH=$(uname -m)
+
+  URL=""
+
+  if [[ $OS == "Darwin" ]]; then
+    if [[ $ARCH == "arm64" ]]; then
+      URL="https://github.com/edgelesssys/constellation/releases/latest/download/constellation-darwin-arm64"
```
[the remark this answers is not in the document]
elchead: you have an idea how to mention the TODO question from above: about how to make this work before we make the next release? (is it good enough to run bazel devbuild in the install script?)

IMO it doesn't really matter since we will publish the module with the release and hence they should coincide. Would just complicate versioning and artifact bundling further.

msanft:
IMO, as we also discussed yesterday, we shouldn't place the module in the CLI directory. Here is my idea:

Move all of our Terraform configurations to a top-level directory called `terraform`. In that directory, place an `assets.go`, similar to what we use in the [debugd](https://github.com/edgelesssys/constellation/blob/feat/cli/state-file-validation/debugd/logstash/assets.go). This file should just embed all the Terraform configurations that we currently embed in the [CLI  Terraform code](https://github.com/edgelesssys/constellation/blob/feat/cli/state-file-validation/cli/internal/terraform/loader.go#L29) directly. Then, in the CLI, we can import this assets-file. What do you think?

cli/internal/terraform/terraform/aws-constellation/main.tf
msanft, line 1: Should we provide the relevant outputs of the Cluster (i.e. values in admin-conf and values set in the state-file during initialization) as Terraform outputs? Just an idea.

cli/internal/terraform/terraform/aws-constellation/variables.tf
```diff
@@ -0,0 +1,110 @@
+variable "name" {
+  type        = string
+  description = "Name of your Constellation"
+  validation {
+    condition     = length(var.name) <= 10
+    error_message = "The length of the name of the Constellation must be <= 10 characters"
+  }
```
msanft, line 7: Generally, having validation is always good, but I think it is sufficient to implement that kind of validation in the CLI here and have a single source of truth for what is valid. Failing earlier is good, but this will require us to update 2 places if we update our validation constraints.

cli/internal/terraform/terraform/constellation-cluster/install-constellation.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+VERSION="latest"
+if ! command -v constellation &> /dev/null; then
+  OS=$(uname -s)
+  ARCH=$(uname -m)
+
+  URL=""
```
msanft, line 7:
```suggestion
if ! command -v constellation &> /dev/null; then
  OS=$(uname -s)
  ARCH=$(uname -m)
  VERSION="latest"
  URL=""
```
nit

cli/internal/terraform/terraform/constellation-cluster/install-constellation.sh
```diff
@@ -0,0 +1,30 @@
...
+  elif [[ $OS == "Linux" ]]; then
+    if [[ $ARCH == "x86_64" ]]; then
+      URL="https://github.com/edgelesssys/constellation/releases/${VERSION}/download/constellation-linux-amd64"
+    elif [[ $ARCH == "arm64" ]]; then
+      URL="https://github.com/edgelesssys/constellation/releases/${VERSION}/download/constellation-linux-arm64"
+    fi
+  fi
+
+  if [ -z "$URL" ]; then
+    echo "Unsupported OS or architecture"
```
msanft, line 24:
```suggestion
    echo "OS \"$OS\" and/or architecture \"$ARCH\" is not supported."
```
nit

cli/internal/terraform/terraform/constellation-cluster/install-constellation.sh
```diff
@@ -0,0 +1,30 @@
...
+    exit 1
+  else
+    curl -o constellation -LO $URL
+    sudo install constellation /usr/local/bin/constellation
```
msanft, line 28: We should rather use the local file here. Installing the file to the user's machine might be an unwanted / unexpected side-effect of applying our module.

cli/internal/terraform/terraform/constellation-cluster/main.tf
```diff
@@ -0,0 +1,89 @@
...
+  // generate config here to only create it once (csp won't change)
+}
+# TODO ensure yq dependency
+
+# TODO how to handle CSP specific config without CSP specific input?
+resource "terraform_data" "csp_config" {
+  count = var.csp == "aws" ? 1 : 0
+  provisioner "local-exec" {
+    command = <<EOT
+      if [ "${var.csp}" == "aws" ]; then
```
msanft, line 28: doesn't the "count" check above already implement this "if"?

msanft (dismissed):
IMO, as we also discussed yesterday, we shouldn't place the module in the CLI directory. Here is my idea:

Move all of our Terraform configurations to a top-level directory called `terraform`. In that directory, place an `assets.go`, similar to what we use in the [debugd](https://github.com/edgelesssys/constellation/blob/feat/cli/state-file-validation/debugd/logstash/assets.go). This file should just embed all the Terraform configurations that we currently embed in the [CLI  Terraform code](https://github.com/edgelesssys/constellation/blob/feat/cli/state-file-validation/cli/internal/terraform/loader.go#L29) directly. Then, in the CLI, we can import this assets-file. What do you think?

cli/internal/terraform/terraform/aws-constellation/main.tf
elchead: i think here it would be good to have business feedback. would add it if people find that useful

cli/internal/terraform/terraform/constellation-cluster/main.tf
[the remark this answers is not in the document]
elchead: EDIT: it is needed for devbuilds of the CLI (since currently only release images are supported).

msanft:
> should we fetch yq like the constellation CLI in the background? It was found that yq has incompatible changes in different minor releases. E.g. one version required - to be present to process stdout input and another didn't. I'm not aware of any incompatibility issues in the current code, but I think it's safer to ensure one yq version.

This is a very valid point. We should grab a working version for the used architecture from their [release page](https://github.com/mikefarah/yq/releases/tag/v4.35.2)

.github/workflows/e2e-test-tf-module.yml
```diff
@@ -0,0 +1,193 @@
...
+        run: |
+          cd terraform-module/aws-constellation
+          cp ../../build/constellation .
+          terraform init
+          terraform apply -var-file=terraform.tfvars  -auto-approve
+
+      - name: Destroy Terraform Cluster
+        if: always() && (steps.apply_terraform.outcome == 'failure' || steps.apply_terraform.outcome == 'success')
+        run: |
+          cd terraform-module/${{ inputs.cloudProvider }}-constellation
```
msanft, line 183:
```suggestion
        working-directory: ./terraform-module/${{ inputs.cloudProvider }}-constellation
        run: |
```

.github/workflows/e2e-test-tf-module.yml
```diff
@@ -0,0 +1,193 @@
...
+          terraform init
+          terraform destroy -var-file=terraform.tfvars -auto-approve
+
+      - name: Verify cleanup
+        run: |
+          cd terraform-module/${{ inputs.cloudProvider }}-constellation
```
msanft, line 189:
```suggestion
        working-directory: ./terraform-module/${{ inputs.cloudProvider }}-constellation
        run: |
```

.github/workflows/e2e-test-tf-module.yml
```diff
@@ -0,0 +1,193 @@
...
+        default: "eu-west-1" # TODO switch to us-east-2?
+      zone:
+        description: "Which zone to use."
+        type: string
+        default: "eu-west-1b" # TODO "us-east-2c"?
+      image:
+        description: "Node image version of the cluster."
+        type: string
+        default: "v2.12.0"
+        required: true
```
msanft, line 67: We should use an image reference here, similar to the other e2e tests. With that, we can ensure image consistency between the runs in an, e.g., weekly test.

.github/workflows/e2e-test-weekly.yml
```diff
@@ -343,3 +343,18 @@ jobs:
...
+      id-token: write
+      contents: read
+      packages: write
+    secrets: inherit
+    uses: ./.github/workflows/e2e-test-tf-module.yml
+    with:
+      cloudProvider: "aws"
+      region: "eu-west-1"
+      zone: "eu-west-1b"
+      name: "e2eTfAws"
```
msanft, line 359:
```suggestion
      name: "e2eTfAws"
```
Can we parametrize this to a workflow run, e.g. by incorporating the run ID in the naming, similar to what we do with Azure resource groups? Otherwise, runs failing to terminate the resource may cause an avalanche of other failing runs due to naming collisions.

terraform/aws-constellation/bootstrapper
msanft, line 1: This landed here by accident, I assume? Including the other files laying around here.

terraform/terraform/azure/main.tf
```diff
@@ -28,6 +28,7 @@ locals {
   }
   ports_node_range      = "30000-32767"
   cidr_vpc_subnet_nodes = "192.168.178.0/24"
+  cidr_vpc_subnet_pods  = "10.10.0.0/16"
```
msanft, line 31: AFAIK, this is a GCP-only input?

terraform/constellation-cluster/variables.tf
```diff
@@ -0,0 +1,89 @@
...
+    error_message = "The role has to be 'control-plane' or 'worker'."
+  }
+}
+
+variable "name" {
+  type        = string
+  description = "Name used in the cluster's named resources / cluster name"
+}
+
+# STATE related
```
msanft, line 31:
```suggestion
```
We should remove these comments here or reword them, as this file will be customer-facing

terraform/constellation-cluster/variables.tf
```diff
@@ -0,0 +1,89 @@
...
+}
+
+variable "apiServerCertSANs" {
+  type        = list(string)
+  description = "List of additional SANs for the Kubernetes API server certificate"
+}
+
+
+# CONFIG related
+
```
msanft, line 64:
```suggestion
```

terraform/constellation-cluster/install-constellation.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
msanft, line 1:
```suggestion
#!/usr/bin/env bash
```

more portable :) Different Unixes put bash in different locations

terraform/aws-constellation/variables.tf
```diff
@@ -0,0 +1,76 @@
...
+
+variable "zone" {
+  type        = string
+  description = "The AWS availability zone name to create the cluster in"
+}
+
+variable "debug" {
+  type        = bool
+  default     = false
+  description = "Enable debug mode. This opens up a debugd port that can be used to deploy a custom bootstrapper."
```
msanft, line 52:
```suggestion
  description = "DON'T USE IN PRODUCTION: Enable debug mode and allow the use of debug images."
```

terraform/aws-constellation/variables.tf
```diff
@@ -0,0 +1,76 @@
+variable "name" {
+  type        = string
+  description = "Name of your Constellation"
+}
+
+variable "image" {
+  type        = string
+  description = "Node image version. Needs to be a release version."
+} # is required resolve the AMI
```
msanft, line 9: As stated above, we should allow to pass an image ref in the same format as the config uses here for compatibility with the current CI and to align with the config inputs. We can use the hack-tool `image-fetch` to get the actual CSP image ref from it.

terraform/aws-constellation/variables.tf
```diff
@@ -0,0 +1,76 @@
...
+variable "debug" {
+  type        = bool
+  default     = false
+  description = "Enable debug mode. This opens up a debugd port that can be used to deploy a custom bootstrapper."
+} # TODO: debug is currently not supported (due to limited ami resolve logic). Remove this variable?
+
+variable "enable_snp" {
+  type        = bool
+  default     = true
+  description = "Enable AMD SEV SNP. Setting this to true sets the cpu-option AmdSevSnp to enable."
```
msanft, line 58:
```suggestion
  description = "Enable AMD SEV-SNP."
```
nit: Second part of the sentence doesn't provide much more information to the user imo.

terraform/aws-constellation/variables.tf
```diff
@@ -0,0 +1,76 @@
...
+}
+
+variable "custom_endpoint" {
+  type        = string
+  default     = ""
+  description = "Custom endpoint to use for the Kubernetes apiserver. If not set, the default endpoint will be used."
```
msanft, line 64:
```suggestion
  description = "Custom endpoint (DNS Name) to use for the Constellation API server. If not set, the default endpoint will be used."
```

terraform/aws-constellation/variables.tf
```diff
@@ -0,0 +1,76 @@
+variable "name" {
+  type        = string
+  description = "Name of your Constellation"
```
msanft, line 3:
```suggestion
  description = "Name of the Constellation cluster."
```
nit

terraform/constellation-cluster/main.tf
```diff
@@ -0,0 +1,110 @@
...
+
+
+resource "null_resource" "config" {
+  provisioner "local-exec" {
+    command = <<EOT
+      if [ "${var.csp}" = "aws" ]; then
+      yq eval '.provider.aws.region = "${var.aws_config.region}"' -i constellation-conf.yaml
+      yq eval '.provider.aws.zone = "${var.aws_config.zone}"' -i constellation-conf.yaml
+      yq eval '.provider.aws.iamProfileControlPlane = "${var.aws_config.iam_instance_profile_control_plane}"' -i constellation-conf.yaml
+      yq eval '.provider.aws.iamProfileWorkerNodes = "${var.aws_config.iam_instance_profile_worker_nodes}"' -i constellation-conf.yaml
```
msanft, line 45: We could indent the blocks here for improved readability.

terraform/constellation-cluster/main.tf
```diff
@@ -0,0 +1,110 @@
...
+resource "null_resource" "apply" {
+  provisioner "local-exec" {
+    command = "./constellation apply --debug --force --yes --skip-phases infrastructure"
+  } // need force to allow release images while using the dev build
+  // TODO use --debug flag?
+
+  provisioner "local-exec" {
+    when    = destroy
+    command = "./constellation terminate --yes && rm constellation-conf.yaml constellation-mastersecret.json ./fetch-ami/ami.txt && rm -r constellation-upgrade"
+  } // TODO cleanup all files?
```
msanft, line 102: Good as it is right now, imo.

terraform/constellation-cluster/main.tf
```diff
@@ -0,0 +1,110 @@
...
+    always_run = timestamp()
+  }
+}
+
+
+resource "null_resource" "apply" {
+  provisioner "local-exec" {
+    command = "./constellation apply --debug --force --yes --skip-phases infrastructure"
+  } // need force to allow release images while using the dev build
+  // TODO use --debug flag?
```
msanft, line 97: Are the logs of these operations exposed to the user? (or more specifically, can they be not shown to the user explicitly?) If yes, we should enable the debug flag here for sure.

terraform/constellation-cluster/variables.tf
```diff
@@ -0,0 +1,89 @@
+variable "csp" {
+  type        = string
+  description = "The CSP to create the cluster in"
+  validation {
+    condition     = var.csp == "aws" || var.csp == "azure" || var.csp == "gcp"
+    error_message = "The CSP must be 'aws', 'azure' or 'gcp'."
```
msanft, line 6:
```suggestion
    condition     = var.csp == "aws" 
    error_message = "The CSP must be 'aws'."
```

Add others once implemented to prevent confusion.

terraform/constellation-cluster/variables.tf
```diff
@@ -0,0 +1,89 @@
...
+}
+
+variable "clusterEndpoint" {
+  type        = string
+  description = "Endpoint of the cluster"
+}
+
+variable "inClusterEndpoint" {
+  type        = string
+  description = "The endpoint the cluster uses to reach itself. This might differ from the ClusterEndpoint in case e.g. an internal load balancer is used."
```
msanft, line 44: nit: We should use a trailing dot in all description texts or in none

terraform/constellation-cluster/variables.tf
```diff
@@ -0,0 +1,89 @@
...
+}
+
+variable "ipCidrNode" {
+  type        = string
+  description = "Node IP CIDR"
+}
+
+variable "apiServerCertSANs" {
+  type        = list(string)
+  description = "List of additional SANs for the Kubernetes API server certificate"
```
msanft, line 59:
```suggestion
  description = "List of additional SANs (Subject Alternative Names) for the Kubernetes API server certificate"
```
nit

msanft (changes requested):

.github/actions/upload_terraform_module/action.yml
```diff
@@ -0,0 +1,23 @@
+name: Upload Terraform module
+description: "Upload the Terraform module as an artifact."
+runs:
+  using: "composite"
+  steps:
+    - name: Copy Terraform module
+      shell: bash
+      run: |
+        cp -r terraform terraform-module
+        rm terraform-module/assets.go terraform-module/BUILD.bazel
```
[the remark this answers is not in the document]
elchead: should we upload the modules separately?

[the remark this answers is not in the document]
msanft: No strong opinion on this, but it was my initial idea. Let's leave it like that for now and wait for other opinions

terraform/constellation-cluster/main.tf
msanft: I rather meant logging to whatever stream is used for the [TF_LOG logging](https://developer.hashicorp.com/terraform/internals/debugging). This doesn't seem possible on module-level though

elchead: so should we keep --debug or remove it?

.github/workflows/e2e-test-weekly.yml
elchead: done, i had to shorten the logic though due to the 10 char limit on AWS.

cli/internal/terraform/terraform/aws-constellation/main.tf
msanft: conclusion: maybe later on, but no req for mvp.

terraform/constellation-cluster/main.tf
msanft: better keep it for now than have no logs at all.

msanft (changes requested):
manual testing works fine, however we should test again once the req. changes are addressed

.github/workflows/e2e-test-tf-module.yml
```diff
@@ -0,0 +1,184 @@
...
+        if: inputs.cloudProvider == 'aws'
+        id: apply_terraform
+        working-directory: ./terraform-module/${{ inputs.cloudProvider }}-constellation
+        run: |
+          cp ../../build/constellation .
+          terraform init
+          terraform apply -var-file=terraform.tfvars  -auto-approve
+
+      - name: Destroy Terraform Cluster
+        if: always() && (steps.apply_terraform.outcome == 'failure' || steps.apply_terraform.outcome == 'success')
```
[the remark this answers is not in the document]
elchead: outcome is part of the steps context: https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context. 
This ensures that destroy is always run when the apply step was exectued

3u13r (changes requested):

.github/workflows/e2e-test-tf-module.yml
```diff
@@ -0,0 +1,206 @@
...
+        type: choice
+        options:
+          - "us-east-2c"
+        default: "us-east-2c"
+        required: true
+      image:
+        description: "Node image version of the cluster."
+        type: string
+        default: "v2.12.0"
+        required: true
```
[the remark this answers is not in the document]
elchead: the weekly test injects the latest version. I remove the default, since the field is required at any rate.

terraform/aws-constellation/fetch-ami/main.tf
```diff
@@ -0,0 +1,28 @@
+locals {
+
+  image_ref         = startswith(var.image, "v") ? "ref/-/stream/stable/${var.image}" : var.image
```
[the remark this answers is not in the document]
elchead: If we want to support image input like "v2.12.0" this is needed. We support image refs too (var.image can be a ref too).
This is not valid:
curl -s https://cdn.confidential.cloud/constellation/v2/v2.12.0/image/info.json

terraform/aws-constellation/install-yq.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+VERSION="v4.35.2"
+if [ -f ./yq ] && ./yq --version | grep -q "${VERSION}"; then
```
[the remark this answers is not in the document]
elchead: this is however not POSIX compliant, meaning this syntax can cause problems

.github/workflows/e2e-test-tf-module.yml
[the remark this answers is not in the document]
msanft: Can we add a comment here? Can be just a link to the docs.

msanft:
Looking good from a first glance! Will give this one more thorough review on monday, then we should be gtm.

terraform/constellation-cluster/install-constellation.sh
```diff
@@ -0,0 +1,33 @@
...
+if [[ -f ./constellation ]]; then
+  echo "constellation CLI is already available."
+  exit 0
+fi
+
+echo "Fetching constellation ${VERSION}"
+OS=$(uname -s)
+ARCH=$(uname -m)
+VERSION="latest"
+URL=""
```
thomasten, line 11: VERSION isn't set when echoed.

Having VERSION as a var here indicates that one could replace `latest` with a version, but that won't work because the URLs are constructed differently.

thomasten, line 11: by convention, variable scoped to your script should be lowercase

terraform/constellation-cluster/install-constellation.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+if [[ -f ./constellation ]]; then
+  echo "constellation CLI is already available."
+  exit 0
+fi
+
+echo "Fetching constellation ${VERSION}"
+OS=$(uname -s)
+ARCH=$(uname -m)
+VERSION="latest"
```
thomasten, line 10: I'd like to discuss what version should be downloaded [again](https://github.com/edgelesssys/constellation/pull/2503#discussion_r1370296925). Is it possible to pass a version value from Terraform to this script?

If we go with latest and one downloads this TF module, it could break if we release a new version that requires tf changes, right?
It also wouldn't be possible to stay with an older release.

elchead: The script would break, if you download the module, don't use it and meanwhile another CLI is released. I will expose the cli version as a TF variable.

elchead: So do you think it's necessary to inject the CLI version during the release process?

.github/workflows/e2e-test-tf-module.yml
```diff
@@ -0,0 +1,191 @@
...
+        with:
+          outputPath: "constellation"
+          enterpriseCLI: true
+          push: true
+
+      - name: Download CLI
+        if: inputs.cliVersion != ''
+        shell: bash
+        run: |
+          mkdir build
```
msanft, line 151: what for?

.github/workflows/e2e-test-tf-module.yml
```diff
@@ -0,0 +1,191 @@
...
+          curl -fsSL -o constellation https://github.com/edgelesssys/constellation/releases/download/${{ inputs.cliVersion }}/constellation-linux-amd64
+          chmod u+x constellation
+          PATH_ADD=$(pwd)
+          export PATH="$PATH:$PATH_ADD"
+          constellation version
```
msanft, line 156:
```suggestion
          chmod u+x ./constellation
          PATH_ADD=$(pwd)
          export PATH="$PATH:$PATH_ADD"
          ./constellation version
```
ensure local file is used in case of collisions with an already-installed CLI

internal/constants/constants.go
```diff
@@ -172,7 +172,7 @@ const (
 	// TerraformIAMUpgradeBackupDir is the directory name being used to backup the pre-upgrade state of iam in an upgrade.
 	TerraformIAMUpgradeBackupDir = "terraform-iam-backup"
 	// TerraformEmbeddedDir is the name of the base directory embedded in the CLI binary containing the Terraform files.
-	TerraformEmbeddedDir = "terraform"
+	TerraformEmbeddedDir = "infrastructure"
```
msanft, line 175: why change this?

terraform/constellation-cluster/variables.tf
```diff
@@ -0,0 +1,86 @@
...
+variable "node_groups" {
+  type = map(object({
+    role          = string
+    initial_count = optional(number)
+    instance_type = string
+    disk_size     = number
+    disk_type     = string
+    zone          = string
+  }))
+  description = "A map of node group names to node group configurations"
```
msanft, line 19: nit: either use trailing dots everywhere or none at all.

terraform/constellation-cluster/variables.tf
```diff
@@ -0,0 +1,86 @@
...
+
+variable "kubernetes_version" {
+  type        = string
+  description = "Kubernetes version"
+}
+
+variable "microservice_version" {
+  type        = string
+  description = "Microservice version"
+}
```
msanft, line 86: nit: either use trailing dots everywhere or none at all.

msanft (commented):
LGTM apart from the comments I made. Still, I feel unsure about approving this alone. I would like if @3u13r or someone else could have another look

.github/workflows/e2e-test-tf-module.yml
elchead: no longer used, removing it.

internal/constants/constants.go
elchead: because @3u13r wanted it :) `terraform/terraform` is a confusing naming to refer to the relevant infra files for the CLI

cli/internal/terraform/terraform/constellation-cluster/install-constellation.sh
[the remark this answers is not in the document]
elchead: this was decided to be fixed to a specific version, closing.

terraform/aws-constellation/main.tf
```diff
@@ -0,0 +1,67 @@
...
+    EOT
+  }
+  triggers = {
+    always_run = timestamp()
+  }
+}
+
+module "fetch_ami" {
+  source              = "./fetch-ami"
+  attestation_variant = "aws-sev-snp"
```
3u13r, line 25: Doesn't this depend on `var.enable_snp` ?

terraform/aws-constellation/variables.tf
```diff
@@ -0,0 +1,71 @@
+variable "name" {
+  type        = string
+  description = "Name of the Constellation cluster."
+}
+
+variable "image" {
+  type        = string
+  description = "Node image reference or semantical release version."
+} # is required resolve the AMI
```
3u13r, line 9: I think we can also put this comment in the description. Doesn't hurt to be verbose here.

terraform/constellation-cluster/main.tf
```diff
@@ -0,0 +1,109 @@
...
+
+
+resource "null_resource" "apply" {
+  provisioner "local-exec" {
+    command = "./constellation apply --debug --yes --skip-phases infrastructure"
+  }
+
+  provisioner "local-exec" {
+    when    = destroy
+    command = "./constellation terminate --yes && rm constellation-conf.yaml constellation-mastersecret.json ./fetch-ami/ami.txt && rm -r constellation-upgrade"
```
3u13r, line 100: I think that we should write down the expected lifecycle of a Constellation created with Terraform. I worry about removing the mastersecret here. This is ok for now but we should create a follow-up ticket for this.

3u13r (changes requested):

terraform/constellation-cluster/main.tf
elchead: I agree, also had a brief exchange with @daniel-weisse on `constellation-upgrade` dir. Creating a ticket

.github/actions/upload_terraform_module/action.yml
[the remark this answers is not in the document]
elchead: I added rules to delete all *.go and all *.bazel. Deleting all but .tf is too dangerous, since there are different file formats used: lock files, bash scripts, xsl file and who knows whatelse in the future.

msanft (approved):

3u13r (approved):
LGTM

PR #2523, opened 2023-10-26
cli: state file validation

msanft:
### Context

We want to use the generic validation framework introduced in #2480 to verify the state file and the Constellation config file.

### Proposed change(s)

- Validate the state file by a set of constraints when read from disk.

### Additional info

- [AB#3503](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3503)

### Checklist

- [x] Link to Milestone


cli/internal/state/state.go
```diff
@@ -20,32 +20,41 @@ import (
...
 func ReadFromFile(fileHandler file.Handler, path string) (*State, error) {
 	state := &State{}
 	if err := fileHandler.ReadYAML(path, &state); err != nil {
 		return nil, fmt.Errorf("reading state file: %w", err)
 	}
+
+	v := validation.NewValidator()
+	if err := v.Validate(state, validation.ValidateOptions{}); err != nil {
+		return nil, fmt.Errorf("validating state file: %w", err)
+	}
```
daniel-weisse, line 43: I think it would be beneficial to not validate the state file directly on read, but have separate functions/methods to do so, similar to how we currently trigger validation on the config file.
This would allow us to run different validation when opening the file for `constellation create` and for `constellation apply`.
The first requiring the file to be completely empty, the later requiring just the infrastructure to be set.
Other commands also rely on values from the cluster part of the file, so they might want to be able to validate that those values are set.

tldr; We should have separate validation functions that check for empty infrastructure or cluster values, as well as functions that do the opposite (check for valid infrastructure or valid cluster values) and call those from the CLI command functions.

cli/internal/state/state.go
```diff
@@ -20,32 +20,52 @@ import (
...
 const (
 	// Version1 is the first version of the state file.
 	Version1 = "v1"
 )
 
-// ReadFromFile reads the state file at the given path and returns the state.
+const (
+	// PreCreate are the constraints that should be enforced when the state file
+	// is validated before "constellation create" is run.
+	PreCreate = iota
```
elchead, line 34: nit:
```suggestion
	PreCreate StateConstraints = iota
```

cli/internal/state/state.go
```diff
@@ -20,32 +20,52 @@ import (
...
+	// is validated before "constellation apply" is run.
+	PreInit
+	// PostInit are the constraints that should be enforced when the state file
+	// is validated after "constellation apply" is run.
+	PostInit
+)
+
+// StateConstraints defines which constraints the state file
+// should be validated against.
+type StateConstraints int
```
elchead, line 45: nit: rename to ValidationConstraint to avoid state.State..

cli/internal/state/state.go
```diff
@@ -186,6 +206,343 @@ func (s *State) Merge(other *State) (*State, error) {
 	return s, nil
 }
 
+// Validate validates the state against the given constraint set, which can be one of
+//   - PreCreate, which is the constraint set that should be enforced before "constellation create" is run.
```
elchead, line 210: nit: I think the ConstraintSet explanation is enough where the enum is defined

cli/internal/state/state.go
```diff
@@ -186,6 +206,343 @@ func (s *State) Merge(other *State) (*State, error) {
...
+func (s *State) preCreateConstraints() []*validation.Constraint {
+	return []*validation.Constraint{
+		// state version needs to be accepted by the parsing CLI.
+		validation.OneOf(s.Version, []string{Version1}).
+			WithFieldTrace(s, &s.Version),
+		// Infrastructure must be empty.
+		// As the infrastructure struct contains slices, we cannot use the
+		// Empty constraint on the entire struct. Instead, we need to check
+		// each field individually.
+		validation.Empty(s.Infrastructure.UID).
```
elchead, line 248: nit / idea: looking at this it seems that the WithFieldTrace seems always wanted. Why not integrate it into the condition and change the syntax to:
```
validation.OneOf(s, &s.Version, []string{Version1})
```
it would have the number of code.

cli/internal/state/state.go
```diff
@@ -186,6 +206,343 @@ func (s *State) Merge(other *State) (*State, error) {
...
+						return validation.And(
+							validation.EvaluateAll,
+							validation.Empty(s.Infrastructure.GCP.ProjectID).
+								WithFieldTrace(s, &s.Infrastructure.GCP.ProjectID),
+							validation.Empty(s.Infrastructure.GCP.IPCidrPod).
+								WithFieldTrace(s, &s.Infrastructure.GCP.IPCidrPod),
+						)
+					},
+				),
+			),
```
elchead, line 328: Do you agree that this would be much more readable? 
This solution would rely on closures but it seems good enough to me for the improved ergonomy. What do you think?

```suggestion
			*validation.Constraint {
				Satisfied: func() *TreeError {
					if s.Infrastructure.GCP == nil {
						return nil
					} else {
						return validation.Empty(s.Infrastructure.GCP)
					}
				}
			}
```
This syntax could probably be further shortened through a validation.CustomFn() helper.

cli/internal/state/state.go
msanft: Of course this would be preferred, but this doesn't work (without reflection in the constraint function) due to the pass-by-value semantics of Go. Theoretically, we could implement it with some hacks in the constraint functions that we would have to implicitly require in all constraints, which seemed like a bad practice to me. Especially for custom constraints. However, this is something I will keep an eye on in the future, I definitely agree with you in terms of ergonomics here.

cli/internal/state/state.go
```diff
@@ -186,6 +206,343 @@ func (s *State) Merge(other *State) (*State, error) {
...
+//
+// The constraints check if the state file version is valid,
+// and if all fields are empty, which is a requirement pre-create.
+func (s *State) preCreateConstraints() []*validation.Constraint {
+	return []*validation.Constraint{
+		// state version needs to be accepted by the parsing CLI.
+		validation.OneOf(s.Version, []string{Version1}).
+			WithFieldTrace(s, &s.Version),
+		// Infrastructure must be empty.
+		// As the infrastructure struct contains slices, we cannot use the
```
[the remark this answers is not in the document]
msanft: Without reflection - no

cli/internal/state/state.go
```diff
@@ -186,6 +206,343 @@ func (s *State) Merge(other *State) (*State, error) {
...
+		validation.NotEmpty(s.ClusterValues.OwnerID).
+			WithFieldTrace(s, &s.ClusterValues.OwnerID),
+		// MeasurementSalt needs to be filled.
+		validation.NotEmptySlice(s.ClusterValues.MeasurementSalt).
+			WithFieldTrace(s, &s.ClusterValues.MeasurementSalt),
+	}
+}
+
+// Constraints is a no-op implementation to fulfill the "Validatable" interface.
+func (s *State) Constraints() []*validation.Constraint {
```
elchead, line 542: I think we should aim to use the interface instead of implementing the "hacky" Validate function above which instantiates its dependency.

One way would be to use types:

```
type PreInitInfraState struct {
	*Infrastructure
}

func (state *PreInitInfraState) Constraints() []*validationConstraint {
	// implements state.preInitConstraints
}
```


Happy to discuss this and get other opinions on this

internal/validation/errors_test.go
```diff
@@ -24,7 +55,7 @@ func TestNewValidationErrorSingleField(t *testing.T) {
 	}
 
 	doc, field := references(t, st, &st.OtherField, "")
-	err := newError(doc, field, assert.AnError)
+	err := newTraceError(doc, field, assert.AnError)
```
elchead, line 58: lets call it traceError or treeError. I advocate trace

internal/validation/validation_test.go
```diff
@@ -30,6 +30,11 @@ func TestValidate(t *testing.T) {
 				},
 				NotEmptyField:   "certainly not.",
 				MatchRegexField: "abc",
+				OneOfField:      "one",
```
elchead, line 33: clever idea with these fields and the validation! I think the fields should be optional though to keep the focus of the test struct on the field that is relevant

cli/internal/cmd/apply.go
```diff
@@ -214,28 +214,28 @@ func runApply(cmd *cobra.Command, _ []string) error {
...
-		cmd.Context(),
-		constants.TerraformWorkingDir,
-		upgradeDir,
-		flags.tfLogLevel,
-		fileHandler,
-	)
-	if err != nil {
-		return fmt.Errorf("setting up cluster upgrader: %w", err)
+
+	newClusterApplier := func(ctx context.Context) (clusterUpgrader, error) {
```
elchead, line 218: nit: to make clear that this is related to the infra phase
```suggestion
	newInfrastructureApplier := func(ctx context.Context) (infrastructureApplier, error) {
```

elchead: was there a reason to make this a factory function? I think we want to avoid this pattern, @daniel-weisse? Afaik, we only do this for the other clients because their instantiation relies on information that is only available at a later stage (e.g. kubeconfig file or some flag argument)

elchead: thinking about it, it might be useful though if we want to avoid failures of the client instantiation (e.g. the terraform dir is in an invalid state) when we ignore the infra phase anyways.

cli/internal/state/state.go
elchead: I'm not sure why this doesn't work. but I do see that it would require to nicely wrap the reflect logic for DRY:

```
// OneOf is a constraint that s is in the set of values p.
func OneOf[T comparable](doc any, s T, p []T) *Constraint {
	c := Constraint{
		Satisfied: func() *TreeError {
			for _, v := range p {
				if s == v {
					return nil
				}
			}
			return NewErrorTree(fmt.Errorf("%v must be one of %v", s, p))
		},
	}
	derefedField := pointerDeref(reflect.ValueOf(s))
	fieldRef := referenceableValue{
		value: derefedField,
		addr:  derefedField.UnsafeAddr(),
		_type: derefedField.Type(),
	}
	derefedDoc := pointerDeref(reflect.ValueOf(doc))
	docRef := referenceableValue{
		value: derefedDoc,
		addr:  derefedDoc.UnsafeAddr(),
		_type: derefedDoc.Type(),
	}
	return c.withTrace(docRef, fieldRef)
}
```

msanft: Both `doc` and `s` get copied by value

cli/internal/state/state.go
msanft: Not a fan of this, as this would require type assertion everywhere this is used

internal/validation/errors_test.go
msanft: Those 2 do a different thing. TreeError formats a tree of errors nicely, while trace-error traverses the struct to find the field

cli/internal/state/state.go
elchead: I don't think type assertions are necessary, since this wrapped data type could also just be used for validation:

```
state...
err := validator.Validate(PreInitInfraState{state.Infrastructure})
```

msanft: I don't see any advantage in wrapping the type in every call. That comes out at basically the same that we do now, isn't it?

daniel-weisse: I agree with Moritz here.
I don't see an advantage for devs if they have to wrap the state in an additional data type to validate it.
This seems more complex and error prone then simply calling a validate method of the struct (and using some argument to specify what portions of the struct to validate)

cli/internal/state/state.go
```diff
@@ -186,6 +206,338 @@ func (s *State) Merge(other *State) (*State, error) {
...
+			WithFieldTrace(s, &s.ClusterValues.MeasurementSalt),
+	}
+}
+
+// postInitConstraints are the constraints on the state that should be enforced
+// *after* a Constellation cluster is initialized. (i.e. before "constellation apply" is run.)
+//
+// The constraints check if the infrastructure state and cluster state
+// is valid, so that the cluster can be used correctly.
+func (s *State) postInitConstraints() []*validation.Constraint {
```
daniel-weisse, line 408: How do we validate `postInitConstraints` contain/not-contain values for a specific CSP?
Right now these constraints would pass validation for GCP if the GCP values are left completely empty, wouldn't they?
They would also pass successfully for GCP if they contain only Azure values.

What about allowing callers to pass the CSP to the `Validate` method, and then conditionally adding constraints based on that?
E.g. for GCP we would enforce GCP values being valid, while Azure values have to be nil/empty.

elchead (approved):

cli/internal/cmd/apply.go
```diff
@@ -419,11 +419,24 @@ func (a *applyCmd) validateInputs(cmd *cobra.Command, configFetcher attestationc
...
+		// has already been initialized and check against the respective constraints.
+		if err := stateFile.Validate(state.PostInit, conf.GetProvider()); err != nil {
+			return nil, nil, err
+		}
+	} else {
+		// The cluster has not been initialized yet, so we check against the pre-init constraints.
+		if err := stateFile.Validate(state.PreInit, conf.GetProvider()); err != nil {
+			return nil, nil, err
+		}
+	}
```
daniel-weisse, line 439: This validation needs to be moved down (to the end of the input validation function), since we don't know what phases are skipped until we parsed all inputs

cli/internal/cmd/create.go
```diff
@@ -202,6 +202,9 @@ func (c *createCmd) create(cmd *cobra.Command, applier cloudApplier, fileHandler
 	if err != nil {
 		return fmt.Errorf("reading state file: %w", err)
 	}
+	if err := stateFile.Validate(state.PreCreate, provider); err != nil {
+		return fmt.Errorf("validating state file: %w", err)
+	}
```
daniel-weisse, line 207: Here we probably want to run validation at the start of the `create` function to prevent creating a cluster only to then fail after everything was created by Terraform because the state file wasn't empty.
This does not matter too much, since this function will be removed/reworked in https://github.com/edgelesssys/constellation/pull/2549

cli/internal/state/state.go
```diff
@@ -186,6 +207,324 @@ func (s *State) Merge(other *State) (*State, error) {
...
+			// OwnerID needs to be filled.
+			validation.NotEmpty(s.ClusterValues.OwnerID).
+				WithFieldTrace(s, &s.ClusterValues.OwnerID),
+			// MeasurementSalt needs to be filled.
+			validation.NotEmptySlice(s.ClusterValues.MeasurementSalt).
+				WithFieldTrace(s, &s.ClusterValues.MeasurementSalt),
+		}
+
+		switch csp {
+		case cloudprovider.Azure:
```
daniel-weisse, line 454: add a default case that verifies both GCP and Azure are empty

daniel-weisse (approved):

PR #2539, opened 2023-10-30
logging: convert logger package from zap to slog

miampf:
### Proposed change(s)

- rewrite the `logger` package in `internal/logger` to use slog instead of zap
- rewrite references to the logger package to be compatible with slog
- add non-formatting functions for similar usage to `fmt.Print(a, b)`

### Additional info

- code in `operators/` was not changed and still uses zap since it didn't depend on the `logger` package

### Checklist

- [X] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)
- [X] Add labels (e.g., for changelog category)
- [X] Is PR title adequate for changelog?
- [X] Link to Milestone

### DIFF
--- a/internal/logger/log.go
+++ b/internal/logger/log.go
@@ -216,7 +216,6 @@ func (l *Logger) With(fields ...any) *Logger {
 }
 
 // Grouped returns a named logger.
-// TODO: rename function to Grouped()
 func (l *Logger) Grouped(name string) *Logger {
 	return &Logger{logger: l.logger.WithGroup(name)}
 }
--- a/internal/logger/log.go
+++ b/internal/logger/log.go
@@ -225,7 +225,7 @@ func (l *Logger) GetClientStreamInterceptor() grpc.DialOption {
 
 func (l *Logger) middlewareLogger() logging.Logger {
 	return logging.LoggerFunc(func(ctx context.Context, lvl logging.Level, msg string, fields ...any) {
-		f := make([]slog.Attr, 0, len(fields)/2)
+		f := make([]interface{}, 0, len(fields)/2)
 
 		for i := 0; i < len(fields); i += 2 {
 			key := fields[i]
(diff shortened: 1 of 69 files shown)


internal/osimage/aws/awsupload.go
```diff
@@ -103,7 +102,7 @@ func (u *Uploader) Upload(ctx context.Context, req *osimage.UploadRequest) ([]ve
 	}
 	defer func() {
 		if err := u.ensureBlobDeleted(ctx, blobName); err != nil {
-			u.log.Errorf("post-cleaning: deleting temporary blob from s3", err)
+			u.log.Error("post-cleaning: deleting temporary blob from s3", err)
```
daniel-weisse, line 105:
```suggestion
			u.log.With(slog.Any("error", err)).Error("post-cleaning: deleting temporary blob from s3 failed")
```

daniel-weisse (commented):
Thanks, looks mostly good.
Just a couple of small things

internal/logger/log.go
```diff
@@ -6,7 +6,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 
 /*
 Package logger provides logging functionality for Constellation services.
-It is a thin wrapper around the zap package, providing a consistent interface for logging.
+It is a thin wrapper around the slog package, providing a consistent interface for logging.
```
katexochen, line 9: Wasn't the purpose of the ticket to get rid of the wrapper and use slog directly?

msanft: cc @miampf 

After internal discussion, we concluded that we rather want to get rid of the wrapper entirely and replace the logging calls we do in our code with using slog directly, feel free to consult me before implementation or if unsure anywhere. This was a misunderstanding and I didn't explain the issue to you properly - sorry for that. However, pivoting here does not mean that the work you did so far is worthless. You should be able to reuse almost everything you have so far.

miampf:
Since it was decided that the wrapper should be completely removed, I will create a new branch for implementing that.

PR #2549, opened 2023-11-02
cli: enable `constellation apply` to create new clusters

daniel-weisse:
### Context
The `constellation apply` command should be the one command to manage a Constellation cluster.
This PR brings the functionality of another command (`constellation create`) to `apply`.

### Proposed change(s)

- Use the logic of the `constellation apply` for creating new cluster, replacing `constellation create`

### Additional info

- [AB#3288](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3288)
- [x] [e2e test MiniConstellation](https://github.com/edgelesssys/constellation/actions/runs/6731720237)
  - Unchanged for now, should be adapted to use just `constellation apply` in the future
  - https://github.com/edgelesssys/constellation/pull/2551
- [x] [e2e test AWS](https://github.com/edgelesssys/constellation/actions/runs/6732121628)
- [x] [e2e test Azure](https://github.com/edgelesssys/constellation/actions/runs/6732122796)
- [x] [e2e test GCP](https://github.com/edgelesssys/constellation/actions/runs/6732124137)
- [x] [e2e self-managed infra GCP](https://github.com/edgelesssys/constellation/actions/runs/6888965366)
- [x] [e2e Terraform module GCP](https://github.com/edgelesssys/constellation/actions/runs/6888983480)

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/cli/internal/cloudcmd/tfplan.go
+++ b/cli/internal/cloudcmd/tfplan.go
@@ -55,7 +55,7 @@ func plan(
 
 	// If we are planning in a new workspace, we don't want to show a diff
 	if isNewWorkspace {
-		return false, nil
+		return hasDiff, nil
 	}
 
 	if hasDiff {
@@ -74,7 +74,7 @@ func restoreBackup(fileHandler file.Handler, workingDir, backupDir string) error
 	if err := fileHandler.CopyDir(
 		backupDir,
 		workingDir,
-	); err != nil {
+	); err != nil && !errors.Is(err, os.ErrNotExist) { // ignore not found error because backup might not exist for new clusters
 		return fmt.Errorf("replacing terraform workspace with backup: %w", err)
 	}
 
--- a/cli/internal/cloudcmd/tfplan.go
+++ b/cli/internal/cloudcmd/tfplan.go
@@ -53,7 +53,7 @@ func plan(
 		return false, fmt.Errorf("terraform plan: %w", err)
 	}
 
-	// If we are planning in a new workspace, we don't want to show a diff
+	// If we are planning in a new workspace, we don't want to show the plan
 	if isNewWorkspace {
 		return hasDiff, nil
 	}
--- a/cli/internal/cloudcmd/tfplan.go
+++ b/cli/internal/cloudcmd/tfplan.go
@@ -67,6 +67,7 @@ func plan(
 }
 
 // restoreBackup replaces the existing Terraform workspace with the backup.
+// If no backup exists, this function simply removes workingDir.
 func restoreBackup(fileHandler file.Handler, workingDir, backupDir string) error {
 	if err := fileHandler.RemoveAll(workingDir); err != nil {
 		return fmt.Errorf("removing existing workspace: %w", err)
@@ -74,7 +75,7 @@ func restoreBackup(fileHandler file.Handler, workingDir, backupDir string) error
 	if err := fileHandler.CopyDir(
 		backupDir,
 		workingDir,
-	); err != nil && !errors.Is(err, os.ErrNotExist) { // ignore not found error because backup might not exist for new clusters
+	); err != nil && !errors.Is(err, os.ErrNotExist) { // ignore not found error because backup does not exist for new clusters
 		return fmt.Errorf("replacing terraform workspace with backup: %w", err)
 	}
 
--- a/cli/internal/cmd/apply.go
+++ b/cli/internal/cmd/apply.go
@@ -60,8 +60,8 @@ const (
 )
 
 // allPhases returns a list of all phases that can be skipped as strings.
-func allPhases() []string {
-	return []string{
+func allPhases(except ...skipPhase) []string {
+	phases := []string{
 		string(skipInfrastructurePhase),
 		string(skipInitPhase),
 		string(skipAttestationConfigPhase),
@@ -70,6 +70,14 @@ func allPhases() []string {
 		string(skipImagePhase),
 		string(skipK8sPhase),
 	}
+
+	var returnedPhases []string
+	for idx, phase := range phases {
+		if !slices.Contains(except, skipPhase(phase)) {
+			returnedPhases = append(returnedPhases, phases[idx])
+		}
+	}
+	return returnedPhases
 }
 
 // formatSkipPhases returns a formatted string of all phases that can be skipped.
@@ -352,7 +360,7 @@ func (a *applyCmd) apply(
 	a.log.Debugf("Running license check")
 	checker := license.NewChecker(quotaChecker, a.fileHandler)
 	if err := checker.CheckLicense(cmd.Context(), conf.GetProvider(), conf.Provider, cmd.Printf); err != nil {
-		cmd.PrintErrf("License check failed: %v", err)
+		cmd.PrintErrf("License check failed: %s", err)
 	}
 	a.log.Debugf("Checked license")
 
--- a/cli/internal/cmd/apply.go
+++ b/cli/internal/cmd/apply.go
@@ -365,7 +365,7 @@ func (a *applyCmd) apply(cmd *cobra.Command, configFetcher attestationconfigapi.
 	// and apply migrations if necessary.
 	if !a.flags.skipPhases.contains(skipInfrastructurePhase) {
 		if err := a.runTerraformApply(cmd, conf, stateFile, upgradeDir); err != nil {
-			return fmt.Errorf("applying Terraform configuration : %w", err)
+			return fmt.Errorf("applying Terraform configuration: %w", err)
 		}
 	}
 
@@ -469,7 +469,8 @@ func (a *applyCmd) validateInputs(cmd *cobra.Command, configFetcher attestationc
 			return nil, nil, err
 		}
 
-		printCreateWarnings(cmd.OutOrStdout(), conf)
+		a.log.Debugf("No Terraform state found in current working directory. Preparing to create a new cluster.")
+		printCreateWarnings(cmd.ErrOrStderr(), conf)
 	}
 
 	// Check if the state file is in a pre-init or pre-create state
@@ -550,13 +551,9 @@ func (a *applyCmd) validateInputs(cmd *cobra.Command, configFetcher attestationc
 
 	// Constellation on QEMU or OpenStack don't support upgrades
 	// If using one of those providers, make sure the command is only used to initialize a cluster
-	if !(conf.GetProvider() == cloudprovider.AWS || conf.GetProvider() == cloudprovider.Azure || conf.GetProvider() == cloudprovider.GCP) {
-		if a.flags.skipPhases.contains(skipInitPhase) {
-			return nil, nil, fmt.Errorf("upgrades are not supported for provider %s", conf.GetProvider())
-		}
-		// Skip Terraform phase
-		a.log.Debugf("Skipping Infrastructure upgrade")
-		a.flags.skipPhases.add(skipInfrastructurePhase)
+	if !(conf.GetProvider() == cloudprovider.AWS || conf.GetProvider() == cloudprovider.Azure || conf.GetProvider() == cloudprovider.GCP) &&
+		(!a.flags.skipPhases.contains(skipImagePhase) && a.flags.skipPhases.contains(skipInitPhase)) {
+		return nil, nil, fmt.Errorf("image upgrades are not supported for provider %s", conf.GetProvider())
 	}
 
 	// Print warning about AWS attestation
--- a/cli/internal/cmd/apply.go
+++ b/cli/internal/cmd/apply.go
@@ -242,7 +242,6 @@ func runApply(cmd *cobra.Command, _ []string) error {
 		log:             log,
 		spinner:         spinner,
 		merger:          &kubeconfigMerger{log: log},
-		quotaChecker:    license.NewClient(),
 		newHelmClient:   newHelmClient,
 		newDialer:       newDialer,
 		newKubeUpgrader: newKubeUpgrader,
@@ -253,7 +252,7 @@ func runApply(cmd *cobra.Command, _ []string) error {
 	defer cancel()
 	cmd.SetContext(ctx)
 
-	return apply.apply(cmd, attestationconfigapi.NewFetcher(), upgradeDir)
+	return apply.apply(cmd, attestationconfigapi.NewFetcher(), license.NewClient(), upgradeDir)
 }
 
 type applyCmd struct {
@@ -263,8 +262,7 @@ type applyCmd struct {
 	log     debugLog
 	spinner spinnerInterf
 
-	merger       configMerger
-	quotaChecker license.QuotaChecker
+	merger configMerger
 
 	newHelmClient   func(kubeConfigPath string, log debugLog) (helmApplier, error)
 	newDialer       func(validator atls.Validator) *dialer.Dialer
@@ -340,24 +338,23 @@ The control flow is as follows:
 	                        │Write success output│
 	                        └────────────────────┘
 */
-func (a *applyCmd) apply(cmd *cobra.Command, configFetcher attestationconfigapi.Fetcher, upgradeDir string) error {
-	// Migrate state file
-	stateFile, err := state.ReadFromFile(a.fileHandler, constants.StateFilename)
-	if err != nil {
-		return fmt.Errorf("reading state file: %w", err)
-	}
-	if err := stateFile.Migrate(); err != nil {
-		return fmt.Errorf("migrating state file: %w", err)
-	}
-	if err := stateFile.WriteToFile(a.fileHandler, constants.StateFilename); err != nil {
-		return fmt.Errorf("writing state file: %w", err)
-	}
-
+func (a *applyCmd) apply(
+	cmd *cobra.Command, configFetcher attestationconfigapi.Fetcher,
+	quotaChecker license.QuotaChecker, upgradeDir string,
+) error {
 	// Validate inputs
 	conf, stateFile, err := a.validateInputs(cmd, configFetcher)
 	if err != nil {
 		return err
 	}
+
+	// Check license
+	a.log.Debugf("Running license check")
+	checker := license.NewChecker(quotaChecker, a.fileHandler)
+	if err := checker.CheckLicense(cmd.Context(), conf.GetProvider(), conf.Provider, cmd.Printf); err != nil {
+		cmd.PrintErrf("License check failed: %v", err)
+	}
+	a.log.Debugf("Checked license")
 
 	// Now start actually running the apply command
 
@@ -441,14 +438,6 @@ func (a *applyCmd) validateInputs(cmd *cobra.Command, configFetcher attestationc
 	if err != nil {
 		return nil, nil, err
 	}
-
-	// Check license
-	a.log.Debugf("Running license check")
-	checker := license.NewChecker(a.quotaChecker, a.fileHandler)
-	if err := checker.CheckLicense(cmd.Context(), conf.GetProvider(), conf.Provider, cmd.Printf); err != nil {
-		cmd.PrintErrf("License check failed: %v", err)
-	}
-	a.log.Debugf("Checked license")
 
 	// Validate the state file and set flags accordingly
 	//
--- a/cli/internal/cmd/apply.go
+++ b/cli/internal/cmd/apply.go
@@ -125,6 +125,7 @@ func NewApplyCmd() *cobra.Command {
 
 	must(cmd.Flags().MarkHidden("timeout"))
 
+	must(cmd.RegisterFlagCompletionFunc("skip-phases", skipPhasesCompletion))
 	return cmd
 }
 
@@ -541,7 +542,7 @@ func (a *applyCmd) validateInputs(cmd *cobra.Command, configFetcher attestationc
 
 	// Validate microservice version (helm versions) in the user's config matches the version of the CLI
 	// This makes sure we catch potential errors early, not just after we already ran Terraform migrations or the init RPC
-	if !a.flags.force && !a.flags.skipPhases.contains(skipHelmPhase) && !a.flags.skipPhases.contains(skipInitPhase) {
+	if !a.flags.force && !a.flags.skipPhases.contains(skipHelmPhase, skipInitPhase) {
 		if err := validateCLIandConstellationVersionAreEqual(constants.BinaryVersion(), conf.Image, conf.MicroserviceVersion); err != nil {
 			return nil, nil, err
 		}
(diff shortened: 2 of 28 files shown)


derpsteb (approved):
Lgtm. I am not very familiar with the TF code right now, so I am requesting @elchead as well.

derpsteb:
Maybe it makes sense at some point to reuse the [pattern](https://github.com/edgelesssys/constellation/blob/main/internal/api/attestationconfigapi/client.go#L142-L170) Adrian introduced in the attestationconfig cli.

It allows unittesting the correct order of operations. And might simplify the apply function. I have sth like this in mind:
```go
type Phase interface {
    func Apply() error
}

func NewPhases(skips map[skipPhase]{}) ([]Phase, error){
	// determine which phases to execute here.
	// call constructor for each phase.
}

func main() {
	phases, err := NewPhases(map[skipPhase]{})
	if err != nil {
		log.Fatal(err)
	}
	for _, phase := range phases {
		if err := phase.Apply(); err != nil {
			log.Fatal(err)
		}
	}
}
```

But I am not sure yet if this is needed. Most phases are just a single command. Imo the main goal would be to separate phase execution from determining which phase to run.

cli/internal/cmd/apply.go
```diff
@@ -603,3 +624,88 @@ func (a *applyCmd) tfStateExists() (bool, error) {
...
+	if printedAWarning {
+		fmt.Fprintln(out, "")
+	}
+}
+
+// skipPhasesCompletion returns suggestions for the skip-phases flag.
+// We suggest completion for all phases that can be skipped.
+// The phases may be given in any order, as a comma-separated list.
+// For example, "skip-phases helm,init" should suggest all phases but "helm" and "init".
+func skipPhasesCompletion(_ *cobra.Command, _ []string, toComplete string) ([]string, cobra.ShellCompDirective) {
```
elchead, line 680: could you add a unit test?

cli/internal/cmd/apply.go
```diff
@@ -394,8 +402,7 @@ func (a *applyCmd) apply(cmd *cobra.Command, configFetcher attestationconfigapi.
 
 	// Upgrade NodeVersion object
 	// This can be skipped if we ran the init RPC, as the NodeVersion object is already up to date
-	if !(a.flags.skipPhases.contains(skipK8sPhase) && a.flags.skipPhases.contains(skipImagePhase)) &&
-		a.flags.skipPhases.contains(skipInitPhase) {
+	if !(a.flags.skipPhases.contains(skipK8sPhase) && a.flags.skipPhases.contains(skipImagePhase)) {
```
elchead, line 405:
```suggestion
	if !(a.flags.skipPhases.contains(skipK8sPhase,skipImagePhase)) {
```

cli/internal/cmd/create.go
```diff
@@ -34,212 +22,24 @@ func NewCreateCmd() *cobra.Command {
...
+			cmd.Flags().Bool("skip-helm-wait", false, "")
+			cmd.Flags().Bool("merge-kubeconfig", false, "")
+			cmd.Flags().Duration("timeout", 5*time.Minute, "")
+			// Skip all phases but the infrastructure phase.
+			cmd.Flags().StringSlice("skip-phases", []string{
+				string(skipInitPhase), string(skipAttestationConfigPhase),
+				string(skipCertSANsPhase), string(skipHelmPhase),
+				string(skipImagePhase), string(skipK8sPhase),
+			}, "")
+			return runApply(cmd, args)
```
elchead, line 36: do we plan to deprecate create?

cli/internal/cmd/init_test.go
```diff
@@ -152,7 +152,7 @@ func TestInitialize(t *testing.T) {
 		/*
 			Tests currently disabled since we don't actually have validation for the state file yet
 			These tests cases only passed in the past because of unrelated errors in the test setup
-			TODO(AB#3492): Re-enable tests once state file validation is implemented
+			TODO(AB#3503): Re-enable tests once state file validation is implemented
```
elchead, line 155: I guess this will enabled once rebased

cli/internal/cmd/upgradeapply_test.go
```diff
@@ -211,7 +211,7 @@ func TestUpgradeApply(t *testing.T) {
 			},
 			fh: fsWithStateFileAndTfState,
 		},
-		"no tf state, skip infrastructure upgrade": {
+		"no tf state, but existing admin conf file": {
```
elchead, line 214: why does this cause an error? the terraform module doesn't store it's TF state in this folder but it will have an admin file

cli/internal/cloudcmd/tfplan.go
```diff
@@ -55,7 +55,7 @@ func plan(
 
 	// If we are planning in a new workspace, we don't want to show a diff
 	if isNewWorkspace {
-		return false, nil
+		return hasDiff, nil
```
elchead, line 58: nit: if the comment still applies it's clearer to just return false no?
```suggestion
		return false, nil
```

cli/internal/cmd/apply.go
```diff
@@ -436,16 +443,15 @@ func (a *applyCmd) validateInputs(cmd *cobra.Command, configFetcher attestationc
 	// Check if we already have a running Kubernetes cluster
 	// by checking if the Kubernetes admin config file exists
 	// If it exist, we skip the init phase
-	// If it does not exist, we need to run the init RPC first
-	// This may break things further down the line
```
elchead, line 440: should we document that the user should not move this file?

cli/internal/cmd/apply.go
```diff
@@ -486,29 +495,33 @@ func (a *applyCmd) validateInputs(cmd *cobra.Command, configFetcher attestationc
...
+	// If using one of those providers, print a warning when trying to upgrade the image
+	if !(conf.GetProvider() == cloudprovider.AWS || conf.GetProvider() == cloudprovider.Azure || conf.GetProvider() == cloudprovider.GCP) &&
+		!a.flags.skipPhases.contains(skipImagePhase) {
+		cmd.PrintErrf("Image upgrades are not supported for provider %s\n", conf.GetProvider())
+		cmd.PrintErrln("Image phase will be skipped")
+		a.flags.skipPhases.add(skipImagePhase)
 	}
 
 	// Check if Terraform state exists
+	// If it does not, assume we are creating a new cluster,
```
elchead, line 514: we cannot make this assumption with the Terraform modules

cli/internal/cmd/apply.go
```diff
@@ -486,29 +495,33 @@ func (a *applyCmd) validateInputs(cmd *cobra.Command, configFetcher attestationc
...
+	// meaning we may need to print some warnings
 	if tfStateExists, err := a.tfStateExists(); err != nil {
 		return nil, nil, fmt.Errorf("checking Terraform state: %w", err)
 	} else if !tfStateExists {
-		a.flags.skipPhases.add(skipInfrastructurePhase)
-		a.log.Debugf("No Terraform state found in current working directory. Assuming self-managed infrastructure. Infrastructure upgrades will not be performed.")
+		if err := a.checkDirClean(); err != nil && !a.flags.skipPhases.contains(skipInfrastructurePhase) {
```
elchead, line 519: For the TF module, we want to skip infra for a new cluster

cli/internal/cmd/apply.go
```diff
@@ -486,29 +495,33 @@ func (a *applyCmd) validateInputs(cmd *cobra.Command, configFetcher attestationc
...
+			return nil, nil, err
+		}
+
+		a.log.Debugf("No Terraform state found in current working directory. Preparing to create a new cluster.")
```
elchead, line 523: this is not true for TF module + self managed usage

cli/internal/cmd/apply.go
elchead, line 519: this change means that a self-managed infra user needs to explicityl skip infra. Let's update the docs in this case (create.md)

cli/internal/cmd/apply.go
```diff
@@ -517,10 +530,18 @@ func (a *applyCmd) validateInputs(cmd *cobra.Command, configFetcher attestationc
 		cmd.PrintErrln("WARNING: Attestation temporarily relies on AWS nitroTPM. See https://docs.edgeless.systems/constellation/workflows/config#choosing-a-vm-type for more information.")
 	}
 
+	if !a.flags.skipPhases.contains(skipInitPhase) {
```
elchead, line 533: nit: is it just me, but I find this easier to read;
skip suggests we should skip it. (feel free to ignore)
```suggestion
	if !a.flags.skipPhases.skip(skipInitPhase) {
```

cli/internal/cmd/apply_test.go
```diff
@@ -167,5 +177,215 @@ func TestSkipPhases(t *testing.T) {
...
+		"no admin config file, but mastersecret file exists": {
+			createConfig:       defaultConfig(cloudprovider.GCP),
+			createState:        defaultState,
+			createMasterSecret: defaultMasterSecret,
+			createAdminConfig:  func(require *require.Assertions, fh file.Handler) {},
+			createTfState:      defaultTfState,
+			flags:              applyFlags{},
+			wantErr:            true,
+		},
+		"no admin config file, no master secret": {
```
elchead, line 300: I wonder how we can make this test easier to read / understand.

```suggestion
		"skip k8s+image when init was not yet called (no admin + mastersecret file)": {
```

cli/internal/cmd/apply_test.go
```diff
@@ -167,5 +177,215 @@ func TestSkipPhases(t *testing.T) {
...
+			createMasterSecret: func(require *require.Assertions, fh file.Handler) {},
+			createAdminConfig:  func(require *require.Assertions, fh file.Handler) {},
+			createTfState:      defaultTfState,
+			flags:              applyFlags{},
+			wantPhases: skipPhases{
+				skipImagePhase: struct{}{},
+				skipK8sPhase:   struct{}{},
+			},
+		},
+		"no tf state, but admin config exists": {
```
elchead, line 312: this should not error, for self-managed so let's add this:
```suggestion
		"error if not skip infra and no tf state, but admin config exists": {
```

elchead: add a test case for self managed

cli/internal/cmd/apply_test.go
```diff
@@ -167,5 +177,215 @@ func TestSkipPhases(t *testing.T) {
...
+			createMasterSecret: func(require *require.Assertions, fh file.Handler) {},
+			createAdminConfig:  func(require *require.Assertions, fh file.Handler) {},
+			createTfState:      func(require *require.Assertions, fh file.Handler) {},
+			flags:              applyFlags{},
+			wantPhases: skipPhases{
+				skipImagePhase: struct{}{},
+				skipK8sPhase:   struct{}{},
+			},
+		},
+		"skip terraform": {
```
elchead, line 333: nit:
```suggestion
		"skip infra": {
```

cli/internal/cmd/apply_test.go
```diff
@@ -167,5 +177,215 @@ func TestSkipPhases(t *testing.T) {
...
+		"no tf state, but admin config exists": {
+			createConfig:       defaultConfig(cloudprovider.GCP),
+			createState:        defaultState,
+			createMasterSecret: defaultMasterSecret,
+			createAdminConfig:  defaultAdminConfig,
+			createTfState:      func(require *require.Assertions, fh file.Handler) {},
+			flags:              applyFlags{},
+			wantErr:            true,
+		},
+		"only config file": {
```
elchead, line 321: add a case with skip init

cli/internal/cmd/apply_test.go
```diff
@@ -167,5 +177,215 @@ func TestSkipPhases(t *testing.T) {
...
+			},
+		},
+		"skip terraform": {
+			createConfig:       defaultConfig(cloudprovider.GCP),
+			createState:        defaultState,
+			createMasterSecret: func(require *require.Assertions, fh file.Handler) {},
+			createAdminConfig:  func(require *require.Assertions, fh file.Handler) {},
+			createTfState:      func(require *require.Assertions, fh file.Handler) {},
+			flags: applyFlags{
+				skipPhases: skipPhases{
```
elchead, line 340: nit: might be good to have a constructor for this to hide the cumbersome init syntax

cli/internal/cmd/apply_test.go
```diff
@@ -167,5 +177,215 @@ func TestSkipPhases(t *testing.T) {
...
+					skipInfrastructurePhase: struct{}{},
+				},
+			},
+			wantPhases: skipPhases{
+				skipInfrastructurePhase: struct{}{},
+				skipImagePhase:          struct{}{},
+				skipK8sPhase:            struct{}{},
+			},
+		},
+	}
```
elchead, line 350: I think we should only support skip infra for self-managed cases (no tf state). Otherwise how do we ensure that a cluster upgrade is safe when a necessary infra change was skipped? Or is this always safe? At any rate this might break if a future CLI relies on a newer version infra state, but the state file migrations where not applied in prior CLI versions because infra was always skipped. Could this happen?

elchead:
> Maybe it makes sense at some point to reuse the [pattern](https://github.com/edgelesssys/constellation/blob/main/internal/api/attestationconfigapi/client.go#L142-L170) Adrian introduced in the attestationconfig cli.
> 
> It allows unittesting the correct order of operations. And might simplify the apply function. I have sth like this in mind:
> 
> ```go
> type Phase interface {
>     func Apply() error
> }
> 
> func NewPhases(skips map[skipPhase]{}) ([]Phase, error){
> 	// determine which phases to execute here.
> 	// call constructor for each phase.
>         // don't use reflection, hardcode each constructor call.
> }
> 
> func main() {
> 	phases, err := NewPhases(map[skipPhase]{})
> 	if err != nil {
> 		log.Fatal(err)
> 	}
> 	for _, phase := range phases {
> 		if err := phase.Apply(); err != nil {
> 			log.Fatal(err)
> 		}
> 	}
> }
> ```
> 
> But I am not sure yet if this is needed. Most phases are just a single command. Imo the main goal would be to separate phase execution from determining which phase to run.

that could indeed help to improve coverage of the cmd (cover order and correct skipping). but I see little value of the test to invest into it right now.

cli/internal/cmd/create.go
daniel-weisse: Yes, this should be deprecated.
I wasn't sure if we want to do it right away.
We might want to keep it around for a while (even if just as a convenience for devs since `constellation create` is a lost quicker to write than `constellation apply --skip-phases helm,init,attestationConfig,certSANs`)

cli/internal/cmd/upgradeapply_test.go
daniel-weisse: Previously, we had the Code set up to simply assume self-managed infrastructure if the workspace was not clean, but no tf state existed.
Since we now want to create new resources if the state does not exist, we can't simply continue if the workspace is not clean.
Maybe this test case would be better if placed in `TestValidateInputs`

cli/internal/cloudcmd/tfplan.go
daniel-weisse: The comment still applies, but its about not _showing_ the diff, i.e. not printing the `terraform plan` output to the user.
The caller still relies on the `hasDiff` information to create new resources (in case no state previously existed) or update them

cli/internal/cmd/apply.go
```diff
@@ -486,29 +495,33 @@ func (a *applyCmd) validateInputs(cmd *cobra.Command, configFetcher attestationc
...
-			return nil, nil, fmt.Errorf("upgrades are not supported for provider %s", conf.GetProvider())
-		}
-		// Skip Terraform phase
-		a.log.Debugf("Skipping Infrastructure upgrade")
-		a.flags.skipPhases.add(skipInfrastructurePhase)
+	// Constellation does not support image upgrades on all CSPs. Not supported are: QEMU, OpenStack
+	// If using one of those providers, print a warning when trying to upgrade the image
+	if !(conf.GetProvider() == cloudprovider.AWS || conf.GetProvider() == cloudprovider.Azure || conf.GetProvider() == cloudprovider.GCP) &&
+		!a.flags.skipPhases.contains(skipImagePhase) {
+		cmd.PrintErrf("Image upgrades are not supported for provider %s\n", conf.GetProvider())
```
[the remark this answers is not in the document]
daniel-weisse: Not really.
While we can update libvirt resources with Terraform, and also update helm charts (which was always possible), we still can't update the image, which imo is the most important part for upgrades.
So I don't think we should advertise this.

cli/internal/cmd/apply.go
daniel-weisse: We have to.
Self-managed infrastructure users should use `--skip-phases=infrastructure`

cli/internal/cmd/apply.go
daniel-weisse: No, TF module + self-managed should use `--skip-phases=infrastructure`

cli/internal/cmd/apply_test.go
daniel-weisse: Yes, skipping phases might break things further down the line, but this is something we have to accept if we want to provide the skip-phases functionality at all.
Similar to `kubeadm`, a user might have a need to skip specific phases, and we should provide the option to do so.
Simple reason why **we** would like to support it is `cdbg`: We want a way to run just the infrastructure phase, then run `cdbg`, then run anything but the infrastructure phase.

cli/internal/cmd/apply_test.go
daniel-weisse: > this should not error, for self-managed

For self-managed skipInfrastructure has to be set, otherwise it should error (which this test case checks)

> add a test case for self managed

See the last test case of this function

cli/internal/cmd/upgradeapply_test.go
daniel-weisse: This test case already exists in `TestValidateInputs`, removing it from here.

cli/internal/cmd/apply_test.go
daniel-weisse: Tried adding some more context to the names of the testcases.

cli/internal/cmd/apply_test.go
elchead: looks much better thanks! can you add a case for self-managed upgrade

docs/docs/architecture/orchestration.md
```diff
@@ -13,10 +13,10 @@ Each workspace is associated with a single cluster and configuration.
 The CLI stores state in the local filesystem making the current directory the active workspace.
 Multiple clusters require multiple workspaces, hence, multiple directories.
 Note that every operation on a cluster always has to be performed from the directory associated with its workspace.
-While files from the workspace may be copied to different locations, for example to create backups,
-they should not moved or deleted while the Constellation cluster is still being used.
+While files from the workspace may be copied to different locations,
+they should'nt be moved or deleted while the Constellation cluster is still being used.
```
elchead, line 17: they shouldn't be moved or deleted while the Constellation cluster is still being used.

docs/docs/architecture/orchestration.md
```diff
@@ -13,10 +13,10 @@ Each workspace is associated with a single cluster and configuration.
...
 The Constellation CLI will take care of managing the workspace.
-Only when a cluster was terminated, and you are sure the files are not needed anymore, should you remove a workspace.
+Only when a cluster was terminated, and you are sure the files are'nt needed anymore, should you remove a workspace.
```
elchead, line 19: Only when a cluster was terminated, and you are sure the files aren't needed anymore, should you remove a workspace.

cli/internal/cmd/create.go
```diff
@@ -34,215 +22,24 @@ func NewCreateCmd() *cobra.Command {
...
 		Args:  cobra.ExactArgs(0),
-		RunE:  runCreate,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cmd.Flags().Bool("conformance", false, "")
+			cmd.Flags().Bool("skip-helm-wait", false, "")
+			cmd.Flags().Bool("merge-kubeconfig", false, "")
+			cmd.Flags().Duration("timeout", 5*time.Minute, "")
+			// Skip all phases but the infrastructure phase.
+			cmd.Flags().StringSlice("skip-phases", []string{
+				string(skipInitPhase), string(skipAttestationConfigPhase),
```
elchead, line 32: was thinking if we should adjust `allPhases(except ...string)` and use here. this way we would be sure that we don't miss any phases if a new one gets added

elchead (approved):
lgtm, but we should merge after the release.

docs/docs/architecture/orchestration.md
```diff
@@ -13,6 +13,10 @@ Each workspace is associated with a single cluster and configuration.
 The CLI stores state in the local filesystem making the current directory the active workspace.
 Multiple clusters require multiple workspaces, hence, multiple directories.
 Note that every operation on a cluster always has to be performed from the directory associated with its workspace.
+While files from the workspace may be copied to different locations,
+they should'nt be moved or deleted while the Constellation cluster is still being used.
+The Constellation CLI will take care of managing the workspace.
+Only when a cluster was terminated, and you are sure the files are'nt needed anymore, should you remove a workspace.
```
thomasten, line 19:
```suggestion
Note that every operation on a cluster always has to be performed from the directory associated with its workspace.

You may copy files from the workspace to other locations,
but you shouldn't move or delete them while the cluster is still being used.
The Constellation CLI takes care of managing the workspace.
Only when a cluster was terminated, and you are sure the files aren't needed anymore, should you remove a workspace.
```
suggest starting a new paragraph and some rewording

thomasten (approved):

daniel-weisse:
Updated the docs to remove `create`.
See the last two commits.

docs/docs/getting-started/first-steps-local.md
```diff
@@ -100,29 +100,22 @@ attaching persistent storage, or autoscaling aren't available.
...
-  ```shell-session
-  $ constellation create
-  Your Constellation cluster was created successfully.
-  ```
-
-3. Initialize the cluster
+2. Now you can create your cluster and its nodes. `constellation apply` uses the options set in `constellation-conf.yaml`.
 
   ```bash
   constellation apply
```
thomasten, line 106:
```suggestion
  constellation apply -y
```

docs/docs/getting-started/first-steps-local.md
```diff
@@ -100,29 +100,22 @@ attaching persistent storage, or autoscaling aren't available.
...
   ```
 
-  This should give the following output:
+  The Output should look like the following::
```
thomasten, line 109:
```suggestion
  The Output should look like the following:
```

docs/docs/getting-started/first-steps-local.md
```diff
@@ -100,29 +100,22 @@ attaching persistent storage, or autoscaling aren't available.
...
 
   ```shell-session
-  $ constellation apply
+  $ constellation apply -y
+  Checking for infrastructure changes
+  The following Constellation cluster will be created:
+    1 control-plane node of type 2-vCPUs will be created.
```
thomasten, line 115:
```suggestion
    3 control-plane nodes of type 2-vCPUs will be created.
```

docs/docs/getting-started/first-steps.md
```diff
@@ -128,26 +128,19 @@ If you encounter any problem with the following steps, make sure to use the [lat
...
-
-    This should give the following output:
-
-    ```shell-session
-    $ constellation apply
+    $ constellation apply -y
+    Checking for infrastructure changes
+    The following Constellation cluster will be created:
+      3 control-plane node of type n2d-standard-4 will be created.
+      2 worker node of type n2d-standard-4 will be created.
```
thomasten, line 141:
```suggestion
      3 control-plane nodes of type n2d-standard-4 will be created.
      1 worker node of type n2d-standard-4 will be created.
```

docs/docs/workflows/create.md
```diff
@@ -8,32 +8,33 @@ This recording presents the essence of this page. It's recommended to read it in
...
+You can use the `--skip-phases` flag to skip specific phases of the process.
+For example, if you created the infrastructure manually, you can skip the cloud resource creation phase.
 
 See the [architecture](../architecture/orchestration.md) section for details on the inner workings of this process.
 
 :::tip
 If you don't have a cloud subscription, you can also set up a [local Constellation cluster using virtualization](../getting-started/first-steps-local.md) for testing.
 :::
 
-## The *create* step
```
thomasten, line 22: There are some links to this and the apply section that need to be fixed. Search for `create.md#`

thomasten (approved):

PR #2551, opened 2023-11-02
cli: use apply command to start mini cluster

daniel-weisse:
### Context
`mini up` was still using separate code to create and initialize the cluster.

### Proposed change(s)

- Use `constellation apply` to set up and initialize miniConstellation clusters

### Additional info

- We could think about deprecating `mini up` and `mini down` as a whole, and simply add a miniConstellation target for `config generate`, since `apply` now takes care of creating and initializing a cluster in one command.
- [x] [e2e test MiniConstellation](https://github.com/edgelesssys/constellation/actions/runs/6734000175)

### DIFF
--- a/cli/internal/cmd/minidown.go
+++ b/cli/internal/cmd/minidown.go
@@ -48,7 +48,7 @@ func checkForMiniCluster(fileHandler file.Handler) error {
 		return fmt.Errorf("reading state file: %w", err)
 	}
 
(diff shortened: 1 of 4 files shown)


elchead (approved):
lgtm, imo we don't need to prioritize merging it with the other cmds atm. I think that most of the broken mini issues we encountered in the past should also be solved now through the common apply code right?

derpsteb (approved):
Testing this locally works fine :+1:

PR #2553, opened 2023-11-03
terraform: Terraform module for GCP

elchead:
### Context

This PR builds on top of #2503 to add the module for GCP.

### Proposed change(s)

- add GCP module

### Additional info

- [AB#3520](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3520)
- [successful GCP E2E run](https://github.com/edgelesssys/constellation/actions/runs/6782739265/job/18435569332)
- [AWS E2E run](https://github.com/edgelesssys/constellation/actions/runs/6783103693)

### Checklist

- [x] Link to Milestone


msanft:
Let's put this on hold until #2503 is merged.

terraform/constellation-cluster/main.tf
```diff
@@ -34,16 +34,48 @@ resource "terraform_data" "config_generate" {
...
+
+resource "null_resource" "gcp_config" {
+  count = var.gcp_config != null ? 1 : 0
+  provisioner "local-exec" {
+    command = <<EOT
+      ./yq eval '.provider.gcp.project = "${var.gcp_config.project}"' -i constellation-conf.yaml
+      ./yq eval '.provider.gcp.region = "${var.gcp_config.region}"' -i constellation-conf.yaml
+      ./yq eval '.provider.gcp.zone = "${var.gcp_config.zone}"' -i constellation-conf.yaml
+      ./yq eval '.provider.gcp.serviceAccountKeyPath = "${var.gcp_config.serviceAccountKeyPath}"' -i constellation-conf.yaml
+
```
msanft, line 63:
```suggestion
```

terraform/gcp-constellation/main.tf
```diff
@@ -0,0 +1,86 @@
...
+  image_id        = module.fetch_image.image
+  name            = var.name
+  node_groups     = var.node_groups
+  region          = local.region
+  zone            = var.zone
+  debug           = var.debug
+  custom_endpoint = var.custom_endpoint
+}
+
+resource "null_resource" "sa_account_file" {
```
msanft, line 46:
```suggestion
resource "null_resource" "service_account_file" {
```
sa_account is semantically duplicate (i.e. serviceaccount_account)

terraform/gcp-constellation/main.tf
```diff
@@ -0,0 +1,86 @@
...
+  provisioner "local-exec" {
+    command = <<EOT
+          #echo "${module.gcp_iam.sa_key}" TODO use base64decode fn
```
msanft, line 49: what exactly does this mean?

msanft (commented):
LGTM from a high level. Will give this a more thorough review once the AWS module is merged.

terraform/gcp-constellation/main.tf
elchead: I originally wanted to try to get rid of the base64 dependency and use the HCL function. However, there were some encoding errors (quotation marks were not decoded) with it and I didn't look too deep since we support Linux and Mac were base64 is built-in and used with compatible flags (`-d`)

msanft:
Regarding release pipeline changes, @malt3 should review that.

terraform/constellation-cluster/main.tf
elchead: I thought it was nice to seperate conf and state configuration

elchead:
Rebased and incorporated feedback. Would you test it also locally on Mac to see if it works as expected for you @msanft?

msanft:
> Rebased and incorporated feedback. Would you test it also locally on Mac to see if it works as expected for you @msanft?

I don't have a MacOS machine set up for development right now unfortunately

.github/workflows/e2e-test-tf-module.yml
```diff
@@ -112,6 +111,37 @@ jobs:
...
+      - name: Create GCP Terraform variable input file
+        if: inputs.cloudProvider == 'gcp'
+        working-directory: ./terraform-module/gcp-constellation
+        run: |
+          cat > terraform.tfvars <<EOF
+          name = "${{ steps.create-prefix.outputs.prefix }}"
+          project = "${{ secrets.GCP_E2E_PROJECT }}"
+          service_account_id = "${{ steps.create-prefix.outputs.prefix }}-sa"
+          image = "${{ inputs.image }}"
+          zone = "${{ inputs.regionZone || 'europe-west3-b' }}"
```
msanft, line 123: I think we should rather add a default value in the `inputs` section, it doesn't need duplication there and is more accessible.

msanft (commented):
TF specific changes LGTM except the comment. Didn't look at release pipeline or Go upgrade. Just ping me once ready for final review

malt3 (approved):
Only reviewed the changes made to the pipeline. They look good to me.
Apologies for the delay in reviewing from my side.

msanft:
One thing that is broken here, but that will be addressed in #2566:

We cannot default to a `latest` version when downloading the CLI. This will break for users that have an outdated Terraform module, say of `v2.12.0`, and try to apply it at a point where `v2.13.0` is already released. `latest` will then point to  `v2.13.0`. If the  `v2.13.0`  CLI is incompatible with  the  `v2.12.0` module, this won't work.

We should rather let the user set the CLI version through a variable in the `constellation-cluster` module and set it during packaging of the module. I will take care of that in #2566.

.github/workflows/e2e-test-tf-module.yml
elchead: Leo wanted exactly this since the default is provider specific and we don't want a region input for each csp.

elchead:
> One thing that is broken here, but that will be addressed in #2566:
> 
> We cannot default to a `latest` version when downloading the CLI. This will break for users that have an outdated Terraform module, say of `v2.12.0`, and try to apply it at a point where `v2.13.0` is already released. `latest` will then point to `v2.13.0`. If the `v2.13.0` CLI is incompatible with the `v2.12.0` module, this won't work.
> 
> We should rather let the user set the CLI version through a variable in the `constellation-cluster` module and set it during packaging of the module. I will take care of that in #2566.

This is already fixed, we decided to inject the version during the release process:
https://github.com/edgelesssys/constellation/blob/cea6204b37c161169bc9675501305b6f2bbb68a8/dev-docs/workflows/release.md#L34-L35

msanft (approved):

PR #2557, opened 2023-11-06
ci: use structured logging for all parts of the malicious-join test

daniel-weisse:
### Context
The malicious join test was using structured logging for some parts, and relied on structured output to parse the final result.
However, large parts of the test used simple `fmt.Print` style logging.
This could result in some inconsistencies depending when the logs are flushed.

### Proposed change(s)

- Use structured logging for all output
- Correctly build, tag, upload and use malicious-join images from the branch the test is run on

### Related issue
- [e2e failure](https://github.com/orgs/edgelesssys/projects/3/views/1?pane=issue&itemId=43626559)
- [e2e test](https://github.com/edgelesssys/constellation/actions/runs/6772885265)

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/e2e/malicious-join/malicious-join.go
+++ b/e2e/malicious-join/malicious-join.go
@@ -20,10 +20,14 @@ import (
 	"github.com/edgelesssys/constellation/v2/internal/grpc/dialer"
 	"github.com/edgelesssys/constellation/v2/internal/logger"
 	"github.com/edgelesssys/constellation/v2/joinservice/joinproto"
+	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 )
 
 func main() {
+	log := logger.New(logger.JSONLog, zapcore.DebugLevel)
+	defer log.Sync()
+
 	jsEndpoint := flag.String("js-endpoint", "", "Join service endpoint to use.")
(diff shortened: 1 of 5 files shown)


e2e/malicious-join/malicious-join.go
```diff
@@ -50,48 +58,44 @@ func main() {
...
 		}
 	}
 
 	testOutput.AllPassed = allPassed
-	out, err := json.Marshal(testOutput)
-	if err != nil {
-		panic(fmt.Sprintf("marshalling test output: %s", err))
-	}
-	fmt.Println(string(out))
+	log.With(zap.Any("result", testOutput)).Infof("Test completed")
```
msanft, line 98: I assume, each zap logger instance has it's own buffer? Otherwise, we might want to flush before here to ensure that it's the last log.

msanft (approved):

e2e/malicious-join/malicious-join.go
daniel-weisse: See https://github.com/edgelesssys/constellation/blob/7cdb7fd4555921eab1c149a0b4321c33067e6916/e2e/malicious-join/malicious-join.go#L29

PR #2566, opened 2023-11-08
terraform: add Terraform module for Azure

msanft:
### Context

We want to have a Terraform module for all CSPs, including Azure.

### Proposed change(s)

- Add the Terraform module for Azure.
- Add CI for the Terraform module for Azure.
- Factor out the image-fetching to be a reusable action in the CI.
- Add a hidden command in the Constellation CLI for patching the MAA policy, and use it in the Terraform module as well as the self-managed infrastructure workflow.
- Automate the release process for the Terraform module, ensuring that a compatible Constellation CLI version is picked.

### Additional info

- [AB#3521](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3521)
- [E2E AWS](https://github.com/edgelesssys/constellation/actions/runs/6852226909)
- [E2E Azure](https://github.com/edgelesssys/constellation/actions/runs/6849644022)
- [E2E GCP](https://github.com/edgelesssys/constellation/actions/runs/6848785096)

### Checklist

- [x] Link to Milestone


.github/workflows/e2e-test-tf-module.yml
```diff
@@ -6,32 +6,28 @@ on:
...
           - "aws"
+          - "azure"
           - "gcp"
         required: true
       regionZone:
         description: "Region or zone to create the cluster in. Leave empty for default region/zone."
         type: string
       image:
-        description: "Node image version of the cluster."
+        description: "OS Image version used in the cluster's VMs. If not set, the latest debug image from main is used."
```
elchead, line 21: I did this in the weekly run, can you rm the image there then and azure there?

elchead: we should make clear that we accept both a reference and a release version

.github/workflows/e2e-test-tf-module.yml
```diff
@@ -40,16 +36,14 @@ on:
         description: "Which zone to use."
         type: string
       image:
-        description: "Node image reference which is compatible with the current dev release version."
+        description: "Node image reference which is compatible with the current dev release version. If not set, the latest debug image from main is used."
```
elchead, line 39: keep same as above description

.github/actions/self_managed_create/action.yml
```diff
@@ -84,7 +84,7 @@ runs:
       working-directory: ${{ github.workspace }}/e2e-infra
       if: ${{ inputs.cloudProvider }} == 'azure'
       run: |
-        bazel run //hack/maa-patch:maa-patch $(terraform output attestationURL | jq -r)
+        ./constellation maa-patch $(terraform output attestationURL | jq -r)
```
elchead, line 87: please also update the docs of create.md

.github/actions/upload_terraform_module/action.yml
```diff
@@ -9,15 +14,23 @@ runs:
         cp -r terraform terraform-module
         find terraform-module -name "*.go" -type f -delete
         find terraform-module -name "*.bazel" -type f -delete
+
+    - name: Stamp version
+      shell: bash
+      run: |
+        sed -i "s/@@CONSTELLATION_VERSION@@/${{ inputs.version }}/g" terraform-module/constellation-cluster/main.tf
```
elchead, line 21: on Mac only this worked: `sed -i '' 's/@@CONSTELLATION_VERSION@@/v2/g' terraform/constellation-cluster/variables.tf`
I suggest to doublecheck this works

```suggestion
        sed -i "s/@@CONSTELLATION_VERSION@@/${{ inputs.version }}/g" terraform-module/constellation-cluster/variables.tf
```

.github/workflows/release-cli.yml
```diff
@@ -112,6 +109,25 @@ jobs:
...
+  build-terraform-module:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout
+        id: checkout
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
+        with:
+          ref: ${{ inputs.ref || github.head_ref }}
+
+      - name: Setup bazel
```
elchead, line 121: why do we need bazel here?

.github/workflows/release-cli.yml
```diff
@@ -112,6 +109,25 @@ jobs:
             build/constellation-${{ matrix.os }}-${{ matrix.arch }}.exe
             build/constellation-${{ matrix.os }}-${{ matrix.arch }}.exe.sig
 
+  build-terraform-module:
```
elchead, line 112: nit:
```suggestion
  upload-terraform-module:
```

.github/workflows/e2e-test-tf-module.yml
```diff
@@ -62,16 +56,30 @@ jobs:
         with:
           ref: ${{ inputs.ref || github.head_ref }}
 
-      - name: Upload module
+      - name: Find Latest Image
```
elchead, line 59: nit: Get latest image (also for ID to suggest that it might not need to find it)

terraform/constellation-cluster/variables.tf
```diff
@@ -1,9 +1,15 @@
+variable "constellation_version" {
+  type        = string
+  description = "Constellation CLI version to use."
```
elchead, line 3: I'm not a big fan of exposing this to the user. The user should not be able to change this default since we don't support it?

terraform/aws-constellation/variables.tf
```diff
@@ -27,7 +27,7 @@ variable "node_groups" {
     instance_type = string
     disk_size     = number
     disk_type     = string
-    zone          = string
+    zone          = optional(string, "")
```
elchead, line 30: Is it correct that this field can be empty in the config? I see it makes sense to keep it optional and use var.zone otherwise though. Please add E2E runs for AWS and GCP also when ready

terraform/gcp-constellation/main.tf
```diff
@@ -43,9 +43,24 @@ module "gcp" {
   custom_endpoint = var.custom_endpoint
 }
 
+resource "null_resource" "sa_account_file" {
+  provisioner "local-exec" {
+    command = <<EOT
+          #echo "${module.gcp_iam.sa_key}" TODO use base64decode fn
```
elchead, line 49: please rebase

terraform/constellation-cluster/variables.tf
```diff
@@ -100,3 +123,15 @@ variable "microservice_version" {
...
   description = "Microservice version."
 }
+
+variable "debug" {
+  type        = bool
+  default     = false
+  description = "DON'T USE IN PRODUCTION: Enable debug mode and allow the use of debug images."
+}
+
+variable "create_maa" {
```
elchead, line 133: nit: should be part of azure config

.github/workflows/e2e-test-weekly.yml
```diff
@@ -376,20 +376,18 @@ jobs:
...
-          - test: "GCP"
-            provider: "gcp"
+          - provider: "gcp"
             regionZone: "europe-west2-a"
-          - test: "AWS"
-            provider: "aws"
+          - provider: "aws"
             regionZone: "us-east-2c"
+          - provider: "azure"
+            regionZone: "northeurope"
```
elchead, line 384: nit: should we rely on the region defaults of the workflow instead of hardcoding them here too?

.github/workflows/release-cli.yml
```diff
@@ -112,6 +109,20 @@ jobs:
             build/constellation-${{ matrix.os }}-${{ matrix.arch }}.exe
             build/constellation-${{ matrix.os }}-${{ matrix.arch }}.exe.sig
 
+  upload-terraform-module:
```
elchead, line 112: since you make it a separate job, the follow up jobs need to depend on it explicitly.

terraform/constellation-cluster/variables.tf
msanft: I've tried checking if `maaURL` in `azure_config` is set. Unfortunately, we cannot access any field on `azure_config` in the `count` attribute, it throws the error:
```console
The "count" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created.
To work around this, use the -target argument to first apply only the resources that the count depends on.
```

I couldn't figure out the exact path the dependencies are resolved in, or, to be more exact, I couldn't find the cause for the error if the dependencies are resolved in the order they should (imo).
With the release in mind I decided not to spend more time on this

.github/workflows/release-cli.yml
msanft: The other jobs shouldn't depend on it, or do I miss something? My whole point was to make it run in parallel.

terraform/constellation-cluster/variables.tf
elchead: you could add this to the resource azure_config (which seems better imo to encapsulate azure config logic in one resource) and make an if check.

.github/workflows/release-cli.yml
elchead: `Download Terraform module` depends on the upload so, `provenance-subjects`,  `provenance-verify` and `release` need it, right?

terraform/azure-constellation/constellation
```diff
@@ -0,0 +1 @@
+Not Found
```
elchead, line 1: you accidentally added the binaries + cluster files

elchead (approved):
lgtm

malt3 (approved):
I reviewed the changes to the release pipeline. Looks good to me.

daniel-weisse (approved):

cli/internal/cmd/maapatch.go
```diff
@@ -0,0 +1,71 @@
...
+func NewMaaPatchCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "maa-patch <attestation-url>",
+		Short: "Patch the MAA's attestation policy",
+		Long:  "Patch the MAA's attestation policy.",
+		Args: cobra.MatchAll(
+			cobra.ExactArgs(1),
+			func(cmd *cobra.Command, args []string) error {
+				if _, err := url.Parse(args[0]); err != nil {
+					return fmt.Errorf("argument %s is not a valid attestion url", args[0])
```
thomasten, line 28:
```suggestion
					return fmt.Errorf("argument %s is not a valid URL: %w", args[0], err)
```

docs/docs/reference/cli.md
```diff
@@ -840,3 +841,30 @@ constellation init [flags]
   -C, --workspace string   path to the Constellation workspace
 ```
 
+## constellation maa-patch
```
thomasten, line 844: change the clidocgen tool so that it skips hidden commands

docs/docs/workflows/create.md
```diff
@@ -252,14 +252,13 @@ management tooling of your choice. You need to keep the essential functionality
...
+  You can apply the update with the following command after creating the infrastructure, with `<URL>` being the URL of the MAA provider. (i.e. `$(terraform output attestationURL | jq -r)`, when using the minimal Terraform configuration.)
 
   ```bash
-  git clone --branch <VERSION> https://github.com/edgelesssys/constellation
-  cd constellation/hack/maa-patch
-  go run . $(terraform output attestationURL | jq -r)
+  constellation maa-patch <URL>
   ```
-
+   <!-- vale on -->
```
thomasten, line 261: nit: this can be above the code block

docs/docs/workflows/create.md
```diff
@@ -252,14 +252,13 @@ management tooling of your choice. You need to keep the essential functionality
 :::info
 
   On Azure, if the enforcement policy is set to `MAAFallback` in `constellation-config.yaml`, a manual update to the MAA provider's policy is necessary.
-  You can apply the update with the following commands, where `<VERSION>` is the version of Constellation that should be set up. (e.g. `v2.12.0`)
+  <!-- vale off -->
+  You can apply the update with the following command after creating the infrastructure, with `<URL>` being the URL of the MAA provider. (i.e. `$(terraform output attestationURL | jq -r)`, when using the minimal Terraform configuration.)
```
thomasten, line 256:
```suggestion
  You can apply the update with the following command after creating the infrastructure, with `<URL>` being the URL of the MAA provider (i.e., `$(terraform output attestationURL | jq -r)`, when using the minimal Terraform configuration).
```

terraform/constellation-cluster/variables.tf
```diff
@@ -84,10 +107,10 @@ variable "gcp_config" {
 
 variable "image" {
   type        = string
-  description = "The node image reference or semantical release version."
+  description = "The node image reference or semantic release version."
```
elchead, line 110: This allows to leave the field empty too and follows the same defaults we have in the CLI config.
```suggestion
  description = "The node image reference or semantic release version."
  default     = "@@CONSTELLATION_VERSION@@"
```

msanft: Any argument in favor of this except one less input? I think we should have an explicit image input to be aligned with the config. But no strong opinion.

docs/docs/workflows/create.md
msanft: It cannot be inside the info block unfortunately, see the [build logs](https://app.netlify.com/sites/constellation-docs/deploys/65523a73195c900009569b53)

terraform/constellation-cluster/variables.tf
elchead: One less input with sensible default. This is what `config generate` provides too.  It can still be configured. How does it not align with the config?

msanft: You're right. I thought we didn't fill it in the config.

thomasten (approved):
cli, docs, and hack LGTM.

PR #2567, opened 2023-11-08
ci: encrypt artifacts

miampf:
### Context

Currently, any artifacts uploaded by the CI are unencrypted which can be insecure (e.g. for logs).

### Proposed change(s)

- add the action `.github/actions/artifact_upload`
- add the action `.github/actions/artifact_download`
- replace all CI references to `actions/upload-artifact`/`actions/download-artifact` with their respective wrappers

### Checklist


msanft:
Removed the security-fix label as this is rather a defensive security measure on our side and since it interferes with the no-changelog label in the changelog generation process

.github/actions/artifact_download/action.yml
```diff
@@ -0,0 +1,30 @@
...
+      shell: bash
+      run: echo "directory=$(mktemp -d)" >> "$GITHUB_OUTPUT"
+    - name: Download the artifact
+      uses: actions/download-artifact@v3
+      with:
+        name: ${{ inputs.name }}
+        path: ${{ steps.tempdir.outputs.directory }}
+    - name: Decrypt and unzip archive
+      shell: bash
+      run: unzip -P '${{ inputs.encryption-secret }}' -o -qq -d ${{ inputs.path }} ${{ steps.tempdir.outputs.directory }}/archive.zip
```
msanft, line 30: I don't know if we actually want to overwrite (`-o`) here. Why did you choose to do so?

.github/actions/artifact_upload/action.yml
```diff
@@ -0,0 +1,50 @@
...
+  using: "composite"
+  steps:
+    - name: Install zip
+      uses: montudor/action-zip@v1
+    - name: Find glob pattern matches
+      uses: tj-actions/glob@v17
+      id: glob
+      with:
+        files: |
+          ${{ inputs.path }}
```
msanft, line 27: For this, I think we should be able to scrap it completely, normal bash globs should work.

.github/actions/artifact_upload/action.yml
```diff
@@ -0,0 +1,50 @@
...
+      run: echo "directory=$(mktemp -d)" >> "$GITHUB_OUTPUT"
+    - name: Create archive
+      shell: bash
+      run: |
+        # Create an archive if files exist.
+        # Don't create an archive file if no files are found
+        # and warn/error if configured that way.
+        if [ -n "${{ steps.glob.outputs.paths }}" ]
+        then
+          zip -e -P '${{ inputs.encryption-secret }}' -qq -r ${{ steps.tempdir.outputs.directory }}/archive.zip ${{ steps.glob.outputs.paths }}
```
msanft, line 40: Won't `-e` require a password to be entered in a prompt? Also, why do we need `-r` here. We would only want to unpack one archive, right?

msanft (changes requested):
One general thing: If making as many commits is part of your standard workflow, awesome! If you do this because you think we expect it. No. Feel free to do less commits, don't feel urged by the "guidelines" saying that you should push your work regularly. This is _really_ regular. However, I really don't wan't you to change it if it's part of your default workflow. Just don't feel urged to do that by us :D

.github/actions/artifact_download/action.yml
```diff
@@ -0,0 +1,30 @@
...
+    description: 'The name of the artifact.'
+    required: true
+  path:
+    description: 'Download to a specified path.'
+    required: false
+    default: ./
+  encryption-secret:
+    description: 'The secret to use for decrypting the artifact.'
+    required: true
+runs:
```
katexochen, line 14:
```suggestion
name: Download artifact
description: Download and decrypt an artifact.

inputs:
  name:
    description: 'The name of the artifact.'
    required: true
  path:
    description: 'Download to a specified path.'
    required: false
    default: ./
  encryption-secret:
    description: 'The secret to use for decrypting the artifact.'
    required: true
    
runs:
```

.github/actions/artifact_download/action.yml
```diff
@@ -0,0 +1,30 @@
...
+  using: "composite"
+  steps:
+    - name: Install unzip
+      uses: montudor/action-zip@v1
+    - name: Create temporary directory
+      id: tempdir
+      shell: bash
+      run: echo "directory=$(mktemp -d)" >> "$GITHUB_OUTPUT"
+    - name: Download the artifact
```
katexochen, line 23:
```suggestion
    - name: Create temporary directory
      id: tempdir
      shell: bash
      run: echo "directory=$(mktemp -d)" >> "$GITHUB_OUTPUT"
      
    - name: Download the artifact
```
empty lines to separate steps

.github/actions/artifact_upload/action.yml
```diff
@@ -0,0 +1,50 @@
...
+      id: tempdir
+      shell: bash
+      run: echo "directory=$(mktemp -d)" >> "$GITHUB_OUTPUT"
+    - name: Create archive
+      shell: bash
+      run: |
+        # Create an archive if files exist.
+        # Don't create an archive file if no files are found
+        # and warn/error if configured that way.
+        if [ -n "${{ steps.glob.outputs.paths }}" ]
```
katexochen, line 38:
```suggestion
        if [[ -n "${{ steps.glob.outputs.paths }}" ]]
```
we use [[ per convention.

.github/actions/artifact_download/action.yml
miampf: Oh, this was an artifact of some tests I ran at the beginning and I overlooked it when I cleaned up the branch. I removed it now.

.github/actions/artifact_upload/action.yml
miampf: I don't think that normal bash globs would be enough since I cannot find a way to reliably apply multiple glob patterns while still allowing the same intuitive usage of them as we did before. For example:
https://github.com/edgelesssys/constellation/blob/e105ab5ced130d84ebb6be3df1d4ea1a61031d18/.github/actions/constellation_create/action.yml#L232-L234
Just appending these patterns to a command will not work, it would still zip the `terraform.log` file. I also wasn't able to find another way of applying both filters.

.github/actions/artifact_upload/action.yml
miampf: `-e` specifies that the archive should be encrypted. However, it does not ask for a password since it's provided with the `-P` flag. The `-r` is there because per default, `zip` doesn't recursively include stuff in directories, so if you give it just a directory it wouldn't include any files in that directory if the `-r` flag isn't given.

msanft: Thanks for the explanation!

.github/actions/artifact_upload/action.yml
miampf: `extglob` patterns are now used and it seems to work fine.

.github/actions/artifact_upload/action.yml
```diff
@@ -0,0 +1,61 @@
...
+
+inputs:
+  path:
+    description: 'The path(s) that should be uploaded. Those are evaluated with bash and the extglob option.'
+    required: true
+  name:
+    description: 'The name of the artifact.'
+    required: true
+  retention-days:
+    description: 'How the artifact should be retained.'
```
msanft, line 12:
```suggestion
    description: 'How long the artifact should be retained for.'
```

.github/actions/artifact_download/action.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    - name: Download the artifact
+      uses: actions/download-artifact@v3
+      with:
+        name: ${{ inputs.name }}
+        path: ${{ steps.tempdir.outputs.directory }}
+
+    - name: Decrypt and unzip archive
+      shell: bash
+      run: |
```
msanft, line 35: See #2605 and rebase to that once merged.

.github/actions/artifact_upload/action.yml
```diff
@@ -0,0 +1,61 @@
...
+    required: false
```
msanft, line 13:
```suggestion
```

.github/actions/e2e_benchmark/action.yml
```diff
@@ -166,11 +168,12 @@ runs:
 
     - name: Upload benchmark results to action run
       if: (!env.ACT)
-      uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+      uses: ./.github/actions/artifact_upload
       with:
-        path: |
+        path: >
```
msanft, line 173:
```suggestion
        path: |
```
assume a typo? :smile:

.github/workflows/e2e-upgrade.yml
```diff
@@ -276,13 +276,14 @@ jobs:
 
       - name: Always upload logs
         if: always()
-        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        uses: ./.github/actions/artifact_upload
         with:
           name: upgrade-logs
-          path: |
+          path: >
```
msanft, line 282:
```suggestion
          path: |
```

s.a.

.github/workflows/release-cli.yml
```diff
@@ -92,22 +92,24 @@ jobs:
           cosignPassword: ${{ inputs.key == 'release' && secrets.COSIGN_PASSWORD || secrets.COSIGN_DEV_PASSWORD }}
 
       - name: Upload CLI as artifact (unix)
-        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        uses: ./.github/actions/artifact_upload
         if : ${{ matrix.os != 'windows' }}
         with:
           name: constellation-${{ matrix.os }}-${{ matrix.arch }}
-          path: |
+          path: >
```
msanft, line 99:
```suggestion
          path: |
```
s.a.

.github/workflows/release-cli.yml
```diff
@@ -92,22 +92,24 @@ jobs:
...
+          encryption-secret: ${{ secrets.ARTIFACT_ENCRYPT_PASSWD }}
 
       - name: Upload CLI as artifact (windows)
-        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        uses: ./.github/actions/artifact_upload
         if : ${{ matrix.os == 'windows' }}
         with:
           name: constellation-${{ matrix.os }}-${{ matrix.arch }}
-          path: |
+          path: >
```
msanft, line 109:
```suggestion
          path: |
```
s.a.

.github/workflows/test-artifact-actions-and-stuff.yml
```diff
@@ -0,0 +1,24 @@
+name: Test artifact actions
+on: push
```
msanft, line 2:
```suggestion
on: push
```

It's super great that you have added a workflow to test your action, but we should limit it to run only when the [action is modified](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-affects-specific-files)

test-folder/file-1.txt
msanft, line 1: remove these extra files

.github/workflows/test-artifact-actions-and-stuff.yml
```diff
@@ -0,0 +1,24 @@
...
+          path: test-folder/file-*.txt
+          encryption-secret: asdf
+      - name: Download
+        uses: ./.github/actions/artifact_download
+        with:
+          name: test-artifact
+          path: /tmp
+          encryption-secret: asdf
+      - name: Check existence
+        run: ls /tmp/test-folder
```
msanft, line 24: We could not only test that the item exists here, but also that it is 
1. Not containing the original content in it's encrypted state, by downloading it through the "normal" artifact download action
2. Containing exactly the original content after decryption

msanft (changes requested):
Changes look really good to me already other than that :)

.github/workflows/test-artifact-actions-and-stuff.yml
miampf: Oh, I actually meant to remove that workflow before merging, should I keep it?

.github/workflows/test-artifact-actions-and-stuff.yml
miampf: I verified both things by manually downloading the artifacts, but I can rewrite it if the test workflow is meant to be kept for merging.

.github/actions/artifact_download/action.yml
```diff
@@ -0,0 +1,30 @@
...
+    required: false
+    default: ./
+  encryption-secret:
+    description: 'The secret to use for decrypting the artifact.'
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Install unzip
+      uses: montudor/action-zip@v1
```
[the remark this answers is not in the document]
miampf: `zip` and `unzip` are now installed over nix with the [setup_bazel_nix](https://github.com/edgelesssys/constellation/blob/main/.github/actions/setup_bazel_nix/action.yml) action.

.github/workflows/test-artifact-actions-and-stuff.yml
msanft: I don't mind keeping this, as this tests some actually vital functionality. @katexochen what's your opinion?

katexochen: I'd opt to remove, code that doesn't run is eventually broken. The code will be tested well enough, as we execute it frequently in many actions.

.github/actions/artifact_upload/action.yml
```diff
@@ -0,0 +1,63 @@
...
+      shell: bash
+      run: |
+        shopt -s extglob
+
+        # Check if any file matches the given pattern(s).
+        something_exists=false
+        for pattern in ${{ inputs.path }}
+        do
+          if compgen -G $pattern > /dev/null
+          then
```
katexochen, line 42:
```suggestion
          if compgen -G $pattern > /dev/null; then
```

.github/actions/artifact_upload/action.yml
```diff
@@ -0,0 +1,63 @@
...
+
+    - name: Create archive
+      shell: bash
+      run: |
+        shopt -s extglob
+
+        # Check if any file matches the given pattern(s).
+        something_exists=false
+        for pattern in ${{ inputs.path }}
+        do
```
katexochen, line 40:
```suggestion
        for pattern in ${{ inputs.path }}; do
```

katexochen (changes requested):

.github/workflows/test-artifact-actions-and-stuff.yml
msanft: Ok, let's remove it as per Paul's comment above.

miampf: I removed everything test related :)

malt3 (approved):
LGTM

.github/actions/artifact_download/action.yml
```diff
@@ -0,0 +1,39 @@
...
+  name:
+    description: 'The name of the artifact.'
+    required: true
+  path:
+    description: 'Download to a specified path.'
+    required: false
+    default: ./
+  encryption-secret:
+    description: 'The secret to use for decrypting the artifact.'
+    required: true
```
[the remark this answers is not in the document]
malt3: Maybe it can be an optional input? This way we get the best of both worlds.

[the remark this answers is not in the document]
miampf: In my experience this doesn't really work since I believe that composite actions are prevented from reading secrets. I originally tried to read it directly in the action and it didn't work, I remember the docs telling me the reason was the one above.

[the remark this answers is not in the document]
miampf: Not the official docs, but [here is a stackoverflow post with the same problem](https://stackoverflow.com/questions/70098241/using-secrets-in-composite-actions-github)

malt3:
I think this can be merged 🚀

.github/actions/constellation_create/action.yml
```diff
@@ -59,6 +59,9 @@ inputs:
   force:
     description: "Set the force-flag on apply to ignore version mismatches."
     required: false
+  encryption-secret:
+    description: 'The secret to use for encrypting the artifact.'
```
daniel-weisse, line 63:
```suggestion
    description: "The secret to use for encrypting the artifact."
```
nit: I don't think we have a set style guide for the quoting style in our actions/workflows, but it should at least be kept consistent in one action.
Also applies to other actions changed by this PR

.github/workflows/draft-release.yml
```diff
@@ -92,22 +92,24 @@ jobs:
           cosignPassword: ${{ inputs.key == 'release' && secrets.COSIGN_PASSWORD || secrets.COSIGN_DEV_PASSWORD }}
 
       - name: Upload CLI as artifact (unix)
-        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        uses: ./.github/actions/artifact_upload
```
daniel-weisse, line 95: nit: We don't technically need to encrypt these artifacts since they'll be published as part of the release anyways

.github/workflows/draft-release.yml
```diff
@@ -340,14 +347,16 @@ jobs:
         uses: ./.github/actions/download_release_binaries
```
daniel-weisse, line 347: This is missing the encryption/decryption key input

daniel-weisse (changes requested):

daniel-weisse (approved):

msanft (approved):

malt3 (approved):

PR #2568, opened 2023-11-08
ci: dont run SNP version upload on v2.12.0 CLI tests

daniel-weisse:
### Context
The sev-guest library used in v2.12.0 of Constellation included a debug `fmt.Println` statement in the verification code for SNP values. This breaks the SNP version upload since the stdout output of `constellation verify -o json` is no longer valid json.
Since v2.12.0 also didn't properly marshal all certificates, I opted to not fix the output file in case we are running a test with a v2.12 CLI, but simply not run the version upload

### Proposed change(s)

- Don't upload TCB versions from Azure e2e verify tests for CLI versions v2.12

### Additional info

- [Here](https://github.com/edgelesssys/constellation/actions/runs/6799481770/job/18486776613#step:6:4825) is the error in a run on main using `v2.12.0` CLI
- [Here](https://github.com/edgelesssys/constellation/actions/runs/6752370284/job/18357662122#step:3:5102) we can see the error occurring in our weekly e2e tests
- [e2e test v2.12.0 CLI with fix](https://github.com/edgelesssys/constellation/actions/runs/6800371931)
- [e2e test v2.13.0-pre with fix](https://github.com/edgelesssys/constellation/actions/runs/6800374308)

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/.github/actions/e2e_verify/action.yml
+++ b/.github/actions/e2e_verify/action.yml
@@ -90,6 +90,11 @@ runs:
         COSIGN_PASSWORD: ${{ inputs.cosignPassword }}
         COSIGN_PRIVATE_KEY: ${{ inputs.cosignPrivateKey }}
       run: |
+        if constellation version | head -n 1 | grep "v2.12.0" > /dev/null
+        then
+          exit 0
+        fi
+
         for file in $(ls snp-report-*.json); do
           path=$(realpath "${file}")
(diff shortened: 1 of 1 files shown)


derpsteb (approved):
Since the upload is only executed on main we will have to merge this to see a successful run. But lgtm.

PR #2577, opened 2023-11-13
docs: add Helm chart for VPN connectivity

burgerdev:
### Context

Constellation users would like to connect their VPNs to their cluster. This howto describes how a simple setup with NATs and Proxies might look like.

### Proposed change(s)

* Helm chart for NATed connection to a Constellation cluster added to the howtos.

### Additional info

- [AB#3540](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3540)

### Checklist
- [x Is PR title adequate for changelog?
- [x] Link to Milestone


dev-docs/howto/vpn/templates/statefulset.yaml
```diff
@@ -0,0 +1,108 @@
...
+          capabilities:
+            add: ["NET_ADMIN"]
+        volumeMounts:
+        - name: wireguard
+          mountPath: "/etc/wireguard"
+          readOnly: true
+      {{- end }}
+      containers:
+      - name: tproxy
+        image: docker.io/burgerdev/tproxy:latest
```
3u13r, line 45: We should put this repo/Dockerfile/Containerfile somewhere publicly. Either your or the companies github. Also, I don't know if there's a difference in the registries (i.e. maybe docker.io has stricter download limitation in comparison to ghcr.io) but I also like that on ghcr.io one can link the container to the repository. This would make the source easily findable.

burgerdev: The docker.io image was deployed from a dirty branch. It's productionized at github.com/burgerdev/go-tproxy now, adjusted the image value accordingly.

dev-docs/howto/vpn/templates/service.yaml
```diff
@@ -0,0 +1,25 @@
...
+    {{- include "..selectorLabels" . | nindent 4 }}
+    component: frontend
+  ports:
+    {{- if .Values.ipsec.enabled }}
+    - name: isakmp
+      protocol: UDP
+      port: 500
+    - name: ipsec-nat-t
+      protocol: UDP
+      port: 4500
```
3u13r, line 19: Interestingly, this creates a LB with ports _from_ 500 _to_ 4500. We should either try to fix this with annotations (if possible) or just deploy 2 different LBs (this is more agnostic than cloud-specific annotations).

dev-docs/howto/vpn/README.md
```diff
@@ -0,0 +1,36 @@
...
+1. Create and populate the configuration.
+
+   ```sh
+   helm inspect values . >config.yaml
+   ```
+
+2. Install the Helm chart.
+
+   ```sh
+   helm install vpn . >config.yaml
```
3u13r, line 16: Overwriting the config looks wrong. Did you mean to
```suggestion
   helm install -f config.yaml vpn . 
```

3u13r (changes requested):

dev-docs/howto/vpn/README.md
burgerdev: Good catch, thanks.

dev-docs/howto/vpn/templates/service.yaml
burgerdev: This seems to be a bug in the load balancer implementation, doesn't it? https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services are a supported concept, and I only saw the note about needing to use the same protocol for all ports.

The correct workaround may be to just not expose 500, and configure on-prem to start with encap right away.

burgerdev: I just double-checked, and while the LB frontend claims to be forwarding the 500-4500 range, only  500 and 4500 are observed on the node.

3u13r: https://github.com/kubernetes/cloud-provider-gcp/blob/master/providers/gce/gce_loadbalancer_external.go#L838
Oh well, this is not how I would have implemented that. 
And they even have a different logic for the firewall rules:
https://github.com/kubernetes/cloud-provider-gcp/blob/master/providers/gce/gce_loadbalancer_internal.go#L868C10-L868C10
This is truly their default way of doing things.

dev-docs/howto/vpn/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
+name: vpn
+description: A VPN server for Constellation
+
+type: application
+
+version: 0.1.0
+
+appVersion: "v2.13.0"  # target Constellation version
```
3u13r, line 9: Since this is likely very loosely coupled and we don't want to confuse people about the compatibility I think that we should drop the optional `appVersion` field.

dev-docs/howto/vpn/templates/statefulset.yaml
```diff
@@ -0,0 +1,108 @@
...
+  template:
+    metadata:
+      labels:
+        {{- include "..selectorLabels" . | nindent 8 }}
+        component: frontend
+    spec:
+      hostNetwork: false
+      initContainers:
+      - name: tproxy-setup
+        image: nixery.dev/busybox/iptables
```
3u13r, line 20: I think in the long run we want to be able to pin images _at least_ by version if not also by hash. We could also think about signing them. 

It should be fine to just use one image with the union of all packages needed.

We should create a follow-up ticket for this.

dev-docs/howto/vpn/files/tproxy-setup.sh
```diff
@@ -0,0 +1,37 @@
...
+  iptables -t nat -A VPN_POST -s "${cidr}" -d "${VPN_POD_CIDR}" -j MASQUERADE
+done
+
+iptables -t nat -C POSTROUTING -j VPN_POST || iptables -t nat -A POSTROUTING -j VPN_POST
+
+### Service IPs ###
+
+# Service IPs need to be connected to locally to trigger the cgroup connect hook, thus we send them to the transparent proxy.
+
+# Packets with mark 1 are for tproxy and need to be delivered locally.
```
3u13r, line 21:
```suggestion
# Packets with mark 1 are for tproxy and need to be delivered locally.
# For more information see: https://www.kernel.org/doc/Documentation/networking/tproxy.txt
```

dev-docs/howto/vpn/templates/statefulset.yaml
```diff
@@ -0,0 +1,108 @@
...
+        env: {{- include "..commonEnv" . | nindent 10 }}
+        securityContext:
+          capabilities:
+            add: ["NET_ADMIN"]
+        volumeMounts:
+        - name: tproxy-setup
+          mountPath: "/entrypoint.sh"
+          subPath: "tproxy-setup.sh"
+          readOnly: true
+      {{- if .Values.wireguard.enabled }}
```
3u13r, line 31: I think that having both WireGuard and IPSec in one file is a bit convoluted. I think it makes sense to split them into 2 statefulsets.

dev-docs/howto/vpn/templates/strongswan-secret.tpl
```diff
@@ -0,0 +1,26 @@
...
+    remote {
+        auth = psk
+    }
+    children {
+        net-net {
+            local_ts  = {{ .Values.podCIDR }},{{ .Values.serviceCIDR }}
+            remote_ts = {{ join "," .Values.peerCIDRs }}
+            start_action = trap
+        }
+    }
```
3u13r, line 17:
```suggestion
      remote_addrs = {{ .Values.ipsec.peer }}
      local {
          auth = psk
      }
      remote {
          auth = psk
      }
      children {
          net-net {
              local_ts  = {{ .Values.podCIDR }},{{ .Values.serviceCIDR }}
              remote_ts = {{ join "," .Values.peerCIDRs }}
              start_action = trap
          }
      }
```

dev-docs/howto/vpn/templates/NOTES.txt
```diff
@@ -0,0 +1,9 @@
+{{- if .Values.ipsec.enabled }}
+Required postinstallation steps (also see README.md):
+
+# Configure the LoadBalancer
+
+1. Find the node hosting the VPN server: 
+   kubectl get pods {{ include "..fullname" . }}-frontend-0 -o jsonpath={.spec.nodeName}
+2. Edit the load balancer resource in GCP and remove all other endpoints.
```
3u13r, line 8: Can we circumvent this by adding `externalTrafficPolicy: Local` to the service (https://cloud.google.com/kubernetes-engine/docs/concepts/service-load-balancer#health_check). This would also improve failover for when we replace the VPN node (e.g. on Constellation upgrades).

3u13r (changes requested):

dev-docs/howto/vpn/templates/service.yaml
burgerdev: I'm considering this resolved, then.

dev-docs/howto/vpn/templates/NOTES.txt
3u13r: fyi I tested this and this seems to work at least on GCP.

burgerdev: Thanks, that sounds like the exact solution I needed!

dev-docs/howto/vpn/templates/statefulset.yaml
burgerdev: Created a follow-up ticket, not going to address it in this PR though.

dev-docs/howto/vpn/templates/statefulset.yaml
burgerdev: Done

3u13r (approved):
LGTM

PR #2580, opened 2023-11-13
state: add migration for `IPCidrNode` and `InClusterEndpoint`

3u13r:
### Context

We added validation for the statefile after I implemented changes which depended on there being no validation (at least before the state is updated/re-crated from the new terrafom).

### Proposed change(s)

- Add a `Migrate` function to the state file to implement all future migrations.
- Add migrations to pass the validation for the new fields.
- We need to also revert https://github.com/edgelesssys/constellation/pull/2545. Otherwise we schedule 30+ nodes during the upgrade of a 1:1 cluster.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/apply.go
+++ b/cli/internal/cmd/apply.go
@@ -337,6 +337,18 @@ The control flow is as follows:
 	                        └────────────────────┘
 */
 func (a *applyCmd) apply(cmd *cobra.Command, configFetcher attestationconfigapi.Fetcher, upgradeDir string) error {
+	// Migrate state file
+	stateFile, err := state.ReadFromFile(a.fileHandler, constants.StateFilename)
+	if err != nil {
+		return fmt.Errorf("reading state file: %w", err)
+	}
+	if err := stateFile.Migrate(); err != nil {
+		return fmt.Errorf("migrating state file: %w", err)
+	}
+	if err := stateFile.WriteToFile(a.fileHandler, constants.StateFilename); err != nil {
+		return fmt.Errorf("writing state file: %w", err)
+	}
+
 	// Validate inputs
 	conf, stateFile, err := a.validateInputs(cmd, configFetcher)
 	if err != nil {
--- a/cli/internal/state/state.go
+++ b/cli/internal/state/state.go
@@ -20,7 +20,9 @@ import (
 
 	"dario.cat/mergo"
 	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
+	"github.com/edgelesssys/constellation/v2/internal/constants"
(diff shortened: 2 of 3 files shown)


msanft (approved):

daniel-weisse:
[e2e upgrade](https://github.com/edgelesssys/constellation/actions/runs/6849031669)

cli/internal/state/state.go
```diff
@@ -555,6 +557,23 @@ func (s *State) Constraints() []*validation.Constraint {
 	return []*validation.Constraint{}
 }
 
+// Migrate migrates the state to the current version.
+// This is mostly done to pass the validation of the current version.
+// The infrastructure will be overwritten by the terraform outputs after the validation.
+func (s *State) Migrate() error {
+	// In v2.13.0 the ClusterEndpoint and InClusterEndpoint fields were added.
+	// So they are expected to be empty when upgrading to this version.
+	if semver.Compare(constants.BinaryVersion().String(), "v2.13.0") != -1 {
```
elchead, line 566: should we add a todo to remove this code in `v2.14`? this would be "wrong" in v2.15

elchead (approved):

cli/internal/state/state.go
3u13r: Yes we should add a TODO. Also I'll add a upper bound to the version check. Then we are "safe" either way. (We are already since the only set the values if they are empty).

elchead (approved):

3u13r:
New e2e upgrade (gcp, 1:1): https://github.com/edgelesssys/constellation/actions/runs/6849532241

cli/internal/cmd/apply.go
```diff
@@ -337,6 +337,18 @@ The control flow is as follows:
...
+	stateFile, err := state.ReadFromFile(a.fileHandler, constants.StateFilename)
+	if err != nil {
+		return fmt.Errorf("reading state file: %w", err)
+	}
+	if err := stateFile.Migrate(); err != nil {
+		return fmt.Errorf("migrating state file: %w", err)
+	}
+	if err := stateFile.WriteToFile(a.fileHandler, constants.StateFilename); err != nil {
+		return fmt.Errorf("writing state file: %w", err)
+	}
```
3u13r, line 350: We always write the state file back, even when we didn't change anything. We could add some "changed" boolean to the `Migrate` function or do the reading and writing inside the migration. The second option could be confusing since one needs to read the statefile to call methods on it.

msanft: I think the current flow (as you've implemented it) is clear. I'd suggest leaving it like that.

PR #2596, opened 2023-11-13
Revert "operator: always delete terminated pending nodes (#2545)"

3u13r:
This reverts commit 5267ad0f08de57e67c2713d8ec45ad2f8041ee3e.

### Proposed change(s)

- We need to revert this. With this patch we've observed 30+ nodes being scheduled during the upgrade of a 1:1 cluster on GCP. This presumably happens because if a node is not found then it's state within the operator is "Terminated". This likely happens in the first seconds of the new node being created. With is patch the "Terminated" node already has reached its goal. It's true goal was joining the cluster though. 

The idea before (and after) this patch is that the joiningNode has a deadline set until it must join the cluster. Otherwise it's goal is set to "Leaving" which is reached if the node never joined the cluster because it was shut down. This should've already handled the AWS case.

### Checklist

- [x] Link to Milestone


malt3:
I agree that reverting is better. However, we should not forget about the AWS bug, where we will now schedule too many nodes

msanft (approved):
approving here while workaround for AWS should be implemented separately

PR #2605, opened 2023-11-15
ci: allow jobs to install tools from pinned nixpkgs

malt3:
### Context

We often need additional tools in GitHub Actions Workflow jobs. This extension to the `setup_bazel_nix` action adds a new parameter `nixTools` which (if provided) will install a list of tools for the duration of the job and add their store paths to the `GITHUB_PATH` env.

### Proposed change(s)

- ci: allow jobs to install tools from pinned nixpkgs

### Additional info

- [test run](https://github.com/edgelesssys/constellation/actions/runs/6891198755/job/18745750123) (prints cowsay and ponysay output)

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/.github/actions/setup_bazel_nix/action.yml
+++ b/.github/actions/setup_bazel_nix/action.yml
@@ -12,6 +12,10 @@ inputs:
   rbePlatform:
     description: "RBE platform to use. If empty, RBE will not be used."
     required: false
+  nixTools:
+    description: "Nix tools to install as list of strings separated by newlines. If empty, no tools will be installed."
+    default: ""
+    required: false
 
 runs:
   using: "composite"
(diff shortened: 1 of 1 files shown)


malt3:
@miampf FYI. This enables you to install tools like zip by specifying `nixTools: '["zip"]'`.

msanft:
This would require us to add the tools in the top-level workflow where the `setup_bazel_nix` action is first called, right. This might add some duplication throughout multiple workflows that all call the same action that requires a specific tool. But I'm not sure yet how to circumvent that the best way.

msanft (approved):
Other than my above comment, changes LGTM

malt3:
> This would require us to add the tools in the top-level workflow where the `setup_bazel_nix` action is first called, right. This might add some duplication throughout multiple workflows that all call the same action that requires a specific tool. But I'm not sure yet how to circumvent that the best way.

Agreed. This is not ideal but probably a good enough approximation. Ideally, any bash scripts used in the CI would be built and run using nix directly and with explicitly list direct dependencies.
Let’s go with this solution until we have a better idea.

.github/actions/setup_bazel_nix/action.yml
```diff
@@ -12,6 +12,10 @@ inputs:
   rbePlatform:
     description: "RBE platform to use. If empty, RBE will not be used."
     required: false
+  nixTools:
+    description: "Nix tools to install as list of strings (e.g `['cowsay' 'ponysay']`). If empty, no tools will be installed."
```
katexochen, line 16: We should simplify the user interface for this:

```yaml
nixTools: |
    cowsay
    ponysay
```

Will give us better diffs and easier use, we can assemble the list in the action using bash.

.github/actions/setup_bazel_nix/action.yml
```diff
@@ -267,3 +271,42 @@ runs:
...
+          --argstr rev "${gitSha}" \
+          --attr tools \
+          "${expressionFile}"
+        EXTRA_PATH=$(nix eval --raw --file "${expressionFile}" \
+          --arg tools "${tools}" \
+          --argstr repository "${repository}" \
+          --argstr rev "${gitSha}" \
+          pathVar)
+        echo "EXTRA_PATH=${EXTRA_PATH}"
+        echo "${EXTRA_PATH}" >> "${GITHUB_PATH}"
```
katexochen, line 311: If a tool isn't installed by us, it will still fall back to the preinstalled tools, right?

katexochen: We should investigate if there is a way to clear the path and ensure only explicitly declared tools are used, but not as part of this PR.

katexochen (approved):

.github/actions/setup_bazel_nix/action.yml
malt3: https://github.com/edgelesssys/constellation/actions/runs/6891198755/job/18745750123

PR #2608, opened 2023-11-16
Support internal load balancer setup on Azure

3u13r:
### Proposed change(s)

- Allow to create a Constellation with internal LB instead of public LB.
- The actual implementation was done in https://github.com/edgelesssys/constellation/pull/2388

### Additional info

- AB#2767

### Checklist

- [x] Link to Milestone


.github/workflows/e2e-test-release.yml
```diff
@@ -176,6 +176,23 @@ jobs:
...
+          - test: "lb"
+            refStream: "ref/main/stream/debug/?"
+            provider: "azure"
+            kubernetes-version: "v1.28"
+            internal-lb: "true"
+          - test: "recover"
+            refStream: "ref/main/stream/debug/?"
+            provider: "aws"
+            kubernetes-version: "v1.28"
+            internal-lb: "true"
```
derpsteb, line 194:
```suggestion
          # internal-lb tests (at least one test on each supported provider)
          - test: "sonobuoy quick"
            refStream: "ref/main/stream/debug/?"
            provider: "gcp"
            kubernetes-version: "v1.28"
            internal-lb: "true"
            runner: "ubuntu-22.04"
          - test: "lb"
            refStream: "ref/main/stream/debug/?"
            provider: "azure"
            kubernetes-version: "v1.28"
            internal-lb: "true"
            runner: "ubuntu-22.04"
          - test: "recover"
            refStream: "ref/main/stream/debug/?"
            provider: "aws"
            kubernetes-version: "v1.28"
            internal-lb: "true"
            runner: "ubuntu-22.04"
```

.github/workflows/e2e-test-weekly.yml
```diff
@@ -260,6 +283,7 @@ jobs:
           selfManagedInfra: ${{ matrix.selfManagedInfra == 'true' }}
           s3AccessKey: ${{ secrets.AWS_ACCESS_KEY_ID_S3PROXY }}
           s3SecretKey: ${{ secrets.AWS_SECRET_ACCESS_KEY_S3PROXY }}
+          internalLoadBalancer: ${{ matrix.internal-lb == 'true' }}
```
derpsteb, line 286:
```suggestion
          internalLoadBalancer: ${{ matrix.internal-lb }}
```

This doesn't work?

3u13r: I just copied what we do for `matrix.selfManagedInfra`. Hence the variable is a string and only truthy. But I don't know if this is equal to your suggestion.

3u13r:
Ah the strict node-to-node routing in combination with Azure internal LB's not doing SNAT breaks this feature for now. I can fix this in our Cilium fork though. We'll need another exception port (as we already do for both etcd ports). 
Converting to draft until then.

PR #2613, opened 2023-11-17
rfc: Terraform provider

msanft:
### Context

We want to determine how the Terraform provider should look in practice. This should not yet discuss implementation details, but rather a high-level overview of what the outcome should be.

### Proposed change(s)

- Add an RFC on how the overall structure of the Terraform provider, mostly user-facing but also on some high-level implementation side, should look like.

### Additional info

- AB#3541

### Checklist

- [x] Link to Milestone


rfc/terraform-provider.md
```diff
@@ -0,0 +1,92 @@
+# Constellation Terraform Provider
+
+The Constellation Terraform Provider should allow it's user to manage the full lifecycle of a Constellation cluster -- namely creation, updates, and deletion -- via Terraform.
```
elchead, line 3:
```suggestion
The Constellation Terraform Provider allows its user to manage the full lifecycle of a Constellation cluster -- namely creation, updates, and deletion -- via Terraform.
```

rfc/terraform-provider.md
```diff
@@ -0,0 +1,92 @@
...
+    region = "us-central1" # optional, for AWS only
+}
+
+# constellation_image provides:
+# constellation_image.foo.reference (CSP-specific image reference)
+```
+
+The `constellation_cluster` resource is the main resource implemented by the provider.
+It resembles a Constellation cluster with a specific configuration.
+Applying it will create the cluster if not existing *or* if the required updates can't be performed in-place (e.g. changing the master secret),
```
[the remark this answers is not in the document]
malt3: We use the CRUD operations of terraform and the terraform state:

- When starting out, without terraform state, we expect the cluster to be uninitialized, so we perform the init
- The result of init is saved as terraform state
- Terraform will now call the update hook instead, so we know that we don't need to intialize the cluster again

rfc/terraform-provider.md
```diff
@@ -0,0 +1,92 @@
...
+    csp = "GCP"
+    region = "us-central1" # optional, for AWS only
+}
+
+# constellation_image provides:
+# constellation_image.foo.reference (CSP-specific image reference)
+```
+
+The `constellation_cluster` resource is the main resource implemented by the provider.
+It resembles a Constellation cluster with a specific configuration.
```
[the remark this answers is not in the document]
msanft: >  What does resemble mean? Is it not the same as a CLI created cluster?

It is, but I intended to say that we have the Terraform resource declaration on one side, and the actual cloud resources & data on the other side.

> What is the behavior of recreate?

Recreate means destroy and create, performed in that order. (i.e. you get a new cluster)

> Is it possible to perform a new init on an initialized cluster?

No

rfc/terraform-provider.md
```diff
@@ -0,0 +1,92 @@
...
+Applying it will create the cluster if not existing *or* if the required updates can't be performed in-place (e.g. changing the master secret),
+updates it with the according configuration if already existing, or deletes it if not present in the configuration but in the state.
```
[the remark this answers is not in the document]
msanft: I don't understand the question. What do you mean exactly? The "deletes it if not" means that if the cluster is existing (in the Terraform state), but not declared in the config, it will be destroyed. (i.e. the normal Terraform behaviour, diffing state against declarations)

rfc/terraform-provider.md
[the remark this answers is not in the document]
elchead: How would we enforce that the infra is recreated when the master secret changes? Atm, the infra is only related to the secret through a tag (only checked AWS). I don't think this would force a recreate. So I understand that we would design the infra module to force a recreate when it changes?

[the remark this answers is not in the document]
elchead: > It is, but I intended to say that we have the Terraform resource declaration on one side, and the actual cloud resources & data on the other side.

I would reword the sentence then. "The resource declares the provisioning / management of a Constellation cluster on the provided infrastructure."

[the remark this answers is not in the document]
msanft: > How would we enforce that the infra is recreated when the master secret changes?

I think we should throw an error [like so](https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/force-new#framework). Changing the underlying infrastructure is not necessary, the disks of the machines can just be overwritten with new images, leaving them in a "fresh" state. The only thing we need to warn the user about are previously created persistent volumes, which are encrypted with a secret derived from the old mastersecret and thus cannot be accessed with the new mastersecret.

rfc/terraform-provider.md
```diff
@@ -0,0 +1,92 @@
...
+    kubernetes_api_endpoint = "10.10.10.10" # should use public ip of LB resource, ideally also provisioned through TF
+    extra_microservices = {
+        csi_driver = true
+        # + more
+        # possiblly also constellation microservices with version and maybe service options,
+        # which would make constellation_microservice_version obsolete.
+        # exact API TBD
+    }
+    master_secret = "foo" # updating this would force recreation of the cluster
+    init_secret = "bar" # maybe derive from master_secret, updating this would force recreation of the cluster
```
burgerdev, line 46: These secrets end up in the Terraform state, right? I believe we should mention that in this RFC.

rfc/terraform-provider.md
```diff
@@ -0,0 +1,92 @@
...
+    network_config = {
+        # TBD
+        # should contain CIDRs for pod network, service cidr, node network... for Cilium
+        # and in-cluster Kubernetes API endpoint, e.g. for Kubelets
+    }
```
burgerdev, line 51: Are users expected to fill this in, or can this be derived from the cloud provider name and the state of the cloud infrastructure?

rfc/terraform-provider.md
```diff
@@ -0,0 +1,92 @@
...
+
+resource "constellation_cluster" "foo" {
+    uid = "bar"
+    name = "baz"
+    image = data.constellation_image.ref # or provide manually crafted values
+    kubernetes_version = "v1.27.6"
+    constellation_microservice_version = "lockstep" # or pinned to a specific version
+    debug = false
+    init_endpoint = "10.10.10.10" # should use public ip of LB resource, ideally also provisioned through TF
+    kubernetes_api_endpoint = "10.10.10.10" # should use public ip of LB resource, ideally also provisioned through TF
```
burgerdev, line 37: Are these the only pieces of information that we need from the cloud infrastructure setup?

rfc/terraform-provider.md
```diff
@@ -0,0 +1,92 @@
...
+    network_config = {
+        # TBD
+        # should contain CIDRs for pod network, service cidr, node network... for Cilium
+        # and in-cluster Kubernetes API endpoint, e.g. for Kubelets
+    }
+    attestation = data.constellation_attestation.attestation # or provide manually crafted values
+}
+
+# constellation_cluster provides:
+# constellation_cluster.foo.kubeconfig
```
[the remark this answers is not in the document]
elchead: I think the kubeconfig would just be stored in the output after init

rfc/terraform-provider.md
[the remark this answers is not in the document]
elchead: Pls add this to the RFC.

rfc/terraform-provider.md
```diff
@@ -0,0 +1,92 @@
...
+It declares a Constellation cluster with a specific configuration.
+Applying it will create the cluster if not existing, upgrade the cluster when the changes can be performed in place (e.g. K8s / node image / microservice update) *or* recreate the resource when the update can't be performed in-place (e.g. changing the master secret).
+updates it with the according configuration if already existing, or deletes it if not present in the configuration but in the state.
+
+The "constellation_attestation" and "constellation_image" objects are [data sources](https://developer.hashicorp.com/terraform/language/data-sources),
+which are objects that should be evaluated by the Provider each time the state is refreshed (i.e. each time any Terraform command that evaluates configuration against state),
+but have no observable side effects. For image and attestation, this is required as the provider need to evaluate `latest` values or map CSP-agnostic image references (e.g. `v2.13.0`)
+to CSP-specific image references (e.g. `/CommunityGalleries/.../Image` for Azure). This is implemented as an idempotent API call and thus has no observable side-effects, but needs
```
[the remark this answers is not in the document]
msanft: These are idempotent. We host an API that transforms CSP-agnostic image references (e.g. `v2.13.0`, `ref/main/stream/nightly/v2.13.0-pre.0.20231019162807-6c0a3b8efa49`) into CSP-specific image references (e.g. AMI-ID for AWS, Community Gallery ID for Azure, ...)

rfc/terraform-provider.md
msanft: Correct.

rfc/terraform-provider.md
msanft: I don't know for sure yet, @3u13r may have more information here?

rfc/terraform-provider.md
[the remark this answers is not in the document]
msanft: This would be the case if we tried to support importing resources. I don't know if we need to or if we want to yet, but in that case, we would have to make the Kubeconfig an input or find another way to pass it when importing the resource. For the init-case, we can make it so that the resource is only stored on init, and just returned from the state on any other operation. We cannot trivially recover the Kubeconfig with the mastersecret.

rfc/terraform-provider.md
```diff
@@ -0,0 +1,97 @@
...
+
+The Constellation Terraform Provider allows its user to manage the full lifecycle of a Constellation cluster -- namely creation, updates, and deletion -- via Terraform.
+
+## Design Goals
+
+- The User needs to be able to perform Creation, Upgrades and Deletion of a Cluster through the provider.
+- Infrastructure provisioning should explicitly **not** be performed by the Constellation Terraform provider.
+- The User needs to receive some sort of Access Token to the Cluster (e.g. a Kubeconfig, whether long- or short-lived)
+when applying a configuration containing the provider and the resources listed below.
+
```
elchead, line 11: How would we include `status` information such as image / k8s upgrades? The user could manually fetch these with the CLI for troubleshooting cases? Same for `recovery`?

elchead (approved):

elchead (approved):

rfc/terraform-provider.md
msanft: The goal of the provider is to provide management and provisioning (i.e. create, update, delete) capabilities imo, not interaction with the cluster itself, like the CLI does. I don't think we need to achieve feature-parity with the CLI. The Terraform model would also not make it very easy to implement such without some hacks.

elchead: I agree

rfc/terraform-provider.md
3u13r: We can get everything from the terraform as we also do today.

3u13r (changes requested):

rfc/terraform-provider.md
```diff
@@ -0,0 +1,97 @@
+# Constellation Terraform Provider
+
+The Constellation Terraform Provider allows its user to manage the full lifecycle of a Constellation cluster -- namely creation, updates, and deletion -- via Terraform.
+
+## Design Goals
+
+- The User needs to be able to perform Creation, Upgrades and Deletion of a Cluster through the provider.
+- Infrastructure provisioning should explicitly **not** be performed by the Constellation Terraform provider.
```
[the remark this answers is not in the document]
msanft: Creation is better worded as "initialization" here, I guess.

rfc/terraform-provider.md
burgerdev: But the values in this resource are the only thing that our provider gets to see, correct? Would we need to reference other Terraform resources, like

```
  network_config {
    flavor = "gcp"
    podCIDR = other_resource.vpc.cidr
    # ...
  }
```

rfc/terraform-provider.md
burgerdev: Thanks, but I'm slightly confused now looking at https://github.com/edgelesssys/constellation/pull/2613#discussion_r1401835204. Resolving this here and let's continue this topic over there.

rfc/terraform-provider.md
msanft: The latter, I think: https://github.com/edgelesssys/constellation/blob/f6c44ee5cc6f1d097c0b2096f6a8e289e2210730/terraform/infrastructure/gcp/outputs.tf#L35-L37

3u13r: We don't have to reference the other resource directly but there should be a direct link between the output of the vpc and the input of the Constellation provider (as Markus mentioned).

msanft: Is this something that needs to be mentioned in the RFC? IMO, it should be agnostic about where the value comes from (output of another resource or directly supplied)

burgerdev: Only if the goal is to replace the TBD with an explicit list, which I don't think is needed. Maybe we could mention that some fields here need to be populated by other resources' outputs, but maybe that is obvious.

rfc/terraform-provider.md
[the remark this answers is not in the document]
burgerdev: Got it: the kubeconfig is persisted in TF state, because it is part of the output, and this is where the provider gets it from for upgrades etc.

daniel-weisse (approved):

3u13r (approved):
LGTM

burgerdev (approved):
lgtm

malt3 (approved):

rfc/terraform-provider.md
```diff
@@ -0,0 +1,99 @@
...
+    kubernetes_api_endpoint = "10.10.10.10" # should use public ip of LB resource, ideally also provisioned through TF
+    extra_microservices = {
+        csi_driver = true
+        # + more
+        # possiblly also constellation microservices with version and maybe service options,
+        # which would make constellation_microservice_version obsolete.
+        # exact API TBD
+    }
+    master_secret = "foo" # updating this would force recreation of the cluster
+    init_secret = "bar" # maybe derive from master_secret, updating this would force recreation of the cluster
```
thomasten, line 46: Do I understand correctly that the user generates these secrets and pastes them into this TF config?
If so, is it common to have such secrets in a TF config? Or should they come from input variables, env variables, etc.?

msanft: You could use the Terraform `random` provider (what we currently do) to generate this, load them from your HashiCorp vault by using the corresponding provider, etc.

I don't think you would expose them in the configuration files usually, but only in the state, which can be stored remotely, etc. and only have a reference to it, e.g. by using the `random` provider, in the configuration files - But the RFC should be agnostic of where the value comes from actually. When documenting our provider we could mention some recommendations of how secrets can be provisioned.

thomasten (approved):

PR #2619, opened 2023-11-20
dev-docs: add on-prem terraform to vpn setup

3u13r:
### Proposed change(s)

- Move the helm files to a helm/ subdir
- Add fake on-prem architecture in terraform

### Additional info

I tested the connectivity and it works like a charm with your helm files.

### Checklist

- [x] Link to Milestone


dev-docs/howto/vpn/on-prem-terraform/main.tf
```diff
@@ -0,0 +1,286 @@
...
+  network_security_group_id = azurerm_network_security_group.security_group.id
+}
+
+# Create virtual machine
+resource "azurerm_linux_virtual_machine" "public_vm" {
+  name                  = "public_vm"
+  location              = azurerm_resource_group.rg.location
+  resource_group_name   = azurerm_resource_group.rg.name
+  network_interface_ids = [azurerm_network_interface.public_nic.id]
+  size                  = "Standard_D2as_v5"
```
burgerdev, line 147: Nit: maybe the `Bs` series is sufficient for the purpose of this deployment.

dev-docs/howto/vpn/on-prem-terraform/main.tf
```diff
@@ -0,0 +1,286 @@
...
+      default = 1
+      ike = 2
+      flush_line = yes
+    }
+  }
+}
+EOT
+
+
+cat <<EOT >> /etc/swanctl/conf.d/constellation.conf
```
burgerdev, line 196: If the secret contains a `$` character it might get interpolated by the shell. I'd suggest quoting the heredoc marker to prevent that, for safety.

dev-docs/howto/vpn/on-prem-terraform/main.tf
```diff
@@ -0,0 +1,286 @@
...
+
+cat <<EOT >> /entrypoint.sh
+#!/bin/sh
+
+while ! swanctl --stats > /dev/null 2> /dev/null; do
+  sleep 1
+done
+swanctl --load-all
+
+wait
```
burgerdev, line 233: This looks like you're starting Charon via systemd. In that case, you can remove the logging configuration and consume from `journalctl -u strongswan`.

burgerdev: Actually, I think we'd not need to do any `swanctl` if we were using the systemd variant.

dev-docs/howto/vpn/on-prem-terraform/main.tf
3u13r: I currently get a 
```
        "message": "Allocation failed. We do not have sufficient capacity for the requested VM size in this zone. Read more about improving likelihood of allocation success at http://aka.ms/allocation-guidance"
```
when using `Standard_B2ts_v2` in westeurope. Let's see which similar instance types work.

dev-docs/howto/vpn/on-prem-terraform/main.tf
3u13r: There is no output for `journalctl -u strongswan`. I also don't install the `charon-systemd` package. While I tried for a few minutes to use the systemd variant, I gave up when this variant worked.

burgerdev: That makes perfect sense. But where is Charon being started here, I wonder?

3u13r: I'd imagine that it is stared when loading the config via swanctl.

burgerdev (approved):
LGTM

Do you want to merge into my feature branch, or into main?

3u13r:
> Do you want to merge into my feature branch, or into main?

I think you can merge into main and then I rebase + cleanup and merge into main also.

PR #2623, opened 2023-11-21
cli: move `cli/internal` libraries

msanft:
### Context

We want to transition to a Constellation library that can be consumed by both the CLI and the Terraform provider. The CLI should only contain code that directly interacts with the user and should consume everything else through the library's API. A first step towards this is moving the packages that do not fall under this category from `cli/internal` to `internal`.

### Proposed change(s)

- Move all packages from `cli/internal` to `internal` that will also be used by the Terraform provider.
- Create the `internal/constellation` package.

### Additional info

- [AB#3544](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3544)
- [E2E Test](https://github.com/edgelesssys/constellation/actions)

### Checklist

- [x] Link to Milestone


internal/featureset/BUILD.bazel
daniel-weisse, line 1: I feel like this could also be kept in `cli/internal/`, but there might be cases where the Terraform provider could also use it, so I'm fine with moving it here for now.

internal/libvirt/BUILD.bazel
daniel-weisse, line 1: Should be kept in `cli/internal`, since the provider wont use this

internal/terraform/BUILD.bazel
daniel-weisse, line 1: Should be kept in `cli/internal`, as the provider wont provision infrastructure using Terraform itself (it should expect the infrastructure to already exist)

daniel-weisse (commented):
Annotated some files of packages I think should be kept in `cli/internal`.

While we are already in the process of moving files, should we consider moving the Helm charts to their own (more globally visible) directory similar to the Terraform files?
I feel that it can sometimes be a bit of hassle to have them deeply nested somewhere in our Go helm code.

daniel-weisse (approved):

elchead (approved):

3u13r (approved):
LGTM

PR #2625, opened 2023-11-21
cli: fix panic in status cmd without conf file

elchead:
### Context

When executing `constellation status` without a present config file, the command panics because the config reading error is not caught.
This issue occurs on 2.13 and also 2.12.

### Proposed change(s)

- return config read error

### Checklist

- [ ] Link to Milestone


elchead:
I think a bug fix release is not necessary. We should create a bug issue though.

daniel-weisse (approved):

PR #2629, opened 2023-11-22
ci: use different GCP project

daniel-weisse:
### Context
We want to use a separate project for running e2e tests on GCP

### Proposed change(s)

- Update service accounts used for infrastructure creation and IAM creation
- Remove unused `gcpInClusterServiceAccountKey` account from `.github/actions/e2e_test/action.yml`
  - This secret should also be deleted from this repo
- Update project used for deploying e2e tests
  - We currently have this saved as a secret, making updating it rather difficult without breaking things
  - For now, I just hard coded the new project. We can either keep it like this, or change back to using a secret and update the secret in this repo before merging
- Update dev-docs

### Additional info

- AB#3552


daniel-weisse:
self-managed resources seem to not be cleaned up properly, but this should unrelated to this PR
Will investigate

malt3 (approved):

PR #2631, opened 2023-11-22
ci: prepare Bazel for integration testing

burgerdev:
### Context

Our integration tests are currently executed with CMake, but we would like to consolidate everything to Bazel.

### Proposed change(s)

Integration test Bazel targets are tagged with `integration`. Tests tagged with `intregration` are not run in the standard configuration, but we introduce a new Bazel config (`integration`) which also includes these tests. For compatibility with legacy `go test`, the new config also applies the `integration` gotag. A second config (`integration-only`) can be used if only integration tests should be run.

This PR also migrates a first integration test, `//internal/license:license_integration_test`, to the new format.

### Additional info

- [AB#3067](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3067)

### Checklist

### DIFF
--- a/internal/license/integration/BUILD.bazel
+++ b/internal/license/integration/BUILD.bazel
@@ -4,6 +4,7 @@ go_test(
     name = "integration_test",
     srcs = ["license_integration_test.go"],
     tags = [
+        "integration",
         "requires-network",
     ],
     deps = [
--- a/internal/license/integration/BUILD.bazel
+++ b/internal/license/integration/BUILD.bazel
@@ -1,0 +1,13 @@
+load("//bazel/go:go_test.bzl", "go_test")
+
+go_test(
+    name = "integration_test",
+    srcs = ["license_integration_test.go"],
+    tags = [
+        "requires-network",
+    ],
+    deps = [
+        "//internal/license",
+        "@com_github_stretchr_testify//assert",
+    ],
+)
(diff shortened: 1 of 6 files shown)


internal/license/integration/BUILD.bazel
```diff
@@ -0,0 +1,14 @@
...
+    srcs = ["license_integration_test.go"],
+    tags = [
+        "integration",
+        "requires-network",
+    ],
+    deps = [
+        "//internal/license",
+        "@com_github_stretchr_testify//assert",
+    ],
+)
```
malt3, line 14:
```suggestion
go_test(
    name = "integration_test",
    srcs = ["license_integration_test.go"],
    gotags = ["integration"],
    tags = [
        "integration",
        "requires-network",
    ],
    deps = [
        "//internal/license",
        "@com_github_stretchr_testify//assert",
    ],
)
```

We need to tell rules_go to set the go build tag for this test. Alternatively, we could set this globally in the `.bazelrc`, but I'm not sure if this would break the other unfinished integration tests.

burgerdev: The idea was to implicitly set the gotag when the integration config is used. That way, I'm not forced to keep tags and gotags for individual targets at sync. Now, if don't want to use the integration gotag in future tests, I agree that it makes sense to list it here. But then the question is if we need it in the first place: if it's only there to exclude the test from `blaze test //...`, we can just remove it in this commit.

malt3: Oh. I totally missed the gotags in the `integration` and `integration-only` configs in bazelrc.
Your solution is better.

malt3 (approved):

thomasten (approved):

PR #2633, opened 2023-11-22
Fix Konnectivity migration

3u13r:
### Proposed change(s)

- Allow Cilium to jump minor versions from v1.12 to v1.15
- If a nodeversion upgrade takes place (from v2.13.x) then remove all the konnectivity setting from the kubeadm config map. Otherwise the KubeAPI pods try to mount e.g. the Konnectivity UDS which we no longer provider through the bootstrapper

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/kubecmd/kubecmd.go
+++ b/internal/kubecmd/kubecmd.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"slices"
 	"sort"
 	"strings"
 	"time"
@@ -128,6 +129,12 @@ func (k *KubeCmd) UpgradeNodeVersion(ctx context.Context, conf *config.Config, f
 		case err != nil:
 			return fmt.Errorf("updating image version: %w", err)
 		}
+
+		// TODO(3u13r): remove `reconcileKubeadmConfigMap` after v2.14.0 has been released.
+		if err := k.reconcileKubeadmConfigMap(ctx); err != nil {
+			return fmt.Errorf("reconciling kubeadm config: %w", err)
+		}
+
 		k.log.Debugf("Updating local copy of nodeVersion image version from %s to %s", nodeVersion.Spec.ImageVersion, imageVersion.Version())
 		nodeVersion.Spec.ImageReference = imageReference
 		nodeVersion.Spec.ImageVersion = imageVersion.Version()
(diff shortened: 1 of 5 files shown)


internal/kubecmd/kubecmd.go
```diff
@@ -393,6 +399,39 @@ func (k *KubeCmd) applyNodeVersion(ctx context.Context, nodeVersion updatev1alph
 	return updatedNodeVersion, err
 }
 
+func (k *KubeCmd) reconcileKubeadmConfigMap(ctx context.Context) error {
```
daniel-weisse, line 402: We should save the kubeconfig to disk after running this function, otherwise we will have to keep it for as long as there are clusters from before v2.14 still running.

internal/kubecmd/kubecmd.go
```diff
@@ -393,6 +399,39 @@ func (k *KubeCmd) applyNodeVersion(ctx context.Context, nodeVersion updatev1alph
...
+		return fmt.Errorf("marshaling ClusterConfiguration: %w", err)
+	}
+
+	kubeadmConfig.Data[constants.ClusterConfigurationKey] = string(newConfigYAML)
+	k.log.Debugf("Triggering kubeadm config update now")
+	if _, err = k.kubectl.UpdateConfigMap(ctx, kubeadmConfig); err != nil {
+		return fmt.Errorf("setting new kubeadm config: %w", err)
+	}
+
+	fmt.Fprintln(k.outWriter, "Successfully reconciled the cluster's kubeadm config")
```
daniel-weisse, line 431: nit: only run this part if there actually were changes to the config, i.e. konnectivity resources were deleted

internal/kubecmd/kubecmd.go
3u13r: I don't fully understand your point. I'm changing the Kubeadm ConfigMap here. We can remove this function after we've released v2.14 since our upgrade path states that the user must go through every Constellation minor release.

daniel-weisse: Ahhh, I thought we were updating the config of the user, not the config present in the cluster.
In that case its fine

internal/kubecmd/kubecmd.go
3u13r: There should always be changes here since this function is only called if the image is a valid upgrade (see where this is called inside `UpgradeNodeVersion`. Also, since this will land in v2.14.0 every Constellation still has the entries that we want to delete.

3u13r: On second thought this is a nice optimization for when we release patch versions for v2.14

daniel-weisse (approved):

PR #2637, opened 2023-11-23
ci: correctly clean up resource in self-managed infra tests

daniel-weisse:
### Context
Resources were not cleaned up correctly since the destroy action was not configured for self-managed infrastructure

### Proposed change(s)

- Correctly clean up resources

### Related issue
- closes https://github.com/edgelesssys/issues/issues/79

### Additional info

- [e2e test](https://github.com/edgelesssys/constellation/actions/runs/6968930299)


msanft (approved):
LGTM, thanks for the fix!

PR #2638, opened 2023-11-23
rfc: add numeric ids to existing RFCs

burgerdev:
### Context

Now that we have a growing number of RFCs, having numeric identifiers for them makes it easier to refer to them and allows an intuitive understanding of the history of changes, without resorting to `git blame`.

### Proposed change(s)

Assign a numeric ID to all existing RFCs. IDs are handed out incrementally, starting from 001, to the RFCs sorted by first appearance in the repo. 999 RFC IDs should be enough for anybody.

This change was generated by the following script:

```bash
cd rfc

birthday() {
    git log --reverse --format=%at -- "$1" | head -n1
}

order() {
  while read -r line
  do
      echo "$(birthday $line) $line"
  done | sort -n | cut -f 2 -d' '
}

n=1
find . -type f | order | \
    while read -r line
    do
        file="$(basename "$line")"
        sed -i "1s/# /# RFC $(printf "%03d" $n): /" "$file"
        git mv "$file" "$(printf "%03d" $n)-$file"
        n=$((n+1))
    done
```

### Checklist

### DIFF
--- a/rfc/001-key-derivation.md
+++ b/rfc/001-key-derivation.md
@@ -1,4 +1,4 @@
-# Key derivation
+# RFC 001: Key derivation
 
 To derive disk-encryption keys and other secret material in a Constellation Kubernetes cluster, we use [HKDF](https://datatracker.ietf.org/doc/html/rfc5869).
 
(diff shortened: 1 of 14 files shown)


thomasten (approved):
Thanks! LGTM

PR #2640, opened 2023-11-23
terraform-provider: add attestation data source

elchead:
### Context

Part of #2613 to make attestation data available for the constellation cluster resource.

### Proposed change(s)

- introduce new VerifyFetcher which is shared between the CLI and the TF provider
- new TF data source

### Additional info

- [AB#3548](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3548)

### Checklist

- [x] Link to Milestone


elchead:
Align with #2601 to support AWS-SEV-SNP already.

daniel-weisse:
Can you rebase this to point to the head of https://github.com/edgelesssys/constellation/tree/feat/terraform-provider/image-datasource ?
Currently, this PR contains some commits from main which shouldnt be in here

terraform-provider-constellation/internal/provider/attestation_data_source.go
```diff
@@ -0,0 +1,277 @@
...
+		if resp.Diagnostics.HasError() {
+			return
+		}
+	}
+	// TODO align with reenable AWS PR
+	if attestationVariant.Equal(variant.AWSSEVSNP{}) {
+		resp.Diagnostics.AddWarning("AWS attestation not supported", "AWS is not supported yet")
+	}
+
+	verifyFetcher, err := measurements.NewVerifyFetcher(d.client)
```
daniel-weisse, line 198: `NewVerifyFetcher` hard codes `sigstore.NewCosignVerifier` as its verifier, meaning we can't test this function.
We should use the same constructor as is used in the `fetchmeasurements` CLI command instead, which allows setting a custom verifier.
The verifier should be set up in `*AttestationDataSource.Configure`, same as the HTTP client already is.

internal/attestation/measurements/fetchmeasurements.go
daniel-weisse, line 1: Can you move the unit tests for the VerifyFetcher from `cli/internal/cmd/configfetchmeasurements_test.go` to here?

terraform-provider-constellation/internal/provider/attestation_data_source.go
```diff
@@ -0,0 +1,277 @@
...
+	if err != nil {
+		resp.Diagnostics.AddError("creating verify fetcher", fmt.Sprintf("Error creating verify fetcher: %s", err))
+		return
+	}
+	fetchedMeasurements, err := verifyFetcher.FetchAndVerifyMeasurements(ctx, data.ImageVersion.ValueString(), csp, attestationVariant)
+	if err != nil {
+		if _, ok := err.(measurements.ErrRekor); ok {
```
daniel-weisse, line 205: Should use proper error checking with `errors.Is` or `errors.As`

daniel-weisse (changes requested):

cli/internal/cmd/configfetchmeasurements.go
```diff
@@ -132,58 +131,15 @@ func (cfm *configFetchMeasurementsCmd) configFetchMeasurements(
...
-			imageVersion,
-			conf.GetProvider(),
-			conf.GetAttestationConfig().GetVariant(),
-		)
-		if err != nil {
-			return fmt.Errorf("fetching and verifying measurements: %w", err)
-		}
-		cfm.log.Debugf("Fetched and verified measurements, hash is %s", hash)
-		if err := sigstore.VerifyWithRekor(cmd.Context(), publicKey, rekor, hash); err != nil {
+		if _, ok := err.(measurements.ErrRekor); ok {
```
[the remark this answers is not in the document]
elchead: I tried, but it didn't work. It only works on wrapped errors. But I can make it a wrapped error. I also added a test case to cover this logic.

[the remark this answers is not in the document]
elchead: done

terraform-provider-constellation/internal/provider/attestation_data_source.go
elchead: these constructors can err however, and configure doesn't return an Error. 
We only have acceptance tests for the data sources at any rate, so we don't need to mock. The VerifyFetcher is testable through the custom constructor.
What do you think?

malt3:
Could you clean this up so the only change left is 0b229a078585bd1279240f19f291f27653e0d55e?

If rebasing this is cumbersome, you could try:

```
git checkout main
git pull
git checkout -b feat/terraform/cleaned-data-attestation
git cherry-pick 0b229a078585bd1279240f19f291f27653e0d55e
git push origin --force feat/terraform/cleaned-data-attestation:feat/terraform/data-attestation
```

elchead:
> Could you clean this up so the only change left is [0b229a0](https://github.com/edgelesssys/constellation/commit/0b229a078585bd1279240f19f291f27653e0d55e)?
> 
> If rebasing this is cumbersome, you could try:
> 
> ```
> git checkout main
> git pull
> git checkout -b feat/terraform/cleaned-data-attestation
> git cherry-pick 0b229a078585bd1279240f19f291f27653e0d55e
> git push origin --force feat/terraform/cleaned-data-attestation:feat/terraform/data-attestation
> ```

I'm incorporating changes and will rebase later. Will ping you then :)

cli/internal/cmd/configfetchmeasurements.go
```diff
@@ -90,19 +99,21 @@ func runConfigFetchMeasurements(cmd *cobra.Command, _ []string) error {
...
 	fetcher := attestationconfigapi.NewFetcherWithClient(http.DefaultClient, constants.CDNRepositoryURL)
-	return cfm.configFetchMeasurements(cmd, sigstore.NewCosignVerifier, rekor, fileHandler, fetcher, http.DefaultClient)
+	return cfm.configFetchMeasurements(cmd, fileHandler, fetcher)
 }
 
 func (cfm *configFetchMeasurementsCmd) configFetchMeasurements(
-	cmd *cobra.Command, newCosignVerifier cosignVerifierConstructor, rekor rekorVerifier,
-	fileHandler file.Handler, fetcher attestationconfigapi.Fetcher, client *http.Client,
+	cmd *cobra.Command,
+	fileHandler file.Handler, fetcher attestationconfigapi.Fetcher,
```
daniel-weisse, line 116:
```suggestion
	cmd *cobra.Command, fileHandler file.Handler, fetcher attestationconfigapi.Fetcher,
```
nit: not need to keep this on 3 lines

terraform-provider-constellation/internal/provider/provider.go
```diff
@@ -71,7 +71,7 @@ func (p *ConstellationProvider) Configure(ctx context.Context, req provider.Conf
 	config := datastruct.ProviderData{}
 
 	// Make the clients available during data source and resource "Configure" methods.
-	resp.DataSourceData = config
+	// resp.DataSourceData = config // TODO check with moritz
```
msanft, line 74: I think it doesn't have any disadvantage to have this set all the time, does it? Don't have any real opinion on this though

terraform-provider-constellation/internal/provider/provider_test.go
```diff
@@ -34,6 +34,10 @@ var testAccProtoV6ProviderFactories = map[string]func() (tfprotov6.ProviderServe
 // bazelSetTerraformBinaryPath sets the path to the Terraform binary for
 // acceptance testing when running under Bazel.
 func bazelSetTerraformBinaryPath(t *testing.T) {
+	if !runsUnderBazel() {
+		return
+	}
+
```
msanft, line 40:
```suggestion
```

We do this check below already:
https://github.com/edgelesssys/constellation/blob/97aea98e7771d916a234707b03ef6b8df9427e3f/terraform-provider-constellation/internal/provider/provider_test.go#L41-L50

cli/internal/cmd/configfetchmeasurements_test.go
```diff
@@ -154,115 +151,13 @@ func newTestClient(fn roundTripFunc) *http.Client {
 }
 
 func TestConfigFetchMeasurements(t *testing.T) {
```
daniel-weisse, line 153: Can you add test cases to check:
* Errors other than Rekor errors are not skipped
* No error actually passes correctly

msanft (changes requested):
Changes look good to me apart from the comments and nits I made so far. Didn't look too much into the Rekor changes, as I'm quite unfamiliar with what they should do. I think we're on a good path here!

internal/attestation/measurements/fetchmeasurements.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/edgelesssys/constellation/v2/internal/api/versionsapi"
+	"github.com/edgelesssys/constellation/v2/internal/attestation/variant"
+	"github.com/edgelesssys/constellation/v2/internal/cloud/cloudprovider"
+	"github.com/edgelesssys/constellation/v2/internal/sigstore"
+	"github.com/edgelesssys/constellation/v2/internal/sigstore/keyselect"
+)
+
+// ErrRekor is returned when verifying measurements with Rekor fails.
+var ErrRekor = errors.New("verifying measurements with Rekor")
```
daniel-weisse, line 23:
```suggestion
// RekorError is returned when verifying measurements with Rekor fails.
type RekorError struct {
	err error
}

// Error returns the error message
func (e *RekorError) Error() string {
	return fmt.Sprintf("verifying measurements with Rekor failed: %s", e.err)
}

// Unwrap returns the wrapped error
func (e *RekorError) Unwrap() error {
	return e.err
}
```

From your code it seems like this is what you actually want to use.
To create the error:
```Go
if err := sigstore.VerifyWithRekor(ctx, publicKey, m.rekor, hash); err != nil {
	return nil, &RekorError{err)
}
```

To check for the error type, e.g. in `cli/internal/cmd/configfetchmeasurements.go`:

```Go
_, err := verifier.FetchAndVerifyMeasurements(...)
var rekErr *RekorError
if errors.As(err, &rekErr) {
	// handle error
}
```

internal/attestation/measurements/fetchmeasurements_test.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+		if req.URL.Path == "/constellation/v2/ref/-/stream/stable/v999.999.999/image/measurements.json.sig" {
+			return &http.Response{
+				StatusCode: http.StatusOK,
+				Body:       io.NopCloser(bytes.NewBufferString(signature)),
+				Header:     make(http.Header),
+			}
+		}
+
+		fmt.Println("unexpected request", req.URL.String())
```
daniel-weisse, line 85:
```suggestion
```
debug statement?

internal/attestation/measurements/measurements_test.go
```diff
@@ -1043,3 +1045,8 @@ func TestMeasurementsCompare(t *testing.T) {
 		})
 	}
 }
+
+func TestRekorErrCheck(t *testing.T) {
+	err := fmt.Errorf("%w: %w", ErrRekor, errors.New("test"))
+	assert.ErrorIs(t, err, ErrRekor)
+}
```
daniel-weisse, line 1052:
```suggestion
```
Should be in `internal/attestation/measurements/fetchmeasurements_test.go`, if at all

cli/internal/cmd/configfetchmeasurements.go
```diff
@@ -90,19 +99,21 @@ func runConfigFetchMeasurements(cmd *cobra.Command, _ []string) error {
 	if err != nil {
 		return fmt.Errorf("constructing Rekor client: %w", err)
 	}
-	cfm := &configFetchMeasurementsCmd{log: log, canFetchMeasurements: featureset.CanFetchMeasurements}
+
+	verifyFetcher := measurements.NewVerifyFetcher(sigstore.NewCosignVerifier, rekor, http.DefaultClient)
```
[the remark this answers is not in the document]
daniel-weisse: Could you elaborate on the issue?
http Clients are meant to be reusable, so this should not be a problem.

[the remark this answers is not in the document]
msanft: If we, or some library that we use, manipulates the default client at any point, we would run into problems: https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/net/http/client.go;l=110

terraform-provider-constellation/internal/provider/attestation_data_source.go
```diff
@@ -0,0 +1,290 @@
...
+
+	fetchedMeasurements, err := verifyFetcher.FetchAndVerifyMeasurements(ctx, data.ImageVersion.ValueString(),
+		csp, attestationVariant, false)
+	if err != nil {
+		if errors.Is(err, measurements.ErrRekor) {
+			resp.Diagnostics.AddWarning("Ignoring Rekor related error", err.Error())
+		} else {
+			resp.Diagnostics.AddError("fetching and verifying measurements", err.Error())
+			return
+		}
```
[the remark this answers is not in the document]
msanft: True, sorry

terraform-provider-constellation/internal/provider/attestation_data_source.go
```diff
@@ -0,0 +1,301 @@
...
+			fmt.Sprintf("Unknown Attestation Variant: %s", data.AttestationVariant.ValueString()))
+		return
+	}
+	if attestationVariant.Equal(variant.AzureSEVSNP{}) || attestationVariant.Equal(variant.AWSSEVSNP{}) {
+		snpVersions, err := d.fetcher.FetchSEVSNPVersionLatest(ctx, attestationVariant)
+		if err != nil {
+			resp.Diagnostics.AddError("Fetching SNP Version numbers", err.Error())
+			return
+		}
+		tfSnpVersions := convertSNPAttestationTfStateCompatible(resp, attestationVariant, snpVersions)
```
msanft, line 200: I don't think we should pass the resp object and modify it in the utility function here. IMO, we should rather return val, err from it and then add the error to the diagnostics in here, staying consistent with the other functions and removing side-effects from the `convertSNPAttestationTfStateCompatible` function.

terraform-provider-constellation/internal/provider/attestation_data_source.go
```diff
@@ -0,0 +1,290 @@
...
+		diags := resp.State.SetAttribute(ctx, path.Root("attestation"), tfSnpVersions)
```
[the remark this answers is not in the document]
elchead: agreed to keep as ist

daniel-weisse (approved):

terraform-provider-constellation/internal/provider/attestation_data_source.go
```diff
@@ -0,0 +1,302 @@
...
+		resp.Diagnostics.AddAttributeError(
+			path.Root("csp"),
+			"Invalid CSP",
+			fmt.Sprintf("Invalid CSP: %s", data.CSP.ValueString()),
+		)
+		return
+	}
+	attestationVariant, err := variant.FromString(data.AttestationVariant.ValueString())
+	if err != nil {
+		resp.Diagnostics.AddError("Unknown Attestation Variant",
```
msanft, line 190: We can use `resp.Diagnostics.AddAttributeError()` here.

msanft (approved):
LGTM besides the nit

msanft:
And we probably want to wait for TTs opinion on the comment above.

malt3 (approved):

terraform-provider-constellation/docs/data-sources/attestation.md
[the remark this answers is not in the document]
thomasten: Adding this hint seems reasonable to me. The detailed explanation should be in our docs then, for both TF provider and CLI. (We have a basic version of that here https://docs.edgeless.systems/constellation/architecture/attestation#cvm-verification)

thomasten (approved):

PR #2649, opened 2023-11-27
ci: only run verify with JSON on v2.14 or newer

daniel-weisse:
### Context
v2.13 doesnt support JSON flag for `constellation verify` on AWS yet.
So we shouldnt run the e2e verify test with that flag set.

### Proposed change(s)

- Don't use JSON output if CLI version == v2.13

### Additional info
- [e2e verify AWS with v2.13 CLI/image](https://github.com/edgelesssys/constellation/actions/runs/7005955389)

### DIFF
--- a/.github/actions/constellation_create/action.yml
+++ b/.github/actions/constellation_create/action.yml
@@ -167,12 +167,7 @@ runs:
       id: constellation-init
       shell: bash
       run: |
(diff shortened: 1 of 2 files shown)


derpsteb (approved):
Whoops. Thanks for catching this.

PR #2651, opened 2023-11-27
terraform: Azure Marketplace image support

msanft:
### Context

We want to let user's choose to use the VM image we distribute on the Azure marketplace.

### Proposed change(s)

- Allow users to set a config variable to use the marketplace image on Azure.

### Additional info

- AB#1234

### Checklist

- [x] Link to Milestone


terraform/infrastructure/azure/modules/scale_set/main.tf
```diff
@@ -46,9 +46,10 @@ resource "azurerm_linux_virtual_machine_scale_set" "scale_set" {
   disable_password_authentication = false
   upgrade_mode                    = "Manual"
   secure_boot_enabled             = var.secure_boot
-  source_image_id                 = var.image_id
-  tags                            = local.tags
-  zones                           = var.zones
+  # specify the image id only if a non-marketplace image is used
+  source_image_id = var.marketplace_image != null ? null : var.image_id
```
m1ghtym0, line 50: Does that mean I can omit the version string in the Constellation config if I set the marketplace image, or is it checked as a required field ahead of time?
I just want to make sure we have a single source of truth for all values and there is no ambiguity.

m1ghtym0:
Docs come in a separate PR?

terraform/infrastructure/azure/modules/scale_set/main.tf
msanft: No. If you want to use the marketplace image, we can't use the `source_image_id` attribute, but the `source_image_reference`:

https://github.com/edgelesssys/constellation/blob/2f8ec23d60534d6afe11fb481e5cf10a1ab64027/terraform/infrastructure/azure/modules/scale_set/main.tf#L91

The `var.marketplace_image.version` variable is set to the version of `.image` in `constellation-conf.yaml`,

https://github.com/edgelesssys/constellation/blob/2f8ec23d60534d6afe11fb481e5cf10a1ab64027/cli/internal/cloudcmd/tfvars.go#L172

i.e. the image version field, that would otherwise be resolved to a CSP-specific reference and passed into `source_image_id`.

msanft:
> Docs come in a separate PR?

Sorry, actually just forgot to push some pending commits.

daniel-weisse (approved):
Haven't tested this, but code changes lgtm

internal/kubecmd/kubecmd.go
```diff
@@ -104,7 +104,8 @@ func (k *KubeCmd) UpgradeNodeVersion(ctx context.Context, conf *config.Config, f
 	provider := conf.GetProvider()
 	attestationVariant := conf.GetAttestationConfig().GetVariant()
 	region := conf.GetRegion()
-	imageReference, err := k.imageFetcher.FetchReference(ctx, provider, attestationVariant, conf.Image, region)
+	imageReference, err := k.imageFetcher.FetchReference(ctx, provider, attestationVariant,
```
daniel-weisse, line 107: If I understand the code correctly, `imageReference` is a URI with a custom format designed by us.
How does this work for image upgrades / auto scaling if we set this URI as the image reference here?
I would think this would either require to do additional transformations on `imageReference` or changes in the node-operator.

terraform/infrastructure/azure/modules/scale_set/main.tf
m1ghtym0: Can you give a quick example of how a config would look like if I want to use the Marketplace with Constellation v2.13.0?

msanft: sure! It would look like:

```yaml
version: v4 # Schema version of this configuration file.
image: v2.13.0 # Machine image version used to create Constellation nodes.
...
provider:
  azure:
   ...
    useMarketplaceImage: true # Use the Azure Marketplace image instead of a custom image.
...
```

version.txt
```diff
@@ -1 +1 @@
-v2.14.0-pre
+v2.15.0-pre
```
3u13r, line 1:
```suggestion
v2.14.0-pre
```

3u13r (changes requested):

msanft:
[v2.14.0 debug image](https://github.com/edgelesssys/constellation/actions/runs/7043536570)
[v2.15.0 debug image](https://github.com/edgelesssys/constellation/actions/runs/7043531263)

Upgrade TODO

image/upload/pack.sh
```diff
@@ -38,7 +38,10 @@ pack() {
 
   azure)
     echo "📥 Packing Azure image..."
-    truncate -s %1MiB "${unpacked_image_dir}/${unpacked_image_filename}"
+    # Disk Images on Azure have to be a multiple of 1MiB in size.
+    # Marketplace images additionally have to be at least 1GiB in size.
+    truncate -s %2G "${unpacked_image_dir}/${unpacked_image_filename}" # at least 1GiB
+    # truncate -s +1M "${unpacked_image_dir}/${unpacked_image_filename}" # ensure > 1 GiB, but multiple of 1MiB
```
malt3, line 44: nit: let's document clearly why we chose 2GiB here:

Marketplace validation says >= 1 GiB, but using an image of size 1 GiB actually fails. Otherwise this looks  very confusing.

malt3 (approved):

image/upload/pack.sh
msanft: Yes, this is only for temporary testing, to ensure the hour-long marketplace pipeline won't fail. I'll test the commented-out approach (rounding up to a multiple of 1GiB and then adding 1MiB to ensure to be ">") afterwards.

dev-docs/workflows/marketplace-images.md
thomasten, line 1: We could also add this to the user docs and add a "contact us" banner (similar to what we do in the marblerun docs). I don't have a strong opinion. What do you think @m1ghtym0?

elchead:
Why isn't this included in the changelog?

terraform/infrastructure/azure/modules/scale_set/variables.tf
```diff
@@ -96,3 +96,14 @@ variable "secure_boot" {
   default     = false
   description = "Whether to deploy the cluster nodes with secure boot."
 }
+
+variable "marketplace_image" {
```
elchead, line 100: we should expose this in our `azure-constellation` module too.

elchead (changes requested):
only looked at the TF related changes

terraform/infrastructure/azure/modules/scale_set/variables.tf
msanft: I don't think so - the module will be deprecated with the next release and we will want new users to use the provider, to which I added support for the marketplace images.

msanft:
> Why isn't this included in the changelog?

What do you mean by that?

elchead:
> > Why isn't this included in the changelog?
> 
> What do you mean by that?

There is a `no-changelog` label

elchead (approved):

m1ghtym0 (approved):

thomasten (approved):

PR #2654, opened 2023-11-28
config: remove AWS SNP warning

thomasten:
#2601 added SNP on AWS again, so the warnings about the fallback are outdated.


derpsteb (approved):
:see_no_evil: Thanks!

PR #2660, opened 2023-11-29
Make Kubernetes serviceCIDR configurable in config

3u13r:
### Context
Kubernetes uses this serviceCIDR to allocate ClusterIPs from it: https://kubernetes.io/docs/concepts/services-networking/service/#type-clusterip
For peered networks it is essential to be able to re-configure all CIDRs as to not conflict with existing allocations in the peered network.

### Proposed change(s)

- Add optional `serviceCIDR` field in `constellation-conf.yaml`
- Pass value through our Terraform files

### Additional info

Terraform test: https://github.com/edgelesssys/constellation/actions/runs/7059236028
E2E test: https://github.com/edgelesssys/constellation/actions/runs/7059241745

### Checklist

- [x] Link to Milestone


internal/config/config.go
```diff
@@ -319,6 +322,7 @@ func Default() *Config {
 		MicroserviceVersion: constants.BinaryVersion(),
 		KubernetesVersion:   versions.Default,
 		DebugCluster:        toPtr(false),
+		ServiceCIDR:         "10.96.0.0/12",
```
derpsteb, line 325: Does this default work on all CSPs?

bootstrapper/initproto/init.proto
```diff
@@ -30,6 +30,8 @@ message InitRequest {
   string cluster_name = 9;
   // ApiserverCertSans is a list of Subject Alternative Names to add to the apiserver certificate.
   repeated string apiserver_cert_sans = 10;
+  // ServiceCIDR is the CIDR to use for Kubernets ClusterIPs.
```
derpsteb, line 33:
```suggestion
  // ServiceCIDR is the CIDR to use for Kubernetes ClusterIPs.
```

derpsteb (approved):

internal/config/config.go
derpsteb: nvm, that's the kubeadm default.

derpsteb:
Maybe add a sentence of context what the serviceCIDR is / how kubeadm uses it? Because this will be part of the changelog.

PR #2661, opened 2023-11-29
docs: add Kubernetes version support list

daniel-weisse:
### Context
We would like to offer a table displaying which versions of Kubernetes are supported in any given Constellation release.

### Proposed change(s)

- Add a new section to `/architecture/versions.md`  which contains a list of supported Kubernetes versions for a given Constellation release
- Automatically update the supported version list on `bazel run //:generate`

### Additional info

- [AB#3198](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3198)
- [AB#3199](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3199)

### DIFF
--- a/hack/versioninfogen/main.go
+++ b/hack/versioninfogen/main.go
@@ -1,0 +1,74 @@
+/*
+Copyright (c) Edgeless Systems GmbH
+
+SPDX-License-Identifier: AGPL-3.0-only
+*/
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/edgelesssys/constellation/v2/internal/semver"
+	"github.com/edgelesssys/constellation/v2/internal/versions"
+)
+
+const fileSeparator = "<!--AUTO_GENERATED_BY_BAZEL-->\n<!--DO_NOT_EDIT-->\n"
+
+func main() {
+	filePath := flag.String("file-path", "../../docs/docs/architecture/versions.md", "path to the version file to update")
+	flag.Parse()
+
+	k8sVersionStrings := versions.SupportedK8sVersions()
+	var k8sVersions []semver.Semver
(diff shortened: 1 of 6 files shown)


hack/version-matrix-gen/main.go
```diff
@@ -0,0 +1,174 @@
...
+
+	return matrix, nil
+}
+
+func parseNewVersion(cliPath string) (semver.Semver, []semver.Semver, error) {
+	// Run the CLI to get
+	// 1. the new version
+	// 2. the supported Kubernetes versions
+
+	out, err := exec.CommandContext(context.Background(), cliPath, "version").Output()
```
3u13r, line 142: Wouldn't an easier way to just import the `versions.go` ? 
Also we could add the matrix to `bazel run //:generate` to verify that everything has the correct state.

daniel-weisse: Not easier, because we would need a lot of extra checks to make sure we dont update the matrix with any pseudo versions.
Also need to do some changes to the release job that updates the docs, since this creates a branch from main, not from the pending release branch.

I could see a way to make this work though, and it could also make it easier to update the matrix for patch releases.

hack/versioninfogen/main.go
```diff
@@ -0,0 +1,74 @@
...
+		k8sVersions = append(k8sVersions, k8sVersion)
+	}
+
+	if err := updateMatrix(*filePath, k8sVersions); err != nil {
+		fmt.Fprintf(os.Stderr, "error updating versions file: %s\n", err)
+		os.Exit(1)
+	}
+}
+
+func updateMatrix(filePath string, supportedVersions []semver.Semver) error {
```
thomasten, line 43: nit: rename this func

thomasten (approved):

PR #2664, opened 2023-11-30
helm: deprioritize Cilium tc filters

burgerdev:
### Context

By default, Cilium's tc filters are added add the highest priority, which makes it impossible to add any tc filters of our own (because the Cilium eBPF programs don't return to the filter chain).

Two near-future use cases would benefit from changing this:

* Network testing could add counting filters to interfaces and observe e.g. violations of encryption policy.
* The VPN Helm chart could add a filter policy that drops packets on the "physical" interface before they can leak to the CSP.

### Proposed change(s)

- Drop the priority of Cilium tc programs to 50.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/helm/values.go
+++ b/internal/helm/values.go
@@ -14,7 +14,7 @@ var ciliumVals = map[string]map[string]any{
 		"endpointRoutes": map[string]any{
 			"enabled": true,
 		},
-		"extraArgs": []string{"--node-encryption-opt-out-labels=invalid.label"},
+		"extraArgs": []string{"--node-encryption-opt-out-labels=invalid.label", "--bpf-filter-priority=50"},
 		"encryption": map[string]any{
 			"enabled":        true,
 			"type":           "wireguard",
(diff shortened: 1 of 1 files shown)


internal/helm/values.go
```diff
@@ -102,7 +102,7 @@ var ciliumVals = map[string]map[string]any{
 		"endpointRoutes": map[string]any{
 			"enabled": true,
 		},
-		"extraArgs": []string{"--node-encryption-opt-out-labels=invalid.label"},
+		"extraArgs": []string{"--node-encryption-opt-out-labels=invalid.label", "--bpf-filter-priority=50"},
```
3u13r, line 105: We likely want this on all cloud, don't we? I know that this map/file likely needs some refactoring since there's a lot of duplicate configuration. For now, the easiest way would be just to copy the config.

3u13r (changes requested):

3u13r (approved):
LGTM

PR #2665, opened 2023-11-30
constellation-lib: refactor init RPC to be shared

msanft:
### Context

In preparation for the Terraform provider, we want to factor out code that will be used by both the CLI and the Terraform provider into a Constellation Go library. This is also the case for the Init RPC.

### Proposed change(s)

- Factor out the Init RPC to live inside the `internal/constellation` Go package and use it from within the CLI.

### Additional info

- [AB#3572](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3572)
- [E2E Test Azure](https://github.com/edgelesssys/constellation/actions/runs/7058189575)
- [E2E Test Azure 2
](https://github.com/edgelesssys/constellation/actions/runs/7085564122)
### Checklist

- [x] Link to Milestone


internal/constellation/applyinit.go
```diff
@@ -0,0 +1,265 @@
...
+func (a *Applier) Init(
+	ctx context.Context,
+	dialer GrpcDialer,
+	state *state.State,
+	payload InitPayload,
+) (
+	*initproto.InitSuccessResponse,
+	[]byte,
+	error,
+) {
```
daniel-weisse, line 54: Instead of returning the logs as `[]byte`, we could pass an `io.Writer` to `Init`.
This would allow us to pass a logging interface for the Terraform provider (we probably want to collect the cluster logs in case of failure and log them), or a buffer which we later write to file for the CLI.

This is mostly an implementation detail tho. There shouldn't be much difference in the behavior between the current implementation to if you change it up.

internal/constellation/apply.go
```diff
@@ -27,9 +31,10 @@ type debugLog interface {
 }
 
 // NewApplier creates a new Applier.
-func NewApplier(log debugLog) *Applier {
+func NewApplier(log debugLog, spinner spinnerInterf) *Applier {
```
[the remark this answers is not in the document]
msanft: agreed to pass some sort of "spinner" implementation that informs the user of the activity in the Terraform logs

cli/internal/cmd/apply.go
```diff
@@ -293,6 +297,23 @@ type applyCmd struct {
 
 type applier interface {
```
elchead, line 298: nit: this name doesn't fit well. How about `initer`?

cli/internal/cmd/apply.go
```diff
@@ -293,6 +297,23 @@ type applyCmd struct {
...
 	CheckLicense(ctx context.Context, csp cloudprovider.Provider, licenseID string) (int, error)
```
elchead, line 299: seems unused
```suggestion
```

cli/internal/cmd/apply.go
daniel-weisse: The goal is to have this interface implement all methods for clients that will be shared by the CLI and the Terraform provider, so `initer` does not fit that description.
See [AB#3557](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/3557) for more details.

internal/constellation/applyinit.go
```diff
@@ -0,0 +1,264 @@
...
+
+// GrpcDialer dials a gRPC server.
+type GrpcDialer interface {
+	Dial(ctx context.Context, target string) (*grpc.ClientConn, error)
+}
+
+// Init performs the init RPC.
+func (a *Applier) Init(
+	ctx context.Context,
+	dialer GrpcDialer,
```
daniel-weisse, line 46: As suggested already, we could move this dialer to be a private attribute of `Applier`.
Setting it up could be done in the Constructor of the struct.

daniel-weisse (approved):

PR #2669, opened 2023-12-01
ref: decouple helm from config

elchead:
### Context

The terraform provider needs helm, so the library should not depend on CLI specific code, such as the config.

### Proposed change(s)

- decouple from the config (except for OpenStack)

### Additional info

- [AB#3575](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3575)

### Checklist

- [x] Link to Milestone


internal/helm/helm.go
```diff
@@ -80,36 +83,37 @@ func NewClient(kubeConfig []byte, log debugLog) (*Client, error) {
...
-// TODO(elchead): remove validK8sVersion by putting ValidK8sVersion into config.Config, see AB#3374.
 func (h Client) PrepareApply(
-	conf *config.Config, stateFile *state.State,
-	flags Options, serviceAccURI string, masterSecret uri.MasterSecret,
+	csp cloudprovider.Provider, attestationVariant variant.Variant, k8sVersion versions.ValidK8sVersion,
+	microserviceVersion semver.Semver, stateFile *state.State, flags Options, serviceAccURI string,
+	masterSecret uri.MasterSecret, openStackCfg *config.OpenStackConfig,
 ) (Applier, bool, error) {
-	releases, err := h.loadReleases(conf, masterSecret, stateFile, flags, serviceAccURI)
+	releases, err := h.loadReleases(csp, attestationVariant, k8sVersion, masterSecret, stateFile, flags, serviceAccURI, openStackCfg)
```
derpsteb, line 98: Should we add these two checks right here because its the public entrypoint for the lib?
- `if csp == cloudprovider.OpenStack && openStackCfg == nil { return err; }`
- `if csp != cloudprovider.OpenStack && openStackCfg != nil { return err; }`

PR #2670, opened 2023-12-01
ci: fix repository name of shellcheck for linux arm64

malt3:
### Proposed change(s)

- ci: fix repository name of shellcheck for linux arm64


3u13r (approved):
LGTM

PR #2678, opened 2023-12-04
terraform-provider: data skeleton for cluster resource

elchead:
### Context

This adds the data parsing for the cluster resource from #2613.
The implementation will be done in a follow-up PR once the library refactoring is finished.

### Proposed change(s)

- adjust RFC to reflect necessary changes
- remove unnecessary ID field from data sources
- add data skeleton for cluster resource
- move `measurements` and `variant` to `attestation` attribute in attestation data source to simplify usage.

### Additional info

- [AB#3550](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3550) (first part)

I recommend to review by commit.

### Checklist

- [ ] Link to Milestone


rfc/014-terraform-provider.md
```diff
@@ -51,7 +51,9 @@ resource "constellation_cluster" "foo" {
         # through Terraform, such as a VPC.
         # and in-cluster Kubernetes API endpoint, e.g. for Kubelets
     }
+    attestation_variant = "azure-sev-snp" # merging with '.attestation' is not possible due to struct parsing (nested data structure and simple string)
```
msanft, line 54: Why isn't this possible? Can you share some more details?

terraform-provider-constellation/internal/provider/attestation_data_source.go
```diff
@@ -79,27 +73,8 @@ func (d *AttestationDataSource) Schema(_ context.Context, _ datasource.SchemaReq
...
-					"  * `aws-nitro-tpm`\n" +
-					"  * `azure-sev-snp`\n" +
-					"  * `gcp-sev-es`\n",
-				Required: true,
-				Validators: []validator.String{
-					stringvalidator.OneOf("aws-sev-snp", "aws-nitro-tpm", "azure-sev-snp", "gcp-sev-es"),
-				},
-			},
+			"csp":                 newCSPAttribute(),
+			"attestation_variant": newAttesationVariantAttribute(),
```
msanft, line 77:
```suggestion
			"attestation_variant": newAttestationVariantAttribute(),
```

terraform-provider-constellation/internal/provider/cluster_resource.go
```diff
@@ -0,0 +1,319 @@
...
+				MarkdownDescription: "The Constellation microservices version to use for the cluster.",
+				Description:         "The Constellation microservices version to use for the cluster. When not set, the latest default version will be used.",
+				Optional:            true,
+			},
+			"debug": schema.BoolAttribute{
+				MarkdownDescription: "DON'T USE IN PRODUCTION: Enable debug mode and allow the use of debug images.",
+				Description:         "DON'T USE IN PRODUCTION: Enable debug mode and allow the use of debug images.",
+				Optional:            true,
+			},
+			"init_endpoint": schema.StringAttribute{ // TODO what is this?
```
msanft, line 101: FYI: This is the endpoint of the node running the bootstrapper

terraform-provider-constellation/internal/provider/cluster_resource.go
```diff
@@ -0,0 +1,319 @@
...
+				Description:         "The Kubernetes version to use for the cluster. When not set, the latest default version will be used.", // TODO(elchead): refer to supported versions; wait for https://github.com/edgelesssys/constellation/pull/2661
+				Optional:            true,
+			},
+			"constellation_microservices_version": schema.StringAttribute{
+				MarkdownDescription: "The Constellation microservices version to use for the cluster.",
+				Description:         "The Constellation microservices version to use for the cluster. When not set, the latest default version will be used.",
+				Optional:            true,
+			},
+			"debug": schema.BoolAttribute{
+				MarkdownDescription: "DON'T USE IN PRODUCTION: Enable debug mode and allow the use of debug images.",
```
msanft, line 97: We should put a markdown warning here:
```suggestion
				MarkdownDescription: "~> **Warning:** Do not enable Debug mode in production environments.\nEnable debug mode and allow the use of debug images.",
```

terraform-provider-constellation/internal/provider/cluster_resource.go
```diff
@@ -0,0 +1,319 @@
...
+				MarkdownDescription: "The master secret to use for the cluster.",
+				Description:         "The master secret to use for the cluster.",
+				Required:            true,
+			},
+			"init_secret": schema.StringAttribute{
+				MarkdownDescription: "The init secret to use for the cluster.",
+				Description:         "The init secret to use for the cluster.",
+				Required:            true,
+			}, // TODO merge / derive from master secret?
+			"attestation_variant": newAttesationVariantAttribute(),
```
msanft, line 133:
```suggestion
			"attestation_variant": newAttestationVariantAttribute(),
```

terraform-provider-constellation/internal/provider/common.go
msanft, line 1: I like your refactoring to remove the duplication here, but I think we should give the file a more meaningful name. `utils.go`, `common.go`, ... are considered bad practices mostly, as Go code should be grouped in a semantically meaningful way.

What about `schemas.go` or putting these functions in the files they are semantically most related to (i.e. attestationVariant to `attestation_data_source.go` and so on)

This might be sorted altogether though if we agree on using custom types instead.

terraform-provider-constellation/internal/provider/convert.go
```diff
@@ -0,0 +1,173 @@
...
+	"strconv"
+
+	"github.com/edgelesssys/constellation/v2/internal/api/attestationconfigapi"
+	"github.com/edgelesssys/constellation/v2/internal/attestation/idkeydigest"
+	"github.com/edgelesssys/constellation/v2/internal/attestation/measurements"
+	"github.com/edgelesssys/constellation/v2/internal/attestation/variant"
+	"github.com/edgelesssys/constellation/v2/internal/config"
+)
+
+func parseAttestationConfig(tfMeasurements map[string]measurement, tfSnpAttestation sevSnpAttestation, attestationVariant variant.Variant) (config.AttestationCfg, error) {
```
msanft, line 23: We should explain what this function does. It's not clear to me at first glance. I see that it converts values coming from Terraform to Constellation-compatible values by looking at the body. Maybe we could add that as a doc comment?

terraform-provider-constellation/internal/provider/convert.go
```diff
@@ -0,0 +1,173 @@
...
+type extraMicroservices struct {
+	CSIDriver bool `tfsdk:"csi_driver"`
+}
+
+type measurement struct {
+	Expected string `tfsdk:"expected"`
+	WarnOnly bool   `tfsdk:"warn_only"`
+}
+
+type sevSnpAttestation struct {
```
msanft, line 154: Don't we need to use Terraform's `types` package here?

elchead: no, thats the magic of the type annotation :)

terraform-provider-constellation/internal/provider/common.go
elchead: How about `shared_attributes.go`?
Splitting seems difficult in light of attributes like `csp`.

terraform-provider-constellation/internal/provider/convert.go
msanft: Possibly we could directly assign these tags to the types we implement in internal then and spare the duplication?

elchead: agreed to keep for now, as we don't want to couple the constellation API to the TF provider. Might be reworked later.

terraform-provider-constellation/internal/provider/attestation_data_source.go
```diff
@@ -191,27 +126,23 @@ func (d *AttestationDataSource) Read(ctx context.Context, req datasource.ReadReq
...
-			resp.Diagnostics.AddError("Converting SNP attestation", err.Error())
-		}
-		diags := resp.State.SetAttribute(ctx, path.Root("attestation"), tfSnpAttestation)
-		resp.Diagnostics.Append(diags...)
-		if resp.Diagnostics.HasError() {
-			return
-		}
 	}
-
+	tfSnpAttestation, err := convertToTfAttestation(attestationVariant, snpVersions)
```
daniel-weisse, line 142: Can you rename all occurrences of `tfSnpAttestation` to simply `tfAttestation` (or similar)?
If I understand this code correctly, this struct is not exclusively used for just SNP based attestation, so it shouldn't be named for that.

terraform-provider-constellation/internal/provider/cluster_resource.go
```diff
@@ -0,0 +1,308 @@
...
+				Optional:            true, // TODO(elchead): use "constell" as default
+			},
+			"image": schema.StringAttribute{
+				MarkdownDescription: "The Constellation OS image to use in the CSP specific reference format. Use the `constellation_image` data source to find the correct image for your CSP.",
+				Description:         "The Constellation OS image to use in the CSP specific reference format. Use the `constellation_image` data source to find the correct image for your CSP. When not set, the latest default version will be used.",
+				Optional:            true,
+			},
+			"kubernetes_version": schema.StringAttribute{
+				MarkdownDescription: "The Kubernetes version to use for the cluster.",
+				Description:         "The Kubernetes version to use for the cluster. When not set, the latest default version will be used.", // TODO(elchead): refer to supported versions; wait for https://github.com/edgelesssys/constellation/pull/2661
```
daniel-weisse, line 87: You could just embed this information into the binary using `versions.SupportedK8sVersions`.
If Terraform uses this Schema function to generate the docs, it should always contain the correct information for a released version

terraform-provider-constellation/internal/provider/cluster_resource.go
```diff
@@ -0,0 +1,308 @@
...
+			"constellation_microservices_version": schema.StringAttribute{
+				MarkdownDescription: "The Constellation microservices version to use for the cluster.",
+				Description:         "The Constellation microservices version to use for the cluster. When not set, the latest default version will be used.",
+				Optional:            true,
+			},
+			"debug": schema.BoolAttribute{
+				MarkdownDescription: "~> **Warning:** Do not enable Debug mode in production environments.\nEnable debug mode and allow the use of debug images.",
+				Description:         "DON'T USE IN PRODUCTION: Enable debug mode and allow the use of debug images.",
+				Optional:            true,
+			},
```
daniel-weisse, line 99: Is there an option to mark attributes as hidden? If so we should do that for the `debug` mode.

terraform-provider-constellation/internal/provider/cluster_resource.go
```diff
@@ -0,0 +1,308 @@
...
+			"master_secret": schema.StringAttribute{
+				MarkdownDescription: "The master secret to use for the cluster.",
+				Description:         "The master secret to use for the cluster.",
+				Required:            true,
+			},
+			"init_secret": schema.StringAttribute{
+				MarkdownDescription: "The init secret to use for the cluster.",
+				Description:         "The init secret to use for the cluster.",
+				Required:            true,
+			}, // TODO merge / derive from master secret?
```
daniel-weisse, line 131: We shouldn't merge this with the master secret, since within a VPC, the init secret is not considered to be secret (it is attached as a tag to the control plane nodes).
Deriving it from the master secret would require it to be generated by the user with a by us specify , so I don't think this is a good approach either.
I would just keep it as separate input

terraform-provider-constellation/internal/provider/cluster_resource.go
```diff
@@ -0,0 +1,308 @@
...
+	// client, ok := req.ProviderData.(*http.Client)
+
+	// if !ok {
+	//	resp.Diagnostics.AddError(
+	//		"Unexpected Resource Configure Type",
+	//		fmt.Sprintf("Expected *http.Client, got: %T. Please report this issue to the provider developers.", req.ProviderData),
+	//	)
+
+	//	return
+	//}
```
daniel-weisse, line 168:
```suggestion
```
And the other commented out example code as well

terraform-provider-constellation/internal/provider/convert.go
```diff
@@ -0,0 +1,198 @@
...
+		case false:
+			valOption = measurements.Enforce
+		}
+		c11nMeasurements[uint32(idx)] = measurements.Measurement{
+			Expected:      expectedBt,
+			ValidationOpt: valOption,
+		}
+	}
+
+	// parser := attestationParser{attestation: attestation}
```
daniel-weisse, line 54:
```suggestion
```

daniel-weisse (dismissed):

terraform-provider-constellation/internal/provider/attestation_data_source.go
```diff
@@ -108,10 +83,6 @@ func (d *AttestationDataSource) Schema(_ context.Context, _ datasource.SchemaReq
 				MarkdownDescription: "For Azure only, the URL of the Microsoft Azure Attestation service",
 				Optional:            true,
 			},
-			"id": schema.StringAttribute{
-				Computed:            true,
-				MarkdownDescription: "The ID of the data source",
-			},
 			"measurements": schema.MapNestedAttribute{
```
msanft, line 86: I think this is superfluous now, as it is contained in the `attestation` output, right?

terraform-provider-constellation/internal/provider/attestation_data_source.go
```diff
@@ -224,83 +155,11 @@ func (d *AttestationDataSource) Read(ctx context.Context, req datasource.ReadReq
 			return
 		}
 	}
-	tfMeasurements := convertMeasurementsTfStateCompatible(fetchedMeasurements)
-	diags := resp.State.SetAttribute(ctx, path.Root("measurements"), tfMeasurements)
+	tfAttestation.Measurements = convertToTfMeasurements(fetchedMeasurements)
```
msanft, line 158:
```suggestion
	tfAttestation.Measurements = convertToTfMeasurements(fetchedMeasurements)

```
nit: makes the separation a little more apparent

terraform-provider-constellation/internal/provider/convert.go
```diff
@@ -0,0 +1,197 @@
...
+
+// convertToTfFirmwareCfg converts the constellation firmware config to the terraform struct.
+func convertToTfFirmwareCfg(firmwareCfg config.SNPFirmwareSignerConfig) (azureSnpFirmwareSignerConfig, error) {
+	keyDigestAny, err := firmwareCfg.AcceptedKeyDigests.MarshalYAML()
+	if err != nil {
+		return azureSnpFirmwareSignerConfig{}, err
+	}
+	keyDigest, ok := keyDigestAny.([]string)
+	if !ok {
+		return azureSnpFirmwareSignerConfig{}, errors.New("reading Accepted Key Digests: could not convert to []string")
```
msanft, line 132:
```suggestion
		return azureSnpFirmwareSignerConfig{}, fmt.Errorf("reading Accepted Key Digests: could not convert %T to []string", keyDigestAny)
```
nit

msanft (changes requested):

terraform-provider-constellation/internal/provider/cluster_resource.go
daniel-weisse: Looked at the code again.
I dont think we need this attribute for the Terraform provider at all.
The flag only changes how we deploy VMs (we use it to configure serial console access) and to print some warnings, so we should be ok with just removing it from the cluster resource

msanft (approved):

internal/attestation/idkeydigest/idkeydigest.go
```diff
@@ -188,23 +190,26 @@ func (d *List) UnmarshalJSON(b []byte) error {
...
 		if err != nil {
-			return fmt.Errorf("decoding digest: %w", err)
+			return nil, fmt.Errorf("decoding digest: %w", err)
 		}
-		*d = append(*d, digest)
+		*res = append(*res, digest)
 	}
-	return nil
+	return res, nil
 }
```
daniel-weisse, line 215:
```suggestion
func UnmarshalHexString(encodedDigests []string) (List, error) {
	res := make(List, 0, len(encodedDigests))
	for idx, encodedDigest := range encodedDigests {
		if len(encodedDigest) != encodedDigestLength {
			return nil, fmt.Errorf("invalid digest length: %d", len(encodedDigest))
		}
		digest, err := hex.DecodeString(encodedDigest)
		if err != nil {
			return nil, fmt.Errorf("decoding digest: %w", err)
		}
		res[idx] = digest
	}
	return res, nil
}
```
nit: this is a bit cleaner
Looking at the rest of the code, we also don't usually work on a pointer to `List`, but on it directly.

terraform-provider-constellation/internal/provider/shared_attributes.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+const (
+	// AttributeInput is the attribute type used for input variables.
+	AttributeInput attributeType = true
+	// AttributeOutput is the attribute type used for output variables.
+	AttributeOutput attributeType = false
```
daniel-weisse, line 20: Do these need to be exported?

terraform-provider-constellation/internal/provider/cluster_resource.go
daniel-weisse: No, I was referring to `versions.SupportedK8sVersions` which returns a list of the Kubernetes versions that are supported by this version of Constellation.
You don't need to refer to our docs for this information, when you can just provide it inline.

terraform-provider-constellation/internal/provider/cluster_resource.go
elchead: deleted

daniel-weisse (approved):

PR #2680, opened 2023-12-05
constellation-lib: add Helm wrapper

daniel-weisse:
### Context
All clients that are shared by the CLI and our Terraform provider should be handled by the Constellation lib, so we don't need to keep multiple clients in different structs and keep their usage in sync with one another.

### Proposed change(s)

- Add a wrapper for the Helm package to the Constellation lib `Applier`
- Use said wrapper in the CLI to load Helm charts
- Move Helm package from `/internal/helm` to `/internal/constellation/helm`

### Additional info

- [AB#3581](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3581)
- [AB#3583](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3583)


cli/internal/cmd/apply.go
```diff
@@ -845,6 +806,42 @@ func (wl warnLogger) Warnf(fmtStr string, args ...any) {
 	wl.cmd.PrintErrf("Warning: %s\n", fmt.Sprintf(fmtStr, args...))
 }
 
+type warnLog interface {
+	Warnf(format string, args ...any)
+	Infof(format string, args ...any)
+}
+
+// applier is used to run the different phases of the apply command.
+type applier interface {
```
msanft, line 815: I'm not a huge fan of having all interfaces at the end of this (very large) file. Personally, I'd rather have them below the struct / func they are used in

msanft (approved):

PR #2682, opened 2023-12-06
constellation-lib: move state and featureset packages

daniel-weisse:
### Context
CLI and provider specific code should be scoped to the constellation-lib.

### Proposed change(s)

- Move `/internal/state` to `/internal/constellation/state`
- Move `/internal/featureset` to `/internal/constellation/featureset`

### Additional info

- [AB#3583](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3583)


elchead:
Can you explain the rationale behind this a little more? 
Afais, almost all of our internal libraries are related to constellation. I think you mean that if we were to make this constellation library public, all required packages to call it (to instantiate the input structs), should be subpackages of `constellation`?

daniel-weisse:
>Can you explain the rationale behind this

The goal is to keep a degree of logical separation between the code that is used by the CLI and the Terraform provider (the constellation-lib in `/internal/constellation`) and other code that is shared inside the Constellation repo.
Generally, the code inside `/internal/constellation` should not be used by components other than the CLI or the provider.
Ideally, we would add subpackages of the constellation-lib to an internal folder, but not all parts of the CLI or the provider exclusively access the subpackages through the applier interface, so that doesn't work.

elchead:
> > Can you explain the rationale behind this
> 
> The goal is to keep a degree of logical separation between the code that is used by the CLI and the Terraform provider (the constellation-lib in `/internal/constellation`) and other code that is shared inside the Constellation repo. Generally, the code inside `/internal/constellation` should not be used by components other than the CLI or the provider. Ideally, we would add subpackages of the constellation-lib to an internal folder, but not all parts of the CLI or the provider exclusively access the subpackages through the applier interface, so that doesn't work.

I see. So these packages should either only be consumed by the library (like helm, which could hence be made internal) or used to consume the library (like state or featureset).

PR #2686, opened 2023-12-07
terraform-provider: create release in provider repo on Constellation release

daniel-weisse:
### Context
On release, we also want to release a new version of the Terraform provider and publish it in the Terraform registry.
Since publishing to the registry requires a specially set up GitHub repository, the release has to be created outside of edgelesssys/constellation.

### Proposed change(s)

- Rename the `release-cli` workflow to `draft-release` since it does more (and has for a while) than build and publish a CLI binary
- Build, package, and sign Terraform provider binaries as part of the release workflow
- Create a (draft) release with the provider binaries in edgelesssys/terraform-provider-constellation as part of the release workflow
  - This release has the same version as the one in edgelesssys/constellation
  - The release links back to the release in edgelesssys/constellation
- Automatically set the target branch when preparing the release to the release branch
  - Previously, the release was tagged on current main, which means that if any changes were pushed to main after the release pipeline was started, the tag was created on the wrong commit (not the commit used by the pipeline to build binaries, images, etc.) causing the changelog to contain incorrect information and the attached source code to be too new.
  - Updated the dev-docs to only mention checking that the tag and target branch are set correctly

### Additional info

- [AB#3551](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3551)
- [Draft release workflow run](https://github.com/edgelesssys/constellation/actions/runs/7142503811)
  - [Draft release in Constellation repo](https://github.com/edgelesssys/constellation/releases/tag/untagged-f142619ae52ff70466a0)
  - [Draft release in Terraform provider repo](https://github.com/edgelesssys/terraform-provider-constellation/releases/tag/untagged-c8d5e9398dc9c8bc568e)

### Checklist

- [x] Link to Milestone


.github/workflows/draft-release.yml
```diff
@@ -109,6 +109,60 @@ jobs:
...
+      - name: Setup bazel
+        uses: ./.github/actions/setup_bazel_nix
+        with:
+          useCache: "false"
+
+      - name: Build Terraform Provider Binary
+        uses: ./.github/actions/build_tf_provider
+        with:
+          targetOS: ${{ matrix.os }}
+          targetArch: ${{ matrix.arch }}
```
msanft, line 148: I didn't verify yet if the provider is usable on windows. We might want to exclude this here until we have a windows test.

.github/workflows/draft-release.yml
```diff
@@ -208,7 +242,12 @@ jobs:
             constellation-linux-arm64 \
             constellation-windows-amd64.exe \
             constellation.spdx.sbom \
-            terraform-module.zip)
+            terraform-module.zip \
+            terraform-provider-constellation-darwin-amd64 \
+            terraform-provider-constellation-darwin-arm64 \
+            terraform-provider-constellation-linux-amd64 \
+            terraform-provider-constellation-linux-arm64 \
+            terraform-provider-constellation-windows-amd64.exe)
```
msanft, line 250: s.a.

msanft (changes requested):
LGTM otherwise

.github/workflows/draft-release.yml
daniel-weisse: Any reason why it wouldn't work?
We don't use any CGo Code, and also don't interact with the filesystem, which are the 2 things that come to mind that could break windows compatibility.

msanft (approved):

PR #2692, opened 2023-12-09
terraform: fix Azure marketplace image defaults

msanft:
### Context

Multiple of the weekly e2e tests failed due to the Azure marketplace image being set to an object with empty fields, but not an null object, which breaks the assumptions of the Terraform infrastructure module.

### Proposed change(s)

- (Unrelated) Make the Config default to `false` instead of `null` for `useMarketplaceImage`, which makes more sense for the user.
- Use a pointer for the marketplace image variable, making it default to `nil` / `null` in Terraform, which satisfies:
https://github.com/edgelesssys/constellation/blob/0247f4f6baa57b19e51bcffd62a26577b7e21a8e/terraform/infrastructure/azure/modules/scale_set/main.tf#L76-L93

### Additional Information
- [Post-fix E2E Test](https://github.com/edgelesssys/constellation/actions/runs/7153418937)

### Checklist

- [x] Link to Milestone


msanft:
Solves [#129](https://github.com/edgelesssys/issues/issues/129)

elchead (approved):

daniel-weisse (approved):

PR #2694, opened 2023-12-11
versions: consolidate various types of Components

burgerdev:
There used to be three definitions of a Component type, and conversion routines between the three. Since the use case is always the same, and the Component semantics are defined by versions.go and the installer, it seems appropriate to define the Component type there and import it in the necessary places.

### Context

Prequel refactoring for RFC 015.

### Proposed change(s)

- Embrace the proto: Define one `components.Component` proto message and refer to it in init/join protos as well as Go code.

### Checklist

- [ ] Link to Milestone


daniel-weisse (approved):

3u13r (approved):
LGTM

PR #2696, opened 2023-12-11
installer: make hash checking optional

burgerdev:
### Context

RFC 015 proposes adding a new type of KubernetesComponent in the form of data URLs. As the content of these is deterministic, it does not make sense to also generate a hash for them and then check it in the installer.

### Proposed change(s)

- Omit the file hash check if the expected hash is empty. This does not change existing behaviour, but will be used by data URLs (to be introduced).

### Checklist

- [x] Link to Milestone


3u13r (approved):
LGTM

PR #2697, opened 2023-12-11
upgrade-agent: set explicit k8s version for kubeadm

burgerdev:
`kubeadm upgrade plan` takes an optional positional argument for the target Kubernetes version [1], which is documented as

> To skip the internet check, pass in the optional [version] parameter

In light of RFC 015, we should not do an "internet check" for latest versions, but stick to the version we're actually targeting with the upgrade. Thus, we explicitly pass the version as an argument.

[1]: https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-upgrade/#cmd-upgrade-plan

### Context

RFC 015

### Proposed change(s)

- explicitly pass target version to `kubeadm upgrade plan`

### Checklist

- [x] Link to Milestone


3u13r (approved):
nice catch, LGTM

PR #2698, opened 2023-12-11
installer: add support for data URLs

burgerdev:
RFC 015 proposes the introduction of data URLs to materialize static content to files on disk. This commit adds support for data URLs to the installer. The corresponding content will be added to versions.go in a subsequent commit.

### Context

RFC 015

### Proposed change(s)

- handle `data:` and `http(s):` URLs separately

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/installer/installer.go
+++ b/internal/installer/installer.go
@@ -9,6 +9,7 @@ package installer
 
 import (
 	"archive/tar"
+	"bytes"
 	"compress/gzip"
 	"context"
 	"crypto/sha256"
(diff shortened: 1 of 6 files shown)


burgerdev:
This requires #2696, so making this a draft for the time being.

internal/installer/installer.go
```diff
@@ -178,36 +182,81 @@ func (i *OsInstaller) retryDownloadToTempDir(ctx context.Context, url string) (f
...
+	}
+
+	if url.Scheme == "data" {
+		err = i.unpackData(u, out)
+	} else {
+		err = i.downloadHTTP(ctx, u, out)
+	}
+	out.Close()
+	if err != nil {
+		i.fs.Remove(out.Name())
```
3u13r, line 258: This returns an error. I'd make the ignored error more explicit (also because our golangci-lint doesn't like it). We could also simply join the error.

3u13r (changes requested):

internal/installer/installer.go
burgerdev: Thanks for the suggestion, implemented with `errors.Join`.

malt3 (approved):
LGTM

3u13r (approved):
LGTM

PR #2703, opened 2023-12-11
terraform: align infrastructure module attributes

elchead:
### Context

We want to publish the infrastructure and IAM modules with the Constellation release, so that users can use them as a reference setup or directly consume them combined with the Terraform providers resources and data sources. Therefore, their inputs / outputs should be properly named (snake case) and described. Furthermore, the names should be semantically correct and aligned with the input names of the provider.

### Proposed change(s)

- Rename and describe fields as necessary.

### Additional info

- [AB#3599](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3599)
- [E2E Test Daily](https://github.com/edgelesssys/constellation/actions/runs/7219194595)
- [E2E Test Self Managed infrastructure](https://github.com/edgelesssys/constellation/actions/runs/7219195852)
- [E2E Test Upgrade AWS](https://github.com/edgelesssys/constellation/actions/runs/7219199191)
- [E2E Test Upgrade Azure](https://github.com/edgelesssys/constellation/actions/runs/7219200994)
- [E2E Test Upgrade GCP](https://github.com/edgelesssys/constellation/actions/runs/7219202348)

### Checklist

- [x] Link to Milestone


terraform/infrastructure/aws/outputs.tf
```diff
@@ -1,9 +1,11 @@
 output "out_of_cluster_endpoint" {
   value = local.out_of_cluster_endpoint
+  description = "The external endpoint for the Kubernetes API server. This only varies from the `in_cluster_endpoint` when using an internal load balancer setup."
```
msanft, line 3:
```suggestion
  description = "External endpoint for the Kubernetes API server. Only varies from the `in_cluster_endpoint` when using an internal load balancer."
```

terraform/infrastructure/aws/outputs.tf
```diff
@@ -1,9 +1,11 @@
...
 }
 
 output "in_cluster_endpoint" {
   value = local.in_cluster_endpoint
+  description = "The internal endpoint for the Kubernetes API server. This only varies from the `in_cluster_endpoint` when using an internal load balancer setup."
```
msanft, line 8:
```suggestion
  description = "Internal endpoint for the Kubernetes API server."
```

terraform/infrastructure/aws/outputs.tf
```diff
@@ -17,21 +19,26 @@ output "api_server_cert_sans" {
       )
     )
   )
+  description = "List of Subject Alternative Names (SANs) for the API server certificate."
```
msanft, line 22:
```suggestion
  description = "List of additional Subject Alternative Names (SANs) for the API server certificate."
```

terraform/infrastructure/aws/outputs.tf
```diff
@@ -1,9 +1,11 @@
 output "out_of_cluster_endpoint" {
   value = local.out_of_cluster_endpoint
+  description = "The external endpoint for the Kubernetes API server. This only varies from the `in_cluster_endpoint` when using an internal load balancer setup."
 }
 
 output "in_cluster_endpoint" {
   value = local.in_cluster_endpoint
+  description = "The internal endpoint for the Kubernetes API server. This only varies from the `in_cluster_endpoint` when using an internal load balancer setup."
 }
 output "api_server_cert_sans" {
```
msanft, line 10:
```suggestion
output "extra_api_server_cert_sans" {
```

terraform/infrastructure/aws/outputs.tf
```diff
@@ -17,21 +19,26 @@ output "api_server_cert_sans" {
...
 }
 
 output "uid" {
   value = local.uid
+  description = "The UID of the cluster."
 }
 
-output "initSecret" {
+output "init_secret" {
   value     = random_password.initSecret.result
```
msanft, line 31:
```suggestion
  value     = random_password.init_secret.result
```

terraform/infrastructure/aws/outputs.tf
```diff
@@ -17,21 +19,26 @@ output "api_server_cert_sans" {
...
   sensitive = true
+  description = "The init secret for the cluster."
 }
 
 output "name" {
   value = local.name
+  description = "The name of the cluster."
```
msanft, line 38:
```suggestion
  description = "Name of the cluster."
```

terraform/infrastructure/aws/outputs.tf
```diff
@@ -17,21 +19,26 @@ output "api_server_cert_sans" {
...
 }
 
 output "ip_cidr_nodes" {
   value = local.cidr_vpc_subnet_nodes
+  description = "The CIDR block for the nodes subnet."
```
msanft, line 43:
```suggestion
  description = "CIDR range of the node network."
```

terraform/infrastructure/aws/variables.tf
```diff
@@ -29,17 +29,17 @@ variable "node_groups" {
...
 variable "iam_instance_profile_control_plane" {
   type        = string
-  description = "Name of the IAM instance profile for control plane nodes"
+  description = "Name of the IAM instance profile for control plane nodes."
 }
 
 variable "ami" {
   type        = string
-  description = "AMI ID"
+  description = "AMI ID."
```
msanft, line 42:
```suggestion
  description = "Amazon Machine Image (AMI) ID for the cluster's nodes."
```

terraform/infrastructure/aws/variables.tf
```diff
@@ -29,17 +29,17 @@ variable "node_groups" {
...
   validation {
     condition     = length(var.ami) > 4 && substr(var.ami, 0, 4) == "ami-"
     error_message = "The image_id value must be a valid AMI id, starting with \"ami-\"."
```
msanft, line 45:
```suggestion
    error_message = "The \"ami\" value must be a valid AMI id, starting with \"ami-\"."
```

terraform/infrastructure/iam/aws/outputs.tf
```diff
@@ -1,7 +1,9 @@
 output "control_plane_instance_profile" {
   value = aws_iam_instance_profile.control_plane_instance_profile.name
+  description = "The name of the instance profile for the control plane."
 }
 
 output "worker_nodes_instance_profile" {
```
msanft, line 6:
```suggestion
output "worker_nodes_instance_profile_name" {
```

terraform/infrastructure/iam/aws/variables.tf
```diff
@@ -1,10 +1,10 @@
 variable "name_prefix" {
   type        = string
-  description = "Prefix for all resources"
+  description = "Prefix for all resources to easily identify related IAM resources."
```
msanft, line 3:
```suggestion
  description = "Name prefix to use on named resources."
```

terraform/infrastructure/aws/variables.tf
```diff
@@ -29,17 +29,17 @@ variable "node_groups" {
 
 variable "iam_instance_profile_worker_nodes" {
   type        = string
-  description = "Name of the IAM instance profile for worker nodes"
+  description = "Name of the IAM instance profile for worker nodes."
 }
 
 variable "iam_instance_profile_control_plane" {
```
msanft, line 35:
```suggestion
variable "iam_instance_profile_name_control_plane" {
```

terraform/infrastructure/aws/variables.tf
```diff
@@ -29,17 +29,17 @@ variable "node_groups" {
 
 variable "iam_instance_profile_worker_nodes" {
```
msanft, line 30:
```suggestion
variable "iam_instance_profile_name_worker_nodes" {
```

msanft (commented):
Only left some incomplete feedback for AWS for now as I couldn't attach suggestions to unedited parts of the code (e.g. modules within the `infrastructure/aws` module). I propose to prepare a commit with some additional changes I'd like to see here so you can cherry-pick it into here.

msanft:
[E2E Test](https://github.com/edgelesssys/constellation/actions/runs/7182231752)

terraform/infrastructure/aws/modules/public_private_subnet/variables.tf
```diff
@@ -30,5 +30,5 @@ variable "cidr_vpc_subnet_internet" {
 
 variable "tags" {
   type        = map(string)
-  description = "The tags to add to the resource."
+  description = "Zags to add to the resource."
```
daniel-weisse, line 33:
```suggestion
  description = "Tags to add to the resource."
```

daniel-weisse:
We should also run upgrade tests on all csps to make sure the renaming doesnt break anything

elchead:
lets also run self-managed infra and mini

msanft:
So we have at least one upgrade failure [here](https://github.com/edgelesssys/constellation/actions/runs/7193529595/job/19592188138) since I renamed some of the `output.tf` files into the canonical Terraform filename `outputs.tf`.  This is probably not easy to fix without a custom migration in the Constellation CLI, which is why I would say we leave that for now and maybe tackle it in a future version to not increase the failure vector of the changes made within this PR too much.

daniel-weisse:
> since I renamed some of the `output.tf` files into the canonical Terraform filename `outputs.tf`

We could fix that by running a pre-upgrade hook to delete the `outputs.tf` file from the Terraform working dir (the file should be backed up anyway) before running the actual upgrades.
On error/rollback we just restore the old state.

msanft:
> We could fix that by running a pre-upgrade hook to delete the `outputs.tf` file from the Terraform working dir (the file should be backed up anyway) before running the actual upgrades. On error/rollback we just restore the old state.

Yeah, that would be the kind of migration I roughly had in mind. Still, for me it is unclear whether we want to further increase the spectrum of possible failures by doing so in this release - also looking at the feature freeze planned on friday.

.github/actions/self_managed_create/action.yml
```diff
@@ -50,11 +50,10 @@ runs:
...
-          echo "iam_instance_profile_control_plane = \"$(yq '.provider.aws.iamProfileControlPlane' constellation-conf.yaml)\"" >> terraform.tfvars
-          echo "iam_instance_profile_worker_nodes = \"$(yq '.provider.aws.iamProfileWorkerNodes' constellation-conf.yaml)\"" >> terraform.tfvars
+          echo "iam_instance_profile_name_control_plane = \"$(yq '.provider.aws.iamProfileControlPlane' constellation-conf.yaml)\"" >> terraform.tfvars
+          echo "iam_instance_profile_name_worker_nodes = \"$(yq '.provider.aws.iamProfileWorkerNodes' constellation-conf.yaml)\"" >> terraform.tfvars
           echo "region = \"$(yq '.provider.aws.region' constellation-conf.yaml)\"" >> terraform.tfvars
           echo "zone = \"$(yq '.provider.aws.zone' constellation-conf.yaml)\"" >> terraform.tfvars
-          echo "ami = \"${{ steps.get_image.outputs.image_ref }}\"" >> terraform.tfvars
           echo "enable_snp = $(yq '.attestation | has("awsSEVSNP")' constellation-conf.yaml)" >> terraform.tfvars
         elif [[ "${{ inputs.cloudProvider }}" == 'azure' ]]; then
           echo "location = \"$(yq '.provider.azure.location' constellation-conf.yaml)\"" >> terraform.tfvars
```
elchead, line 59: I noticed that we call the attribute `region` in the IAM module, but location here.

elchead:
I agree we should not further increase the bug surface and get this merged quickly.

cli/internal/cloudcmd/iam.go
```diff
@@ -95,7 +95,7 @@ type GCPIAMConfig struct {
 
 // AzureIAMConfig holds the necessary values for Azure IAM configuration.
 type AzureIAMConfig struct {
-	Region           string
+	Location         string
```
daniel-weisse, line 98: I think the reason we had this named as `Region` before was to have it in line with AWS and GCP, both of which use region instead of Location.
Not sure if we want to keep this though, especially when its already called `Location` in the config.

daniel-weisse (approved):

cli/internal/cloudcmd/iam.go
elchead: Seems reasonable to name everything region then

malt3 (approved):
The renaming looks good to me.
Does this introduce breakage for the Constellation CLI? Or to rephrase: is it possible to upgrade from an older version of Constellation if we rename terraform variables and outputs?
I'm not that familiar with the exact interface between CLI and terraform code anymore so I'm genuinely not sure if this might be a problem or not.

msanft:
> is it possible to upgrade from an older version of Constellation if we rename terraform variables and outputs?

Yes, this is safe as long as you don't rename blocks, which we don't do here. In that case, we would have to use `moved` blocks. In our case, where we only rename variables and outputs, they are simply replaced as long as the underlying reference doesn't change (i.e. as long as they keep the same value).
Regarding the CLI / Terraform interface, this is safe as long as the interface is properly adjusted on the CLI side, which I tried to do in this PR. Also manual upgrade tests didn't show any unexpected behaviour here. Still trying to figure out why the e2e tests get stuck though.

PR #2712, opened 2023-12-13
ci: Terraform provider e2e tests

daniel-weisse:
### Context
We want to test our Terraform provider the same way we test the CLI.
Instead of writing new test, we want to adapt the existing tests to allow creation of the Constellation cluster using Terraform.

### Proposed change(s)

- Add a new GitHub workflow: Terraform e2e test
  - Same as manual e2e test (or self-managed-infra e2e test), but uses Terraform for cluster initialization
  - Uses the same e2e test template as other tests, which allows us to easily add Terraform provider tests to daily/weekly test runs

### Additional info

- [AB#3547](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3547)
- [quick e2e test](https://github.com/edgelesssys/constellation/actions/runs/7195061381)
- [e2e gcp](https://github.com/edgelesssys/constellation/actions/runs/7195429715)
- [e2e azure](https://github.com/edgelesssys/constellation/actions/runs/7195431527)
- [e2e aws](https://github.com/edgelesssys/constellation/actions/runs/7196873735)

### DIFF
--- a/.github/actions/terraform_apply/action.yml
+++ b/.github/actions/terraform_apply/action.yml
@@ -60,6 +60,7 @@ runs:
           attestation_variant = "${attestationVariant}"
           image_version       = "$(yq '.image' constellation-conf.yaml)"
           maa_url             = "$(yq '.infrastructure.azure.attestationURL' constellation-state.yaml)"
+          insecure            = true
         }
 
         data "constellation_image" "con_image" {
@@ -97,7 +98,7 @@ runs:
           gcp = {
             count = "$(yq '.provider | keys | .[0]' constellation-conf.yaml)" == "gcp" ? 1 : 0
             project_id          = "$(yq '.infrastructure.gcp.projectID' constellation-state.yaml)"
-            service_account_key = "$(cat $(yq '.provider.gcp.serviceAccountKeyPath' constellation-conf.yaml) | base64 -w0)"
+            service_account_key = sensitive("$(cat $(yq '.provider.gcp.serviceAccountKeyPath' constellation-conf.yaml) | base64 -w0)")
           }
           network_config = {
             ip_cidr_node    = "$(yq '.infrastructure.ipCidrNode' constellation-state.yaml)"
--- a/.github/actions/terraform_apply/action.yml
+++ b/.github/actions/terraform_apply/action.yml
@@ -1,0 +1,156 @@
+name: Terraform provider apply
+description: "Create/Apply a Constellation cluster using the Terraform provider."
+
+inputs:
+  cloudProvider:
+    description: "The cloud provider the test runs on."
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Create Terraform file
+      shell: bash
+      run: |
+        attestationVariant=""
+        case "$(yq '.attestation | keys | .[0]' constellation-conf.yaml)" in
+          "awsSEVSNP")
+            attestationVariant="aws-sev-snp"
+            ;;
+          "azureSEVSNP")
+            attestationVariant="azure-sev-snp"
+            ;;
+          "gcpSEVES")
+            attestationVariant="gcp-sev-es"
+            ;;
+          *)
+            echo "Unknown attestation variant: $(yq '.attestation | keys | .[0]' constellation-conf.yaml)"
+            exit 1
+            ;;
+        esac
+
+        cat << EOF > main.tf
+        terraform {
+          required_providers {
+            constellation = {
+              source  = "edgelesssys/constellation"
+              version = "$(yq '.microserviceVersion' constellation-conf.yaml | sed 's/^v//')"
+            }
+            random = {
+              source  = "hashicorp/random"
+              version = "3.6.0"
+            }
+          }
+        }
+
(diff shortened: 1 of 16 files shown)


.github/workflows/e2e-upgrade.yml
```diff
@@ -182,7 +182,7 @@ jobs:
           awsOpenSearchDomain: ${{ secrets.AWS_OPENSEARCH_DOMAIN }}
           awsOpenSearchUsers: ${{ secrets.AWS_OPENSEARCH_USER }}
           awsOpenSearchPwd: ${{ secrets.AWS_OPENSEARCH_PWD }}
-          selfManagedInfra: "false"
+          clusterCreation: "cli"
```
elchead, line 185: I think we should run an upgrade test on the terraform provider

.github/workflows/e2e-test-weekly.yml
```diff
@@ -291,7 +321,7 @@ jobs:
           test: ${{ matrix.test }}
           kubernetesVersion: ${{ matrix.kubernetes-version }}
           provider: ${{ matrix.provider }}
-          selfManagedInfra: ${{ matrix.selfManagedInfra == 'true' }}
+          clusterCreation: ${{ matrix.clusterCreation }}
```
elchead, line 324: we should remove the old tf-module test (at the bottom) and add the provider to the matrix.

elchead (commented):
the e2e test still calls the CLI for iam + create: https://github.com/edgelesssys/constellation/actions/runs/7195061381/job/19596929843#step:6:3685

.github/workflows/e2e-upgrade.yml
daniel-weisse: I don't think we currently support this (without a decent amount of extra work) as you would imagine it.
The main "upgrade test" is run by a [Go test](https://github.com/edgelesssys/constellation/tree/main/e2e/internal/upgrade), which expects to use a CLI binary.
We *could* create the initial Constellation cluster using Terraform, but I'm not sure if this would assert anything helpful

elchead:
also we were thinking of keeping the provider usage docs updated by sharing / referring to the code of the e2e test do you think it would make sense to extract a template (TF has its own template format: https://developer.hashicorp.com/terraform/language/functions/templatefile, but we could use any template language).

daniel-weisse:
>we were thinking of keeping the provider usage docs updated by sharing / referring to the code of the e2e test

Probably possible, but I would like to keep it separate from this PR.
Not sure if Terraforms templatefile function would be the right fit tho

msanft:
> also we were thinking of keeping the provider usage docs updated by sharing / referring to the code of the e2e test. Do you think it would make sense to extract a template (TF has its own template format: https://developer.hashicorp.com/terraform/language/functions/templatefile, but we could use any template language)?

I regard this as a very good idea, and generally sharing it is a great idea (as we also agreed on it) - but juding now, I think this isn't something we will be able to do until the release. I think for now we should have a somewhat reasonable and meaningful e2e test to test the functionality, and a "polished" example (See #2713) in the docs. Later on, we should work from both sides towards something which can be both used in the e2e tests, but is also polished and generic enough to use in the docs. I'll create a separate ticket for this.

.github/actions/terraform_apply/action.yml
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        resource "random_bytes" "measurement_salt" {
+          length = 32
+        }
+
+        data "constellation_attestation" "con_attestation" {
+          csp                 = "${{ inputs.cloudProvider }}"
+          attestation_variant = "${attestationVariant}"
+          image_version       = "$(yq '.image' constellation-conf.yaml)"
```
elchead, line 61: the MAAURL must be set azure sev snp

daniel-weisse: Must or should?
The e2e test on Azure passes without it being set: https://github.com/edgelesssys/constellation/actions/runs/7195431527
Will add it anyways for completeness, but if it is not actually a hard requirement, we shouldn't say so in the docs.

daniel-weisse:
>the e2e test still calls the CLI for iam + create

This is intentional. Since we want to test the Constellation cluster creation using Terraform (not infrastructure creation, which is left to the user), the easiest way is to just use our CLI to do the job.
This also makes it easier for us since we don't have to maintain an extra Terraform script for the infrastructure + IAM part and can easily set constellation config and statefile which might be required by other parts of the test.

elchead:
> > the e2e test still calls the CLI for iam + create
> 
> This is intentional. Since we want to test the Constellation cluster creation using Terraform (not infrastructure creation, which is left to the user), the easiest way is to just use our CLI to do the job. This also makes it easier for us since we don't have to maintain an extra Terraform script for the infrastructure + IAM part and can easily set constellation config and statefile which might be required by other parts of the test.

I disagree since we need to maintain the examples of #2713 at any rate, so it's better we test that they keep working.

elchead (commented):
changes lgtm

elchead (approved):

elchead (approved):
lgtm

PR #2715, opened 2023-12-13
terraform: use correct security group option for jump host

3u13r:
### Proposed change(s)

- Align security group naming for jump host
- use `vpc_security_group_ids` to set the security group since this is the correct setting (cf: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/instance#security_groups)

Otherwise on every `constellation apply` the are terraform changes which are applied, forcing the re-creating of the jump host.

### Additional info

### Checklist

- [x] Link to Milestone


msanft:
Do we need migrations here for upgrades?

3u13r:
> Do we need migrations here for upgrades?

No, the jump_host is only used for our debug clusters and not by customers.

daniel-weisse (approved):

msanft (approved):

PR #2718, opened 2023-12-14
cli: fix upgrades when using outdated Kubernetes patch version

daniel-weisse:
### Context
Users may choose to not upgrade their Kubernetes version when running a Constellation upgrade.
This includes not upgrading a patch version of Kubernetes.
We use the Kubernetes version set in a user's config to determine which image to use for the Constellation deployed cloud-controller-manager and cluster-autoscaler.
Currently, the CLI incorrectly sets the k8s version to an empty string if a user chooses to apply an upgrade with a no longer supported Kubernetes patch version.

### Proposed change(s)

- Fix the issue by translating an outdated user's Kubernetes version to a currently supported Kubernetes version
  - Kubernetes upgrades are still skipped
  - This version is only used to determine which image to use for the cloud-controller-manager and cluster-autoscaler. Both of which are independent of the Kubernetes patch version

### Additional info
- [e2e test targeting Kubernetes v1.27.7](https://github.com/edgelesssys/constellation/actions/runs/7222740237)

### DIFF
--- a/cli/internal/cmd/apply.go
+++ b/cli/internal/cmd/apply.go
@@ -41,6 +41,7 @@ import (
 	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	xsemver "golang.org/x/mod/semver"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 )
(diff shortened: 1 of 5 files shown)


msanft (approved):

cli/internal/cmd/apply.go
```diff
@@ -545,6 +546,16 @@ func (a *applyCmd) validateInputs(cmd *cobra.Command, configFetcher attestationc
...
+
+			validVersionString, err := versions.ResolveK8sPatchVersion(xsemver.MajorMinor(string(conf.KubernetesVersion)))
+			if err != nil {
+				return nil, nil, fmt.Errorf("resolving Kubernetes patch version: %w", err)
+			}
+			validVersion, err = versions.NewValidK8sVersion(validVersionString, true)
+			if err != nil {
+				return nil, nil, fmt.Errorf("parsing Kubernetes version: %w", err)
+			}
+
```
elchead, line 558: This needs to be moved outside of the skipK8s if block. Otherwise it does not solve the issue. This is one of the reasons I wanted to rename the `contains` method to `skip`;)

elchead (changes requested):

cli/internal/cmd/apply.go
elchead: https://github.com/edgelesssys/constellation/commit/3ef8cee8612b86ef5b1ef64901d43b749393b76a

msanft: I thought the contents of the KubernetesVersion field in the config are irrelevant otherwise, since the upgrade phase is skipped. However, thinking about it again and looking at the code, it seems to me that we should, as Adrian mentioned, adjust it in both cases. The config needs to contain a correct version once the service versions are selected

elchead: yes, this is relevant for helm since it uses the K8s version to determine the chart config.

cli/internal/cmd/apply_test.go
```diff
@@ -396,6 +398,28 @@ func TestValidateInputs(t *testing.T) {
 			},
 			wantPhases: newPhases(skipInfrastructurePhase, skipImagePhase, skipK8sPhase),
 		},
+		"[update] k8s patch version no longer supported, user confirms upgrade": {
```
elchead, line 401:
```suggestion
		"[upgrade] k8s patch version no longer supported, user confirms to skip k8s and continue upgrade. Valid K8s patch version is used in config afterwards": {
```

elchead (approved):

elchead (approved):

PR #2721, opened 2023-12-14
feat: pin cert-manager image to sha256 checksum

elchead:
### Context

Pin the cert-manager related images to their sha256 checksum. 

How I tested it:
upgraded a GCP, v2.13 cluster with
```
./constellation apply --debug --skip-phases infrastructure,k8s,image --force
```

### Proposed change(s)

-

### Additional info

-  [AB#3612](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3612)
Had to rebase and cherry pick to be able to test it locally.

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/internal/constellation/helm/generateCertManager.sh
+++ b/internal/constellation/helm/generateCertManager.sh
@@ -35,21 +35,19 @@ get_sha256_hash() {
 }
 
 echo "Pinning cert-manager images..."
-yq=$(realpath @@YQ@@)
-stat "${yq}" >> /dev/null
 v=$(get_sha256_hash "cert-manager-controller")
-${yq} eval -i '.image.digest = "sha256:'"$v"'"' charts/cert-manager/values.yaml
+yq eval -i '.image.digest = "sha256:'"$v"'"' charts/cert-manager/values.yaml
 
 v=$(get_sha256_hash "cert-manager-webhook")
-${yq} eval -i '.webhook.image.digest = "sha256:'"$v"'"' charts/cert-manager/values.yaml
+yq eval -i '.webhook.image.digest = "sha256:'"$v"'"' charts/cert-manager/values.yaml
 
 v=$(get_sha256_hash "cert-manager-cainjector")
-${yq} eval -i '.cainjector.image.digest = "sha256:'"$v"'"' charts/cert-manager/values.yaml
+yq eval -i '.cainjector.image.digest = "sha256:'"$v"'"' charts/cert-manager/values.yaml
 
 v=$(get_sha256_hash "cert-manager-acmesolver")
-${yq} eval -i '.acmesolver.image.digest = "sha256:'"$v"'"' charts/cert-manager/values.yaml
+yq eval -i '.acmesolver.image.digest = "sha256:'"$v"'"' charts/cert-manager/values.yaml
 
 v=$(get_sha256_hash "cert-manager-ctl")
-${yq} eval -i '.startupapicheck.image.digest = "sha256:'"$v"'"' charts/cert-manager/values.yaml
+yq eval -i '.startupapicheck.image.digest = "sha256:'"$v"'"' charts/cert-manager/values.yaml
 
 echo # final newline
--- a/internal/constellation/helm/generateCertManager.sh
+++ b/internal/constellation/helm/generateCertManager.sh
@@ -5,40 +5,33 @@ set -o errtrace
 shopt -s inherit_errexit
 
 echo "Pulling cert-manager Helm chart..."
+version="1.12.6"
 
 function cleanup {
-  rm -r "charts/cert-manager/README.md" "charts/cert-manager-v1.12.6.tgz"
+  rm -r "charts/cert-manager/README.md" "charts/cert-manager-v${version}.tgz"
 }
 
 trap cleanup EXIT
 
-version="1.12.6"
 helm pull cert-manager \
   --version ${version} \
   --repo "https://charts.jetstack.io" \
   --untar \
   --untardir "charts"
 
-
 get_sha256_hash() {
   local component="$1"
   local url="https://quay.io/v2/jetstack/${component}/manifests/v${version}"
   local hash
-  set -e  # Exit immediately if a command fails
   has_error=$(curl -s -L -H "Accept: application/vnd.docker.distribution.manifest.v2+json" "$url" | jq 'has("errors")')
-
 
   # Check if 'errors' attribute exists
   if [ "$has_error" = "true" ]; then
-      echo "Errors attribute found. Exiting with status 1."
-      exit 1
+    echo "Errors attribute found. Exiting with status 1."
+    exit 1
   fi
 
-  hash=$(curl -s -L -H "Accept: application/vnd.docker.distribution.manifest.v2+json" "$url" | sha256sum | awk '{print $1}')
-
-  set +e  # Disable the 'set -e' option to avoid exiting on subsequent commands
-
-  echo "$hash"
+  curl -s -L -H "Accept: application/vnd.docker.distribution.manifest.v2+json" "$url" | sha256sum | awk '{print $1}'
 }
 
 echo "Pinning cert-manager images..."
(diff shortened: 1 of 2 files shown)


internal/constellation/helm/generateCertManager.sh
```diff
@@ -18,4 +18,20 @@ helm pull cert-manager \
   --untar \
   --untardir "charts"
 
+echo "Pinning cert-manager images..."
```
3u13r, line 21: I assume that yq already formats the helm charts. Since those values are exposed as helm values, I favor just setting them like all the other helm values.

3u13r (dismissed):

internal/constellation/helm/generateCertManager.sh
elchead: I don't like that approach because we then separate the helm version (set in the script) from the according hashes.

daniel-weisse: Setting the hashes later (i.e. "like all the other helm values") would allow us do update the hashes using renovate.
We already do this for the cluster-autoscaler and cloud-controller-manager images, and some others.
See our [versions.go](https://github.com/edgelesssys/constellation/blob/a1dbd13f9577e23d99012416ab1b97c83ecfdedb/internal/versions/versions.go) file.
Not 100% how/if we want that for charts we pull in as third party dependencies since the versions should usually not be updated for one version of a chart, but it could make things easier since we wouldn't need to manually recalculate the image hashes every time we update the chart

elchead: I assume @burgerdev motivation was to protect against a malicious version overwrite in the registry?
I agree that it would be good to use renovate.

internal/constellation/helm/generateCertManager.sh
```diff
@@ -18,4 +18,21 @@ helm pull cert-manager \
...
+# Manually copy the sha256 digest from the corresponding repository:
+# quay.io/jetstack/cert-manager-controller
+yq -i '(.image.digest) = "sha256:fb2546fe51e49206dbf72bb0d6f909a0018eda0c2b024547b03d3f3d604e4c5e"' charts/cert-manager/values.yaml
```
burgerdev, line 24: We could auto-generate the hashes along the lines of

```bash
curl -L -H "Accept: application/vnd.docker.distribution.manifest.v2+json" https://quay.io/v2/jetstack/${component}/manifests/v${version} | sha256sum
```

internal/constellation/helm/generateCertManager.sh
burgerdev: I'm facing more or less the same issue in my other hash pinning attempts, and this is the approach that I feel most comfortable with so far:

1. We want to specify a (chart|k8s|.*) version `$SOMEWHERE`, which ideally would be renovated.
2. As a consequence of that version choice, a list of required container images emerges.
3. We automatically derive the image hashes for those
4. and put them `$SOMEWHERE`.

The place for (1) is currently this script. I think that it would make sense to do (2-4) here as well, at least in the short term. If we can come up with 

* a place for storing (1) 
  - _globally_
* a generator with input (1) and output (2) 
  - _per use case_
* a generator with input (2) and output (3) 
  - _globally_
* a generator with input (3) 
  - _for a couple of different use cases_, e.g. `versions.go`, `values.go`

life would be a lot easier, but I think that's outside the scope of this change. However, I do think it's worthwile considering.

Long story short, +1 to keeping the chart version and the image hashes close together.

daniel-weisse:
Since finding the hashes to pin for images not controlled by us seems to be a recurring issue, and is not easily solved:
How about just keeping it simple for this PR, and then open an RFC to discuss options to have the hash finding centralized at some place?

internal/constellation/helm/generateCertManager.sh
elchead: thanks, implemented that now

internal/constellation/helm/generateCertManager.sh
```diff
@@ -12,10 +12,51 @@ function cleanup {
...
+  if [ "$has_error" = "true" ]; then
+      echo "Errors attribute found. Exiting with status 1."
+      exit 1
+  fi
+
+  hash=$(curl -s -L -H "Accept: application/vnd.docker.distribution.manifest.v2+json" "$url" | sha256sum | awk '{print $1}')
+
+  set +e  # Disable the 'set -e' option to avoid exiting on subsequent commands
+
+  echo "$hash"
```
burgerdev, line 41: `set -e` is inherited from line 3, and `set +e` resets globally. I don't think you need to change this at all.

```suggestion
  has_error=$(curl -s -L -H "Accept: application/vnd.docker.distribution.manifest.v2+json" "$url" | jq 'has("errors")')

  # Check if 'errors' attribute exists
  if [ "$has_error" = "true" ]; then
      echo "Errors attribute found. Exiting with status 1."
      exit 1
  fi

  curl -s -L -H "Accept: application/vnd.docker.distribution.manifest.v2+json" "$url" | sha256sum | awk '{print $1}'
```

internal/constellation/helm/generateCertManager.sh
```diff
@@ -12,10 +12,51 @@ function cleanup {
 
 trap cleanup EXIT
 
+version="1.12.6"
```
burgerdev, line 15: This variable needs to be used in the cleanup function at line 10.

burgerdev (changes requested):

burgerdev (approved):

internal/constellation/helm/generateCertManager.sh
```diff
@@ -5,17 +5,51 @@ set -o errtrace
...
+  local hash
+  has_error=$(curl -s -L -H "Accept: application/vnd.docker.distribution.manifest.v2+json" "$url" | jq 'has("errors")')
+
+  # Check if 'errors' attribute exists
+  if [ "$has_error" = "true" ]; then
+    echo "Errors attribute found. Exiting with status 1."
+    exit 1
+  fi
+
+  curl -s -L -H "Accept: application/vnd.docker.distribution.manifest.v2+json" "$url" | sha256sum | awk '{print $1}'
```
3u13r, line 34:
```suggestion
  curl -fsSL -H "Accept: application/vnd.docker.distribution.manifest.v2+json" "$url" | sha256sum | awk '{print $1}'
```

internal/constellation/helm/generateCertManager.sh
```diff
@@ -5,17 +5,49 @@ set -o errtrace
...
 function cleanup {
-  rm -r "charts/cert-manager/README.md" "charts/cert-manager-v1.12.6.tgz"
+  rm -r "charts/cert-manager/README.md" "charts/cert-manager-v${version}.tgz"
 }
 
 trap cleanup EXIT
 
 helm pull cert-manager \
-  --version 1.12.6 \
+  --version ${version} \
```
3u13r, line 17:
```suggestion
  --version "${version}" \
```

internal/constellation/helm/generateCertManager.sh
```diff
@@ -5,17 +5,49 @@ set -o errtrace
...
   --untardir "charts"
 
+get_sha256_hash() {
+  local component="$1"
+  local url="https://quay.io/v2/jetstack/${component}/manifests/v${version}"
+  local hash
+  has_error=$(curl -s -L -H "Accept: application/vnd.docker.distribution.manifest.v2+json" "$url" | jq 'has("errors")')
+
+  # Check if 'errors' attribute exists
+  if [ "$has_error" = "true" ]; then
```
3u13r, line 29:
```suggestion
  if [[ "$has_error" = "true" ]]; then
```

internal/constellation/helm/generateCertManager.sh
```diff
@@ -5,17 +5,49 @@ set -o errtrace
...
+yq eval -i '.webhook.image.digest = "sha256:'"$v"'"' charts/cert-manager/values.yaml
+
+v=$(get_sha256_hash "cert-manager-cainjector")
+yq eval -i '.cainjector.image.digest = "sha256:'"$v"'"' charts/cert-manager/values.yaml
+
+v=$(get_sha256_hash "cert-manager-acmesolver")
+yq eval -i '.acmesolver.image.digest = "sha256:'"$v"'"' charts/cert-manager/values.yaml
+
+v=$(get_sha256_hash "cert-manager-ctl")
+yq eval -i '.startupapicheck.image.digest = "sha256:'"$v"'"' charts/cert-manager/values.yaml
```
3u13r, line 51:
```suggestion
yq eval -i ".startupapicheck.image.digest = \"sha256:${v}\"" charts/cert-manager/values.yaml
```
Also see [SC2250](https://www.shellcheck.net/wiki/SC2250) everywhere a variable is referenced.

3u13r (dismissed):

PR #2724, opened 2023-12-15
ci: parallelize upgrade e2e test

msanft:
### Context

While working a lot with the e2e test, it became apparent that the curent start-to-feedback time is quite slow. Therefore, the test could be sped up a little.

### Proposed change(s)

- Parallelize building the upgrade target CLI and spinning up the upgrade origin cluster to be concurrent, and then perform the actual e2e test afterwards.

### Additional info

- [AB#3619](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3619)
- [Upgrade E2E Test](https://github.com/edgelesssys/constellation/actions/runs/7264085929)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/actions/artifact_download/action.yml
+++ b/.github/actions/artifact_download/action.yml
@@ -9,7 +9,7 @@ inputs:
     description: 'Download to a specified path.'
     required: false
     default: ./
-  encryptionSecret:
+  encryption-secret:
     description: 'The secret to use for decrypting the artifact.'
     required: true
 
@@ -37,4 +37,4 @@ runs:
       shell: bash
       run: |
         mkdir -p ${{ inputs.path }}
-        unzip -P '${{ inputs.encryptionSecret }}' -qq -d ${{ inputs.path }} ${{ steps.tempdir.outputs.directory }}/archive.zip
+        unzip -P '${{ inputs.encryption-secret }}' -qq -d ${{ inputs.path }} ${{ steps.tempdir.outputs.directory }}/archive.zip
--- a/.github/actions/artifact_download/action.yml
+++ b/.github/actions/artifact_download/action.yml
@@ -9,7 +9,7 @@ inputs:
     description: 'Download to a specified path.'
     required: false
     default: ./
-  encryption-secret:
+  encryptionSecret:
     description: 'The secret to use for decrypting the artifact.'
     required: true
 
@@ -37,4 +37,4 @@ runs:
       shell: bash
       run: |
         mkdir -p ${{ inputs.path }}
-        unzip -P '${{ inputs.encryption-secret }}' -qq -d ${{ inputs.path }} ${{ steps.tempdir.outputs.directory }}/archive.zip
+        unzip -P '${{ inputs.encryptionSecret }}' -qq -d ${{ inputs.path }} ${{ steps.tempdir.outputs.directory }}/archive.zip
--- a/.github/workflows/e2e-upgrade.yml
+++ b/.github/workflows/e2e-upgrade.yml
@@ -240,6 +240,7 @@ jobs:
             constellation-conf.yaml
             constellation-state.yaml
             constellation-mastersecret.json
+            gcpServiceAccountKey.json
           encryptionSecret: ${{ secrets.ARTIFACT_ENCRYPT_PASSWD }}
 
   e2e-upgrade:
@@ -400,6 +401,7 @@ jobs:
             constellation-conf.yaml
             constellation-state.yaml
             constellation-mastersecret.json
+            gcpServiceAccountKey.json
           encryptionSecret: ${{ secrets.ARTIFACT_ENCRYPT_PASSWD }}
 
   clean-up:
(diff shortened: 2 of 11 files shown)


.github/workflows/e2e-upgrade.yml
```diff
@@ -212,6 +289,12 @@ jobs:
         with:
           azure_credentials: ${{ secrets.AZURE_E2E_IAM_CREDENTIALS }}
 
+      - name: Download CLI
+        uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
+        with:
+          name: cli
```
daniel-weisse, line 295:
```suggestion
          name: constellation
```
The following steps assume the CLI is at `build/constellation` not `build/cli`

daniel-weisse (commented):
We need to upload the working dir from the `create-cluster` job, and download it in the `e2e-upgrade` job, otherwise we are missing some files to run the upgrade.

We should also either add a terminate-on-failure step to the `create-cluster` job in case anything goes wrong in that job, or split the terminate step into its own job, which runs if either `create-cluster`, or `e2e-upgrade` fails.
If ran as its own job, the terminate action requires access to the working directory, which means we would need to unconditionally upload the working dir from `create-cluster`, even if a step fails.

.github/workflows/e2e-upgrade.yml
```diff
@@ -184,12 +216,77 @@ jobs:
...
-        uses: ./.github/actions/build_cli
+      - name: Remove Terraform plugin cache
+        if: always()
+        run: |
+          rm -rf constellation-terraform/.terraform
+          rm -rf constellation-iam-terraform/.terraform
+
+      - name: Upload Working Directory
+        if: always()
+        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
```
daniel-weisse, line 227: Why not zip the working directory and upload as a single file?

.github/workflows/e2e-upgrade.yml
```diff
@@ -303,7 +487,7 @@ jobs:
 
       - name: Notify about failure
         if: |
-          failure() &&
+          always() &&
```
daniel-weisse, line 490: not sure if this is for debugging, but we should only create failure tickets if the job actually failed

.github/workflows/e2e-upgrade.yml
msanft: I wanted to have a small archive, zipping the directory would just add extra steps here, as we would have to delete all the files we don't want to zip & upload (e.g. binaries) and then unpack the archive afterwards

msanft:
Another E2E Test: https://github.com/edgelesssys/constellation/actions/runs/7273685348

daniel-weisse (approved):
Should merge after https://github.com/edgelesssys/constellation/pull/2567

msanft:
Blocked until we can encrypt multiple-file artifacts

.github/actions/artifact_download/action.yml
```diff
@@ -9,7 +9,7 @@ inputs:
     description: 'Download to a specified path.'
     required: false
     default: ./
-  encryptionSecret:
+  encryption-secret:
```
daniel-weisse, line 12: fix name

.github/workflows/e2e-upgrade.yml
```diff
@@ -111,53 +111,85 @@ jobs:
...
       id-token: write
       checks: write
       contents: read
       packages: write
-    needs: [split-nodeCount]
     steps:
       - name: Checkout
         if: inputs.gitRef == 'head'
-        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
```
daniel-weisse, line 125: Dont downgrade the checkout action please

.github/workflows/e2e-upgrade.yml
```diff
@@ -183,14 +215,80 @@ jobs:
...
+      contents: read
+      packages: write
+    needs:
+      - split-nodeCount
+      - build-target-cli
+      - create-cluster
+    steps:
+      - name: Checkout
+        if: inputs.gitRef == 'head'
+        uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
```
daniel-weisse, line 255: Old checkout action version

.github/workflows/e2e-upgrade.yml
```diff
@@ -183,14 +215,80 @@ jobs:
...
+        uses: ./.github/actions/versionsapi
+        with:
+          command: latest
+          ref: main
+          stream: nightly
+
+      - name: Simulate patch upgrade
+        if: inputs.simulatedTargetVersion != ''
+        run: |
+          echo ${{ inputs.simulatedTargetVersion }} > version.txt
```
daniel-weisse, line 291: Shouldnt this be done in the `build-target-cli` job?

msanft:
Upgrade tests suceeded for all clouds:
- https://github.com/edgelesssys/constellation/actions/runs/7581234827
- https://github.com/edgelesssys/constellation/actions/runs/7581236626
- https://github.com/edgelesssys/constellation/actions/runs/7581238230

PR #2713, opened 2023-12-13
terraform-provider: add usage examples

msanft:
### Context

We want to provide plug-and-play examples for using the Constellation Terraform provider in combination with our infrastructure referece modules. Therefore, we need parametrized examplary configurations for each CSP to which we can link in the documentation.

### Proposed change(s)

- Add provider usage examples for all CSPs.

### Additional Information
- [E2E Test AWS](https://github.com/edgelesssys/constellation/actions/runs/7225560294/job/19689269676)
- [E2E Test Azure](https://github.com/edgelesssys/constellation/actions/runs/7225559392/job/19689264016)
- [E2E Test GCP](https://github.com/edgelesssys/constellation/actions/runs/7225558351/job/19689258974)

### Checklist

- [x] Add labels (e.g., for changelog category)
- [x] Is PR title adequate for changelog?
- [x] Link to Milestone

PR #2730, opened 2023-12-18
terraform-provider: warn about microservice version changes

msanft:
### Context

In the Constellation CLI, we ask the user to confirm microservice upgrades due to potentially breaking CRD changes.
In the Terraform-provider, we cannot prompt for a custom confirmation (and probably don't want to either), so we want to have a warning being shown to the user when a microservice upgrade is involved in a pre-apply/destroy plan. 

### Proposed change(s)

- Check if the planned resource change contains a microservice version change and warn the user about the potential consequences. Theoretically, this check could be more sophisticated by checking for custom CRDs or if a cert-manager upgrade is to be performed, but this would require a lot of client-setup which probably goes overboard for the use-case here.
- Update the data source examples in the docs with the new image passing. (#2713)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/terraform-provider-constellation/internal/provider/cluster_resource.go
+++ b/terraform-provider-constellation/internal/provider/cluster_resource.go
@@ -365,9 +365,7 @@ func (r *ClusterResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
 		}
 
 		// Warn the user about possibly destructive changes in case microservice changes are to be applied.
-		if currentState.MicroserviceVersion.ValueString() != "" && // current version is set (i.e. no create operation)
-			plannedState.MicroserviceVersion.ValueString() != "" && // planned version is set (i.e. no destroy operation)
-			currentState.MicroserviceVersion.ValueString() != plannedState.MicroserviceVersion.ValueString() {
+		if currentState.MicroserviceVersion.ValueString() != plannedState.MicroserviceVersion.ValueString() {
 			resp.Diagnostics.AddWarning("Microservice version change",
 				"Changing the microservice version can be a destructive operation.\n"+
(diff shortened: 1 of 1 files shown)


daniel-weisse (approved):
lgtm

terraform-provider-constellation/internal/provider/cluster_resource.go
```diff
@@ -344,6 +345,37 @@ func (r *ClusterResource) Configure(_ context.Context, req resource.ConfigureReq
...
+
+		// Read plannedState supplied by Terraform runtime into the model
+		var plannedState ClusterResourceModel
+		resp.Diagnostics.Append(req.Plan.Get(ctx, &plannedState)...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+
+		// Warn the user about possibly destructive changes in case microservice changes are to be applied.
+		if currentState.MicroserviceVersion.ValueString() != "" && // current version is set (i.e. no create operation)
```
elchead, line 368: the microservice version is allowed to be empty (lockstep). what does the comment mean?

msanft: True. This is now obsolete with the whole check only running on an upgrade, and wouldn't have worked with the lock-step behaviour you've mentioned anyways.

malt3 (approved):

terraform-provider-constellation/internal/provider/cluster_resource.go
elchead: Since the microservice version is optional and K8s and image are also lockstepped, we should to that.

elchead (approved):

PR #2732, opened 2023-12-18
ci: don't run unit tests in integration test workflow

burgerdev:
### Context

The `integration` config runs all tests - unit and integration. Since we already have a workflow for unit tests, we should not be running these again in the integration test workflow. There's a dedicated config for running only integration tests: `integration-only`.

### Proposed change(s)

- Use the `integration-only` config in the integration test workflow.

### Checklist
- [x] Link to Milestone


malt3 (approved):
Thanks!

PR #2738, opened 2023-12-19
operators: fix flaky env test

burgerdev:
### Context

The test for `PendingNode` objects includes a step to accelerate reconciliation by modifying the spec. Since the step we want to accelerate is deletion of the object, there's a race between the operator deleting the node and the test runner modifying the spec. The operator rarely wins the race, but it happened at least once. 

### Proposed change(s)
- Allow the acceleration step to fail with a `NotFound` error.

### Related issue
- Fixes edgelesssys/issues#141
- Fixes edgelesssys/issues#3

### Checklist

- [x] Link to Milestone


3u13r (approved):
nice catch, LGTM

PR #2739, opened 2023-12-19
terraform: add module deprecation notice

msanft:
### Context

With Constellation v2.14.0, we introduced the Constellation Terraform provider, which is now the recommended way to manage Constellation clusters through Terraform. With that, the legacy modules can be deprecated (our docs don't link to them anymore anyways) and should be removed with or before v2.15.0.

### Proposed change(s)

- Add a deprecation notice in the directory of the legacy module. 

### Additional info

- ~~TODO: Leave a link to the import-docs of the `constellation_cluster` resource, once provider documentation is published.~~

### Checklist

- [x] Link to Milestone

### DIFF
--- a/terraform/legacy-module/README.md
+++ b/terraform/legacy-module/README.md
@@ -3,6 +3,4 @@
 > [!WARNING]
 > The Constellation Terraform modules are deprecated, and support will be discontinued in v2.15.0.
 > To continue managing Constellation clusters through Terraform, you can use the [Constellation Terraform provider](https://docs.edgeless.systems/constellation/workflows/terraform-provider).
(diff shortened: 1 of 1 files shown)


elchead:
Regarding the TODO, it seems better to me to put the migration (import) instructions close to the deprecation notice. But I think a simple mention in the release notes is enough?

msanft:
> Regarding the TODO, it seems better to me to put the migration (import) instructions close to the deprecation notice. But I think a simple mention in the release notes is enough?

See the most recent commit

elchead (approved):

thomasten (approved):

PR #2746, opened 2023-12-20
docs: mention TF provider more prominently

thomasten:
As the TF provider is a major feature, it should be a bit more visible.


elchead (approved):
we should wait with the promotion until it is actually available in the registry

PR #2719, opened 2023-12-14
Pin Kubernetes container image hashes 

burgerdev:
### Context

The container images pulled by `kubeadm` should be verified by Constellation. This PR pins the images for `kube-apiserver`, `kube-scheduler` and `kube-controller-manager`.

### Proposed change(s)

* Add patches that pin image hashes to `versions.go`.
* Configure the bootstrapper to pass patches to `kubeadm`.
* Configure the operator to pass patches to the upgrade-agent.
* Configure the upgrade-agent to pass patches to `kubeadm`.

In an upgrade from 2.13 to 2.14, the upgrade-agent won't immediately pin the image because it does not understand the proto field with the additional components and does not pass on the patches. However, the VM will eventually get replaced, and the new upgrade-agent and bootstrapper will now be able to pin the image correctly.

### Additional info

- [AB#3565](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3565)
- [RFC 015](https://github.com/edgelesssys/constellation/blob/37580009fe02010dd862bd7ff95633cf9814a955/rfc/015-trusted-kubernetes-images.md)

### Checklist

- [x] Add labels (e.g., for changelog category)
- [x] Is PR title adequate for changelog?
- [x] Link to Milestone
- [x] Run e2e upgrade test
  * https://github.com/edgelesssys/constellation/actions/runs/7247085799

PR #2764, opened 2023-12-21
versions: generate k8s image patches (incl etcd)

burgerdev:
### Context

#2719 introduced Kubernetes component patches that are to be passed to `kubeadm` during init and upgrade. These patches were manually generated and did not include the `etcd` component.

### Proposed change(s)

- Add library support for fetching image hashes and generating JSON patches.
- Use the kubeadm Golang libraries to derive the etcd version for a given k8s version.
- Include patches in  `hash-generator`.

### Additional info

- [AB#3616](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3616)
- RFC 015

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/versions/hash-generator/generate.go
+++ b/internal/versions/hash-generator/generate.go
@@ -4,12 +4,20 @@ Copyright (c) Edgeless Systems GmbH
 SPDX-License-Identifier: AGPL-3.0-only
 */
 
+// hash-generator updates the binary hashes and kubeadm patches in versions.go in place.
+//
+// This binary is usually invoked by the //bazel/ci:go_generate target, but you can run it
+// manually, too. Clear a hash or a data URL in versions.go and execute
+//
+//	bazel run //internal/versions/hash-generator -- --update=false $PWD/internal/versions/versions.go
 package main
 
 import (
 	"bytes"
 	"context"
 	"crypto/sha256"
+	"encoding/json"
+	"flag"
 	"fmt"
 	"go/ast"
 	"go/parser"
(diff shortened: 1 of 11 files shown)


internal/versions/hash-generator/generate.go
```diff
@@ -19,52 +27,111 @@ import (
...
+	}})
+	if err != nil {
+		return "", err
+	}
+	return dataurl.New(content, "application/json").String(), nil
+}
+
+// hashURLContent downloads a binary blob from the given URL and calculates its SHA256 hash.
+//
+// URLs passed to this function are expected to have upstream signatures with a .sha256 suffix. This upstream signatue
```
malt3, line 105:
```suggestion
// URLs passed to this function are expected to have upstream signatures with a .sha256 suffix. This upstream signature
```

internal/versions/hash-generator/generate.go
```diff
@@ -19,52 +27,111 @@ import (
...
+// will be verified, too.
+func hashURLContent(url string) (string, error) {
+	resp, err := http.Get(url)
```
malt3, line 108: The linter wants to use a version with ctx

malt3 (approved):
Otherwise looks good to me!

internal/versions/hash-generator/generate.go
burgerdev: Fixed

internal/versions/hash-generator/generate.go
burgerdev: I fixed that by ignoring the linter, because I believe it does not apply in our situtation here. Please let me know whether you agree.

PR #2765, opened 2023-12-21
ci: fix artifact upload in image build pipeline

daniel-weisse:
### Context
https://github.com/edgelesssys/constellation/pull/2567 introduced some minor issues in the image build pipeline.
This PR aims to fix the issues and clean up some things

### Proposed change(s)

- Fix encrypted uploading of multiple artifacts in one action
- Fix downloading of encrypted artifacts if the directory we want to extract to does not yet exist on the actions runner
- Fix archives containing directory structure that shouldn't be there
  - If we specify we want a file at `someDir/someSubDir/targetFile`, the action would not create an archive containing `targetFile`, but `someDir/someSubDir/targetFile`. This is different behavior from the regular artifact upload action
  - We can't use the `-j` (junk paths) option from `unzip`, because that would destroy **all** directory structure on unpacking. This is a problem if we want to specifically upload directories, e.g. `somePath/dir1` and `somePath/dir2`, each containing multiple files. If we unpack an archive of those directories using the `-j` option, the output directory would contain all files of the directories, but not the directories themselves (`dir1` and `dir2`)
- Rename `encryption-secret` input to `encryptionSecret` to stay consistent with input naming formats across our actions
- Remove access to the `ARTIFACT_ENCRYPT_PASSWD` from `.github/actions/e2e_test`, since actions are not allowed to access secrets

### Related issue
- Fixes https://github.com/edgelesssys/issues/issues/146

### Additional info
- [image build pipeline](https://github.com/edgelesssys/constellation/actions/runs/7286853467)

### DIFF
--- a/.github/actions/artifact_download/action.yml
+++ b/.github/actions/artifact_download/action.yml
@@ -36,4 +36,5 @@ runs:
     - name: Decrypt and unzip archive
       shell: bash
       run: |
+        mkdir -p ${{ inputs.path }}
         unzip -P '${{ inputs.encryption-secret }}' -qq -d ${{ inputs.path }} ${{ steps.tempdir.outputs.directory }}/archive.zip
--- a/.github/actions/artifact_download/action.yml
+++ b/.github/actions/artifact_download/action.yml
@@ -9,7 +9,7 @@ inputs:
     description: 'Download to a specified path.'
     required: false
     default: ./
-  encryption-secret:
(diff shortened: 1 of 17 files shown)


malt3 (approved):

malt3:
[Run is green](https://github.com/edgelesssys/constellation/actions/runs/7290484190)

PR #2771, opened 2023-12-22
ci: don't use disallowed action when uploading scorecard

msanft:
### Context

The Scorecard uploading workflow broke (on main), causing CI checks to fail and our scorecard not to update. This got caused by the artifact encryption, which is a wrapper to GitHub's `upload_artifact` action, that was previously used directly in the scorecard workflow. The scorecard workflow has a restrictive [set of actions](https://github.com/ossf/scorecard-action#restrictions-on-the-job-containing-ossfscorecard-action) that may be called in the workflow the scorecard is created in, which includes the `upload_artifact` action from GitHub, but naturally not our custom wrapper around it. Therefore, we can fix the problem by going back to using GitHub's action for this workflow.

### Proposed change(s)

- Use GitHub's `upload_artifact` action in the scorecard workflow, instead of our custom wrapper.

### Additional Information
- Unsure how to test this, since the scorecard action seems to only be able to be ran from the default branch. (`main`)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -30,12 +30,11 @@ jobs:
           publish_results: true
 
       - name: Upload artifact
-        uses: ./.github/actions/artifact_upload
+        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
         with:
           name: SARIF file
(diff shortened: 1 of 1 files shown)


burgerdev (approved):
lgtm, thanks!

PR #2772, opened 2023-12-22
ci: selectively remove artifact encryption

burgerdev:
### Context

The encrypted artifact actions introduced in #2567 are not drop-in replacements for the regular actions: if you use the artifact upload action multiple times for the same artifact, the default action will append to the outputs, while our action replaces the artifact.

This is only a problem when we actually upload multiple times to the same artifact, usually from matrix runs. I did a quick search for call sites of the upload action, and the only places where the artifact name is fixed but the paths are matrix-dependent were `build-os-image` and `reproducible-builds`. This PR fixes these call sites.

cc @miampf 

### Proposed change(s)

- Revert problematic call sites to the default upload action.
- Revert consumers of the problematic artifacts to the default download action.

I don't see an issue with keeping the affected artifacts unencrypted.

### Related issue

- Fixes edgelesssys/issues#155

### Checklist

- [x] Link to Milestone
- [x] test-drive affected workflows
  * [Reproducible Builds](https://github.com/edgelesssys/constellation/actions/runs/7301296268)
  * [Build and Upload OS Image](https://github.com/edgelesssys/constellation/actions/runs/7300237509)

### DIFF
--- a/.github/workflows/reproducible-builds.yml
+++ b/.github/workflows/reproducible-builds.yml
@@ -53,18 +53,16 @@ jobs:
         run: shasum -a 256 "${binary}" | tee "${binary}.sha256"
 
       - name: Upload binary artifact
-        uses: ./.github/actions/artifact_upload
+        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
         with:
           name: "binaries-${{ matrix.target }}"
           path: "${{ env.binary }}"
-          encryptionSecret: ${{ secrets.ARTIFACT_ENCRYPT_PASSWD }}
 
       - name: Upload hash artifact
-        uses: ./.github/actions/artifact_upload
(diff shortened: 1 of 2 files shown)


.github/workflows/reproducible-builds.yml
```diff
@@ -111,10 +110,11 @@ jobs:
         run: shasum -a 256 "${binary}" | tee "${binary}.sha256"
 
       - name: Upload binary artifact
-        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        uses: ./.github/actions/artifact_upload
         with:
           name: "osimages-${{ matrix.target }}"
           path: "${{ env.binary }}"
+          encryptionSecret: ${{ secrets.ARTIFACT_ENCRYPT_PASSWD }}
```
msanft, line 117: Is there a reason for encrypting here? It seems that a binary is uploaded, right?

msanft (approved):
Thanks for the fix! Code changes LGTM. Approving under the premise of successful runs of the affected CI parts.

PR #2775, opened 2023-12-27
ci: add upgrade to provider example test

elchead:
### Context

This extends #2745 by testing cluster upgrades with the provider.

### Proposed change(s)

- add new upgrade section to the e2e test
- refactor `e2e/internal/upgrade` to reuse the upgrade assertion from the e2e upgrade test in the provider
- adjust the upgrade test assert to skip the target image assertion when the value is empty

### Additional info

- [AB#3693](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3693)

How tested:
```
gh workflow run e2e-test-provider-example.yml -r feat/ci/upgrade-provider-example -f cloudProvider=gcp -f providerVersion=v2.14.2 -f image=v2.14.2 -f toImage=ref/main/stream/nightly/v2.15.0-pre.0.20231229110237-07c884b94501 -f toKubernetes=v1.28.5
```

- ✅ [E2E run on GCP](https://github.com/edgelesssys/constellation/actions/runs/7359410699/job/20034139195)
- ✅ [E2E run on AWS](https://github.com/edgelesssys/constellation/actions/runs/7359412499/job/20034143664)
- ✅ [E2E run on Azure](https://github.com/edgelesssys/constellation/actions/runs/7394026401/job/20114784858)

### Checklist

- [x] Link to Milestone


3u13r:
Please fix the PR description as well as the commit history. 
* There should be no blank bullet point in the PR description 
* For me the "proposed changes" is a _must_ category to fill out
* It would be great if commits were prefixed with a category
* There should be no "fix up", 2x "fix upgrade" commits when opening the PR
* The Checklist has checkboxes to be filled out before or right after opening the PR

elchead:
@derpsteb what was the reason to make the status run in go routine? This is problematic because the go routine might still run after the test has failed.

https://github.com/edgelesssys/constellation/actions/runs/7356364135/job/20026389166#step:27:186
```
--- FAIL: TestUpgradeSuccessful (10861.72s)
    upgrade.go:120: 
        	Error Trace:	e2e/internal/upgrade/upgrade.go:120
        	            				src/runtime/asm_amd64.s:1650
        	Error:      	Should be false
        	Test:       	TestUpgradeSuccessful
        	Messages:   	expected at least 5 nodes, got 4
    upgrade.go:114: 
        	Error Trace:	e2e/internal/upgrade/upgrade.go:114
        	            				e2e/internal/upgrade/upgrade.go:58
        	            				e2e/provider-upgrade/upgrade_test.go:66
        	Error:      	Condition never satisfied
        	Test:       	TestUpgradeSuccessful
K8s config path empty. Using environment variable KUBECONFIG=/home/runner/work/constellation/constellation/cluster/constellation-admin.conf.
stdout: Target versions:
```

Also the test reruns even when it passes. I added prints to show this behavior:
```
ivj6skkHxuzcxbk1xv6ZGxrteJxVH7KlX7YRdZ6eARKwLe4	    AFZEAwoKCQ==	    -----END CERTIFICATE-----
Status is available.
Microservices are upgraded.
Nodes are upgraded.
2023/12/29 16:10:09 Node status (2023-12-29 16:10:09.665193 +0000 UTC m=+71.211386042):
2023/12/29 16:10:09 Setting image version: ref/main/stream/nightly/v2.15.0-pre.0.20231229110237-07c884b94501
2023/12/29 16:10:09 Setting K8s version: v1.28.4
2023/12/29 16:10:09 Setting microservice version: v2.15.0-pre.0.20231229125039-57f8888b02cf
```

Any idea why this happens?

elchead:
For the upgrade test to work, https://github.com/edgelesssys/constellation/pull/2788 is required

.github/workflows/e2e-test-provider-example.yml
```diff
@@ -254,11 +278,137 @@ jobs:
...
+          resource "constellation_cluster" "${{ inputs.cloudProvider }}_example" {
+            kubernetes_version = "${{ inputs.toKubernetes }}"
+          }
+          EOF
+          fi
+
+          prefixed_version=${{ steps.build.outputs.build_version }}
+          version=${prefixed_version#v} # remove v prefix
+
+          # needs to be explicitly set to upgrade
```
elchead, line 326: As suggested by @3u13r, I'm working on a PR to enforce to set the microservice version explicitly. cc: @msanft

elchead:
Executing a new [run](https://github.com/edgelesssys/constellation/actions/runs/7412423993/job/20169181751) after rebase

msanft:
Postponing this for now, until general upgrades are back to working

elchead:
> Postponing this for now, until general upgrades are back to working

This test is not affected, and all tests are green. The test only has 1 control plane and 1 worker.

thomasten:
@msanft @daniel-weisse Not sure if you both or just one of you should review this.

.github/workflows/e2e-test-provider-example.yml
```diff
@@ -197,15 +227,23 @@ jobs:
               }
             }
           }
+
+          locals {
+            control_plane_count = 1
+            worker_count        = 1
+          }
```
daniel-weisse, line 234: Is it intended for this test to only run on 1 worker + 1 control-plane node?
Seems to be easily changeable to a configurable input

.github/workflows/e2e-test-provider-example.yml
```diff
@@ -186,7 +212,11 @@ jobs:
...
           # take the middle (2nd) supported Kubernetes version (default)
-          kubernetes_version="$(../build/constellation config kubernetes-versions | awk 'NR==3{print $1}')"
+          if [[ "${{ inputs.providerVersion }}" != "" ]]; then
+            kubernetes_version="$(../release/constellation config kubernetes-versions | awk 'NR==3{print $1}')"
+          else
+            kubernetes_version="$(../build/constellation config kubernetes-versions | awk 'NR==3{print $1}')"
+          fi
 
           cat > _override.tf <<EOF
           terraform {
```
daniel-weisse, line 222: nit: the whitespaces in this generated Terraform file are very inconsistent (they were already before this PR)
e.g.
```terraform
terraform {
  required_providers {
    constellation = {
      source  = "edgelesssys/constellation"
      version            = "${version}"
    }
  }
}
```

.github/workflows/e2e-test-provider-example.yml
```diff
@@ -259,11 +297,137 @@ jobs:
...
+        working-directory: ${{ github.workspace }}/cluster
+        shell: bash
+        run: |
+          if [[ "${{ inputs.toImage }}" != "" ]]; then
+          cat >> _override.tf <<EOF
+          locals {
+            image_version = "${{ inputs.toImage && inputs.toImage || steps.find-latest-image.outputs.image }}"
+          }
+          EOF
+          fi
```
daniel-weisse, line 332: I think this breaks the "upgrade to latest main if not set" behavior
The conditional should be removed, since we either have `inputs.toImage` as non empty, or `steps.find-latest-image.outputs.image`
(Need to double check the template, though I think it should be correct to assume it picks the second, if `inputs.toImage` is an empty string)
```suggestion
          cat >> _override.tf <<EOF
          locals {
            image_version = "${{ inputs.toImage || steps.find-latest-image.outputs.image }}"
          }
          EOF
```

.github/workflows/e2e-test-provider-example.yml
```diff
@@ -259,11 +297,137 @@ jobs:
...
             ../build/constellation maa-patch "$(terraform output -raw maa_url)"
-            TF_LOG=INFO terraform apply -target constellation_cluster.azure_example -auto-approve
+            terraform apply -target constellation_cluster.azure_example -auto-approve
           else
-            TF_LOG=INFO terraform apply -auto-approve
+            terraform apply -auto-approve
+          fi
+
+      - name: Cleanup Terraform Cluster on failure
+        # cleanup here already on failure, because the subsequent TF overrides might make the TF config invalid and thus the destroy would fail later
```
daniel-weisse, line 306: We should move this to just before the `Update cluster configuration` step to avoid lingering clusters if building the provider fails.
We may still run into problems if updating the cluster configuration itself fails, but I think thats acceptable for now

daniel-weisse (commented):
Only the workflow part for now

.github/workflows/e2e-test-provider-example.yml
msanft: We could easily address this by running terraform fmt ... after writing the file.

.github/workflows/e2e-test-provider-example.yml
```diff
@@ -259,11 +297,137 @@ jobs:
...
+
+      - name: Add Provider to local Terraform registry # needed if release version was used before
+        if: inputs.providerVersion != ''
+        working-directory: ${{ github.workspace }}/build
+        shell: bash
+        run: |
+          bazel run //:devbuild --cli_edition=enterprise
+
+      - name: Update cluster configuration # for duplicate variable declaration, the last one is used
+        working-directory: ${{ github.workspace }}/cluster
```
msanft, line 323:
```suggestion
        working-directory: ${{ github.workspace }}/cluster
        if: inputs.toImage != "" && inputs.toKubernetes != ""
```
nit: Would save us some of the bash below

.github/workflows/e2e-test-provider-example.yml
```diff
@@ -259,11 +297,137 @@ jobs:
...
+          fi
+
+          # cfg must be in same dir as KUBECONFIG
+          ../build/constellation config generate "${{ inputs.cloudProvider }}"
+          # make cfg valid with fake data
+          # IMPORTANT: zone needs to be correct because it is used to resolve the CSP image ref
+          if [[ "${{ inputs.cloudProvider }}" == "azure" ]]; then
+            location="${{ inputs.regionZone || 'northeurope' }}"
+            yq e ".provider.azure.location = \"${location}\"" -i constellation-conf.yaml
+
```
msanft, line 399:
```suggestion
```

.github/workflows/e2e-test-provider-example.yml
```diff
@@ -259,11 +297,137 @@ jobs:
...
+            MICROSERVICES_FLAG="--target-microservices=${MICROSERVICES}"
+          fi
+          if [[ -n "${KUBERNETES}" ]]; then
+            KUBERNETES_FLAG="--target-kubernetes=${KUBERNETES}"
+          fi
+          if [[ -n "${IMAGE}" ]]; then
+            IMAGE_FLAG="--target-image=${IMAGE}"
+          fi
+
+          # cfg must be in same dir as KUBECONFIG
```
msanft, line 392: This shouldn't be the case

.github/workflows/e2e-test-provider-example.yml
daniel-weisse: This was more about the formatting in GitHub actions.
I don't think there is a need to have our e2e actions use a file that has pretty formatting.
But having consistent whitespaces helps with reading the test and editing it in the future.

.github/workflows/e2e-test-provider-example.yml
elchead: what do you mean? There is no way that this step will fail due to missing Terraform, because Terraform was executed in the step before.

.github/workflows/e2e-test-provider-example.yml
elchead: it should be possible to only update one of them.

.github/workflows/e2e-test-provider-example.yml
elchead: good catch

.github/workflows/e2e-test-provider-example.yml
daniel-weisse: Not suggesting to change the default, just thought we might want to leave it configurable if that can be done easily (e.g. we havent hit the inputs limit on the workflow yet).

If not, just ignore.

e2e/internal/upgrade/upgrade.go
```diff
@@ -17,3 +19,313 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+// AssertUpgradeSuccessful tests that the upgrade to the target version is successful.
+func AssertUpgradeSuccessful(t *testing.T, cli string, targetVersions VersionContainer, k *kubernetes.Clientset, wantControl, wantWorker int, timeout time.Duration) {
+	wg := queryStatusAsync(t, cli)
+	require.NotNil(t, k)
+
+	testMicroservicesEventuallyHaveVersion(t, targetVersions.Microservices, timeout)
+	fmt.Println("Microservices are upgraded.")
+
+	testNodesEventuallyHaveVersion(t, k, targetVersions, wantControl+wantWorker, timeout)
+	fmt.Println("Nodes are upgraded.")
```
daniel-weisse, line 71:
```suggestion
	testMicroservicesEventuallyHaveVersion(t, targetVersions.Microservices, timeout)
	log.Println("Microservices are upgraded.")

	testNodesEventuallyHaveVersion(t, k, targetVersions, wantControl+wantWorker, timeout)
	log.Println("Nodes are upgraded.")
```
nit: the rest of the functions seem to use `log.PrintX` instead of `fmt.PrintX`

Ignore if this was done intentionally for some formatting purposes

.github/workflows/e2e-test-provider-example.yml
elchead: what do you mean? i tested this a few weeks ago and it was complaining in the upgrade test when the cfg was put in the same directory as the binary.

.github/workflows/e2e-test-provider-example.yml
msanft: Then replace the && with a ||😉

elchead: see comment from Daniel (new code). So the if condition would be incorrect.

msanft (commented):
Skimmed over the CI and code changes, which look alright to me. Regarding the logical relations of the upgrade, I can't say really much.

e2e/provider-upgrade/upgrade_test.go
```diff
@@ -0,0 +1,60 @@
...
+	targetImage         = flag.String("target-image", "", "Image (shortversion) to upgrade to.")
+	targetKubernetes    = flag.String("target-kubernetes", "", "Kubernetes version (MAJOR.MINOR.PATCH) to upgrade to. Defaults to default version of target CLI.")
+	targetMicroservices = flag.String("target-microservices", "", "Microservice version (MAJOR.MINOR.PATCH) to upgrade to. Defaults to default version of target CLI.")
+	// When executing the test as a bazel target the CLI path is supplied through an env variable that bazel sets.
+	// When executing via `go test` extra care should be taken that the supplied CLI is built on the same commit as this test.
+	// When executing the test as a bazel target the workspace path is supplied through an env variable that bazel sets.
+	workspace   = flag.String("workspace", "", "Constellation workspace in which to run the tests.")
+	cliPath     = flag.String("cli", "", "Constellation CLI to run the tests.")
+	wantWorker  = flag.Int("want-worker", 0, "Number of wanted worker nodes.")
+	wantControl = flag.Int("want-control", 0, "Number of wanted control nodes.")
```
daniel-weisse, line 35:
```suggestion
	wantWorker  = flag.Int("want-worker", 1, "Number of wanted worker nodes.")
	wantControl = flag.Int("want-control", 1, "Number of wanted control nodes.")
```
very nit picky: setting 1 as the expected default seems more reasonable for the test, since we expect at least 1 node to upgrade for both control-plane and worker

daniel-weisse (commented):
Go code looks good to me

elchead:
@msanft / @daniel-weisse  thanks for the review! But one of you needs to approve to merge this :D I will start a new E2E run before merging

.github/workflows/e2e-test-provider-example.yml
daniel-weisse: Not relevant anymore
When I made the comment there were builds steps after this cleanup which could fail without having a dedicated cleanup of the cluster.

daniel-weisse (approved):
Good to merge from my side if tests pass

PR #2776, opened 2023-12-27
ci: remove artifact encryption for public artifacts

3u13r:
### Proposed change(s)

- Remove artifact encryption/decryption where the artifacts might as well be public (e.g. the windows Constellation CLI)
- Remove artifact encryption/decryption where it just didn't make sense (the Terraform modules)

This should fix our build-os-image pipeline so the images are also usable. Currently, the pipeline is green but misses all-1 entries in the v2.15.0.pre-... --> CSP image ref lookup table.

Test: https://github.com/edgelesssys/constellation/actions/runs/7338017372

Note that the uploaded map (https://cdn.confidential.cloud/constellation/v2/ref/fix-ci-remove-artifact-encryption/stream/debug/v2.15.0-pre.0.20231227125133-35eca6437c37/image/info.json) now contains all image references again.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/actions/upload_terraform_module/action.yml
+++ b/.github/actions/upload_terraform_module/action.yml
@@ -20,11 +20,10 @@ runs:
         zip -r terraform-module.zip terraform-module
 
     - name: Upload artifact
-      uses: ./.github/actions/artifact_upload
+      uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
       with:
         name: terraform-module
         path: terraform-module.zip
-        encryptionSecret: ${{ inputs.encryptionSecret }}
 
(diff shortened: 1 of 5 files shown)


.github/actions/upload_terraform_module/action.yml
```diff
@@ -20,11 +20,10 @@ runs:
         zip -r terraform-module.zip terraform-module
 
     - name: Upload artifact
-      uses: ./.github/actions/artifact_upload
+      uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
```
elchead, line 23: we should rename the actions to artifact_encrypted_upload / download

elchead (approved):
lgtm

PR #2781, opened 2023-12-28
Add pod disruption budgets so the cluster-autoscaler is able to move kube-admin namespaced resources

3u13r:
### Proposed change(s)

- Refactor the cilium helm values so that every CSP is based off of shared config and only overwrites CSP specific values.
- Enable the pod disruption budget for cert-manafger components 
- Enable the pod disruption budget for the cilium operator
- Manually add a podDisruptionBudget for coreDNS if the cluster-autoscaler if deployed

How to test:
* Start a Constellation cluster with initially 1 control-plane and 2 worker nodes
* Wait for all nodes to be up
* Restart the cert-manager components to make sure that some of them are scheduled on worker nodes
* Also make sure that a coreDNS pod is scheduled on a worker node
* Manually increase the control-plane count to 3 by increasing the count in the CSP console
* Wait for the other 2 control-plane node to join (Those nodes are joined to that the autoscaler has enough nodes to re-schedule all the e.g. cert-manager pods)
* Enable autoscaling and set the `min` value to `0` in the worker scaling group CR
* Have a look at the logs of the cluster-autoscaler
* After maximum of 25 minutes both worker nodes should be scaled down

### Additional info

- [AB#3692](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3692)
- https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-types-of-pods-can-prevent-ca-from-removing-a-node

### Checklist

- [x] Link to Milestone


internal/constellation/helm/BUILD.bazel
```diff
@@ -457,6 +457,7 @@ go_library(
         "charts/cert-manager/templates/controller-config.yaml",
         "charts/cert-manager/templates/poddisruptionbudget.yaml",
         "charts/cert-manager/templates/webhook-poddisruptionbudget.yaml",
+        "charts/edgeless/constellation-services/charts/autoscaler/templates/coredns-pdb.yaml",
```
burgerdev, line 460: This appears to be out of order.

internal/constellation/helm/charts/edgeless/constellation-services/charts/autoscaler/templates/coredns-pdb.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: coredns-pdb
+  namespace: {{ .Release.Namespace }}
+spec:
+  maxUnavailable: 1
+  selector:
+    matchLabels:
+      k8s-app: kube-dns
```
burgerdev, line 10: Please add trailing newline.

internal/constellation/helm/charts/edgeless/constellation-services/charts/autoscaler/templates/coredns-pdb.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: coredns-pdb
+  namespace: {{ .Release.Namespace }}
```
burgerdev, line 5: It might be better to hard-code `kube-system` here, since CoreDNS is not deployed by Helm and thus the release namespace does not apply.

internal/constellation/helm/loader.go
```diff
@@ -232,9 +231,17 @@ func (i *chartLoader) loadCertManagerValues() map[string]any {
 		"tolerations": controlPlaneTolerations,
 		"webhook": map[string]any{
 			"tolerations": controlPlaneTolerations,
+			"podDisruptionBudget": map[string]any{
+				"enabled": true,
+			},
+			"replicaCount": 2,
```
burgerdev, line 237: Cert-Manager has guidance on PDB: https://cert-manager.io/docs/installation/best-practice/#poddisruptionbudget. Is there a reason why we chose different values? Isn't it wasteful to scale to two replicas, if it would be enough to launch an additional one during drains?

3u13r: I just enable the pdb and therefore use the default from their helmcharts, don't I? Also I increase the replica count according to their own warning as stated in the docs you linked.

internal/constellation/helm/BUILD.bazel
3u13r: I didn't edit this manually. The question is if we want to edit this manually or if we just leave this fully autogenerated.

internal/constellation/helm/loader.go
burgerdev: Thanks, I did not realize that `minAvailable: 1` is the default in values.yaml. Ack on the HA comment, this is fine.

internal/constellation/helm/BUILD.bazel
burgerdev: Well, if there is no buildifier step in our CI I and automation pastes stuff in this order, you are off the hook :)

burgerdev (approved):

malt3 (approved):

PR #2788, opened 2024-01-02
node-operator: fix setting Azure node image

elchead:
### Context

The Azure upgrade is currently broken in both the CLI and the Terraform provider. Azure must have recently changed its API spec, basically requiring to set the community image ID in the communityID field. This path is however broken, since the image path /communityGalleries/... does not match the case:
https://github.com/edgelesssys/constellation/blob/c15e4efef6844aad55d344089d5e40ab1a75b184/operators/constellation-node-operator/internal/cloud/azure/client/scalinggroup.go#L168
 
This led to this error in the operator:
```
ERROR    Unable to set ScalingGroup NodeImage    {"controller": "scalinggroup", "controllerGroup": "update.
edgeless.systems", "controllerKind": "ScalingGroup", "ScalingGroup": {"name":"e2e-222-7d279332-worker-44bdb38a"}, "namespace": "", "name":
"e2e-222-7d279332-worker-44bdb38a", "reconcileID": "55fb92f7-4625-4b43-bf3a-4c3dcef839dd", "error": "PATCH
https://management.azure.com/sub
scriptions/0d202bbb-4fa7-4af8-8125-58c269a05435/resourceGroups/e2e-222-test-rg/providers/Microsoft.Compute/virtualMachineScaleSets/e2e-222-
7d279332-worker-44bdb38a\n--------------------------------------------------------------------------------\nRESPONSE 400: 400 Bad Request\n
ERROR CODE: LinkedInvalidPropertyId\n--------------------------------------------------------------------------------\n{\n  \"error\": {\n
   \"code\": \"LinkedInvalidPropertyId\",\n    \"message\": \"Property id '/communityGalleries/ConstellationCVM-728bd310-e898-4450-a1ed-21c
f2fb0d735/images/main-nightly/versions/2023.1229.102745' at path 'properties.virtualMachineProfile.storageProfile.imageReference.id' is inv
alid. Expect fully qualified resource Id that start with '/subscriptions/{subscriptionId}' or '/providers/{resourceProviderNamespace}/'.\"\
n  }\n}\n--------------------------------------------------------------------------------\n"}
 ```
 
This affects any node upgrade, hence requiring a patch release.
The Azure upgrade was still working during the 2.14.0 release, so the change must have happened in the last 2 weeks.

### Proposed change(s)

- change the matching case to lowercase

### Additional info

- [AB#1234](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/1234)

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/operators/constellation-node-operator/internal/cloud/azure/client/scalinggroup.go
+++ b/operators/constellation-node-operator/internal/cloud/azure/client/scalinggroup.go
@@ -165,7 +165,7 @@ func imageReferenceFromImage(img string) (*armcompute.ImageReference, error) {
 	}
 
 	// expecting image to not be a marketplace image
-	if strings.HasPrefix(img, "/CommunityGalleries") {
+	if strings.HasPrefix(img, "/communityGalleries") {
 		ref.CommunityGalleryImageID = to.Ptr(img)
 	} else {
 		ref.ID = to.Ptr(img)
--- a/operators/constellation-node-operator/internal/cloud/azure/client/scalinggroup.go
+++ b/operators/constellation-node-operator/internal/cloud/azure/client/scalinggroup.go
@@ -165,7 +165,7 @@ func imageReferenceFromImage(img string) (*armcompute.ImageReference, error) {
 	}
 
 	// expecting image to not be a marketplace image
-	if strings.HasPrefix(img, "/communityGalleries") {
+	if strings.HasPrefix(strings.ToLower(img), "/communitygalleries") {
 		ref.CommunityGalleryImageID = to.Ptr(img)
 	} else {
 		ref.ID = to.Ptr(img)
--- a/operators/constellation-node-operator/internal/cloud/azure/client/scalinggroup_test.go
+++ b/operators/constellation-node-operator/internal/cloud/azure/client/scalinggroup_test.go
@@ -49,13 +49,13 @@ func TestGetScalingGroupImage(t *testing.T) {
 					VirtualMachineProfile: &armcompute.VirtualMachineScaleSetVMProfile{
 						StorageProfile: &armcompute.VirtualMachineScaleSetStorageProfile{
 							ImageReference: &armcompute.ImageReference{
-								CommunityGalleryImageID: to.Ptr("/CommunityGalleries/gallery-name/Images/image-name/Versions/1.2.3"),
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by msanft — APPROVED** [review]

File: operators/constellation-node-operator/internal/cloud/azure/client/scalinggroup.go [hunk 1]
```diff
@@ -165,7 +165,7 @@ func imageReferenceFromImage(img string) (*armcompute.ImageReference, error) {
 	}
 
 	// expecting image to not be a marketplace image
-	if strings.HasPrefix(img, "/CommunityGalleries") {
+	if strings.HasPrefix(img, "/communityGalleries") {
```
malt3, line 168: Can we make this a case insensitive compare? This way, the code still works if Azure decides to change the API under our feet.

burgerdev: Is there even a use case for the `else` part anymore?

elchead: I think we only have community images and market images so we should be able to remove it.

malt3: I'm pretty sure that future security types on Azure will have restrictions regarding the use of community galleries. This was already the case for SEV-SNP in the past.

burgerdev: And then we'd need to resort to private images? Maybe it's just not clear to me what one would use the third option for.

burgerdev (approved):

operators/constellation-node-operator/internal/cloud/azure/client/scalinggroup.go
malt3: That's correct in simplified terms. The `ID` field is used for most image references that are not community (public to the world) or direct shared. Azure has a lot of other types of images and sharing that are less convenient.

malt3: For instance, there is this alternative sharing option that still uses the `ID` field: https://learn.microsoft.com/en-us/azure/virtual-machines/share-gallery

malt3 (approved):

PR #2790, opened 2024-01-03
helm: remove konnectivity agents

3u13r:
### Proposed change(s)

Konnectivity wasn't used anymore, since the Konnectivity server's are unused since v2.14.0. Now we can also safely remove the agents. 

### Checklist

- [x] Link to Milestone


msanft (approved):
Can't say much regarding the overall implications of this change, but the removal of the daemonset itself looks good to me

PR #2801, opened 2024-01-05
ci: reproducible builds test on artifacts v2

malt3:
### Context

GitHub Actions [has changed the API for Artifacts](https://github.com/actions/toolkit/tree/main/packages/artifact#v2---whats-new). This limits the amount of artifacts per workflow, doesn't allow overwriting artifacts in a Workflow run and changed the way globs are handled.
Since the old API will likely be disabled soon, we have to refactor the reproducible builds test.

### Proposed change(s)

- upgrade reproducible builds test to new artifacts api (see what breaks)
- fix matrix up/downloads using `pattern` field

### Additional info

- [CI run](https://github.com/edgelesssys/constellation/actions/runs/7423223631)

### DIFF
--- a/.github/workflows/reproducible-builds.yml
+++ b/.github/workflows/reproducible-builds.yml
@@ -1,4 +1,4 @@
-# Build Constellation CLI and check for reproducible builds
+# Build Constellation CLI + OS images and check for reproducible builds
 name: Reproducible Builds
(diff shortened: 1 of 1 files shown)


burgerdev (approved):
Very nice!

PR #2804, opened 2024-01-08
ci: fix empty Kubernetes version attribute in Terraform provider e2e test

daniel-weisse:
### Context
Required field was missing

### Proposed change(s)

- Add the required field using value from config
- Add the required field to the example in our Terraform docs

### Additional info

- [e2e test](https://github.com/edgelesssys/constellation/actions/runs/7447154608)


terraform-provider-constellation/docs/resources/cluster.md
msanft, line 1: We should not edit the docs directly, as they are generated. We need to fix the resource schema (in the Go file).

msanft (approved):

terraform-provider-constellation/docs/resources/cluster.md
msanft: Nevermind actually, this part of the example is not generated.

elchead (approved):

PR #2807, opened 2024-01-09
deps: upgrade github.com/aws/aws-sdk-go-v2/...

burgerdev:
### Context

The attestationconfig API test is currently broken on main, because the AWS SDK returns errors along the lines of `not found, ResolveEndpointV2`: https://github.com/edgelesssys/constellation/actions/runs/7450533078. According to https://github.com/aws/aws-sdk-go-v2/issues/2370, the problem is an incompatibility of AWS SDK subcomponents and the solutions is to upgrade all of them. 

This went unnoticed because the workflow is not invoked on dependency updates, and the`internal/api` code has not been touched recently.

### Proposed change(s)

- Upgrade AWS SDK with `go get -u github.com/aws/aws-sdk-go-v2/...`.
- Adapt to struct type changes in the libraries (versions don't seem to be semantic ☹️ ).
- Add `go.mod` to the workflow triggers so that we detect similar problems on the culprit PRs.

### Checklist

- [x] Link to Milestone


internal/api/attestationconfigapi/cli/azure.go
```diff
@@ -27,6 +27,8 @@ func deleteAzure(ctx context.Context, client *attestationconfigapi.Client, cfg d
 	return client.DeleteSEVSNPVersion(ctx, variant.AzureSEVSNP{}, cfg.version)
 }
 
+func ptr[A any](a A) *A { return &a }
+
```
3u13r, line 31: This is our canonical ptr implementation: 
```go
func toPtr[T any](v T) *T {
	return &v
}
```
Please also move this to the bottom of the file as it is almost always the least important thing in a file.

3u13r (commented):
I think you could've also tried opening the renovate PR for "deps: update AWS SDK "

internal/api/attestationconfigapi/cli/azure.go
burgerdev: Thanks for the suggestion, adopted in the other PR.

burgerdev:
Closing in favour of #2809.

PR #2810, opened 2024-01-09
cli: fix AWS SEV-SNP latest version resolution in cluster

elchead:
### Context

When setting "latest" for the SNP versions, the numerical values should be persisted in the `join-config` on `apply`, so that the user knows the currently deployed latest version value.
This was missing in AWS SEV-SNP, because of a hidden code path.

### Proposed change(s)

- make the need for latest version resolution more visible through an interface, which needs to be assigned to every new SEV-SNP type

### Additional info

Tested that `constellation status` shows the resolved versions after `constellation apply`

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -629,15 +629,13 @@ func (c *Config) GetProvider() cloudprovider.Provider {
 // GetAttestationConfig returns the configured attestation config.
 func (c *Config) GetAttestationConfig() AttestationCfg {
 	if c.Attestation.AWSSEVSNP != nil {
-		return c.Attestation.AWSSEVSNP
+		return c.Attestation.AWSSEVSNP.getToMarshallLatestWithResolvedVersions()
 	}
 	if c.Attestation.AWSNitroTPM != nil {
 		return c.Attestation.AWSNitroTPM
 	}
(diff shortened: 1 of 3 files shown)


internal/config/config.go
```diff
@@ -1114,17 +1112,15 @@ type AzureSEVSNP struct {
...
 type AzureTrustedLaunch struct {
 	// description: |
 	//   Expected TPM measurements.
 	Measurements measurements.M `json:"measurements" yaml:"measurements" validate:"required,no_placeholders"`
 }
+
+// sevsnpMarshaller is used to marshall "latest" versions with resolved version numbers.
+type sevsnpMarshaller interface {
+	// getToMarshallLatestWithResolvedVersions brings the attestation config into a state where marshalling uses the numerical version numbers for "latest" versions.
+	getToMarshallLatestWithResolvedVersions() AttestationCfg
```
daniel-weisse, line 1125: Do we maybe want to use this getter for all attestation types to make it clearer to other devs that we dont want to directly pass the structs in `GetAttestationConfig`?

daniel-weisse (approved):

PR #2813, opened 2024-01-10
ci: fix OpenSearch link for e2e notifications

burgerdev:
### Context

Issues created by our CI contain an OpenSearch link. Unfortunately, that link never shows the correct logs: it's pointing to the wrong index, and it's not using the right cloud provider field (`metadata.github.e2e-test-provider` instead of `cloud.provider`). Also, the default time window is `last 15 minutes`, which is usually not what we need. 

### Proposed change(s)

- Construct OpenSearch URL from parts.
- Fix filter query for cloud provider.
- Fix index to `logs-*`.
- Remove column pre-selection (defaults to `time` and `source`) - there are no good default columns because we have both pod logs and systemd logs.

### Additional info

- TODO

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/.github/actions/gh_create_issue/create_issue.sh
+++ b/.github/actions/gh_create_issue/create_issue.sh
@@ -33,17 +33,15 @@ function flagsFromInput() {
 
 function createIssue() {
   flags=(
-    "assignee"
     "body"
     "body-file"
     "label"
     "milestone"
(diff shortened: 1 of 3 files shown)


msanft (approved):
Beautiful, thank you!

PR #2814, opened 2024-01-10
terraform-provider: validate microservice and image version during plan

elchead:
### Context

This fixes primarily an issue for an invalid microservice version, but also makes it more convenient for the user by showing invalid versions during `terraform plan`.

When creating a cluster with invalid microservice version for the current provider version, cluster creation gets stuck, because the helm chart apply is skipped due to an invalid version, but the further constellation apply steps depend on the helm chart deployments.
 
### Proposed change(s)

- check that the microservice version is the same as the provider version

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/internal/compatibility/compatibility.go
+++ b/internal/compatibility/compatibility.go
@@ -149,16 +149,6 @@ func NextMinorVersion(version string) (string, error) {
 	return fmt.Sprintf("v%d.%d", major, minor+1), nil
 }
 
-// PriorMinorVersion returns the prior minor version for a given canonical semver.
-// The returned format is vMAJOR.MINOR.
-func PriorMinorVersion(version string) (string, error) {
-	major, minor, err := parseCanonicalSemver(EnsurePrefixV(version))
-	if err != nil {
-		return "", err
-	}
-	return fmt.Sprintf("v%d.%d", major, minor-1), nil
-}
-
 func parseCanonicalSemver(version string) (major int, minor int, err error) {
 	version = semver.MajorMinor(version) // ensure version is in canonical form (vX.Y.Z)
 	if version == "" {
--- a/internal/compatibility/compatibility.go
+++ b/internal/compatibility/compatibility.go
@@ -149,6 +149,16 @@ func NextMinorVersion(version string) (string, error) {
 	return fmt.Sprintf("v%d.%d", major, minor+1), nil
 }
 
+// PriorMinorVersion returns the prior minor version for a given canonical semver.
+// The returned format is vMAJOR.MINOR.
+func PriorMinorVersion(version string) (string, error) {
+	major, minor, err := parseCanonicalSemver(EnsurePrefixV(version))
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("v%d.%d", major, minor-1), nil
+}
+
 func parseCanonicalSemver(version string) (major int, minor int, err error) {
 	version = semver.MajorMinor(version) // ensure version is in canonical form (vX.Y.Z)
 	if version == "" {
--- a/terraform-provider-constellation/internal/provider/cluster_resource.go
+++ b/terraform-provider-constellation/internal/provider/cluster_resource.go
@@ -460,7 +460,8 @@ func (r *ClusterResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
 			"For details, see https://docs.edgeless.systems/constellation/overview/license")
 	}
 
-	// Validate during plan. must be done in ModifyPlan to read provider data.
+	// Validate during plan. Must be done in ModifyPlan to read provider data.
+	// See https://developer.hashicorp.com/terraform/plugin/framework/resources/configure#define-resource-configure-method.
 	_, diags := r.getMicroserviceVersion(&plannedState)
 	resp.Diagnostics.Append(diags...)
 
--- a/terraform-provider-constellation/internal/provider/cluster_resource.go
+++ b/terraform-provider-constellation/internal/provider/cluster_resource.go
@@ -460,6 +460,17 @@ func (r *ClusterResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
 			"For details, see https://docs.edgeless.systems/constellation/overview/license")
 	}
 
+	// Validate during plan. must be done in ModifyPlan to read provider data.
+	_, diags := r.getMicroserviceVersion(&plannedState)
+	resp.Diagnostics.Append(diags...)
+
+	_, _, diags = r.getImageVersion(ctx, &plannedState)
+	resp.Diagnostics.Append(diags...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
 	// Checks running on updates to the resource. (i.e. state and plan != nil)
 	if !req.State.Raw.IsNull() {
 		// Read currentState supplied by Terraform runtime into the model
--- a/terraform-provider-constellation/internal/provider/cluster_resource_test.go
+++ b/terraform-provider-constellation/internal/provider/cluster_resource_test.go
@@ -12,7 +12,6 @@ import (
 	"regexp"
 	"testing"
 
-	"github.com/edgelesssys/constellation/v2/internal/compatibility"
 	"github.com/edgelesssys/constellation/v2/internal/semver"
 	"github.com/edgelesssys/constellation/v2/internal/versions"
 	"github.com/edgelesssys/constellation/v2/terraform-provider-constellation/internal/data"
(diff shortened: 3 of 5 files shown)


terraform-provider-constellation/internal/provider/cluster_resource.go
```diff
@@ -460,6 +460,19 @@ func (r *ClusterResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
...
 	}
 
+	// Check microservice version is set to the provider version. // must be done in modify plan to read provider data.
+	plannedVer, diags := r.getMicroserviceVersion(&plannedState)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+	if !strings.EqualFold(plannedVer.String(), r.providerData.Version.String()) {
+		resp.Diagnostics.AddAttributeError(
```
msanft, line 470: I assume we'll want to return after this, as it is an error?

terraform-provider-constellation/internal/provider/cluster_resource.go
```diff
@@ -460,6 +460,19 @@ func (r *ClusterResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
...
+			path.Root("constellation_microservice_version"),
+			"Microservice version mismatch", fmt.Sprintf("The microservice version must be set to '%s'.", r.providerData.Version.String()),
```
msanft, line 472: This is not a helpful error message. We should state that the Microservice version needs to equal the provider version, if this is the case. 

Something like:
> **Invalid microservice version**
> The microservice version must be set to the provider version (%s).

terraform-provider-constellation/internal/provider/cluster_resource.go
```diff
@@ -460,6 +460,19 @@ func (r *ClusterResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
 			"For details, see https://docs.edgeless.systems/constellation/overview/license")
 	}
 
+	// Check microservice version is set to the provider version. // must be done in modify plan to read provider data.
```
msanft, line 463: I don't understand this. Why wouldn't we be able to access `r.providerData` in the creation or config validation?

elchead:
@msanft sorry this was supposed to be a draft when you were reviewing.

terraform-provider-constellation/internal/provider/cluster_resource.go
elchead: Because TF has its quirks :/ 
Note that Terraform calling the [ValidateResourceConfig](https://developer.hashicorp.com/terraform/plugin/framework/internals/rpcs#validateresourceconfig-rpc) RPC would not call the [ConfigureProvider](https://developer.hashicorp.com/terraform/plugin/framework/internals/rpcs#configureprovider-rpc) RPC first, so implementations need to account for that situation. Configuration validation in Terraform occurs without provider configuration ("offline").

internal/compatibility/compatibility.go
```diff
@@ -149,6 +149,16 @@ func NextMinorVersion(version string) (string, error) {
 	return fmt.Sprintf("v%d.%d", major, minor+1), nil
 }
 
+// PriorMinorVersion returns the prior minor version for a given canonical semver.
+// The returned format is vMAJOR.MINOR.
+func PriorMinorVersion(version string) (string, error) {
```
daniel-weisse, line 154: nit: inline the functionality of this function in `cluster_resource_test.go`, since that is the only place it is used (don't export other functions, just copy the needed functionality).

terraform-provider-constellation/internal/provider/cluster_resource_test.go
```diff
@@ -23,9 +24,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-var providerVersion = semver.NewFromInt(2, 15, 0, "")
+var providerVersion string
```
daniel-weisse, line 27:
```suggestion
```
If we dont assign a value, we should not have it package scoped to prevent accidentally setting a value and in that way breaking other tests

elchead: this value needs to be injected in the build (see bazel) to make the microservice version compatible with the provider build

daniel-weisse: But this is in the unit test.
If the value being injected by bazel is required, then this breaks running Go the tests outside of bazel.

And as far as I can tell, we simply override the value in some functions, e.g. [here](https://github.com/edgelesssys/constellation/blob/7ce339d343e28e58b0ce64694a3ed23c38b60cdd/terraform-provider-constellation/internal/provider/cluster_resource_test.go#L30).
Whats stopping us from doing the same in other tests as well?

terraform-provider-constellation/internal/provider/cluster_resource.go
msanft: We should then remove that validation step to the top of create imo. I think the `ModifyPlan` semantics are quite special, and as a developer, I would not expect validation to happen here. But no strong opinion.

terraform-provider-constellation/internal/provider/cluster_resource.go
```diff
@@ -460,6 +460,17 @@ func (r *ClusterResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
 			"For details, see https://docs.edgeless.systems/constellation/overview/license")
 	}
 
+	// Validate during plan. must be done in ModifyPlan to read provider data.
```
msanft, line 463:
```suggestion
	// Validate during plan. Must be done in ModifyPlan to read provider data.
	// See https://developer.hashicorp.com/terraform/plugin/framework/resources/configure#define-resource-configure-method.
```

terraform-provider-constellation/internal/provider/cluster_resource_test.go
```diff
@@ -23,9 +23,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-var providerVersion = semver.NewFromInt(2, 15, 0, "")
+const providerVersion string = "v0.0.0"
```
msanft, line 26: Why do we need this here, when we overwrite it in the function scope?

terraform-provider-constellation/internal/provider/cluster_resource_test.go
```diff
@@ -447,8 +448,8 @@ func TestAccClusterResource(t *testing.T) {
 							kubernetes_version = "%s"
 							constellation_microservice_version = "%s"
 					  }
-				`, versions.Default, providerVersion.String()),
-					ExpectError: regexp.MustCompile(".*When csp is set to 'gcp', 'ip_cidr_pod' must be set.*"),
+				`, versions.Default, providerVersion),
+					ExpectError: regexp.MustCompile(".*When csp is set to 'gcp', 'ip_cidr_pod' must be set.*|.*Image version \\(v[0-9]+\\.[0-9]+\\.[0-9]+\\) incompatible with provider version*"),
```
msanft, line 452: We only expect the first error here. The latter would be a bug.

msanft (changes requested):

terraform-provider-constellation/internal/provider/cluster_resource_test.go
elchead: no its not a bug. I can explain the trickiness tomorrow.

terraform-provider-constellation/internal/provider/cluster_resource_test.go
elchead: its not overwritten, see tests below

terraform-provider-constellation/internal/provider/cluster_resource_test.go
elchead: I fixed it

terraform-provider-constellation/internal/provider/cluster_resource_test.go
msanft:
```go
const providerVersion string = "v2.14.0"

func TestMicroserviceConstraint(t *testing.T) {
	providerVersion := semver.NewFromInt(2, 15, 0, "")
```

Do I misunderstand something here?

msanft: If the const is just used in the function below, I think we should move it there.

daniel-weisse (approved):

terraform-provider-constellation/internal/provider/cluster_resource_test.go
elchead: its used in several tests, keeping as constant therefore

PR #2821, opened 2024-01-15
ci: use sonobuoy 0.57.1 and skip cilium conformance problem

malt3:
### Context

While this doesn't seem to resolve all issues, it does improve the situation and makes remaining issue easily discoverable.

### Proposed change(s)

- use sonobuoy 0.57.1 
- print test failures more prominently
- skip conformance test that is [known to be broken with cilium and k8s 1.29](https://github.com/cilium/cilium/issues/29913)

### Additional info

- [e2e](https://github.com/edgelesssys/constellation/actions/runs/7528645597/job/20491290770)

### DIFF
--- a/.github/actions/e2e_sonobuoy/action.yml
+++ b/.github/actions/e2e_sonobuoy/action.yml
@@ -21,7 +21,7 @@ runs:
     - name: Install sonobuoy
       shell: bash
       env:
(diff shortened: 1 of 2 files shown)


3u13r (approved):
LGTM

PR #2825, opened 2024-01-15
deps: convert zap to slog

miampf:
### Context

Rewrite packages to use slog instead of zap.

### Proposed change(s)

- change occurences of zap to slog
- don't use our own logger struct anymore; slog should be used directly
- reduce the logger package to a minimum of functionality often used together with slog

### Additional info

- `operators` was not rewritten
- Since slog has no `Fatal()` function it was replaced by a call to `Error()` followed by `os.Exit(1)`
    - because of that logs will contain `level=ERROR` instead of `level=FATAL`
    - This can be fixed with an `replaceAttr()` function that could be written in the logger package, however this would cause more boilerplate in the logger package and accross the whole project, so I decided against it

### Checklist

- [x] Link to Milestone


dev-docs/conventions.md
```diff
@@ -35,43 +35,41 @@ Further we try to adhere to the following guidelines:
...
 
   Example:
 
   ```Go
-  log.With(zap.Error(someError), zap.String("ip", "192.0.2.1")).Errorf("Connecting to IP failed")
+  log.WithAttrs(slog.Any("error" someError), slog.String("ip", "192.0.2.1")).Error("Connecting to IP failed")
   ```
 
-* Log messages may use format strings to produce human readable messages. However, the information should also be present as structured context fields if it might be valuable for debugging purposes.
+* Slog does not support format strings out of the box. You have to wrap any format strings in `fmt.Sprintf()`.
```
msanft, line 51: IMO, with slog, we should rather advise to not use format strings, as these are not machine-readable in case of switching to JSON loggers. Instead, where a format string was used previously, the input should now be supplied as an attribute. E.g.:
```go
log.withAttrs(slog.String("address", addr), slog.Int("port", port)).Info("Starting server")
```

internal/logger/log.go
```diff
@@ -5,223 +5,160 @@ SPDX-License-Identifier: AGPL-3.0-only
...
-
-Use [Logger.Warnf] to log information that may indicate unwanted behavior, but is not an error.
-
-Use [Logger.Errorf] to log information about any errors that occurred.
-
-Use [Logger.Fatalf] to log information about any errors that occurred and then exit the program.
+You can create a new logger with a new log level by creating a new slog.Logger with the LevelHandler in this package
+and passing the handler of the other logger. As an example, if you have a slog.Logger named `log` you can create a
+new logger with an increased log level (here slog.LevelWarn) like this:
+`slog.New(logger.NewLevelHandler(slog.LevelWarn, log.Handler()))`
```
msanft, line 21:
```suggestion

    slog.New(logger.NewLevelHandler(slog.LevelWarn, log.Handler()))
```

This works better for godoc.

internal/logger/log.go
```diff
@@ -5,223 +5,160 @@ SPDX-License-Identifier: AGPL-3.0-only
...
 		case logging.LevelWarn:
-			logger.Warn(msg)
+			r = slog.NewRecord(time.Now(), slog.LevelWarn, fmt.Sprintf(msg, fields...), pcs[0])
 		case logging.LevelError:
-			logger.Error(msg)
+			r = slog.NewRecord(time.Now(), slog.LevelError, fmt.Sprintf(msg, fields...), pcs[0])
 		default:
 			panic(fmt.Sprintf("unknown level %v", lvl))
 		}
+		_ = l.Handler().Handle(context.Background(), r)
```
msanft, line 106: We can already panic above, so I think we should panic here too if it throws an error. If there's other reasoning behind this, we need to add a [nolint directive](https://golangci-lint.run/usage/false-positives/#nolint-directive) here.

internal/attestation/aws/snp/validator_test.go
```diff
@@ -161,7 +162,7 @@ func TestValidateSNPReport(t *testing.T) {
 			require.NoError(err)
 
 			v := awsValidator{httpsGetter: newStubHTTPSGetter(&urlResponseMatcher{}, nil), verifier: tc.verifier, validator: tc.validator}
-			err = v.validate(vtpm.AttestationDocument{InstanceInfo: infoMarshalled}, ask, ark, [64]byte(hash), config.DefaultForAWSSEVSNP(), logger.NewTest(t))
+			err = v.validate(vtpm.AttestationDocument{InstanceInfo: infoMarshalled}, ask, ark, [64]byte(hash), config.DefaultForAWSSEVSNP(), slog.New(slog.NewTextHandler(logger.TestWriter{T: t}, nil)))
```
msanft, line 165: nit: I'd prefer to have some kind of shorthand here, as we'll need to use this quite often. Something like:
```go
func NewTest(t *testing.T) {
    return slog.New(slog.NewTextHandler(logger.TestWriter{T: t}, nil))
}
```

Then in the tests, we could simply call `logger.NewTest(t)`. Specialized cases requiring a JSON handler or passing extra options could then just instantiate the logger themselves.

internal/attestation/azure/snp/validator.go
```diff
@@ -214,18 +214,18 @@ func (v *Validator) checkIDKeyDigest(ctx context.Context, report *spb.Attestatio
 	// the MAA if necessary.
 	switch v.config.FirmwareSignerConfig.EnforcementPolicy {
 	case idkeydigest.MAAFallback:
-		v.log.Infof(
+		v.log.Info(fmt.Sprintf(
```
msanft, line 217: As mentioned above, I'd advocate for using fields here instead of the formatter. This doesn't need to be done within this PR though.

joinservice/cmd/main.go
```diff
@@ -49,35 +51,39 @@ func main() {
 	verbosity := flag.Int("v", 0, logger.CmdLineVerbosityDescription)
 	flag.Parse()
 
-	log := logger.New(logger.JSONLog, logger.VerbosityFromInt(*verbosity))
+	log := slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: logger.VerbosityFromInt(*verbosity)}))
```
msanft, line 54: While choosing the right log output stream is a [quite controversal topic](https://github.com/ipfs/go-log/issues/60#issuecomment-484140608), I'd personally advocate for writing to stderr. Would like to hear more opinions on this though.

terraform-provider-constellation/internal/provider/cluster_resource.go
```diff
@@ -1233,15 +1233,15 @@ type tfContextLogger struct {
 	ctx context.Context // bind context to struct to satisfy interface
 }
 
-func (l *tfContextLogger) Debugf(format string, args ...any) {
+func (l *tfContextLogger) Debug(format string, args ...any) {
```
msanft, line 1236: I think we should keep these at `Debugf, Infof, ...`, as they do not use slog and perform formatting.

msanft (changes requested):
Went over `internal/logger`, `dev-docs/conventions.md`, `internal/attestation`, `joinservice`, `terraform-provider-constellation`, `internal/constellation`. Mostly some usability nit-picks. Overall structure looks good to me!

dev-docs/conventions.md
```diff
@@ -35,43 +35,41 @@ Further we try to adhere to the following guidelines:
...
-  * `Debugf()` for log low level and detailed information. This may include variable dumps, but should not disclose sensitive information, e.g. keys or secret tokens.
-  * `Infof()` for general information.
-  * `Warnf()` for information that may indicate unwanted behavior, but is not an application error. Commonly used by retry loops.
-  * `Errorf()` to log information about any errors that occurred.
-  * `Fatalf()` to log information about any errors that occurred and then exit the program. Should only be used in the main package of an application.
+  * `Debug()` for log low level and detailed information. This may include variable dumps, but should not disclose sensitive information, e.g. keys or secret tokens.
+  * `Info()` for general information.
+  * `Warn()` for information that may indicate unwanted behavior, but is not an application error. Commonly used by retry loops.
+  * `Error()` to log information about any errors that occurred.
+  * `Fatal()` to log information about any errors that occurred and then exit the program. Should only be used in the main package of an application.
```
daniel-weisse, line 65:
```suggestion
```
No fatal log messages in `slog`

internal/logger/log.go
```diff
@@ -5,223 +5,160 @@ SPDX-License-Identifier: AGPL-3.0-only
...
-Use this package to implement logging for your Constellation services.
+Package logger provides helper functions that can be used in combination with slog to increase functionality or make
+working with slog easier.
 
-# Usage
+1. Logging in unit tests
 
-1. Create a logger using New().
+To log in unit tests you can create a new slog logger that uses logger.TestWriter as its writer. This can be constructed
+by creating a logger like this: `slog.New(slog.NewTextHandler(logger.TestWriter{T: t}, nil))`.
```
daniel-weisse, line 14: We should just provide a convenience function for this, and make `logger.TestWriter` private instead

E.g.
```Go
type testWriter struct {
	t *testing.T
}

func (t testWriter) Write(p []byte) (int, error) {
	t.t.Helper()
	t.t.Log(string(p))
	return len(p), nil
}

func NewTest(t *testing.T) *slog.Logger {
	return slog.New(slog.NewTextHandler(testWriter{t: t}, nil))
}
```

elchead (approved):
there we no relevant changes in my CODEOWNER files except for the provider. lgtm

dev-docs/conventions.md
```diff
@@ -35,43 +35,41 @@ Further we try to adhere to the following guidelines:
...
   ```
 
-* Use the `With()` method to add structured context to your log messages. The context tags should be easily searchable to allow for easy log filtering. Try to keep consistent tag naming!
+* Use the `WithAttrs()` method to add structured context to your log messages. The context tags should be easily searchable to allow for easy log filtering. Try to keep consistent tag naming!
 
   Example:
 
   ```Go
-  log.With(zap.Error(someError), zap.String("ip", "192.0.2.1")).Errorf("Connecting to IP failed")
+  log.WithAttrs(slog.Any("error" someError), slog.String("ip", "192.0.2.1")).Error("Connecting to IP failed")
```
thomasten, line 48: The idiomatic way seems to be
```go
log.Error("Connecting to IP failed", "error", someError, "ip", "192.0.2.1")
```
Any reason we should deviate from this? (Except for it's more similar to what we did before.)

dev-docs/conventions.md
```diff
@@ -35,43 +35,41 @@ Further we try to adhere to the following guidelines:
...
-  * `Errorf()` to log information about any errors that occurred.
-  * `Fatalf()` to log information about any errors that occurred and then exit the program. Should only be used in the main package of an application.
+  * `Debug()` for log low level and detailed information. This may include variable dumps, but should not disclose sensitive information, e.g. keys or secret tokens.
+  * `Info()` for general information.
+  * `Warn()` for information that may indicate unwanted behavior, but is not an application error. Commonly used by retry loops.
+  * `Error()` to log information about any errors that occurred.
+  * `Fatal()` to log information about any errors that occurred and then exit the program. Should only be used in the main package of an application.
 
-* Loggers passed to subpackages of an application may use the `Named()` method for better understanding of where a message originated.
+* Loggers passed to subpackages of an application may use the `WithGroup()` method for better understanding of where a message originated.
```
thomasten, line 67: WithGroup doesn't seem the ideal fit for this. Maybe this can better be implemented with a wrapping handler? Didn't look to deep into it. Opinions?

internal/logger/log.go
```diff
@@ -5,223 +5,160 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+			r = slog.NewRecord(time.Now(), slog.LevelDebug, fmt.Sprintf(msg, fields...), pcs[0])
 		case logging.LevelInfo:
-			logger.Info(msg)
+			r = slog.NewRecord(time.Now(), slog.LevelInfo, fmt.Sprintf(msg, fields...), pcs[0])
 		case logging.LevelWarn:
-			logger.Warn(msg)
+			r = slog.NewRecord(time.Now(), slog.LevelWarn, fmt.Sprintf(msg, fields...), pcs[0])
 		case logging.LevelError:
-			logger.Error(msg)
+			r = slog.NewRecord(time.Now(), slog.LevelError, fmt.Sprintf(msg, fields...), pcs[0])
```
thomasten, line 102: nit: just set a slog.Level variable in these cases to not duplicate the whole NewRecord call

internal/logger/log.go
```diff
@@ -5,223 +5,160 @@ SPDX-License-Identifier: AGPL-3.0-only
...
 		default:
 			panic(fmt.Sprintf("unknown level %v", lvl))
 		}
+		_ = l.Handler().Handle(context.Background(), r)
 	})
 }
+
+// LevelHandler copied from the official LevelHandler example in the slog package documentation.
```
thomasten, line 110: nit: move the LevelHandler to its own file

internal/logger/grpclogger.go
```diff
@@ -7,71 +7,114 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+	logger    *slog.Logger
 	verbosity int
 }
 
 func (l *grpcLogger) Info(args ...interface{}) {
-	l.logger.Info(fmt.Sprint(args...))
+	var pcs [1]uintptr
+	runtime.Callers(2, pcs[:])
+	r := slog.NewRecord(time.Now(), slog.LevelInfo, fmt.Sprint(args...), pcs[0])
+	_ = l.logger.Handler().Handle(context.Background(), r)
```
thomasten, line 37: nit: you may reduce duplication with some helper func like
 ```go
func (l *grpcLogger) log(level slog.Level, msg string) {
	var pcs [1]uintptr
	runtime.Callers(3, pcs[:])
	r := slog.NewRecord(time.Now(), level, msg, pcs[0])
	_ = l.logger.Handler().Handle(context.Background(), r)
}
```

File: internal/grpc/grpclog/grpclog.go [hunk 15]
```diff
@@ -49,5 +50,5 @@ type getStater interface {
 }
 
 type debugLog interface {
-	Debugf(format string, args ...any)
+	Debug(format string, args ...any)
```
thomasten, line 53:
```suggestion
	Debug(msg string, args ...any)
```

internal/grpc/grpclog/grpclog_test.go
```diff
@@ -88,7 +88,7 @@ type spyLog struct {
 	msgs []string
 }
 
-func (f *spyLog) Debugf(format string, args ...any) {
+func (f *spyLog) Debug(format string, args ...any) {
 	f.msgs = append(f.msgs, fmt.Sprintf(format, args...))
```
thomasten, line 92:
```suggestion
func (f *spyLog) Debug(msg string, args ...any) {
	f.msgs = append(f.msgs, msg)
```

thomasten (changes requested):

internal/logger/log.go
thomasten: It's common to ignore errors relating to log or console output.
The panic above is a logic error, while the panic here would be a runtime error.
Is `_ = ` not sufficient to be ignored by the linter?

joinservice/cmd/main.go
thomasten: I'm also in favor of stderr. What did the old implementation write to?

dev-docs/conventions.md
miampf: Not really, if this is preferred I can change it accordingly across the codebase.

dev-docs/conventions.md
miampf: It would surely be possible to implement this with a wrapping handler somehow, but I don't really see the benefit. Is there a specific reason why `WithGroup()` isn't ideal?

terraform-provider-constellation/internal/provider/cluster_resource.go
miampf: While that is true, there are multiple places where we use `slog.Logger` as a `tfContextLogger` and since `slog` doesn't implement any formatting functions we would have to wrap it, which I think would be unintended? Alternatively, I could convert the functions to non-formatting functions since you mentioned that not using formatting with `slog` would be preferred anyways.

dev-docs/conventions.md
thomasten: I'd like to have at least one more opinion on this before you invest time in changing this :)
(Another option would be to recommend the new style in conventions.md without adapting all the existing code.)

dev-docs/conventions.md
thomasten: Sorry, nevermind. I only read half of the documentation on slog groups and it seemed to me that the use case is a different one. But they also mention the exact use case we have here.

internal/logger/log.go
miampf: What is the consensus here? I am in favor of panicking as few times as possible.

msanft: imho, we shouldn't panic in both cases. As TT said, the underscore assignment is probably sufficient to satisfy the linter. @thomasten what do you think regarding the panics?

terraform-provider-constellation/internal/provider/cluster_resource.go
msanft: > there are multiple places where we use `slog.Logger` as a `tfContextLogger`

`tfContextLogger` is not an interface. What do you mean by this?

dev-docs/conventions.md
msanft: I'm in favor of Thomas' suggested solution, while also proposing to not adapt this in this PR, but rather do a follow-up if desired.

thomasten: Let's adapt the recommendation in conventions.md, but not change all existing code. (At least not in this PR.)

terraform-provider-constellation/internal/provider/cluster_resource.go
miampf: Im sorry, this was badly formulated from my side. `tfContextLogger` has to implement `debugLog` which `slog.Logger` needs to implement as well for it to be used in parts of the code. Since `slog.Logger` has no formatting functions I had to rewrite the `debugLog` interface which in turn caused me to rewrite `tfContextLogger`.

internal/logger/log.go
miampf: Seeing that it may be that services pass levels into this function that aren't handled by `slog` out of the box and would have to be added ourselves (for example with a [replaceAttr function](https://pkg.go.dev/log/slog#example-HandlerOptions-CustomLevels)) I would argue that we should go with the last proposal and log on a hardcoded level in default. If we just cast the level I believe `slog` logs that as `<level>+<diff-to-int>`, for example casting 2 would result in the `slog` level "INFO+2", and I don't know how searchable that is.

msanft: Agreed

terraform-provider-constellation/internal/provider/cluster_resource.go
```diff
@@ -1240,15 +1240,15 @@ type tfContextLogger struct {
...
 	tflog.Debug(l.ctx, fmt.Sprintf(format, args...))
```
thomasten, line 1244: I don't fully see through this. Are args here still values to be inserted into a format string, or are they key-value pairs of slog?

cli/internal/cmd/apply.go
```diff
@@ -807,18 +806,18 @@ type warnLogger struct {
 
 // Infof messages are reduced to debug messages, since we don't want
 // the extra info when using the CLI without setting the debug flag.
-func (wl warnLogger) Infof(fmtStr string, args ...any) {
-	wl.log.Debugf(fmtStr, args...)
+func (wl warnLogger) Info(fmtStr string, args ...any) {
+	wl.log.Debug(fmtStr, args...)
```
thomasten, line 810:
```suggestion
func (wl warnLogger) Info(msg string, args ...any) {
	wl.log.Debug(msg, args...)
```

cli/internal/cmd/apply.go
```diff
@@ -807,18 +806,18 @@ type warnLogger struct {
...
 }
 
 // Warnf prints a formatted warning from the validator.
-func (wl warnLogger) Warnf(fmtStr string, args ...any) {
+func (wl warnLogger) Warn(fmtStr string, args ...any) {
 	wl.cmd.PrintErrf("Warning: %s\n", fmt.Sprintf(fmtStr, args...))
```
thomasten, line 815:
```suggestion
func (wl warnLogger) Warn(msg string, args ...any) {
	wl.cmd.PrintErrf("Warning: %s\n", fmt.Sprintf(fmtStr, args...))
```
this also needs some code for printing correctly as args are not format string values anymore

cli/internal/cmd/apply.go
```diff
@@ -807,18 +806,18 @@ type warnLogger struct {
...
 }
 
 type warnLog interface {
-	Warnf(format string, args ...any)
-	Infof(format string, args ...any)
+	Warn(format string, args ...any)
+	Info(format string, args ...any)
```
thomasten, line 820:
```suggestion
	Warn(msg string, args ...any)
	Info(msg string, args ...any)
```

cli/internal/cmd/log.go
```diff
@@ -7,25 +7,25 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+	"os"
+
 	"github.com/spf13/cobra"
-	"go.uber.org/zap/zapcore"
 )
 
 type debugLog interface {
-	Debugf(format string, args ...any)
-	Sync()
+	Debug(format string, args ...any)
```
thomasten, line 17:
```suggestion
	Debug(msg string, args ...any)
```
please also find all other places where "format string" needs to be replaced with "msg string"

thomasten (changes requested):

terraform-provider-constellation/internal/provider/cluster_resource.go
miampf: The arguments here are still inserted into a format string as it was done before since the `tfContextLogger` doesn't use `slog`. This code is weird because `tfContextLogger` has to implement `debugLog`, which also has to be implemented by the `slog.Logger` in other parts of the code.

thomasten: Got it, thanks. Please add a short comment to avoid confusion.

internal/logger/cmdline.go
```diff
@@ -7,25 +7,24 @@ SPDX-License-Identifier: AGPL-3.0-only
...
-		return zap.InfoLevel
-	case verbosity == 1:
-		return zap.WarnLevel
-	case verbosity >= 2:
-		return zap.ErrorLevel
+		return slog.LevelInfo
+	case verbosity == 4:
+		return slog.LevelWarn
+	case verbosity >= 8:
+		return slog.LevelError
```
thomasten, line 26: @msanft @daniel-weisse What do you think about this change? Is it ok or should we stay compatible with the previous log level values?

3u13r (approved):
LGTM

derpsteb (approved):

malt3 (approved):
My owned files are looking good.

internal/logger/grpclogger.go
```diff
@@ -7,71 +7,92 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+func (l *grpcLogger) log(level slog.Level, args ...interface{}) {
+	var pcs [1]uintptr
+	runtime.Callers(3, pcs[:])
+	r := slog.NewRecord(time.Now(), level, fmt.Sprint(args...), pcs[0])
+	_ = l.logger.Handler().Handle(context.Background(), r)
+}
+
+func (l *grpcLogger) logf(level slog.Level, format string, args ...interface{}) {
+	var pcs [1]uintptr
+	runtime.Callers(3, pcs[:])
```
daniel-weisse, line 37:
```suggestion
	runtime.Callers(7, pcs[:])
	r := slog.NewRecord(time.Now(), level, fmt.Sprint(args...), pcs[0])
	_ = l.logger.Handler().Handle(context.Background(), r)
}

func (l *grpcLogger) logf(level slog.Level, format string, args ...interface{}) {
	var pcs [1]uintptr
	runtime.Callers(7, pcs[:])
```
Caller skip of 7 was needed to get the actual line in the grpc package that was making the logging call.
I'm however unsure if we want to bother logging the exact line for grpc, since the caller depth may change on new releases of the grpc library.
Caller skip of 3 would make sure we skip just our own functions and log the location of the top level grpc code.

daniel-weisse (approved):
Two things, otherwise looks good

msanft (approved):

internal/logger/grpclogger.go
miampf: Is there a preferred option? I chose three since it should be more consistent with the code before the conversion to slog, but I can see why it could be beneficial to log the actual line instead of the top level code.

thomasten (approved):

PR #2826, opened 2024-01-15
image: Azure TDX variant

malt3:
### Context

For Azure TDX support, we need OS images with a mainline kernel.

### Proposed change(s)

- image: add mainline kernel and azure tdx image target
- config: allow Azure TDX instance types

### Checklist

- [x] Link to Milestone
- [x] [image build](https://github.com/edgelesssys/constellation/actions/runs/7541457903)

### DIFF
--- a/internal/config/instancetypes/azure.go
+++ b/internal/config/instancetypes/azure.go
@@ -6,9 +6,9 @@ SPDX-License-Identifier: AGPL-3.0-only
 
 package instancetypes
 
-// AzureCVMInstanceTypes are valid Azure CVM instance types.
-var AzureCVMInstanceTypes = []string{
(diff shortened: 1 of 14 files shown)


internal/config/instancetypes/azure.go
```diff
@@ -43,6 +43,40 @@ var AzureCVMInstanceTypes = []string{
 	"Standard_EC48ads_v5",
 	"Standard_EC64ads_v5",
 	"Standard_EC96ads_v5",
+
+	// TDX CVMs
+	// DCesv5-series
+	"Standard_DC4es_v5",
```
daniel-weisse, line 49: Create a new variable for this so we can more easily differentiate between SEV and TDX instances

daniel-weisse (approved):

PR #2830, opened 2024-01-17
image upload: use unique blob name for AWS images uploaded to S3

malt3:
### Context

When uploading images to AWS, they need to be uploaded to S3 first. Since blob names were not unique between attestation variants, there was a possibility for one S3 upload to be used for the wrong AMI.

### Proposed change(s)

- [image upload: use unique blob name for AWS images uploaded to S3](https://github.com/edgelesssys/constellation/commit/db645ee4adee89877f5628166979050f184f1683)

### Additional info

- Fixes https://github.com/edgelesssys/issues/issues/272


burgerdev (approved):
Thanks for the quick fix!

PR #2838, opened 2024-01-22
ci: fix pod ip cidr setting when not on GCP

msanft:
### Context

In the E2E tests for the Terraform provider, we currently set the pod IP CIDR, even if the CSP isn't GCP. This causes problems in the validation, that requires the field to be unset.

### Proposed change(s)

- Validate the `ip_cidr_pod` field only if a GCP cluster is being created.

### Related issue
- Fixes https://github.com/edgelesssys/issues/issues/217

### Additional information
- [E2E test GCP](https://github.com/edgelesssys/constellation/actions/runs/7613735976)
- [E2E test Azure](https://github.com/edgelesssys/constellation/actions/runs/7613734445)

### Checklist

- [x] Link to Milestone


elchead (approved):
lgtm when tests are green

PR #2840, opened 2024-01-22
ci: add TF provider example test to weekly

elchead:
### Context

We do not run the Terraform provider example tests regularly yet, though they replicate the user's path the best. Hence, they should be run in any of the regularly run pipelines. (i.e. daily or weekly)

### Proposed change(s)

- Run the E2E test using the Terraform provider examples in the weekly E2E pipeline.

### Checklist

- [x] Link to Milestone


elchead:
Adding @thomasten to decide if we want to do this / if for all providers.

thomasten (approved):
yes, I think we should do this for all providers

elchead:
@msanft can you have a quick look?

msanft (approved):

PR #2849, opened 2024-01-24
helm: remove kube-rbac-proxy

burgerdev:
### Context

This is the quick alternative to #2837 that only removes kube-rbac-proxy while ignoring all of the kubebuilder legacy.

### Proposed change(s)
- Remove kube-rbac-proxy sidecar from the constellation operator. We don't have any sensitive metrics to guard.
- There are a lot of defunct resources lying around (some new, some existing) that should be cleaned up as part of #2837.

### Additional info

- [AB#3611](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3611)

### Checklist

- [x] Link to Milestone


malt3 (approved):
Thanks!

PR #2850, opened 2024-01-24
ci: fix e2e_benchmark comparison

m1ghtym0:
### Context

The e2e_benchmark test correctly compares the current benchmark with the previous one.
However, there are two problems:
1. The comparison results do not lead to a failed test due to wrong/missing error handling
2. The test ALWAYS overwrites the stored results in the S3 bucket with the new results, leading to a degradation of our baseline

### Proposed change(s)

- Update comparison to fail when the new values are outside the threshold (80% of the previous benchmark)
- Only upload new results if the comparison succeeds -> Baseline can still slowly degrade, but I'd suspect the performance to either be stable or drop significantly in a single step

### Additional info

- https://github.com/edgelesssys/constellation/actions/runs/7637200058
- Keeping a longer history of results is out of scope for this PR
- We could introduce fixed baseline values if we want to
### Checklist

### DIFF
--- a/.github/actions/e2e_benchmark/.gitignore
+++ b/.github/actions/e2e_benchmark/.gitignore
@@ -1,0 +1,2 @@
+benchmarks/
+out/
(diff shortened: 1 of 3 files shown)


thomasten (approved):

m1ghtym0:
Test with failure case failed correctly: https://github.com/edgelesssys/constellation/actions/runs/7652366550

PR #2853, opened 2024-01-24
ci: Upload e2e terraform state as artifact

miampf:
### Context

Upload the terraform state of E2E tests (and other actions that use `constellation_create/action.yml`) as a github artifact. This allows to manually clean up resources and makes automation later on easy.

### Proposed change(s)

- add a step to upload `constellation-terraform` and `constellation-iam-terraform` in `.github/actions/constellation_create/action.yml`

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/actions/constellation_create/action.yml
+++ b/.github/actions/constellation_create/action.yml
@@ -282,7 +282,6 @@ runs:
       with:
         name: terraform-state-${{ inputs.artifactNameSuffix }}
         path: >
-          constellation-terraform/!(.terraform|plan.zip)
-          constellation-iam-terraform/!(.terraform)
+          constellation*-terraform/!(.terraform|plan.zip)
         encryptionSecret: ${{ inputs.encryptionSecret }}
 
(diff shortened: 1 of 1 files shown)


miampf:
E2E Test: https://github.com/edgelesssys/constellation/actions/runs/7641854082

katexochen (changes requested):
We should remove the `.terraform/` directories (which contain the provider binaries) as well as the `plan.zip` from `constellation-terraform/` to safe some space.

miampf:
The glob patterns need to be matched in a previous step and written to a variable. This variable can then be used in the artifact upload step.

miampf:
> Seems to work as expected 👍 https://github.com/edgelesssys/constellation/actions/runs/7711725301

Unfortunately it does not work as expected. I would expect a `constellation-terraform` and a `constellation-iam-terraform` folder, each with their own states. I will look into a fix today.

miampf:
Everything seems to work as expected now: https://github.com/edgelesssys/constellation/actions/runs/7727511372

PR #2864, opened 2024-01-29
ci: safely set attestation variant in OpenSearch URL

daniel-weisse:
### Proposed change(s)

- Quote all inputs to make sure no issues arise from empty strings in the URL
- Set attestation variant in notify action for all tests that support it


.github/workflows/trigger_notify.yml
msanft, line 1: Remove

.github/actions/notify_e2e_failure/action.yml
```diff
@@ -61,13 +61,40 @@ runs:
 
         # Query with filters for this workflow run
         # TODO(msanft): Add Self-managed param once logcollection is fixed.
```
msanft, line 63:
```suggestion
```

msanft (approved):

.github/actions/notify_e2e_failure/action.yml
```diff
@@ -59,15 +59,42 @@ runs:
         # Default window: last 7 days
         g='(time:(from:now-7d,to:now))'
 
+        # Query construction
+        # Omit empty fields since OpenSearch will otherwise only display results where the field is empty
+        if [[ -n "${{ inputs.provider }}" ]]
+        then
+          providerQry="(query:(match_phrase:(cloud.provider:${{ inputs.provider }}))),"
+        fi
+        if [[ -n "${{ inputs.refStream }}" ]]
```
burgerdev, line 68: Maybe this would be easier to maintain as

```sh
query() {
  key=$1
  val=$2
  if [[ -n "$key" ]];  then
    printf "(query:(match_phrase:(%s:'%s')))," "$key" "$val"
  fi
}

q=$(echo "(filters:!(
  $(query metadata.github.ref-stream "${{ inputs.refStream }}")
  # ...
  ))
```

burgerdev (approved):
Thanks for fixing this!

PR #2865, opened 2024-01-29
terraform-provider: only print license warning once in provider

elchead:
### Context

The output is duplicated at the moment:
```
╷
│ Warning: Constellation license not found.
│ 
│   with constellation_cluster.aws_example,
│   on main.tf line 96, in resource "constellation_cluster" "aws_example":
│   96: resource "constellation_cluster" "aws_example" {
│ 
│ Using community license.
│ For details, see
│ https://docs.edgeless.systems/constellation/overview/license
│ 
│ (and one more similar warning elsewhere)
╵
╷
│ Warning: Using community license.
│ 
│   with constellation_cluster.aws_example,
│   on main.tf line 96, in resource "constellation_cluster" "aws_example":
│   96: resource "constellation_cluster" "aws_example" {
│ 
│ For details, see
│ https://docs.edgeless.systems/constellation/overview/license
╵
```

### Proposed change(s)

- Only print one of the warnings.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/terraform-provider-constellation/internal/provider/cluster_resource.go
+++ b/terraform-provider-constellation/internal/provider/cluster_resource.go
@@ -451,8 +451,7 @@ func (r *ClusterResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
 	if licenseID == "" {
 		resp.Diagnostics.AddWarning("Constellation license not found.",
 			"Using community license.\nFor details, see https://docs.edgeless.systems/constellation/overview/license")
-	}
-	if licenseID == license.CommunityLicense {
+	} else if licenseID == license.CommunityLicense {
 		resp.Diagnostics.AddWarning("Using community license.",
 			"For details, see https://docs.edgeless.systems/constellation/overview/license")
 	}
--- a/terraform-provider-constellation/internal/provider/cluster_resource.go
+++ b/terraform-provider-constellation/internal/provider/cluster_resource.go
@@ -449,9 +449,10 @@ func (r *ClusterResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
 
 	licenseID := plannedState.LicenseID.ValueString()
 	if licenseID == "" {
-		resp.Diagnostics.AddWarning("Constellation license not found.",
-			"Using community license.\nFor details, see https://docs.edgeless.systems/constellation/overview/license")
-	} else if licenseID == license.CommunityLicense {
+		resp.Diagnostics.AddWarning("Constellation license ID not set.",
+			"Will use community license.")
+	}
+	if licenseID == license.CommunityLicense {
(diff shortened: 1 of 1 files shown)


msanft (changes requested):
`license.CommunityLicense` is not an empty string, so the suggested change is not right yet. Furthermore, I think we want to keep this as-is. The first message tells you that no license was found (i.e. user doesn't supply a license at all), the second message, that happens later on, tells you that the community license is used, when normally the license would be checked at that step. We can shorten or omit either of the warnings. @thomasten might want to decide that

thomasten:
Yes, we need both. As Moritz said, the first one is validation and the second one is the license check.
I'd change the first one to something like "license_id not set. Will use community license."

elchead:
> Yes, we need both. As Moritz said, the first one is validation and the second one is the license check. I'd change the first one to something like "license_id not set. Will use community license."

My point was that the first warning contains the exact same information like the second one. If we want to keep both, I'd just close the PR.

At the moment, the first warning prints this information "license_id not set. Will use community license.", right?

thomasten:
> My point was that the first warning contains the exact same information like the second one. If we want to keep both, I'd just close the PR.

The first one is printed in ModifyPlan, the second one in apply. I would have thought that there may be TF actions where not both functions are involved.

> At the moment, the first warning prints this information "license_id not set. Will use community license.", right?

Not sure what you mean. It obviously doesn't print that sentence at the moment. But I think it's more helpful. (Strictly speaking, this is not related to this PR.)

daniel-weisse:
> The first one is printed in ModifyPlan, the second one in apply. 

This is not entirely true.
Both warnings will be printed by `terraform plan`, but this was intended since they are warnings for two distinct cases.

The first warning should inform the user that no license was found / configured for this Terraform deployment, as a subtext it also informs the user that in this case the community license will be used.
The second warning informs the user that a community license is in use, and that this has some licensing implications.

thomasten (approved):

terraform-provider-constellation/internal/provider/cluster_resource.go
```diff
@@ -449,8 +449,8 @@ func (r *ClusterResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
 
 	licenseID := plannedState.LicenseID.ValueString()
 	if licenseID == "" {
-		resp.Diagnostics.AddWarning("Constellation license not found.",
-			"Using community license.\nFor details, see https://docs.edgeless.systems/constellation/overview/license")
+		resp.Diagnostics.AddWarning("Constellation license ID not set.",
+			"Will use community license.")
```
msanft, line 453:
```suggestion
			"Continuing with community license.")
```
nit: Sounds less anthropomorphized imo

PR #2867, opened 2024-01-29
ci: set board fields for tf example test

elchead:
### Context

Solves https://github.com/edgelesssys/issues/issues/325. 

### Proposed change(s)

- Set the corresponding fields in the test caller.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/e2e-test-provider-example.yml
+++ b/.github/workflows/e2e-test-provider-example.yml
@@ -31,6 +31,15 @@ on:
         description: Kubernetes version to target for the upgrade, empty for no upgrade.
         type: string
         required: false
+      attestationVariant:
+        description: "Attestation variant to use."
+        type: choice
+        options:
+          - "aws-sev-snp"
+          - "azure-sev-snp"
+          - "azure-tdx"
+          - "gcp-sev-es"
+        required: true
   workflow_call:
     inputs:
       ref:
@@ -57,6 +66,10 @@ on:
         description: Kubernetes version to target for the upgrade, empty for target's default version.
         type: string
         required: false
+      attestationVariant:
+        description: "Attestation variant to use."
+        type: string
+        required: true
 
 jobs:
   provider-example-test:
@@ -243,6 +256,7 @@ jobs:
             image_version = "${image_version}"
             microservice_version = "${prefixed_version}"
             kubernetes_version = "${kubernetes_version}"
+            attestation_variant = "${{ inputs.attestationVariant }}"
           }
 
           module "${{ inputs.cloudProvider }}_iam" {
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-test-provider-example.yml
```diff
@@ -444,4 +449,7 @@ jobs:
         with:
           projectWriteToken: ${{ secrets.PROJECT_WRITE_TOKEN }}
           test: "terraform-provider-example"
+          refStream: ${{ inputs.ref}}
           provider: ${{ inputs.cloudProvider }}
+          kubernetesVersion: ${{ steps.kubernetes.outputs.version }}
+          clusterCreation: "terraform"
```
elchead, line 455: I'm not setting attestationVariant since the value would have to be parsed from the example tf file. Since we rarely change the default, I think its okay to leave it empty.

.github/workflows/e2e-test-provider-example.yml
```diff
@@ -444,4 +449,8 @@ jobs:
...
+          attestationVariant: ${{ inputs.cloudProvider != 'gcp' && format('{0}-sev-snp', inputs.cloudProvider) || 'gcp-sev-es' }}
```
msanft, line 456: This won't be working for Azure-TDX. If we want to support this in the example test, we should pass it here

.github/actions/notify_e2e_failure/action.yml
msanft, line 1: I think this is handled correctly with #2864

.github/workflows/e2e-test-provider-example.yml
elchead: I made it an input parameter.

.github/actions/notify_e2e_failure/action.yml
elchead: reverted the commit and based the PR on this one

msanft (approved):
Thanks!

.github/workflows/e2e-test-provider-example.yml
daniel-weisse: `attestationVariant` should replace `cloudProvider` as an input parameter, since there are many combinations of the two which result in invalid test setups.

You can look at the manual e2e test code for how we derive `cloudProvider` from `attestationVariant` input.

Eventually this test should support multiple attestationVariants for one CSP, but for this PR its ok to limit the runs to `azure-sev-snp`, `aws-sev-snp` and `gcp-sev-es`.

.github/workflows/e2e-test-weekly.yml
```diff
@@ -494,12 +494,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        cloudProvider: ["gcp", "azure", "aws"]
+        attestationVariant: ["gcp-sev-es", "azure-tdx", "aws-sev-snp"]
```
daniel-weisse, line 497:
```suggestion
        attestationVariant: ["gcp-sev-es", "azure-sev-snp", "azure-tdx", "aws-sev-snp"]
```

PR #2869, opened 2024-01-29
ci: run provider upgrade test only once

daniel-weisse:
### Context
[Terraform provider ugrade tests are run with our default Go test policy of `count=3`](https://github.com/edgelesssys/constellation/blob/6d4a8d594eed031246ef1c0283a1e600b81bf14e/e2e/internal/upgrade/BUILD.bazel#L35)
This causes them to run for ~6 hours in case of failures.

### Proposed change(s)

- Run the test only once


3u13r (approved):
LGTM

malt3 (approved):

PR #2873, opened 2024-01-29
deps: update go-uefi and use new authenticode package

malt3:
### Proposed change(s)

- deps: update go-uefi and use new authenticode package

### Checklist

- [x] Link to Milestone


3u13r:
If I'm correct then this should be tested by running an image build and having a look at the measurements file, since in `upload_os_images.sh.in` the `measured-boot` binary is used.
Run: https://github.com/edgelesssys/constellation/actions/runs/7748423826/job/21131013495

3u13r (approved):
LGTM (see test run above)

PR #2877, opened 2024-01-30
ci: fix upload CLI path line splitting

elchead:
### Context

The action for encrypting artifacts merged in #2567 and it's reversal introduced a bug here, as one requires multiple upload targets to split by spaces, and one by newlines. The reversal only changed the action path, but not the operator concatenating multiple lines in the upload targets, causing the upload to [fail](https://github.com/edgelesssys/constellation/actions/runs/7707757698).

### Proposed change(s)

- Use the previously used operator `|`, which concatenates the lines with `\n`s, as required by [actions/upload-artifact](https://github.com/actions/upload-artifact?tab=readme-ov-file#upload-using-multiple-paths-and-exclusions).

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -55,17 +55,17 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          #- arch: amd64
-          #  os: linux
+          - arch: amd64
+            os: linux
 
-          #- arch: amd64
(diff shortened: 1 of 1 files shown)


msanft (approved):
Approving under the premise of https://github.com/edgelesssys/constellation/actions/runs/7707908660 being successful

PR #2893, opened 2024-02-05
ci: make instance type configurable for provider sample

daniel-weisse:
### Context
1. Terraform provider test for Azure TDX was using the default instance type for SEV-SNP resulting in failed initialization of the cluster.
2. Terraform provider test tries to upgrade TDX images to SEV-SNP images

### Proposed change(s)

- Make instance type configurable for provider sample
- Select the correct instance type for TDX e2e test
- TODO: Fix image upgrade issue

### Additional info
- [Azure e2e TDX](https://github.com/edgelesssys/constellation/actions/runs/7785773077)

### DIFF
--- a/.github/workflows/e2e-test-provider-example.yml
+++ b/.github/workflows/e2e-test-provider-example.yml
@@ -420,7 +420,7 @@ jobs:
           fi
 
           # cfg must be in same dir as KUBECONFIG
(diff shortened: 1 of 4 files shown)


msanft (approved):

elchead (approved):
Lgtm when the test is green.

PR #2901, opened 2024-02-09
terraform: provide required configuration for yawol on OpenStack

malt3:
### Context

The yawol loadbalancer requires values that were deleted in 26305e8f80c0f875334338d74671a589eb0f9076. Here we are adding them back.

### Proposed change(s)

- terraform: provide required configuration for yawol on OpenStack

### Checklist

- [x] Link to Milestone


internal/constellation/state/state.go
```diff
@@ -168,6 +171,13 @@ type Azure struct {
 	AttestationURL string `yaml:"attestationURL"`
 }
 
+// OpenStack describes the infra state related to OpenStack.
+type OpenStack struct {
+	// description: |
+	//   ID of the network
+	NetworkID string `yaml:"networkID"`
```
daniel-weisse, line 178: Is there any validation we should run on this field (CLI and Terraform provider), or is this mostly optional?

malt3: I don't think validation would be useful. This is currently output generated by the embedded terraform for OpenStack. There is no Constellation terraform provider support for OpenStack.

daniel-weisse (approved):

PR #2906, opened 2024-02-13
cli: collect debug logs in file

miampf:
### Context

Collect debug information of `constellation apply` in a file called `constellation-debug.log`, even when `--debug` is not set. This is useful for debugging purposes if `--debug` wasn't set.

### Proposed change(s)

- add a function `logDebug` in `apply.go` that logs to stdout if `--debug` is set and always logs to `constellation-debug.log`

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/apply.go
+++ b/cli/internal/cmd/apply.go
@@ -588,7 +588,8 @@ func (a *applyCmd) validateInputs(cmd *cobra.Command, configFetcher attestationc
 // applyJoinConfig creates or updates the cluster's join config.
 // If the config already exists, and is different from the new config, the user is asked to confirm the upgrade.
 func (a *applyCmd) applyJoinConfig(cmd *cobra.Command, newConfig config.AttestationCfg, measurementSalt []byte,
-debugFileHandler file.Handler) error {
+	debugFileHandler file.Handler,
+) error {
 	clusterAttestationConfig, err := a.applier.GetClusterAttestationConfig(cmd.Context(), newConfig.GetVariant())
 	if err != nil {
 		logDebug(a.log, debugFileHandler, fmt.Sprintf("Getting cluster attestation config failed: %s", err))
@@ -860,7 +861,7 @@ type imageFetcher interface {
 }
 
 func logDebug(logger debugLog, filehandler file.Handler, msg string) {
-  logger.Debug(msg)
+	logger.Debug(msg)
 
-  filehandler.Write(constants.InitDebugLogFile, []byte(msg + "\n"), file.OptAppend)
+	filehandler.Write(constants.InitDebugLogFile, []byte(msg+"\n"), file.OptAppend)
 }
--- a/cli/internal/cmd/apply.go
+++ b/cli/internal/cmd/apply.go
@@ -223,7 +223,8 @@ func runApply(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
-	fileHandler := file.NewHandler(afero.NewOsFs())
+  fileHandler := file.NewHandler(afero.NewOsFs())
+  logger := debugFileLogger{fileHandler: file.NewHandler(afero.NewOsFs()), log: log}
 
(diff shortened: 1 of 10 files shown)


cli/internal/cmd/apply.go
```diff
@@ -858,3 +859,9 @@ type imageFetcher interface {
 		image, region string, useMarketplaceImage bool,
 	) (string, error)
 }
+
+func logDebug(logger debugLog, filehandler file.Handler, msg string) {
+	logger.Debug(msg)
+
+	_ = filehandler.Write(constants.InitDebugLogFile, []byte(msg+"\n"), file.OptAppend)
+}
```
daniel-weisse, line 867: Why not create a new debugFileLogger struct which satisfies the `debugLog` interface and writes messages to a file?
That would easily ensure all messages are logged correctly, and also makes it more easily usable for other commands.
e.g. :
```Go
type debugFileLogger struct {
	fileHandler file.Handler
	log         debugLog
}

func (l *debugFileLogger) Debug(msg string, args ...any) {
	l.log.Debug(msg, args...)
	_ = l.fileHandler.Write("file", []byte(msg+"\n"), file.OptAppend)
}
```

miampf: Yes, that's a good idea, I should've thought of that myself.

cli/internal/cmd/apply.go
```diff
@@ -858,3 +859,14 @@ type imageFetcher interface {
...
+type debugFileLogger struct {
+	fileHandler file.Handler
+	log         debugLog
+}
+
+func (l debugFileLogger) Debug(msg string, args ...any) {
+	l.log.Debug(msg, args...)
+
+	_ = l.fileHandler.Write(constants.InitDebugLogFile, []byte(msg+"\n"), file.OptAppend)
+}
```
daniel-weisse, line 872: Have you considered using something like [slogmulti](https://github.com/samber/slog-multi?tab=readme-ov-file#broadcast-slogmultifanout) to handle log distribution to multiple sinks?
With the current approach, we lose a decent amount of information, e.g. no timestamps or additional fields added to the logger.

Threw something together that seems to work fine:
```Go
func newDebugFileLogger(cmd *cobra.Command, fileHandler file.Handler) (debugLog, error) {
	logLvl := slog.LevelInfo
	debugLog, err := cmd.Flags().GetBool("debug")
	if err != nil {
		return nil, err
	}
	if debugLog {
		logLvl = slog.LevelDebug
	}

	fileWriter := &fileWriter{
		fileHandler: fileHandler,
	}
	return slog.New(
		slogmulti.Fanout(
			slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{AddSource: true, Level: logLvl}),           // first handler: stderr at log level
			slog.NewJSONHandler(fileWriter, &slog.HandlerOptions{AddSource: true, Level: slog.LevelDebug}), // second handler: debug JSON log to file
		),
	), nil
}

type fileWriter struct {
	fileHandler file.Handler
}

// Write satisfies the io.Writer interface by writing a message to file.
func (l *fileWriter) Write(msg []byte) (int, error) {
	err := l.fileHandler.Write("file", msg, file.OptAppend)
	return len(msg), err
}
```

This writes any log message into a file in JSON format.
Alternatively, we could use a regular text handler to have message parity with the CLI debug output.

miampf: I don't really think that this issue is worth bringing in a new dependency. However, I could maybe write a new Handler for those cases myself if you want?

daniel-weisse: I don't think bringing in a new dependency for this is an issue, especially since the library itself has very few dependencies (only four, three of which we already depend on).
I also don't think putting in extra effort to re-implement and maintain similar code is worth it, when the library already does exactly what we want in an easy to use way.

miampf: Okay, I've added your code and tested it :)

daniel-weisse (approved):

malt3 (approved):

PR #2909, opened 2024-02-14
vpn: ship our own container image

burgerdev:
### Context

The container image used in the VPN chart should be reproducible and stable.  The resulting image should be handled similar to s3proxy: it's built as a part of the Constellation release process and then consumed from a Helm chart in our registry.

### Proposed change(s)

* Build an image with nix, extract a single layer tarball.
* Include the layer in a Bazel OCI rule.
* Integrate the image into the publishing process.

### Additional info

- [AB#3554](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3554)

### Checklist

- [x] Link to Milestone


nix/container/vpn/strongswan.sh
```diff
@@ -1,6 +1,11 @@
 #!/bin/sh
 
-# The charon binary is not included in the PATH generated by nixery.dev, find it manually.
+set -eux
```
malt3, line 3: Is the -x intended? Feel free to keep.

nix/container/vpn/vpn.nix
```diff
@@ -0,0 +1,83 @@
...
+in
+
+stdenv.mkDerivation {
+  name = "image";
+
+  src = image;
+
+  buildInputs = with pkgs; [ gnutar jq ];
+
+  # unpackPhase = "true";
```
malt3, line 70:
```suggestion
```

malt3 (approved):
LGTM

nix/container/vpn/strongswan.sh
burgerdev: Yes, `-x` is currently set in the deployment and only moves to the script here. I wanted to have some logs for what happens before charon is ready, and `-x` is the poor man's solution for that :)

PR #2939, opened 2024-02-22
helm: avoid waiting for non-essential services

malt3:
### Context

In our e2e tests, we see a lot of "etcd-leader changed" errors while deploying non-essential helm charts.
If this transient error occurs, helm gets into a broken state where it cannot uninstall cleanly and thus any retry attempts fail. By not waiting for the installation of helm charts to succeed, we can avoid making most of the kubernetes API calls while control-plane nodes are joining.
This makes "constellation apply" faster and more resilient.

### Proposed change(s)

- helm: avoid waiting for non-essential services


burgerdev (approved):
Thanks, lgtm!

daniel-weisse (approved):

PR #2942, opened 2024-02-22
docs: remove broken links and publish removal of cloud logging

thomasten:


msanft (approved):
Thanks! I wasn't aware of the links in these sections. Should've checked better there.

PR #2971, opened 2024-03-06
bootstrapper: wipe disk and reboot on non-recoverable error

daniel-weisse:
### Context
Constellation's bootstrapper may fail at multiple stages.
Some of these failures are transient, like joining an existing cluster using `kubeadm` and may be retried.
Others are not recoverable, meaning one can no longer re-run `constellation apply` on the same node, or the node can no longer retry joining an existing cluster.
To "fix" a VM that is in this failed state, the VM needs to be rebooted, and the state-disk wiped.

### Proposed change(s)

- Change how `joinClient` and `initServer` are run
  - Both routines now have return values and are run in parallel
  - If any error occurs in `run()`, the VM is rebooted
- Adapt the flow of `joinClient` to more easily differentiate between transient and non recoverable errors
- In both `joinClient` and `initServer`, mark the state disk as not initialized if a non recoverable error occurs (any error after the VM was marked as initialized)
- Remove the ability to call `JoinClient.Start()` and `JoinClient.Stop()` idempotently
  - This was and is not required for the bootstrapper
  - This was unnecessarily complicating the set up and testing of the client

### Related issue
- Fixes https://github.com/edgelesssys/issues/issues/421

### Additional info

- [AB#3930](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3930)
- Why not use `systemd` native ways to restart the VM when `constelllation-bootstrapper.service` fails?
  - On debug images, the service is expected to fail until a bootstrapper is uploaded using `cdbg`. This makes using debug images impossible
- How to test?
  - This PR currently contains a commit marked with `[drop]` that returns an error in the `joinClient` which causes a fatal error and causes the VM to reboot. The state disk should then be marked as non initialized and be wiped again on start.

### Checklist

  - [x] [e2e Test (Azure SNP) to make sure regular behavior is still as expected](https://github.com/edgelesssys/constellation/actions/runs/8187559867)
- [x] Link to Milestone

### DIFF
--- a/bootstrapper/cmd/bootstrapper/run.go
+++ b/bootstrapper/cmd/bootstrapper/run.go
@@ -12,6 +12,7 @@ import (
 	"log/slog"
 	"net"
 	"os"
+	"sync"
 
 	"github.com/edgelesssys/constellation/v2/bootstrapper/internal/clean"
 	"github.com/edgelesssys/constellation/v2/bootstrapper/internal/diskencryption"
@@ -67,12 +68,26 @@ func run(issuer atls.Issuer, openDevice vtpm.TPMOpenFunc, fileHandler file.Handl
 	go cleaner.Start()
 	defer cleaner.Done()
 
-	joinClient.Start(cleaner)
+	var wg sync.WaitGroup
 
-	if err := initServer.Serve(bindIP, bindPort, cleaner); err != nil {
-		log.With(slog.Any("error", err)).Error("Failed to serve init server")
-		os.Exit(1)
-	}
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		if err := joinClient.Start(cleaner); err != nil {
+			log.With(slog.Any("error", err)).Error("Failed to join cluster")
+			os.Exit(1)
+		}
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		if err := initServer.Serve(bindIP, bindPort, cleaner); err != nil {
+			log.With(slog.Any("error", err)).Error("Failed to serve init server")
+			os.Exit(1)
+		}
+	}()
+	wg.Wait()
 
 	log.Info("bootstrapper done")
 }
--- a/bootstrapper/cmd/bootstrapper/run.go
+++ b/bootstrapper/cmd/bootstrapper/run.go
@@ -13,6 +13,7 @@ import (
 	"net"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/edgelesssys/constellation/v2/bootstrapper/internal/clean"
 	"github.com/edgelesssys/constellation/v2/bootstrapper/internal/diskencryption"
@@ -74,7 +75,8 @@ func run(issuer atls.Issuer, openDevice vtpm.TPMOpenFunc, fileHandler file.Handl
 	go func() {
 		defer wg.Done()
 		if err := joinClient.Start(cleaner); err != nil {
-			log.With(slog.Any("error", err)).Error("Failed to join cluster")
+			log.With(slog.Any("error", err)).Error("Failed to join cluster. Rebooting...")
+			time.Sleep(20 * time.Second) // ensure log message is written
 			reboot()
 		}
 	}()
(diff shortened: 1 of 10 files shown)


3u13r (approved):
LGTM

bootstrapper/internal/joinclient/joinclient.go
```diff
@@ -93,146 +90,134 @@ func New(lock locker, dial grpcDialer, joiner ClusterJoiner, meta MetadataAPI, l
...
-
 	c.log.Info("Stopping")
 
 	c.stopC <- struct{}{}
 	<-c.stopDone
 
-	c.stopC = nil
-	c.stopDone = nil
+	c.stopC = make(chan struct{}, 1)
+	c.stopDone = make(chan struct{}, 1)
```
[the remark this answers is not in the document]
daniel-weisse: I removed all code that was explicitly there to handle starting multiple `JoinClient.Start()` routines, since this was not (and should not be) done.
Resetting the channels here is mostly being done so calling `JoinClient.Start()` after `JoinClient.Stop()` does not panic.
This is because `c.stopC` and `c.stopDone` channels are no longer created during `JoinClient.Start()`, but during `joinclient.New()`, so setting them to `nil` during `JoinClient.Stop()` causes errors if `JoinClient.Start` is called after.

I think simply removing this code section would be functionally the same, and better highlights that we don't intend to reuse the client.

bootstrapper/cmd/bootstrapper/run.go
```diff
@@ -67,12 +70,26 @@ func run(issuer atls.Issuer, openDevice vtpm.TPMOpenFunc, fileHandler file.Handl
...
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		if err := initServer.Serve(bindIP, bindPort, cleaner); err != nil {
+			log.With(slog.Any("error", err)).Error("Failed to serve init server")
+			reboot(fmt.Errorf("serving init server: %w", err))
+		}
+	}()
+	wg.Wait()
```
burgerdev, line 92: It seems like both initserver and joinclient now have internal logic for differentiating non-recoverable errors. Would it make sense to check the error type here in run instead so that we only have one place that deals with disk reset?

bootstrapper/internal/initserver/initserver_test.go
```diff
@@ -402,6 +406,10 @@ func (d *stubDisk) UpdatePassphrase(string) error {
 	return d.updatePassphraseErr
 }
 
+func (d *stubDisk) MarkDiskForReset() error {
+	return nil
+}
```
burgerdev, line 411: Could we add a test case that asserts MarkDisk is called (or that the right error type is returned, see other thread) on non-recoverable error states?

bootstrapper/internal/joinclient/joinclient.go
```diff
@@ -313,11 +289,12 @@ func (c *JoinClient) startNodeAndJoin(ticket *joinproto.IssueJoinTicketResponse,
 
 	// We currently cannot recover from any failure in this function. Joining the k8s cluster
 	// sometimes fails transiently, and we don't want to brick the node because of that.
-	for i := 0; i < 3; i++ {
+	for i := range 3 {
 		err = c.joiner.JoinCluster(ctx, btd, c.role, ticket.KubernetesComponents, c.log)
-		if err != nil {
-			c.log.Error("failed to join k8s cluster", "role", c.role, "attempt", i, "error", err)
+		if err == nil {
+			break
```
burgerdev, line 295: Good catch!

burgerdev (approved):

bootstrapper/cmd/bootstrapper/run.go
daniel-weisse: Decided to move the disk wiping into the `run` function in case `InitServer` or `JoinClient` return an error.
We can do this unconditionally in this case, as we are sure the node was not previously bootstrapped, and setting the token to `false` is safe to do, regardless if it was set to `true` or not at all.
This takes out a bit of complexity when it comes to determining failure states.

PR #2984, opened 2024-03-13
helm: manually retry uninstalling a failed release

burgerdev:
### Context

Helm installations can fail halfway through due to transient problems (e.g. edgelesssys/issues#30). If we're unlucky, the subsequent uninstall also fails, which makes future installation attempts fail because there is already an existing release.

### Proposed change(s)

- After a failure in `helm install`, check whether uninstalling worked and uninstall manually if not.

### Related issue

- Fixes edgelesssys/issues#367

### Manual Test

1. Created a cluster on Azure from `main` and applied up to excluding the `helm` phase.
2. Added an iptables rule to drop all etcd traffic.
3. Ran the `helm` phase, which expectedly failed with the symptoms of edgelesssys/issues#367.
4. Removed the iptables rule, confirmed that Kubernetes is serving again.
5. Re-ran the `helm` phase, expectedly observing the same failure.
6. Built from this PRs HEAD and ran the `helm` phase successfully.

### Checklist

  * [x] [Azure TDX](https://github.com/edgelesssys/constellation/actions/runs/8271318869)
- [x] Link to Milestone

### DIFF
--- a/internal/constellation/helm/action.go
+++ b/internal/constellation/helm/action.go
@@ -8,8 +8,10 @@ package helm
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/edgelesssys/constellation/v2/internal/constants"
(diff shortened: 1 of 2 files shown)


burgerdev:
`helmAction.Force` might be an easier alternative, but I don't know whether that would have other, undesirable consequences.

burgerdev:
Thanks! Yes, I ran into unrelated flakes three times in a row :(

PR #2994, opened 2024-03-19
ci: fix pvc clean-up on non deletable namespaces

daniel-weisse:
### Context
"default" and "kube-system" are protected namespaces in Kubernetes and can't be deleted.
This causes problems if a PVC was create in one of those namespaces and the `constellation_destroy` action tries to clean up the cluster, since that action just tries to delete any namespace with a PVC in it.

### Proposed change(s)

- Don't try to delete non deletable namespaces ("default", "kube-system")
  - Instead, just try to delete all pvcs in that namespace
- Don't abort the `constellation_destroy` action if pvc deletion fails
  - Lingering PVCs are annoying and costs money, but lingering clusters even more so

### Related issue
- See https://github.com/edgelesssys/issues/issues/465 for an e2e run that caused this failure
- Fixes https://github.com/edgelesssys/issues/issues/402

### DIFF
--- a/.github/actions/constellation_destroy/action.yml
+++ b/.github/actions/constellation_destroy/action.yml
@@ -24,6 +24,7 @@ runs:
     - name: Delete persistent volumes
       if: inputs.kubeconfig != ''
       shell: bash
+      continue-on-error: true
       env:
         KUBECONFIG: ${{ inputs.kubeconfig }}
         PV_DELETION_TIMEOUT: "120" # 2 minutes timeout for pv deletion
--- a/.github/actions/constellation_destroy/action.yml
+++ b/.github/actions/constellation_destroy/action.yml
@@ -34,6 +34,14 @@ runs:
         # Scrap namespaces that contain PVCs
         for namespace in `kubectl get namespace --no-headers=true -o custom-columns=":metadata.name"`; do
           if [[ `kubectl get pvc -n $namespace --no-headers=true -o custom-columns=":metadata.name" | wc -l` -gt 0 ]]; then
+            if [[ "${namespace}" == "default" ]]; then
+              kubectl delete all --all --namespace "default" --wait
+              continue
+            fi
+            if [[ "${namespace}" == "kube-system" ]]; then
+              kubectl delete pvc --all --namespace "kube-system" --wait
+              continue
+            fi
             kubectl delete namespace $namespace --wait
           fi
         done


.github/actions/constellation_destroy/action.yml
```diff
@@ -51,13 +55,13 @@ runs:
         echo "::endgroup::"
 
     - name: Login to GCP (Cluster service account)
-      if: inputs.cloudProvider == 'gcp'
+      if: always() && inputs.cloudProvider == 'gcp'
```
burgerdev, line 58: Nit: using [`continue_on_error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) may be a bit more elegant.

.github/actions/constellation_destroy/action.yml
```diff
@@ -34,6 +34,10 @@ runs:
         # Scrap namespaces that contain PVCs
         for namespace in `kubectl get namespace --no-headers=true -o custom-columns=":metadata.name"`; do
           if [[ `kubectl get pvc -n $namespace --no-headers=true -o custom-columns=":metadata.name" | wc -l` -gt 0 ]]; then
+            if [[ "${namespace}" == "default" ]] || [[ "${namespace}" == "kube-system" ]]; then
+              kubectl delete pvc --namespace "${namespace}" --all --wait
+              continue
+            fi
```
burgerdev, line 40: Do we rely on other side effects, like cleanup of LoadBalancer services? If not, we could always delete PVCs only.

burgerdev (approved):

burgerdev:
This should fix https://github.com/edgelesssys/issues/issues/402

.github/actions/constellation_destroy/action.yml
daniel-weisse: IMO, deleting the namespace is preferred for this clean up, since deleting a PVC may fail if there is still a pod using that PVC. If we just delete the entire namespace, that is no problem, if we *only* delete the pvc, we may run into some issues.

Something potentially more robust could be deleting all objects in the "default" namespace.
Cleaning up the "kube-system" namespace can be a bit more "best effort", since we shouldn't deploy any PVCs in there anyway.

I think the following should do the trick:

```bash
kubectl delete all --all --namespace default
```

burgerdev (approved):

PR #2997, opened 2024-03-20
logging: unify debug log message format

miampf:
### Context

The current debug messages all have different formats. For example, some use `%s` while others use `%q` for strings. This PR unifies those formats and removes most uses of `%v` (and similair format strings for structs) in exchange for better logging.

### Proposed change(s)

- Use `%q` instead of `%s` for all strings in debug messages
- When printing flags, don't use `%+v`. Instead, log the important flags individually.
- When debug logging structs focus on the important values of the struct instead of printing it with `%+v` or something similair.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/cli/internal/cmd/configfetchmeasurements.go
+++ b/cli/internal/cmd/configfetchmeasurements.go
@@ -152,7 +152,7 @@ func (cfm *configFetchMeasurementsCmd) configFetchMeasurements(
 			return fmt.Errorf("fetching and verifying measurements: %w", err)
 		}
 	}
-	cfm.log.Debug(fmt.Sprintf("Measurements:\n%s", fetchedMeasurements.String()))
+	cfm.log.Debug(fmt.Sprintf("Measurements: %s", fetchedMeasurements.String()))
 
 	cfm.log.Debug("Updating measurements in configuration")
 	conf.UpdateMeasurements(fetchedMeasurements)
--- a/cli/internal/cmd/configfetchmeasurements.go
+++ b/cli/internal/cmd/configfetchmeasurements.go
@@ -105,12 +105,7 @@ func runConfigFetchMeasurements(cmd *cobra.Command, _ []string) error {
 	if err := cfm.flags.parse(cmd.Flags()); err != nil {
 		return fmt.Errorf("parsing flags: %w", err)
 	}
-	cfm.log.Debug(fmt.Sprintf(
-		`Using flags:
-  insecure: %t
-  measurements-url: %q
-  signature-url: %q
-`, cfm.flags.insecure, cfm.flags.measurementsURL, cfm.flags.signatureURL))
+	cfm.log.Debug("Using flags", "insecure", cfm.flags.insecure, "measurementsURL", cfm.flags.measurementsURL, "signatureURL", cfm.flags.signatureURL)
 
 	fetcher := attestationconfigapi.NewFetcherWithClient(http.DefaultClient, constants.CDNRepositoryURL)
 	return cfm.configFetchMeasurements(cmd, fileHandler, fetcher)
--- a/cli/internal/cmd/configfetchmeasurements.go
+++ b/cli/internal/cmd/configfetchmeasurements.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"net/url"
 	"time"
+	"encoding/hex"
 
 	"github.com/edgelesssys/constellation/v2/internal/api/attestationconfigapi"
 	"github.com/edgelesssys/constellation/v2/internal/api/versionsapi"
@@ -157,7 +158,11 @@ func (cfm *configFetchMeasurementsCmd) configFetchMeasurements(
 			return fmt.Errorf("fetching and verifying measurements: %w", err)
 		}
 	}
-	cfm.log.Debug(fmt.Sprintf("Measurements: %#v\n", fetchedMeasurements))
+	measurementsToPrint := "";
+	for i, measurement := range fetchedMeasurements {
+		measurementsToPrint += fmt.Sprintf("\t%d: 0x%s\n", i, hex.EncodeToString(measurement.Expected))
+	}
+	cfm.log.Debug(fmt.Sprintf("Measurements:\n%s", measurementsToPrint))
 
 	cfm.log.Debug("Updating measurements in configuration")
 	conf.UpdateMeasurements(fetchedMeasurements)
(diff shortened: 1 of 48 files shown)


internal/constellation/kubecmd/kubecmd.go
```diff
@@ -266,7 +266,7 @@ func (k *KubeCmd) ExtendClusterConfigCertSANs(ctx context.Context, alternativeNa
 		k.log.Debug("No new SANs to add to the cluster's apiserver SAN field")
 		return nil
 	}
-	k.log.Debug(fmt.Sprintf("Extending the cluster's apiserver SAN field with the following SANs: %s\n", strings.Join(missingSANs, ", ")))
+	k.log.Debug(fmt.Sprintf("Extending the cluster's apiserver SAN field with the following SANs: %q\n", strings.Join(missingSANs, ", ")))
```
daniel-weisse, line 269:
```suggestion
	k.log.Debug(fmt.Sprintf("Extending the cluster's apiserver SAN field with the following SANs: %q", strings.Join(missingSANs, ", ")))
```

internal/constellation/applyinit.go
```diff
@@ -85,12 +85,17 @@ func (a *Applier) Init(
...
 	}
 
 	// Perform the RPC
-	a.log.Debug(fmt.Sprintf("Making initialization call, doer is %+v", doer))
+	a.log.Debug(fmt.Sprintf(
+		`Making initialization call, doer has the following parameters:
+  endpoint: %q
+  req.KmsUri: %q
+  req.StorageUri: %q
+`, doer.endpoint, doer.req.KmsUri, doer.req.StorageUri))
```
daniel-weisse, line 98:
```suggestion
	a.log.Debug("Making initialization call", "endpoint", doer.endpoint, "kmsUri", doer.req.KmsUri, "storageUri")
```

We should just use log attributes to log this information

internal/api/versionsapi/cli/rm.go
```diff
@@ -631,7 +648,7 @@ func (a *azureClient) deleteImage(ctx context.Context, image string, dryrun bool
 	}
 
 	if dryrun {
-		log.Debug(fmt.Sprintf("DryRun: delete image %v", azImage))
+		log.Debug(fmt.Sprintf("DryRun: delete image with gallery %q, image definition %q, resource group %q and version %q", azImage.gallery, azImage.imageDefinition, azImage.resourceGroup, azImage.version))
```
daniel-weisse, line 651:
```suggestion
		log.Debug(fmt.Sprintf("DryRun: delete image: resource group: %q, gallery: %q, image definition: %q, version: %q"", azImage.resourceGroup, azImage.gallery, azImage.imageDefinition, azImage.version))
```

nit: adhere to the log format of the other message later on in this function

e2e/malicious-join/malicious-join.go
```diff
@@ -172,7 +172,11 @@ func (j *maliciousJoiner) join(ctx context.Context) (*joinproto.IssueJoinTicketR
 		IsControlPlane:     false,
 	}
 	res, err := protoClient.IssueJoinTicket(ctx, req)
-	j.logger.Debug(fmt.Sprintf("Got join ticket response: %+v", res))
+	j.logger.Debug(fmt.Sprintf(
+		`Got join ticket response:
+  ApiServerEndpoint: %q
+  KubernetesVersion: %q
+`, res.ApiServerEndpoint, res.KubernetesVersion))
```
daniel-weisse, line 179:
```suggestion
	j.logger.Debug("Got join ticket response", "apiServerEndpoint", res.ApiServerEndpoint, "kubernetesVersion", res.KubernetesVersion)
```

should just use attributes

daniel-weisse (commented):
Overall looks good to me.
Two issues I have with the changes:

1. Custom formatting of log messages.

This mostly happens when logging flags.
I don't think we should introduce custom formatting as this may break log processing and also breaks the "one message per line" we currently have for debug messages.
IMO, we should just use log attributes for all of the cases in this PR.
I annotated the code parts where this happened outside of flag printing.
Flag printing should also use attributes similar to my annotations, e.g.
```Go
cfm.log.Debug("Using flags", "insecure", cfm.flags.insecure, "measurementsURL", cfm.flags.measurementsURL, "signatureURL", cfm.flags.signatureURL)
```

2. Measurement printing

I personally think the few cases we do print measurements as part of debug logging are helpful and should stay in the code.
Since we do this more than once, adding a simple `.String()` method to the `measurements.M` type would be helpful.
I'm also ok with keeping the `%v` formatting for these parts, as the output was readable enough for their use case,

internal/attestation/measurements/measurements.go
```diff
@@ -330,6 +330,15 @@ func (m *M) UnmarshalYAML(unmarshal func(any) error) error {
 	return nil
 }
 
+// String returns a string representation of the measurements.
+func (m M) String() string {
+	returnString := ""
+	for i, measurement := range m {
+		returnString += fmt.Sprintf("%d: 0x%s\n", i, hex.EncodeToString(measurement.Expected))
+	}
+	return returnString
```
daniel-weisse, line 339:
```suggestion
	var returnString string
	for i, measurement := range m {
		returnString = strings.Join([]string{returnString, fmt.Sprintf("%d: 0x%s", i, hex.EncodeToString(measurement.Expected)}, ",")
	}
	return returnString
```
As mentioned before, I would prefer to keep the logging to one message per line.

joinservice/internal/watcher/validator.go
```diff
@@ -79,7 +79,7 @@ func (u *Updatable) Update() error {
 	if err != nil {
 		return fmt.Errorf("unmarshaling config: %w", err)
 	}
-	u.log.Debug(fmt.Sprintf("New expected measurements: %+v", cfg.GetMeasurements()))
+	u.log.Debug(fmt.Sprintf("New expected measurements:\n%s", cfg.GetMeasurements().String()))
```
daniel-weisse, line 82:
```suggestion
	u.log.Debug(fmt.Sprintf("New expected measurements: %s", cfg.GetMeasurements().String()))
```

cli/internal/cmd/configfetchmeasurements.go
```diff
@@ -152,14 +152,14 @@ func (cfm *configFetchMeasurementsCmd) configFetchMeasurements(
 			return fmt.Errorf("fetching and verifying measurements: %w", err)
 		}
 	}
-	cfm.log.Debug(fmt.Sprintf("Measurements: %#v\n", fetchedMeasurements))
+	cfm.log.Debug(fmt.Sprintf("Measurements:\n%s", fetchedMeasurements.String()))
```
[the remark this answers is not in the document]
miampf: Do you want to log all the measurements on the same line? If yes, I would adjust the `.String()` method for `measurements.M` to not use newlines as well.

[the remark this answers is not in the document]
miampf: Never mind, I overlooked your other comment.

malt3:
The change looks good to me from a high level.

PR #2999, opened 2024-03-21
ci: add a delete artifact action

miampf:
### Context

To correctly reflect terraform states of e2e actions in uploaded artifacts, terraform state artifacts should be deleted if `constellation destroy` and `constellation iam destroy` ran correctly since then, no state exists anymore. This PR adds an action that can be used to delete artifacts.

### Proposed change(s)

- add an action called `artifact_delete` that can delete artifacts from workflows

### Checklist

- [x] Link to Milestone


miampf:
A successful test run of the action can be found [here](https://github.com/edgelesssys/constellation/actions/runs/8376696851).

.github/actions/artifact_delete/delete_artifact.sh
daniel-weisse, line 1: CI wants you to run `bazel run //:tidy` on this

daniel-weisse (approved):
lgtm
Any particular reason for keeping this as a separate bash script?
You could just inline the two `gh api` invocations in the `action.yml`

miampf:
> lgtm Any particular reason for keeping this as a separate bash script? You could just inline the two `gh api` invocations in the `action.yml`

Not really, I wrote the bash script first to test if everything works with an automated script and then integrated it with the CI. But if you think it would be cleaner, I can inline the calls.

PR #3007, opened 2024-04-02
terraform: update terraform provider STACKIT

malt3:
### Context

We received a hint saying that newer versions work more reliably.

### Proposed change(s)

- terraform: update terraform provider STACKIT


3u13r (approved):
LGTM

PR #3025, opened 2024-04-12
snp: ensure we never use ARK supplied by Issuer

daniel-weisse:
### Context

Our SNP certificate parsing code potentially sets the ARK to one returned by the Issuer.
While the calling code never actually used this ARK, it still requires a lot of attention by the developer to not accidentally use this potentially unsafe ARK to verify an SNP attestation.

### Proposed change(s)

- Ensure the ARK used by the Validator is never passed through from the Issuer
  - We now completely ignore any ARKs supplied by the Issuer and try to fetch them in the following order:
    - From the user's config / JoinService's cache
    - From the AMD KDS
- Pass the pre-fetched ASK on GCP to the Validator
- Use the correct reportData when validating SNP reports on GCP

### Additional info
- Tested this manually on GCP with SEV-SNP to make sure attestation passes and the ASK is actually present in the report.
  - The code is built error safe in case an ASK should not be there for some reason

### Checklist

  - [x] [Azure SEV-SNP e2e test to make sure I didn't break things](https://github.com/edgelesssys/constellation/actions/runs/8660214205)
  - [x] [AWS SEV-SNP e2e test to make sure I didn't break things](https://github.com/edgelesssys/constellation/actions/runs/8660216124)


internal/attestation/snp/snp.go
```diff
@@ -97,7 +96,7 @@ func (a *InstanceInfo) addReportSigner(att *spb.Attestation, report *spb.Report,
 
 // AttestationWithCerts returns a formatted version of the attestation report and its certificates from the instanceInfo.
 // Certificates are retrieved in the following precedence:
-// 1. ASK or ARK from issuer. On Azure: THIM. One AWS: not prefilled.
+// 1. ASK from issuer. On Azure: THIM. One AWS: not prefilled. On GCP: prefilled.
```
msanft, line 99:
```suggestion
// 1. ASK from issuer. On Azure: THIM. One AWS: not prefilled. (Go to option 2) On GCP: prefilled.
```

PR #3033, opened 2024-04-18
cli: allow tagging cloud resources with custom tags

miampf:
### Context

Some users might want to use custom tags to group resources together.

### Proposed change(s)

- Add a flag `-t`/`--tags` to `constellation config generate` that writes additional tags into `constellation-conf.yaml`
  - Those tags will be passed to their respective cloud providers in `constellation apply`
  - In GCP, labels are used, in Azure and AWS tags

### Additional Information
- Since OpenStack does not support tagging their resources with key-value pairs, the key-value pairs in the constellation configuration are converted to strings in the form of `key=value` which are then applied as the tags.

### Checklist

- [x] Link to Milestone


malt3:
Tested this on OpenStack (y)

cli/internal/cmd/configgenerate.go
```diff
@@ -221,3 +230,30 @@ func parseAttestationFlag(flags *pflag.FlagSet) (variant.Variant, error) {
...
+	}
+
+	tagsStringSplit := strings.Split(tagsString, ",")
+
+	tags := make(cloudprovider.Tags)
+	for _, tag := range tagsStringSplit {
+		tagSplit := strings.Split(tag, "=")
+		if len(tagSplit) != 2 {
+			return nil, fmt.Errorf("wrong format of tags")
+		}
```
daniel-weisse, line 253:
```suggestion
		if len(tagSplit) != 2 {
			return nil, fmt.Errorf("wrong format of tags: expected \"key=value\", got: %q", tag)
		}
```

Only use `fmt.Errorf` when the error requires format strings, otherwise just use `errors.New`.
In this case, I think it helps to tell the user why the format is wrong

internal/config/config.go
```diff
@@ -67,6 +67,9 @@ type Config struct {
 	// description: |
 	//   Name of the cluster.
 	Name string `yaml:"name" validate:"valid_name,required"`
+	// description : |
+	//	 Custom tags to be applied to created cloud resources.
```
daniel-weisse, line 71:
```suggestion
	//   Custom tags to be applied to created cloud resources.
```
I think there is a tab character in here that breaks comment generation

As a nit, I would also prefer to have this moved down to just before `Provider ProviderConfig`

daniel-weisse:
Tested successfully on GCP

internal/config/config.go
miampf: Unfortunately, the tab is not what messes up comment generation. I'll try to find a fix.

daniel-weisse: Removing the tab fixed it for me

daniel-weisse (approved):
Looks good from my side.
Did someone already test this on Azure and AWS?

PR #3035, opened 2024-04-19
ci: tag resources created by e2e tests with the run name

miampf:
### Context

If cleanups fail, it should be easier to locate resources created by an e2e test.

### Proposed change(s)

- add an input to `constellation_iam_create/action.yml` that allows the addition of additional tags.
- tag all resources created through `e2e_test/action.yml` with the generated prefix

### Additional info

- Since windows e2e tests don't use `e2e_test/action.yml`, resources are tagged with the resource group name instead.

### Checklist

### DIFF
--- a/.github/actions/e2e_test/action.yml
+++ b/.github/actions/e2e_test/action.yml
@@ -126,8 +126,7 @@ runs:
       shell: bash
       run: |
         if [[ "${{ inputs.cosignPassword }}" == '' || "${{ inputs.cosignPrivateKey }}" == '' ]]; then
-          echo "::error::e2e test verify requires cosignPassword and cosignPrivateKey to be set."
-          exit 1
+          echo "::error::e2e test verify requires cosignPassword and cosignPrivateKey to be set." exit 1
         fi
 
     - name: Determine build target
@@ -258,6 +257,7 @@ runs:
         gcpProjectID: ${{ inputs.gcpProject }}
         gcpZone: ${{ inputs.regionZone || 'europe-west3-b' }}
         kubernetesVersion: ${{ inputs.kubernetesVersion }}
+        additionalTags: "run-name=${{ steps.create-prefix.outputs.prefix }}"
 
     - name: Login to GCP (Cluster service account)
       if: inputs.cloudProvider == 'gcp'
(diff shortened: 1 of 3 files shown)


.github/actions/e2e_test/action.yml
```diff
@@ -258,6 +257,7 @@ runs:
         gcpProjectID: ${{ inputs.gcpProject }}
         gcpZone: ${{ inputs.regionZone || 'europe-west3-b' }}
         kubernetesVersion: ${{ inputs.kubernetesVersion }}
+        additionalTags: "run-name=${{ steps.create-prefix.outputs.prefix }}"
```
daniel-weisse, line 260:
```suggestion
        additionalTags: "workflow=${{ github.workflow }}"
```
I think a better tag would be using the workflow run id
Resources are already named using `steps.create-prefix.outputs.prefix`.
If we tag them with their workflow ID, its easier to know where the resources came from.

.github/workflows/e2e-windows.yml
```diff
@@ -84,7 +84,7 @@ jobs:
         run: |
           $uid = Get-Random -Minimum 1000 -Maximum 9999
           $rgName = "e2e-win-${{ github.run_id }}-${{ github.run_attempt }}-$uid"
-          .\constellation.exe config generate azure
+          .\constellation.exe config generate azure -t "run-name=$rgName"
```
daniel-weisse, line 87:
```suggestion
          .\constellation.exe config generate azure -t "workflow=${{ github.workflow }}"
```
Same here

.github/actions/e2e_test/action.yml
```diff
@@ -126,7 +126,7 @@ runs:
       shell: bash
       run: |
         if [[ "${{ inputs.cosignPassword }}" == '' || "${{ inputs.cosignPrivateKey }}" == '' ]]; then
-          echo "::error::e2e test verify requires cosignPassword and cosignPrivateKey to be set."
+          echo "::error::e2e test verify requires cosignPassword and cosignPrivateKey to be set."
```
daniel-weisse, line 129:
```suggestion
          echo "::error::e2e test verify requires cosignPassword and cosignPrivateKey to be set."
```

daniel-weisse (approved):

PR #3038, opened 2024-04-22
attestation: dont set a default for TDX MRSEAM

daniel-weisse:
### Context
TDX attestation verifies the TDX module using various versions and hashes.
One optional value for user validation is `MRSEAM`, the ["measurement of the Intel TDX module. (48 Bytes hash)"](https://cdrdv2-public.intel.com/733580/TDX%20Migration%20TD%20Design%20Guide%20Rev%200.8.9.pdf).
Setting this value ensures attestation only passes if the CVM is running with an Intel TDX module which matches this exact hash.
While this blocks older versions of the module from running (this is also prevented by other versions checks), it also effectively blocks using newer versions.

### Proposed change(s)

- Remove the default value for `MRSEAM` from Constellation's config
  - Users may still set a value if desired

### Additional info

- I think this should go into the changelog of the next release. However, we are don't have a label for "other changes", and I don't think this is a bug fix or a new feature. Leaving it open for now and marking for backport


burgerdev (approved):
Thanks, lgtm!

PR #3045, opened 2024-04-26
terraform: simplify additional tagging logic to ensure they are always applied

daniel-weisse:
### Context
Thought something was broken with the additional tag logic, but turned out I was just looking at the wrong instances on Azure.
Ended up going over the tagging logic nonetheless and should be more robust in applying them to all resources now

### Proposed change(s)

- Instead of merging additional tags at the resource with the predefined "constant" local tags, merge additional tags directly into the local tags.
  - This makes it a lot harder to forget adding the additional tags to new resources added to the terraform files
  - Adds the local tags to a couple resources that were missing them (They only had the additional tags previosuly)

### Checklist

  - [x] [AWS](https://github.com/edgelesssys/constellation/actions/runs/8844909547)
  - [ ] [Azure](https://github.com/edgelesssys/constellation/actions/runs/8845251556)
  - [x] [GCP](https://github.com/edgelesssys/constellation/actions/runs/8844912879)

### DIFF
--- a/terraform/infrastructure/aws/main.tf
+++ b/terraform/infrastructure/aws/main.tf
@@ -48,9 +48,10 @@ locals {
   // example: given "name-1234567890.region.elb.amazonaws.com" it will return "*.region.elb.amazonaws.com"
   wildcard_lb_dns_name = replace(aws_lb.front_end.dns_name, "/^[^.]*\\./", "*.")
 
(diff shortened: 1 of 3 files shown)


malt3 (approved):
Code LGTM. Feel free to merge once e2e tests are green!

PR #3048, opened 2024-04-26
snp: don't print warning if no ASK is present

msanft:
### Context

On AWS, there is no ASK coming from the issuer, so it will always result in a warning.

### Proposed change(s)

- Don't error if no PEM blocks are there.

- [x] Link to Milestone


malt3 (approved):

daniel-weisse (approved):

daniel-weisse:
Can you add a note that the function will return nil if no certificates are returned?

PR #3051, opened 2024-04-29
rfc: node access

burgerdev:
### Context

This RFC proposes a mechanism for accessing Constellation nodes for maintenance.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/rfc/016-node-access.md
+++ b/rfc/016-node-access.md
@@ -1,0 +1,188 @@
+---
+status: approved, not implemented
+---
+
+# RFC 016: Node Access
+
+## Background
+
+A production Constellation cluster is currently configured not to allow any kind of remote administrative access.
+This choice is deliberate: any mechanism for remote accesss can potentially be exploited, or may leak sensitive data.
+
+However, some operations on a Kubernetes cluster require some form of access to the nodes.
+A good class of examples are etcd cluster maintenance tasks, like backup and recovery, or emergency operations like removing a permanently failed member.
+Some kubeadm operations, like certificate rotation, also require some form of cluster access.
+
+While some tasks can be accomplished by DaemonSets, CronJobs and the like, relying on Kubernetes objects is insufficient.
+Executing commands in a Kubernetes pod may fail because Kubernetes is not healthy, etcd is bricked or the network is down.
+Administrative access to the nodes through a side channel would greatly help remediate, or at least debug, those situations.
+
+## Requirements
+
+Constellation admins can log into Constellation nodes for maintenance, subject to the following restrictions:
+
+* Access must be encrypted end-to-end to protect from CSP snooping.
+* Access must be possible even if the Kubernetes API server is down.
+
+Nice-to-have:
+
+* The method of access should not require long-term storage of a second secret.
+* The method of access should be time-limited.
+
+## Proposed Design
+
+Core to the proposal is [certificate-based authentication for OpenSSH](https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Certificate-based_Authentication).
+We can derive a valid SSH key from the Constellation master secret.
+An OpenSSH server on the node accepts certificates issued by this CA key.
+Admins can derive the CA key from the master secret on demand, and issue certificates for arbitrary public keys.
+An example program is in the [Appendix](#appendix).
+
+### Key Details
+
+We use an HKDF to derive an ed25519 private key from the master secret.
+This private key acts as an SSH certificate authority, whose signed certs allow access to cluster nodes.
+Since the master secret is available to both the cluster owner and the nodes, no communication with the cluster is needed to mint valid certificates.
+The choice of curve allows to directly use the derived secret bytes as key.
+This makes the implementation deterministic, and thus the CA key recoverable.
+
+### Server-side Details
+
+An OpenSSH server is added to the node image software stack.
+It's configured with a `TrustedUserCAKeys` file and a `RevokedKeys` file, both being empty on startup.
+All other means of authentication are disabled.
+
+After initialization, the bootstrapper fills the `TrustedUserCAKeys` file with the derived CA's public key.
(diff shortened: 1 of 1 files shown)


rfc/016-node-access.md
```diff
@@ -0,0 +1,176 @@
...
+### Enable Serial Console
+
+Serial consoles for cloud VMs are tunnelled through the CSP in the clear.
+To make this solution secure, an encrypted channel would need to be established on top of the serial connection.
+The author is not aware of any software providing such a channel.
+
+### SSH with Authorized Keys
+
+We could ask users to add a public key to their `constellation-conf.yaml` and add that to `/root/.ssh/authorized_keys` after joining.
+This would require the cluster owner to permanently manage a second secret, and there would be no built-in way to revoke access.
```
daniel-weisse, line 81: We actually used to do this, but decided to remove it in [v2.2.0](https://github.com/edgelesssys/constellation/releases/tag/v2.2.0).
I like the proposed design a lot better, since it makes distributing the SSH access (i.e. the SSH CA) to other nodes of the cluster a lot easier.

rfc/016-node-access.md
```diff
@@ -0,0 +1,176 @@
...
+This private key acts as an SSH certificate authority, whose signed certs allow access to cluster nodes.
+Since the master secret is available to both the cluster owner and the nodes, no communication with the cluster is needed to mint valid certificates.
+The choice of curve allows to directly use the derived secret bytes as key.
+This makes the implementation deterministic, and thus the CA key recoverable.
+
+### Server-side Details
+
+An OpenSSH server is added to the node image software stack.
+It's configured with a `TrustedUserCAKeys` file that is empty on startup.
+After initialization, the bootstrapper fills that file with the derived CA's public key.
```
daniel-weisse, line 48: Just as a side note: Worker nodes don't have access to the master secret, so we will need to generate and distribute the CA during the join process.

rfc/016-node-access.md
```diff
@@ -0,0 +1,176 @@
...
+All other means of authentication are disabled.
+
+### Client-side Details
+
+A new subcommand `constellation credentials ssh` is added to the CLI.
```
daniel-weisse, line 53: nit: We may want to use a different parent command name, e.g. `key-gen` or `key` so we can bundle things like data encryption key generation under this command.
`credentials` also works, but seems a bit less on topic for the other key gen operations.

rfc/016-node-access.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Key Details
+
+We use an HKDF to derive an ed25519 private key from the master secret.
+This private key acts as an SSH certificate authority, whose signed certs allow access to cluster nodes.
+Since the master secret is available to both the cluster owner and the nodes, no communication with the cluster is needed to mint valid certificates.
+The choice of curve allows to directly use the derived secret bytes as key.
+This makes the implementation deterministic, and thus the CA key recoverable.
+
+### Server-side Details
```
daniel-weisse, line 44: We should also enable SSH host certificates on the Constellation nodes.
This should ensure a user is actually talking to a Constellation node over SSH, not just any VM that accepted the SSH key.

daniel-weisse (approved):
Thanks, this design looks like a good approach to me.

3u13r (approved):
LGTM

thomasten (approved):
Thanks, this looks solid to me.

malt3 (approved):

m1ghtym0 (approved):

elchead (approved):

rfc/016-node-access.md
```diff
@@ -0,0 +1,176 @@
...
+
+An OpenSSH server is added to the node image software stack.
+It's configured with a `TrustedUserCAKeys` file that is empty on startup.
```
msanft, line 47: Should we also have some CRL-alike thing that'd enable an administrator to revoke access for a specific key? In theory, you'd have to throw away your cluster anyway, as it might've been backdoored in such a case. But for real-world scenarios, revoking keys might be handy.

msanft (approved):
Love this! Thanks for taking the time to spec all of this out. It will also save us some time in development if we don't have to use CSP Serial Consoles :)

rfc/016-node-access.md
burgerdev: I left the exact placement of the subcommand open for now.

rfc/016-node-access.md
burgerdev: Proposing to only keep the CA in the join server, ptal

rfc/016-node-access.md
burgerdev: Added KRL as an option for the admin - no real complexity added for us, but as mentioned it's probably not useful in general.

rfc/016-node-access.md
burgerdev: Added host certs to the proposal.

burgerdev:
This is good to go as far as I'm concerned. Please let me know until 2024-05-21 if you'd like to see modifications.

rfc/016-node-access.md
```diff
@@ -0,0 +1,184 @@
+# Node Access
```
burgerdev, line 1:
```suggestion
# RFC 016: Node Access
```

rfc/016-node-access.md
```diff
@@ -0,0 +1,184 @@
+# RFC 016: Node Access
+
```
burgerdev, line 2:
```suggestion

---
status: approved, not implemented
---

```

rfc/016-node-access.md
```diff
@@ -0,0 +1,188 @@
...
+---
+status: approved, not implemented
+---
```
burgerdev, line 5:
```suggestion
---
status: approved, not implemented
---

# RFC 016: Node Access
```

PR #3058, opened 2024-05-02
ci: enable gcp-sev-snp for daily tests

daniel-weisse:
### Context
gcp-sev-snp was disabled on daily e2e tests since we currently don't have stable images for this variant.
However, we the test was disabled for both debug and stable images.

### Proposed change(s)

- Enable gcp-sev-snp for daily test runs
  - Currently excludes stable images from the test
- Fix verify e2e test for gcp-sev-snp
  - Properly create the JSON file for gcp-sev-snp and upload it
  - Remove a check for backwards compatibility with old CLI versions (v2.13)
  - [e2e test](https://github.com/edgelesssys/constellation/actions/runs/8922428735/job/24504575569)

### Related issue
- Fixes https://github.com/edgelesssys/issues/issues/547

### DIFF
--- a/.github/actions/e2e_verify/action.yml
+++ b/.github/actions/e2e_verify/action.yml
@@ -66,12 +66,16 @@ runs:
           forwarderPID=$!
           sleep 5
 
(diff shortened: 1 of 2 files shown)


burgerdev (approved):
lgtm, thanks!

PR #3062, opened 2024-05-03
helm: Restore the ability to start a cluster in conformance mode by disabling the cilium ipmasq agent when in conformance mode

3u13r:
### Proposed change(s)

- disable Cilium's ip masq agent when running in conformance mode since the ip masq agent requires bpf masquerading

I tested this on AWS.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/constellation/helm/overrides.go
+++ b/internal/constellation/helm/overrides.go
@@ -33,14 +33,6 @@ import (
 // Also, the charts are not rendered correctly without all of these values.
 func extraCiliumValues(provider cloudprovider.Provider, conformanceMode bool, output state.Infrastructure) map[string]any {
 	extraVals := map[string]any{}
-	if conformanceMode {
-		extraVals["kubeProxyReplacementHealthzBindAddr"] = ""
-		extraVals["kubeProxyReplacement"] = "partial"
-		extraVals["sessionAffinity"] = true
-		extraVals["cni"] = map[string]any{
-			"chainingMode": "portmap",
-		}
-	}
 
 	strictMode := map[string]any{}
 	// TODO(@3u13r): Once we are able to set the subnet of the load balancer VMs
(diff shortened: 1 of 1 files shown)


miampf:
I currently get the following error when trying to set up a cluster on GCP: `Error: applying Helm charts: applying constellation-operators: helm install: release constellation-operators failed, and has been uninstalled due to atomic being set: client rate limiter Wait returned an error: context deadline exceeded`

burgerdev:
It looks like you're not only disabling the ip-masq-agent, but also the entire eBPF service resolution mechanism (kubeProxyReplacement) - could you please document why this is necessary?

3u13r:
> could you please document why this is necessary?

What kind of documentation do you want? Do you want a user-facing documentation of the `--conformance` flag? 
We need to disable kube-proxy replacement since we must use the portmap feature of Cilium go gain conformance regarding the differentiation of UDP and TCP traffic on the same port. 
The idea is that nobody requires (or maybe even uses) that feature since upstream Cilium in their default configuration is also not compliant. They want to fix this eventually (since ~2 years or so). There have been recent efforts which were abandoned even more recently.

3u13r:
> I currently get the following error when trying to set up a cluster on GCP:

Hm, I don't get any error when initializing a Constellation on gcp.

burgerdev:
> We need to disable kube-proxy replacement since we must use theportmap feature of Cilium go gain conformance regarding the differentiation of UDP and TCP traffic on the same port. 

This kind of documentation, as.a comment right here where we set it. Future readers may be interested in the reason for the different configuration in conformance mode, or whether/when it's ok to change it back.

I still don't understand why we need to go from partial replacement to no replacement, unless the previous configuration (minus ip-masq-agent) never worked to begin with.

3u13r:
> I still don't understand why we need to go from partial replacement to no replacement,

Using `partial` is deprecated and using `false` is essentially the `partial` option now, see: https://docs.cilium.io/en/stable/network/kubernetes/kubeproxy-free/#kube-proxy-hybrid-modes

> comment right here where we set it. 

Will do hopefully later today.

burgerdev:
I deployed this branch on GCP without problems.

burgerdev:
However, the conformance test did not pass:

```yaml
   - name: '[It] [sig-network] Services should serve endpoints on same port and different
        protocols [Conformance]'
      status: failed
```

burgerdev (approved):
lgtm, thanks for the detailed warning!

PR #3064, opened 2024-05-03
terraform: Allow nodes to join the cluster when using a jump host by removing the `constellation-uid` tag

3u13r:
### Context

```
May 03 12:39:20 fedora bootstrapper[3378]: {"time":"2024-05-03T12:39:20.374587495Z","level":"ERROR","source":{"function":"github.com/edgelesssys/constellation/v2/bootstrapper/internal/joinclient.(*JoinClient).getControlPlaneIPs","file":"bootstrapper/internal/joinclient/joinclient.go","line":357},"msg":"Failed to list instances from metadata API","bootstrapper":{"join-client":{"error":"retrieving tag for instance i-03976588af8c28747: tag \"constellation-role\" not found"}}}
```

### Proposed change(s)

- Remove the `constellation-uid` tag from the jump host since the tag signals that the instance belongs to the Constellation cluster which the jump host does not.

Note that the jump host is only ever applied when using a load balancer and the debug mode.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/terraform/infrastructure/aws/main.tf
+++ b/terraform/infrastructure/aws/main.tf
@@ -213,5 +213,5 @@ module "jump_host" {
   ports                = [for port in local.load_balancer_ports : port.port]
   security_groups      = [aws_security_group.security_group.id]
   iam_instance_profile = var.iam_instance_profile_name_worker_nodes
-  additional_tags      = local.tags
+  additional_tags      = var.additional_tags
 }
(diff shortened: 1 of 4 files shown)


3u13r:
> Otherwise all instances in the VPC must have the "constellation-role" tag set.

This is not true (at least for AWS). All instances which have the same `constellation-uid` tag set must also have a `constellation-role` which makes sense. Therefore removing the `constellation-uid` tag from the jump host.

daniel-weisse (approved):

PR #3066, opened 2024-05-06
Rewrite "chain of trust" section

flxflx:
I rewrote the "Chain of trust" section to explain E2E how trust is established into a cluster. This is the first step of rewriting the entire attestation.md and potentially other parts of the docs.


m1ghtym0 (approved):
Reads much clearer now, thanks!
Not sure why vale is unhappy with "CLI's" and was happy with "CLI" before.
Maybe we need to add "CLI" to https://github.com/edgelesssys/constellation/blob/main/docs/styles/config/vocabularies/edgeless/accept.txt#L14

docs/docs/architecture/attestation.md
```diff
@@ -341,32 +341,50 @@ When an initialized node tries to join another cluster, its measurements inevita
...
+  D[Edgeless Systems]-- "signs" -->C([Runtime measurements])
+  B[CLI]-- "verifies (remote attestation)" -->E[First node]
+  E[First node]-- "verifies (remote attestation)" -->F[Other nodes]
+  C([Runtime measurements]) -.-> E[First node]
+  C([Runtime measurements]) -.-> F[Other nodes]
 ```
 
+### Upgrades
+
+Whenever a cluster is [upgraded](../workflows/upgrade.md) to a new version of the node image, the CLI sends the corresponding runtime measurements via the Kubernetes API server. The new runtime measurements are stored in etcd within the cluster and replace any previous runtime measurements. The new runtime measurements are then used automatically by the JoinServer for the verification of new nodes.
```
3u13r, line 386:
```suggestion
Whenever a cluster is [upgraded](../workflows/upgrade.md) to a new version of the node image, the CLI sends the corresponding runtime measurements via the Kubernetes API server. The new runtime measurements are stored in etcd within the cluster and replace any previous runtime measurements. The new runtime measurements are then used automatically by the JoinService for the verification of new nodes. 
```

docs/docs/architecture/attestation.md
```diff
@@ -341,32 +341,50 @@ When an initialized node tries to join another cluster, its measurements inevita
...
+* [Signature](https://cdn.confidential.cloud/constellation/v2/ref/-/stream/stable/v2.16.3/image/measurements.json.sig)
 
-The CLI contains the public key required to verify signed runtime measurements from Edgeless Systems.
-When a cluster is [created](../workflows/create.md) or [upgraded](../workflows/upgrade.md), the CLI automatically verifies the measurements for the selected image.
+The CLI contains the long-term public key of Edgeless Systems to verify the signature of downloaded runtime measurements.
 
-Thus, there's a chain of trust based on cryptographic signatures, which goes from CLI to runtime measurements to images. This is illustrated in the following diagram.
+### Cluster creation
+
+When a cluster is [created](../workflows/create.md), the CLI automatically verifies the runtime measurements of the *first node* (also known as "bootstrapper") using remote attestation. Based on this, the CLI and the first node set up a temporary TLS connection. This [aTLS](#attested-tls-atls) connection is used for three things:
```
3u13r, line 359:
```suggestion
When a cluster is [created](../workflows/create.md), the CLI automatically verifies the runtime measurements of the *first node* (also known as "bootstrapper") using remote attestation. Based on this, the CLI and the first node set up a temporary TLS connection. This [aTLS](#attested-tls-atls) connection is used for three things:
```
The bootstrapper is an explicit program inside the VM/node. Since it is part of every VM and also used on every VM, I'd drop/rephrase the `(also known as "bootstrapper)"`

docs/docs/architecture/attestation.md
```diff
@@ -341,32 +341,50 @@ When an initialized node tries to join another cluster, its measurements inevita
...
-The CLI contains the public key required to verify signed runtime measurements from Edgeless Systems.
-When a cluster is [created](../workflows/create.md) or [upgraded](../workflows/upgrade.md), the CLI automatically verifies the measurements for the selected image.
+The CLI contains the long-term public key of Edgeless Systems to verify the signature of downloaded runtime measurements.
 
-Thus, there's a chain of trust based on cryptographic signatures, which goes from CLI to runtime measurements to images. This is illustrated in the following diagram.
+### Cluster creation
+
+When a cluster is [created](../workflows/create.md), the CLI automatically verifies the runtime measurements of the *first node* using remote attestation. Based on this, the CLI and the first node set up a temporary TLS connection. This [aTLS](#attested-tls-atls) connection is used for three things:
+1. The CLI sends the runtime measurements for the applicable node image to the first node.
+2. The first node sends the [master secret](../architecture/keys.md#master-secret) of the to-be-created cluster to the CLI. The master secret is generated by the first node.
```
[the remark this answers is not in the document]
daniel-weisse: Originally we generated the master secret on node, we no longer do this, so it should be 
"The CLI sends the master secret to the first node"

Technically, we also send the runtime measurements to the first node later on once the Kubernetes is actually set up,.
Though at this point only a single node cluster exists, so the statement is still correct

malt3 (approved):
Looks good otherwise

PR #3068, opened 2024-05-07
ci: use 7zip for creating archives

daniel-weisse:
### Proposed change(s)

- Use 7zip for creating archives

### Related issue
- Amends https://github.com/edgelesssys/issues/issues/575

### DIFF
--- a/.github/actions/artifact_download/action.yml
+++ b/.github/actions/artifact_download/action.yml
@@ -16,7 +16,7 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: Install unzip
+    - name: Install 7zip
       uses: ./.github/actions/setup_bazel_nix
       with:
         nixTools: |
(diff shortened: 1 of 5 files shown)


.github/actions/artifact_download/action.yml
```diff
@@ -20,7 +20,7 @@ runs:
       uses: ./.github/actions/setup_bazel_nix
```
thomasten, line 20: nit: change the name of this step in the line above

.github/actions/artifact_upload/action.yml
```diff
@@ -26,9 +26,47 @@ runs:
       uses: ./.github/actions/setup_bazel_nix
```
thomasten, line 26: nit: change the name of this step in the line above

.github/actions/artifact_upload/action.yml
```diff
@@ -26,9 +26,47 @@ runs:
...
+        # and warn.
+        if ! ${something_exists}
+        then
+          echo "::warning:: No files/directories found with the provided path(s): ${paths}. No artifact will be uploaded."
+          exit 0
+        fi
+        for target in ${paths}
+        do
+          pushd "$(dirname "${target}")" || exit 1
+          7zz a -p'${{ inputs.encryptionSecret }}' -t7z -ms=on -mhe=on "${{ steps.tempdir.outputs.directory }}/archive.zip" "$(basename "${target}")"
```
thomasten, line 61: Should we use the correct file extension instead of using -t7z, or would that cause too many changes in other places?

.github/actions/artifact_upload/action.yml
thomasten, line 1: e2e-cleanup also needs to be changed to 7z

thomasten (changes requested):

thomasten (approved):

PR #3092, opened 2024-05-14
ci: explicitly set bazel test timeout to four hours for e2e tests

malt3:
### Context

The default test timeout was five minutes. Due to an upgrade of rules_go, this timeout is now enforced.

### Proposed change(s)

- ci: explicitly set bazel test timeout to four hours for e2e tests

### DIFF
--- a/.github/actions/e2e_lb/action.yml
+++ b/.github/actions/e2e_lb/action.yml
@@ -21,7 +21,7 @@ runs:
       run: |
         kubectl apply -f ns.yml
         kubectl apply -f lb.yml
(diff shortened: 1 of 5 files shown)


daniel-weisse (commented):
We will also need this in [`.github/workflows/e2e-test-provider-example.yml:l456`](https://github.com/edgelesssys/constellation/blob/547d435acab343abcafe95f9f5a9fa50250c66eb/.github/workflows/e2e-test-provider-example.yml#L456)

daniel-weisse (approved):
thanks

PR #3113, opened 2024-05-21
operators: ignore node deletion errors on absence

burgerdev:
### Context

If a node deletion operation fails transiently after the instance was deleted at the CSP, we need to ensure that the retry succeeds, even if the instance we attempt to delete is already gone at the CSP.

Example operator logs: https://gist.github.com/burgerdev/c80de86ae44552a383364ef031243a41

### Proposed change(s)

- AWS: Ignore the specific error message for missing images. 
- Azure: nothing - deleting absent instance IDs is no error here.
- GCP: turn on `SkipInstancesOnValidationError`, which is designed for this use case.

### Additional info
- [AB#3533](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3533)

### Checklist

   - [x] [upgrade test on GCP](https://github.com/edgelesssys/constellation/actions/runs/9187208968) 
- [x] Link to Milestone


operators/constellation-node-operator/internal/cloud/aws/client/nodeimage.go
```diff
@@ -217,3 +218,10 @@ func (c *Client) DeleteNode(ctx context.Context, providerID string) error {
 func toPtr[T any](v T) *T {
 	return &v
 }
+
+func isInstanceNotFoundError(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(err.Error(), "Instance Id not found")
+}
```
daniel-weisse, line 227:
```suggestion
func isInstanceNotFoundError(err error) bool {
	var respErr *awshttp.ResponseError // using "awshttp" for package github.com/aws/aws-sdk-go-v2/aws/transport/http
	return errors.As(err, &respErr) && respErr.HTTPStatusCode() == http.StatusNotFound
}

```
From https://github.com/aws/aws-sdk-go-v2/blob/main/CHANGELOG.md#error-handling
Did not tests this, but if it works it would be preferable  over string comparisons.

burgerdev: That would be preferable indeed, but AWS does not return 404s in that case, but 400s. I meant to also link to the logs but forgot - fixed that now (see above). I would not want to infer nonexistence from a 400, because that might hide other failure conditions, so I'm back at string comparison :/ 

```xml
<ErrorResponse xmlns="http://autoscaling.amazonaws.com/doc/2011-01-01/">
  <Error>
    <Type>Sender</Type>
    <Code>ValidationError</Code>
    <Message>Instance Id not found - No managed instance found for instance ID: i-061c63c5eb45f0416</Message>
  </Error>
  <RequestId>702f44ea-6d55-480b-8dfa-3da9e62f1634</RequestId>
</ErrorResponse>
```

malt3 (approved):

daniel-weisse (approved):

PR #3114, opened 2024-05-21
bootstrapper: prioritize etcd disk I/O

msanft:
### Context

The etcd issues like the notorious "etcdserver: leader changed" are most likely caused due to etcd's disk activity conflicting with other disk-heavy operations during cluster initialization / joining such as the pulling of container images. To mitigate this, we can prioritize the disk operations of the etcd process over those of other processes by giving it an I/O priority in the kernel. It will not consider pod restarts of etcd, which are out of scope for this PR.

**How to move forward:**
We'll need to see (over a course of, say, 2 weeks) if this change is enough to get rid of the issues. If not, we'll probably need to use faster disks per default on Azure. This would probably require switching to the (~90€/M) 512Gb P20 disk, which satisfies the [requirements of etcd](https://etcd.io/docs/v3.6/op-guide/hardware/#example-hardware-configurations). This should be done in a separate testing period to gather differential results.

### Proposed change(s)

- Upon initializing / joining / re-joining a Constellation cluster, find the etcd process on the node and set it's I/O priority to the highest possible value (i.e. RT (real-time) and a priority value of 0 (highest priority)).

### Additional info

- [AB#3812](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3812)

### Checklist

  - [AWS](https://github.com/edgelesssys/constellation/actions/runs/9176378637)
  - [Azure SEV-SNP](https://github.com/edgelesssys/constellation/actions/runs/9176353527)
  - [Azure TDX](https://github.com/edgelesssys/constellation/actions/runs/9176356709)
  - [GCP](https://github.com/edgelesssys/constellation/actions/runs/9176363691)
- [x] Link to Milestone

### DIFF
--- a/bootstrapper/internal/etcdio/etcdio.go
+++ b/bootstrapper/internal/etcdio/etcdio.go
@@ -56,7 +56,7 @@ func NewClient(log *slog.Logger) *Client {
 // Since it might be possible that the process just started (if this method is called
 // right after the kubelet started), it retries to do its work each second
 // until it succeeds or the timeout of 10 seconds is reached.
-func (c *Client) PrioritizeIO() error {
+func (c *Client) PrioritizeIO() {
 	ticker := time.NewTicker(1 * time.Second)
 	defer ticker.Stop()
 	timeout, cancel := context.WithTimeout(context.Background(), 10*time.Second)
@@ -67,19 +67,19 @@ func (c *Client) PrioritizeIO() error {
 		err := c.setIOPriority()
 		if err == nil {
 			// Success, return directly
-			return nil
+			return
 		} else if errors.Is(err, ErrNoEtcdProcess) {
 			c.log.Info("No etcd process found, retrying")
 		} else {
 			c.log.Warn("Prioritizing etcd I/O failed", "error", err)
-			return fmt.Errorf("prioritizing etcd I/O: %w", err)
+			return
 		}
 
 		select {
 		case <-ticker.C:
 		case <-timeout.Done():
 			c.log.Warn("Timed out waiting for etcd to start")
-			return nil
+			return
 		}
 	}
 }
--- a/bootstrapper/internal/etcdio/etcdio.go
+++ b/bootstrapper/internal/etcdio/etcdio.go
@@ -19,7 +19,9 @@ import (
 )
 
 var (
-	ErrNoEtcdProcess         = errors.New("no etcd process found on node")
+	// ErrNoEtcdProcess is returned when no etcd process is found on the node.
+	ErrNoEtcdProcess = errors.New("no etcd process found on node")
+	// ErrMultipleEtcdProcesses is returned when multiple etcd processes are found on the node.
 	ErrMultipleEtcdProcesses = errors.New("multiple etcd processes found on node")
 )
 
@@ -38,18 +40,18 @@ const (
 	targetPrio  = 0 // Highest priority within the class
 )
 
-// EtcdIOClient is a client for managing etcd I/O.
-type EtcdIOClient struct {
+// Client is a client for managing etcd I/O.
+type Client struct {
 	log *slog.Logger
 }
 
 // NewClient creates a new etcd I/O management client.
-func NewClient(log *slog.Logger) *EtcdIOClient {
-	return &EtcdIOClient{log: log}
+func NewClient(log *slog.Logger) *Client {
+	return &Client{log: log}
 }
 
 // PrioritizeIO tries to find the etcd process on the node and prioritizes its I/O.
-func (c *EtcdIOClient) PrioritizeIO() error {
+func (c *Client) PrioritizeIO() error {
 	// find etcd process(es)
 	pid, err := c.findEtcdProcess()
 	if err != nil {
@@ -70,7 +72,7 @@ func (c *EtcdIOClient) PrioritizeIO() error {
 }
 
 // findEtcdProcess tries to find the etcd process on the node.
-func (c *EtcdIOClient) findEtcdProcess() (int, error) {
+func (c *Client) findEtcdProcess() (int, error) {
 	procDir, err := os.Open("/proc")
 	if err != nil {
 		return 0, fmt.Errorf("opening /proc: %w", err)
--- a/bootstrapper/internal/etcdio/etcdio.go
+++ b/bootstrapper/internal/etcdio/etcdio.go
@@ -1,0 +1,120 @@
+/*
+Copyright (c) Edgeless Systems GmbH
+
+SPDX-License-Identifier: AGPL-3.0-only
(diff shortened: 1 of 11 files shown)


bootstrapper/internal/kubernetes/kubernetes.go
```diff
@@ -237,11 +257,21 @@ func (k *KubeWrapper) JoinCluster(ctx context.Context, args *kubeadm.BootstrapTo
...
 	}
 
+	k.log.Info("Prioritizing etcd I/O")
+	err = k.etcdIOPrioritizer.PrioritizeIO()
+	if errors.Is(err, etcdio.ErrNoEtcdProcess) {
+		k.log.Warn("Skipping etcd I/O prioritization as etcd process is not running. " +
+			"This is expected if this node is a non-control-plane node.")
+	} else if err != nil {
+		return fmt.Errorf("prioritizing etcd I/O: %w", err)
+	}
```
daniel-weisse, line 273:
```suggestion
	if peerRole == role.ControlPlane {
		k.log.Info("Prioritizing etcd I/O")
		if err := k.etcdIOPrioritizer.PrioritizeIO(); err != nil {
			return fmt.Errorf("prioritizing etcd I/O: %w", err)
		}
	}
```
Instead of checking the error, we can just skip the priotization if we arent running on a control plane node

bootstrapper/internal/kubernetes/kubernetes.go
```diff
@@ -301,6 +331,15 @@ func (k *KubeWrapper) StartKubelet() error {
...
 	}
 
+	k.log.Info("Prioritizing etcd I/O")
+	err := k.etcdIOPrioritizer.PrioritizeIO()
+	if errors.Is(err, etcdio.ErrNoEtcdProcess) {
+		k.log.Warn("Skipping etcd I/O prioritization as etcd process is not running. " +
+			"This is expected if this node is a non-control-plane node.")
+	} else if err != nil {
+		return fmt.Errorf("prioritizing etcd I/O: %w", err)
+	}
```
daniel-weisse, line 341:
```suggestion
	if peerRole == role.ControlPlane {
		k.log.Info("Prioritizing etcd I/O")
		if err := k.etcdIOPrioritizer.PrioritizeIO(); err != nil {
			return fmt.Errorf("prioritizing etcd I/O: %w", err)
		}
	}
```
Same thing here

msanft: We don't know about our role here. I think it's easier to check for the error in this case

bootstrapper/internal/kubernetes/kubernetes.go
```diff
@@ -237,11 +257,23 @@ func (k *KubeWrapper) JoinCluster(ctx context.Context, args *kubeadm.BootstrapTo
...
 	}
 
+	// If on control plane (and thus with etcd), try to prioritize etcd I/O.
+	if peerRole == role.ControlPlane {
+		k.log.Info("Prioritizing etcd I/O")
+		err = k.etcdIOPrioritizer.PrioritizeIO()
+		if errors.Is(err, etcdio.ErrNoEtcdProcess) {
+			k.log.Warn("No etcd process found, skipping I/O prioritization. Is this a single-node cluster?")
+		} else if err != nil {
+			return fmt.Errorf("prioritizing etcd I/O: %w", err)
```
burgerdev, line 273: Do we really want to fail hard in this case? The way we detect etcd processes is inherently racy, so it might be better to just warn in that case.

msanft: That's a good point.

bootstrapper/internal/kubernetes/kubernetes.go
```diff
@@ -154,7 +153,7 @@ func (k *KubeWrapper) InitCluster(
 
 	k.log.Info("Prioritizing etcd I/O")
 	if err := k.etcdIOPrioritizer.PrioritizeIO(); err != nil {
-		return nil, fmt.Errorf("prioritizing etcd I/O: %w", err)
+		k.log.Warn("prioritizing etcd I/O", "error", err)
```
daniel-weisse, line 156:
```suggestion
		k.log.Warn("Prioritizing etcd I/O failed", "error", err)
```
Log messages should start uppercase

daniel-weisse (approved):

daniel-weisse:
Lets get this merged so we can see if it fixes things during our daily and weekly e2e tests

burgerdev (approved):

bootstrapper/internal/kubernetes/kubernetes.go
```diff
@@ -329,9 +330,30 @@ func (k *KubeWrapper) StartKubelet() error {
...
 
-	k.log.Info("Prioritizing etcd I/O")
-	if err := k.etcdIOPrioritizer.PrioritizeIO(); err != nil {
-		k.log.Warn("Prioritizing etcd I/O failed", "error", err)
+	// We just started the kubelet service here, so let's give it some time to start etcd.
+	ticker := time.NewTicker(1 * time.Second)
+	defer ticker.Stop()
+	timeout, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	for {
```
daniel-weisse, line 338: Wouldn't it be easier to handle this wait and retry logic in the etcdIO client itself?
I also think in this case it would be best to have `PrioritizeIO` not return any errors at all, but simply log potential failures.

bootstrapper/internal/etcdio/etcdio.go
```diff
@@ -0,0 +1,156 @@
...
+// NewClient creates a new etcd I/O management client.
+func NewClient(log *slog.Logger) *Client {
+	return &Client{log: log}
+}
+
+// PrioritizeIO tries to prioritize the I/O of the etcd process.
+// Since it might be possible that the process just started (if this method is called
+// right after the kubelet started), it retries to do its work each second
+// until it succeeds or the timeout of 10 seconds is reached.
+func (c *Client) PrioritizeIO() error {
```
daniel-weisse, line 59:
```suggestion
func (c *Client) PrioritizeIO() {
```
Lets just drop any errors in this function.
Setting IO priority is not critical for the functionality of the cluster, and it clears up the caller side, which doesnt return any errors anyway.

msanft:
Some more E2E tests:
- https://github.com/edgelesssys/constellation/actions/runs/9190836227
- https://github.com/edgelesssys/constellation/actions/runs/9190843281
- https://github.com/edgelesssys/constellation/actions/runs/9190849893

daniel-weisse (approved):

burgerdev (approved):

PR #3141, opened 2024-06-04
Update conditions on when to pick reviewer

daniel-weisse:
### Context
Lots of dependency PRs don't get assigned a reviewer.
I hope this PR fixes some of the cases where the review assignment  is skipped.

### Proposed change(s)

- Add more triggers to the assign_reviewer workflow
- Only run the workflow if no reviewers are assigned
- Update the name we check against to `renovate[bot]` following this comment: https://github.com/renovatebot/renovate/discussions/13704#discussioncomment-7079564

### DIFF
--- a/.github/workflows/assign_reviewer.yml
+++ b/.github/workflows/assign_reviewer.yml
@@ -2,7 +2,12 @@ name: Assign Reviewer
 
 on:
   pull_request:
-    types: ["labeled"]
+    types:
+      - opened
+      - reopened
+      - edited
+      - review_request_removed
+      - labeled
 
 permissions:
   pull-requests: write
@@ -10,7 +15,7 @@ permissions:
 jobs:
   assign_reviewer:
     runs-on: ubuntu-latest
-    if: github.event.label.name == 'dependencies' && github.event.pull_request.user.login == 'app/renovate'
+    if: contains(github.event.pull_request.labels.*.name, 'dependencies') && github.event.pull_request.user.login == 'app/renovate' && github.event.pull_request.requested_reviewers == null
     steps:
     - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
     - name: Pick assignee
--- a/.github/workflows/assign_reviewer.yml
+++ b/.github/workflows/assign_reviewer.yml
@@ -15,7 +15,7 @@ permissions:
 jobs:
   assign_reviewer:
     runs-on: ubuntu-latest
-    if: contains(github.event.pull_request.labels.*.name, 'dependencies') && github.event.pull_request.user.login == 'app/renovate' && github.event.pull_request.requested_reviewers == null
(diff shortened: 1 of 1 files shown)


.github/workflows/assign_reviewer.yml
```diff
@@ -2,15 +2,20 @@ name: Assign Reviewer
...
+      - labeled
 
 permissions:
   pull-requests: write
 
 jobs:
   assign_reviewer:
     runs-on: ubuntu-latest
-    if: github.event.label.name == 'dependencies' && github.event.pull_request.user.login == 'app/renovate'
+    if: contains(github.event.pull_request.labels.*.name, 'dependencies') && github.event.pull_request.requested_reviewers == null && (github.event.pull_request.user.login == 'renovate[bot]')
```
burgerdev, line 18: How does `github.event.pull_request.requested_reviewers == null` interact with auto-assigned code owners? The idea was that the workflow assigns a reviewer in addition to the code owners.

burgerdev, line 18: I believe `app/renovate` is the right login name - note that the linked discussion compares `github.actor` and see the output of 

```sh
gh pr list --repo edgelesssys/constellation --json author | jq | grep renovate
```

burgerdev (changes requested):

.github/workflows/assign_reviewer.yml
daniel-weisse: I have not actually tested this, but from what I can gather `requested_reviewers` is a list of assigned reviewers, which is not empty if there are auto assigned reviewers.

daniel-weisse: `github.actor` and `github.event.pull_request.user.login` both refer to the username of an entity.
Since all workflows triggered after https://github.com/edgelesssys/constellation/pull/3091 was merged were skipped, I would assume that `app/renovate` is not the correct value here.

daniel-weisse: Actually, let me test if I can trigger these workflows on existing renovate PRs and see what the needed value is

daniel-weisse: https://github.com/edgelesssys/constellation/actions/runs/9368485665/job/25790643346?pr=3138#step:2:8

burgerdev: Thanks for checking, I'm surprised but convinced.

burgerdev: We could also try matching on `user.type == Bot` (https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request).

daniel-weisse:
Managed to fix the workflow:
* For some reason `github.event.pull_request.user.login` is `renovate[bot]` not `app/renovate` like the `gh` CLI would suggest
* `github.event.pull_request.requested_reviewers` by itself simply evaluates to `Array`. We need to call `toJson` on it to parse it to a JSON object, at which point we can evaluate if that structure is an empty array. But not actually, because its JavaScript we check if it is a string equal to `'[]'`

burgerdev (approved):

PR #3152, opened 2024-06-07
docs: adjust MAA updating

msanft:
### Context

The docs said that the permissions for updating the MAA policy are only required if the enforcement policy is set to use the MAA. However, this does not reflect our current code, which creates and patches the MAA unconditionally. This is necessary to ensure that a cluster created without using the MAA can switch to using the MAA later on.

### Proposed change(s)

- Remove the parts of the docs saying that the MAA-related permissions are only necessary if the enforcement policy is set to `MAAFallback`.

### Checklist

- [x] Link to Milestone


daniel-weisse (approved):

thomasten (approved):
pls backport to versioned_docs before merge

PR #3153, opened 2024-06-07
cli: return a more helpful error message on MAA patch failure

msanft:
### Context

The updating of the MAA policy is not done via Terraform, hence failures cannot be debugged through the Terraform logs. We only return the status code currently, which is not very helpful for users running into a failure here. The response body should inform the user about which permissions are missing, hence it should also be shown to the user.

### Proposed change(s)

- Return both status code and response body in case of a policy patching failure.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/maa/patch.go
+++ b/internal/maa/patch.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"io"
 	"net/http"
 
 	"github.com/Azure/azure-sdk-for-go/profiles/latest/attestation/attestation"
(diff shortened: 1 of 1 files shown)


internal/maa/patch.go
```diff
@@ -55,10 +56,14 @@ func (p AzurePolicyPatcher) Patch(ctx context.Context, attestationURL string) er
...
-	resp.Body.Close()
+	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		return fmt.Errorf("updating attestation policy: unexpected status code: %s", resp.Status)
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return fmt.Errorf("updating attestation policy: unexpected status code: %s\nread response body: %w", resp.Status, err)
+		}
+		return fmt.Errorf("updating attestation policy: unexpected status code: %s\nresponse body: %s", resp.Status, string(body))
```
daniel-weisse, line 66:
```suggestion
		body, _ := io.ReadAll(resp.Body)
		return fmt.Errorf("updating attestation policy: unexpected status code: %s: %s", resp.Status, string(body))
```
nit

daniel-weisse (approved):

PR #3159, opened 2024-06-10
debugd: return systemd logs if restarting a unit fails

daniel-weisse:
### Context
We have observed multiple cases of `debugd` failing when the `bootstrapper` unit should be restarted.
This is tracked in https://github.com/edgelesssys/issues/issues/609
Since this happens rather early in the deployment process, we don't have any logs pushed to OpenSearch at this point

### Proposed change(s)

- Read and return systemd logs of the unit that failed to be restarted on failure
- Wait until logcollection containers are running before continuing
  - This doesn't really solve the issue of no logs being pushed to OpenSearch since logs are not pushed on creation, but aggregated every 30 seconds and then pushed to OpenSearch, but I think its a good practice regardless

### DIFF
--- a/debugd/internal/debugd/deploy/wrappers.go
+++ b/debugd/internal/debugd/deploy/wrappers.go
@@ -8,6 +8,7 @@ package deploy
 
 import (
 	"context"
+	"os/exec"
 
 	"github.com/coreos/go-systemd/v22/dbus"
 )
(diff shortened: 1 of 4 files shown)


debugd/internal/debugd/deploy/wrappers.go
```diff
@@ -48,3 +49,82 @@ func (c *dbusConnWrapper) ReloadContext(ctx context.Context) error {
...
 	c.conn.Close()
 }
+
+type journalctlWrapper struct{}
+
+func (j *journalctlWrapper) readJournal(unit string) string {
+	out, _ := exec.CommandContext(context.Background(), "journalctl", "-u", unit, "--no-pager").CombinedOutput()
+	return string(out)
+}
+
```
msanft, line 59: We could add a comment here that this is infeasible due to cgo and library deps

msanft (approved):

PR #3164, opened 2024-06-12
cli: enable JSON output for `constellation verify` on Azure TDX

daniel-weisse:
### Context
Our CI parses the JSON output of `constellation verify` in order to upload the latest CC claims to Constellation's CDN.
For now, this was only enabled for SEV-SNP attestation variants.

### Proposed change(s)

- Enabled JSON output for Azure TDX
  - The TDX library we use also handles JSON marshalling of the quote, so I didn't implement any extra methods, unlike we did for SNP. This means we won't have pretty formatted text output for TDX variants, but this can always be added at a later date since it doesn't have high priority: [AB#4266](https://dev.azure.com/Edgeless/Edgeless/_workitems/edit/4266)
- Simplify the `constellation verify` code a bit by removing the formatter factory

### Additional info

- [AB#3798](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3798)

### DIFF
--- a/cli/internal/cmd/verify.go
+++ b/cli/internal/cmd/verify.go
@@ -21,10 +21,9 @@ import (
 	"strconv"
 	"strings"
 
-	tpmProto "github.com/google/go-tpm-tools/proto/tpm"
-
 	"github.com/edgelesssys/constellation/v2/internal/api/attestationconfigapi"
 	"github.com/edgelesssys/constellation/v2/internal/atls"
+	azuretdx "github.com/edgelesssys/constellation/v2/internal/attestation/azure/tdx"
 	"github.com/edgelesssys/constellation/v2/internal/attestation/choose"
 	"github.com/edgelesssys/constellation/v2/internal/attestation/measurements"
 	"github.com/edgelesssys/constellation/v2/internal/attestation/snp"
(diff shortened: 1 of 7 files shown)


cli/internal/cmd/verify.go
```diff
@@ -255,82 +248,91 @@ func (c *verifyCmd) validateEndpointFlag(cmd *cobra.Command, stateFile *state.St
...
 
-	instanceInfo, err := extractInstanceInfo(doc)
-	if err != nil {
+	if (attestationCfg.GetVariant().Equal(variant.AWSSEVSNP{}) ||
+		attestationCfg.GetVariant().Equal(variant.AzureSEVSNP{}) ||
+		attestationCfg.GetVariant().Equal(variant.GCPSEVSNP{})) {
+		return snpFormatJSON(ctx, doc.InstanceInfo, attestationCfg, log)
+	}
+
+	return tdxFormatJSON(doc.InstanceInfo, attestationCfg)
```
msanft, line 265: Should we make this a switch case, with an error message in case another method is used?

msanft (approved):

msanft (approved):

PR #3166, opened 2024-06-13
config: allow "latest" pseudo-version for Azure TDX config values

daniel-weisse:
### Context
Attestation values in the config for Azure TDX currently use default values that require manual updates by users if newer versions are published.
Instead, TDX attestation variants should offer the same usability feature as SEV-SNP variants, where these values can be set to "latest", which causes them to be fetched and replaced by the latest recorded attestation values from our API.

### Proposed change(s)

- Enable Azure TDX config values to use "latest" as placeholder
- Refactor the code in `/internal/api/attestationconfigapi/` to allow for more attestation variants than just SEV-SNP

### Additional info

- [AB#3798](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3798)

### Checklist

  - [x] [attestation config API e2e test](https://github.com/edgelesssys/constellation/actions/runs/9548424286)
  - [x] [e2e verify Azure TDX](https://github.com/edgelesssys/constellation/actions/runs/9517574347)
- [ ] Link to Milestone


internal/api/attestationconfigapi/version.go
```diff
@@ -0,0 +1,127 @@
...
+	// XFAM is the latest XFAM field.
+	XFAM [8]byte `json:"xfam"`
+}
+
+// VersionAPIEntry is the request to get the version information of the specific version in the config api.
+//
+// TODO: Because variant is not part of the marshalled JSON, fetcher and client methods need to fill the variant property.
+// In API v2 we should embed the variant in the object and remove some code from fetcher & client.
+// That would remove the possibility of some fetcher/client code forgetting to set the variant.
+type VersionAPIEntry struct {
```
derpsteb, line 53: I would use a different name here as there already is the versionapi. Which is a distinct thing. Maybe one of `SVNVersion`, `AttestationVersion`, `TCBVersion`, `FirmwareVersion`. Or the plural of those.

internal/api/attestationconfigapi/cli/upload.go
```diff
@@ -92,56 +92,73 @@ func runUpload(cmd *cobra.Command, args []string) (retErr error) {
...
 		}
+
+		newVersion = convertQuoteToTDXVersion(report)
+		log.Info(fmt.Sprintf("Input TDX report: %+v", newVersion))
+
+		//if err := apiClient.UploadTDXVersionLatest(
+		//	ctx, cfg.variant, newVersion, latestVersionInAPI.TDXVersion, cfg.uploadDate, cfg.force,
+		//); err != nil && !errors.Is(err, client.ErrNoNewerVersion) {
+		//	return fmt.Errorf("updating latest version: %w", err)
+		//}
```
derpsteb, line 153:
```suggestion
```
?

internal/api/attestationconfigapi/cli/upload.go
```diff
@@ -92,56 +92,73 @@ func runUpload(cmd *cobra.Command, args []string) (retErr error) {
...
+		}
+
+		newVersion = convertTCBVersionToSNPVersion(report.SNPReport.LaunchTCB)
+		log.Info(fmt.Sprintf("Input SNP report: %+v", newVersion))
+
+		//if err := apiClient.UploadSEVSNPVersionLatest(
+		//	ctx, cfg.variant, newVersion, latestVersionInAPI.SEVSNPVersion, cfg.uploadDate, cfg.force,
+		//); err != nil && !errors.Is(err, client.ErrNoNewerVersion) {
+		//	return fmt.Errorf("updating latest version: %w", err)
+		//}
```
derpsteb, line 135:
```suggestion
```
?

internal/api/attestationconfigapi/version.go
```diff
@@ -0,0 +1,127 @@
...
+func (i VersionAPIEntry) Validate() error {
+	return nil
+}
+
+// VersionList is the request to retrieve of all versions in the API for one attestation variant.
+//
+// TODO: Because variant is not part of the marshalled JSON, fetcher and client methods need to fill the variant property.
+// In API v2 we should embed the variant in the object and remove some code from fetcher & client.
+// That would remove the possibility of some fetcher/client code forgetting to set the variant.
+type VersionList struct {
```
derpsteb, line 83: Following the above comment this would have to be adapted as well.

internal/api/attestationconfigapi/cli/client/reporter.go
```diff
@@ -0,0 +1,358 @@
...
+// the latest version in the API if there is an update.
+// force can be used to bypass the validation logic against the cached versions.
+func (c Client) UploadLatestVersion(
+	ctx context.Context, attestationVariant variant.Variant,
+	inputVersion, latestVersionInAPI any,
+	now time.Time, force bool,
+) error {
+	// Validate input versions against configured attestation variant
+	// This allows us to skip these checks in the individual variant implementations
+	var err error
```
burgerdev, line 45: Should this error be checked?

internal/api/attestationconfigapi/cli/client/reporter.go
```diff
@@ -0,0 +1,358 @@
...
+
+	return *minimalVersion, minimalDate, nil
+}
+
+func isInputNewerThanOtherVersion(variant variant.Variant, inputVersion, otherVersion any) bool {
+	var result bool
+	actionForVariant(variant,
+		func() {
+			input := inputVersion.(attestationconfigapi.TDXVersion)
+			other := otherVersion.(attestationconfigapi.TDXVersion)
```
burgerdev, line 215: Are we sure at this point that the versions can't have the wrong type?

internal/api/attestationconfigapi/cli/client/reporter.go
```diff
@@ -0,0 +1,358 @@
...
+		},
+		func() {
+			a.snp = version.(attestationconfigapi.SEVSNPVersion)
+		},
+	)
+}
+
+func actionForVariant(
+	attestationVariant variant.Variant,
+	tdxAction func(), snpAction func(),
```
burgerdev, line 306: Why did you choose this over, say, an interface implemented by structs `tdxAction` and `snpAction`?

internal/api/attestationconfigapi/cli/client/reporter.go
```diff
@@ -0,0 +1,358 @@
...
+func (a apiVersionObject) getVersion() any {
+	var res any
+	actionForVariant(a.variant,
+		func() {
+			res = a.tdx
+		},
+		func() {
+			res = a.snp
+		},
+	)
```
burgerdev, line 289: Would be clearer with a switch statement.

internal/api/attestationconfigapi/cli/client/reporter.go
```diff
@@ -0,0 +1,358 @@
...
+			res, err = json.Marshal(a.snp)
+		},
+	)
+	return res, err
+}
+
+func (a *apiVersionObject) UnmarshalJSON(data []byte) error {
+	errTDX := json.Unmarshal(data, &a.tdx)
+	errSNP := json.Unmarshal(data, &a.snp)
+	if errTDX == nil || errSNP == nil {
```
burgerdev, line 252: Should this be XOR?

internal/api/attestationconfigapi/cli/client/reporter.go
```diff
@@ -0,0 +1,358 @@
...
+	case variant.AWSSEVSNP{}, variant.AzureSEVSNP{}, variant.GCPSEVSNP{}:
+		snpAction()
+	case variant.AzureTDX{}:
+		tdxAction()
+	default:
+		panic(fmt.Sprintf("unsupported attestation variant: %s", attestationVariant))
+	}
+}
+
+// isInputNewerThanOtherSEVSNPVersion compares all version fields and returns true if any input field is newer.
```
burgerdev, line 318: Is this actually what the function checks? If one version is newer and one is older, it would return false. No idea whether this could actually occur, though.

burgerdev (commented):
Although I'm only halfway through, I wanted to leave a few comments.

internal/api/attestationconfigapi/cli/client/reporter.go
daniel-weisse: Yes, if this panics the code is incorrect.

internal/api/attestationconfigapi/cli/client/reporter.go
daniel-weisse: There are many different cases where we need to execute slightly different code paths depending on whether we are uploading a TDX or an SEV-SNP version.
These code paths do not always work with the same data types, hence adding interfaces for this would require lots of extra code.
The goal here was to reduce code changes that have to be done when a new TDX or SEV-SNP is introduced.
With the current approach, adding a new variant should only require adding it to the switch case inside this function.

internal/api/attestationconfigapi/cli/client/reporter.go
daniel-weisse: See my other comment.
Using the same `actionForVariant` function in all cases where we need to perform different actions depending on the variant allows us to more easily add new versions without having to perform many code changes.

internal/api/attestationconfigapi/cli/client/reporter.go
daniel-weisse: I mostly copied the comments from the existing files. Imo a more correct comment would be "returns false if any version is older or the versions are the same"

internal/api/attestationconfigapi/cli/client/reporter.go
daniel-weisse: For our use case it does not matter if both versions could be unmarshalled.
It does make this a bit more forwardcompatible though, if we ever decided to include SEV and TDX versions in the same object in s3.

internal/api/attestationconfigapi/cli/client/reporter.go
burgerdev: If a new variant is introduced, you need to change all the call sites of `actionForVariant` and add a new func, don't you? I don't think that's less work than adding a new implementation of an interface, but maybe I'm missing something.

daniel-weisse: Only if we introduce a attestation variant that is not based on TDX or SEV-SNP.
E.g. adding `gcp-tdx` or `aws-tdx` would only require adding it to the type switch in `actionForVariant`.
Adding a completely new variant, e.g. something like `arm-realms`, would require more work of course.

internal/api/attestationconfigapi/version.go
daniel-weisse: What about simply going with `Entry`?

internal/api/attestationconfigapi/version.go
daniel-weisse: How do you feel about simply using `List`?

elchead (approved):
I looked at the attestationconfigapi, upload tool and the TF provider changes.

burgerdev (approved):

PR #3171, opened 2024-06-17
cli: Fix unmarshalling of sev-snp attestation documents in "constellation …

daniel-weisse:
…verify"

### Proposed change(s)

-

### Additional info

- AB#1234

### Checklist

- [ ] Link to Milestone


cli/internal/cmd/verify.go
```diff
@@ -191,16 +193,22 @@ func (c *verifyCmd) verify(cmd *cobra.Command, verifyClient verifyClient, config
...
+		if err != nil {
+			return fmt.Errorf("printing attestation document: %w", err)
+		}
+
 	default:
 		return fmt.Errorf("invalid output value for formatter: %s", c.flags.output)
 	}
-	if err != nil {
-		return fmt.Errorf("printing attestation document: %w", err)
-	}
```
burgerdev, line 203: This is functionally equivalent, right? Why did you decide to inline this?

cli/internal/cmd/verify_test.go
```diff
@@ -208,18 +208,20 @@ func TestVerify(t *testing.T) {
...
 	testCases := map[string]struct {
-		doc     string
+		doc     []byte
+		attCfg  config.AttestationCfg
 		wantErr bool
 	}{
 		"invalid doc": {
-			doc:     "invalid",
+			doc:     []byte("invalid"),
+			attCfg:  &config.AzureSEVSNP{},
```
burgerdev, line 217: Do we anticipate adding more test cases? If not, I'd make this one case and remove the loop.

cli/internal/cmd/verify.go
daniel-weisse: I think this way is cleaner, since not all cases set an error.
I would rather handle those errors in the case they originate.

cli/internal/cmd/verify_test.go
daniel-weisse: There should be more cases for this test.
Currently it doesn't do anything helpful

burgerdev (approved):

PR #3177, opened 2024-06-19
helm: upgrade cert-manager from v1.12.6 to v1.15.0

daniel-weisse:
### Proposed change(s)

- Upgrade cert-manager to v1.15.0

### Checklist

  - [x] [Upgrade test](https://github.com/edgelesssys/constellation/actions/runs/9583195374)
  - [x] [e2e test](https://github.com/edgelesssys/constellation/actions/runs/9583190722)
- [x] Link to Milestone


burgerdev (approved):
Chart import lgtm.

1. This probably needs a changelog entry because of the [breaking changes from 1.12 to 1.13](https://cert-manager.io/docs/releases/upgrading/upgrading-1.12-1.13).
2. Do we need to consider the [upgrade instructions](https://cert-manager.io/docs/installation/upgrade/#crds-managed-using-helm) (`reset-then-reuse-values`)? I wonder whether we might miss something when we just apply the charts.

daniel-weisse:
We don't have any `featureGates` set, so the breaking change is not an issue.
As for the `reset-then-reuse-values`, I think @derpsteb probably knows best how the helm charts are configured for upgrade, but I know that we disable value reuse, so we should always use the values from the embedded chart when applying upgrades

daniel-weisse:
We can add some notes in the changelog

Issue #3181, opened 2024-06-19
Incorrect docs: verifying third-party clusters with CLI


cdata:
### Issue description

The docs found at https://docs.edgeless.systems/constellation/workflows/verify-cluster#custom-arguments claim that any cluster can be verified with two pieces of information: the IP of the `VerificationService` and the cluster's `clusterID`. However, this is not true; the CLI gives the following error:

```
Error: loading config file: unable to find constellation-conf.yaml - use `constellation config generate` to generate it first
```

As an aside: we would like to enable third-parties to perform a verification step against our cluster. Is it feasible to do this, and if so how should we do it?

### Steps to reproduce the behavior

1. Set up a cluster and make the `VerificationService` visible on the network
2. In a separate workspace, run `constellation verify -e $THE_SERVICE_IP --cluster-id $THE_CLUSTER_ID`

### Version

Version:        v2.16.4 (Enterprise build; see documentation for license agreement)
GitCommit:      9603dcd09e96475a7378cc4ec7b0126a2d0a5ed5
GitTreeState:   clean
BuildDate:      2024-05-14T13:30:56
GoVersion:      go1.22.3 X:nocoverageredesign
Compiler:       bazel/gc
Platform:       linux/amd6

### Constellation Config

N/A

daniel-weisse:
>the CLI gives the following error

`constellation verify` requires a config file to obtain the measurements and configuration to verify the cluster with, as described in the [featch measurements section](https://docs.edgeless.systems/constellation/workflows/verify-cluster#fetch-measurements) at the start of the page.
There is actually a bug in the CLI however, which requires you to also have a correctly initialized `constellation-state.yaml` file in your working directory. This should not be necessary when setting the `--endpoint` and `--cluster-id` flags
I'll update the docs to reflect this also in the "custom arguments" section and fix the relevant code section.

>we would like to enable third-parties to perform a verification step against our cluster. Is it feasible to do this, and if so how should we do it?

You first need to establish with what goal the third party is verifying your cluster.
Since Constellation does not lock out the operator of the cluster (you), the operator is essentially free to do anything to an existing and initialized cluster.
This means the third party has to trust you first and foremost when connecting to the cluster.
If this is not given, any additional verification of the cluster through the Constellation CLI or services is essentially meaningless.

If this trust is given, but the third party wants to go the extra mile, they can use `constellation verify`.
[...]

PR #3184, opened 2024-06-20
cli: fix `constellation verify` depending on an initialized `constellation-state.yaml` file

daniel-weisse:
### Context
`constellation verify` should be usable with only a config file, the IP of the service, and the cluster ID of a Constellation cluster

### Proposed change(s)

- Remove dependency on an initialized `constellation-state.yaml` when executing `constellation verify` with `--endpoint` and `--cluster-id` flags set
- More prominently mention requirement for `constellation-conf.yaml` for `constellation verify` in our docs

### Related issue
- fixes https://github.com/edgelesssys/constellation/issues/3181

### DIFF
--- a/docs/docs/workflows/verify-cluster.md
+++ b/docs/docs/workflows/verify-cluster.md
@@ -88,6 +88,7 @@ The `verify` command also allows you to verify any Constellation deployment that
 
 * The IP address of a running Constellation cluster's [VerificationService](../architecture/microservices.md#verificationservice). The `VerificationService` is exposed via a `NodePort` service using the external IP address of your cluster. Run `kubectl get nodes -o wide` and look for `EXTERNAL-IP`.
 * The cluster's *clusterID*. See [cluster identity](../architecture/keys.md#cluster-identity) for more details.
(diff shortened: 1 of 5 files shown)


docs/docs/workflows/verify-cluster.md
```diff
@@ -88,6 +88,7 @@ The `verify` command also allows you to verify any Constellation deployment that
 
 * The IP address of a running Constellation cluster's [VerificationService](../architecture/microservices.md#verificationservice). The `VerificationService` is exposed via a `NodePort` service using the external IP address of your cluster. Run `kubectl get nodes -o wide` and look for `EXTERNAL-IP`.
 * The cluster's *clusterID*. See [cluster identity](../architecture/keys.md#cluster-identity) for more details.
+* A `constellation-conf.yaml` file with the expected measurements of the cluster.
```
burgerdev, line 91: I think we should either explicitly add the state file to the command below, or mention that it must be in the current directory (?).

burgerdev (approved):

PR #3223, opened 2024-07-02
ci: only commit measurements if changes are present

msanft:
### Context

We previously tried to commit measurements in the release CI even if no changes were present: https://github.com/edgelesssys/constellation/actions/runs/9746960273/job/26899959187

### Proposed change(s)

- Only try to commit the measurements if changes are present.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -250,8 +250,12 @@ jobs:
         run: |
           git config --global user.name "edgelessci"
           git config --global user.email "edgelessci@users.noreply.github.com"
-          git commit -m "attestation: hardcode measurements for ${VERSION}"
-          git push
+          if git diff-index --quiet HEAD --; then
+            echo "No changes to commit"
+          else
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -250,8 +250,12 @@ jobs:
         run: |
           git config --global user.name "edgelessci"
           git config --global user.email "edgelessci@users.noreply.github.com"
-          git commit -m "attestation: hardcode measurements for ${VERSION}"
-          git push
+          if git status --porcelain | grep measurements_enterprise.go; then
```
daniel-weisse, line 253:
```suggestion
          if git status --porcelain | grep -e measurements_enterprise.go -e image_enterprise.go; then
```
The workflow also updates the default image, so we should check that as well

daniel-weisse (approved):

.github/workflows/release.yml
msanft: We do this in a separate job of the workflow already

PR #3225, opened 2024-07-03
deps: bump Go version to v1.22.5

daniel-weisse:
### Context
A new Go version was released which addresses https://pkg.go.dev/vuln/GO-2024-2963

### Proposed change(s)

- Bump Go version to v1.22.5
- Add instructions to update version in `go.mod` and `go.work` files

### Checklist

  - [ ] [e2e test](https://github.com/edgelesssys/constellation/actions/runs/9773255994)


burgerdev (approved):
I'm surprised that we have this many non-Bazel uses of Go. Since we're using the Bazel toolchain in other scripts, maybe we should use it here, too. I'll add something to the backlog.

PR #3231, opened 2024-07-03
ci: run versionsapi through Bazel instead of building a container

daniel-weisse:
### Context
Our CI copies the versionsapi CLI from a container image we regularly build through our CI.
This requires us to keep a container image update to date, and also update the Go version used in this image.

### Proposed change(s)

- Remove the container image for the versionsapi CLI and use Bazel to build and run the CLI instead

### Related issue
- Fixes https://github.com/edgelesssys/constellation/actions/runs/9773923498/job/26981078229

### Additional info
- This slows down the first invocation of the versionsapi CLI in a job by 2-3 minutes. I think this is acceptable to get rid of the extra container image maintenance

### Checklist

  - [ ] [e2e test (which uses the CLI)](https://github.com/edgelesssys/constellation/actions/runs/9779669154)


burgerdev (approved):
Thanks, that's great!

PR #3236, opened 2024-07-08
helm: manage CoreDNS addon as Helm chart

burgerdev:
### Context

kubeadm switched[^1] its strategy for upgrading cluster addons, which are now upgraded after the last control plane node upgraded. This never happens in Constellation upgrades, because the nodes are replaced and upgrade is only called once.

[^1]: https://github.com/kubernetes/website/blob/5255ca6bfca4ab44d87a6c65701d75c07e092fed/content/en/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade.md?plain=1#L186-L196

### Proposed change(s)

- Disable the `addons/dns` phase while installing a new cluster.
- Add a Helm chart that corresponds to the resources embedded in kubeadm.
- (For upgrading clusters) Migrate the existing coredns resources to Helm management.

We're now applying the DNS upgrade before the cluster upgrade, which is exactly the scenario that motivated the change on k8s' side. However, I'd argue that testing of k8s upgrade should spot such problems, and that it's better to potentially deal with those rather than not to get any upgrades.

### Additional info

- [AB#4330](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/4330)
- [AB#3609](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/3609)

### Checklist

  - [x] [Upgrade @ Azure](https://github.com/edgelesssys/constellation/actions/runs/9838745012)
  - [x] [Sonobuouy Quick @ GCP](https://github.com/edgelesssys/constellation/actions/runs/9841444644)
  - [x] [Sonobuoy Quick @ AWS](https://github.com/edgelesssys/constellation/actions/runs/9852597579)
  - [x] [Upgrade @ AWS](https://github.com/edgelesssys/constellation/actions/runs/9852609496)
  - [x] [Azure, k8s 1.28](https://github.com/edgelesssys/constellation/actions/runs/9858180822)
  - [x] [Terraform test](https://github.com/edgelesssys/constellation/actions/runs/9872665297)
- [x] Link to Milestone


derpsteb:
If we want to go crazy and implement the new behavior at some point we could let the operator apply charts at the end of the upgrade process. But that would be a lot more involved. I don't want to propose this as something we should do here, just a thought I had.

internal/constellation/helm/corednsgen/corednsgen.go
daniel-weisse, line 1: How does this handle generation of CoreDNS charts for different Kubernetes versions?
Wouldn't we need a different chart for each Kubernetes version we support?

burgerdev: It doesn't. The CoreDNS addon has been stable since 1.29, so _right now_ it does not make a difference whether we use it from 1.29 or 1.30. I started a test on 1.28 to confirm it works there.

If k8s authors decide to introduce a backwards-incompatible CoreDNS in a future release, we should catch that in our upgrade tests. Then we can go deploy something else - https://github.com/coredns/helm would seem like a natural candidate. However, I'd prefer to stick closer to the current setup in this PR, to just make it work as a Helm chart at all. Replacing the Helm chart with a different one should then become pretty easy.  

My secret hope is that all this incompatibility sandbagging is for kube-proxy, though.

internal/constellation/helm.go
```diff
@@ -7,13 +7,61 @@ SPDX-License-Identifier: AGPL-3.0-only
...
+	{Group: "rbac.authorization.k8s.io", Version: "v1", Resource: "clusterroles"}:        "system:coredns",
+	{Group: "rbac.authorization.k8s.io", Version: "v1", Resource: "clusterrolebindings"}: "system:coredns",
+}
+
+// AnnotateCoreDNSResources imports existing CoreDNS resources into the Helm release.
+//
+// This is only required when CoreDNS was installed by kubeadm directly.
+// TODO(burgerdev): remove after v2.19 is released.
+func (a *Applier) AnnotateCoreDNSResources() error {
+	ctx := context.Background()
```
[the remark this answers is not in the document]
burgerdev: Context is passed now. Yes, that was my thinking, too.

daniel-weisse (commented):
We will need to do this in the Terrform provider as well.
Probably just before https://github.com/edgelesssys/constellation/blob/391deeae37949658004e563ae39f686cf7a072c4/terraform-provider-constellation/internal/provider/cluster_resource.go#L1064

burgerdev:
Added, but is there an e2e test that I can run?

burgerdev:
Terraform e2e test was successful, too.

daniel-weisse (approved):
thanks

burgerdev:
I'm rebasing this PR instead of squashing it, in order to faciliate revert of [4fa9d85](https://github.com/edgelesssys/constellation/pull/3236/commits/4fa9d85ea63fe5887208a9b0c00e85a182c93cd7).

PR #3237, opened 2024-07-09
ci: fix TF attestation version assertion in integration test

elchead:
### Context

The attestation version numbers were hard-coded in the test but they are dynamically updated.

### Proposed change(s)

- soften version value assertion to be non-zero where applicable

### Additional info

- [AB#4381](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/4381)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/terraform-provider-constellation/internal/provider/attestation_data_source_test.go
+++ b/terraform-provider-constellation/internal/provider/attestation_data_source_test.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: AGPL-3.0-only
 package provider
 
 import (
+	"regexp"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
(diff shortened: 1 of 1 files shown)


terraform-provider-constellation/internal/provider/attestation_data_source_test.go
```diff
@@ -16,6 +17,21 @@ func TestAccAttestationSource(t *testing.T) {
 	// Set the path to the Terraform binary for acceptance testing when running under Bazel.
 	bazelPreCheck := func() { bazelSetTerraformBinaryPath(t) }
 
+	assertNonZeroValue := func(attr string) resource.TestCheckFunc {
+		return resource.TestMatchResourceAttr(
+			"data.constellation_attestation.test",
+			attr,
+			regexp.MustCompile(`^[1-9]\d*$`),
```
burgerdev, line 24: Could we rather use https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestCheckResourceAttrWith? I fear that regexp matching of numbers is somewhat error prone.

elchead: I guess everyone hates regex :D it matches only digits. How is this error prone?
But I can also rewrite the logic in Go.

burgerdev: This one does not match `012`, which may or may not be a valid uint8, depending on who you ask. The other one matches `foo1bar`, which is too permissive I believe.

I'm not against regexps per se, but https://regex.info/blog/2006-09-15/247.

burgerdev (approved):
Great, thanks!

PR #3238, opened 2024-07-09
attestationconfigapi: revise upload frequency

daniel-weisse:
### Context
SVNs are getting promoted to stable more quickly than we intended.
The main issue comes from uploading to many reports from the weekly tests.

### Proposed change(s)

- Add a `compare` command to the attestationconfigapi CLI
  - Compares two or more SNP/TDX reports and returns the one with the lowest SVNs
- Use `compare` command to only upload the report with the lowest SVNs during the `e2e verify` test
- Adjust the window for promoting cached versions to stable from 15 to 6 to keep the intended behavior of updating a version after it has been stable for 3 weeks.

### Additional info

- [AB#4382](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/4382)
- Before merging, I will go over the existing cached versions and see to removing all but the 2 lowest versions from the past 6 weeks

### DIFF
--- a/.github/actions/e2e_verify/action.yml
+++ b/.github/actions/e2e_verify/action.yml
@@ -94,13 +94,11 @@ runs:
         COSIGN_PASSWORD: ${{ inputs.cosignPassword }}
         COSIGN_PRIVATE_KEY: ${{ inputs.cosignPrivateKey }}
       run: |
-        reports=(attestation-report-*.json)
-        if [ -z ${#reports[@]} ]; then
-            exit 1
-        fi
+        reports=attestation-report-*.json
 
-        for file in "${reports[@]}"; do
-          path=$(realpath "${file}")
-          cat "${path}"
-          bazel run //internal/api/attestationconfigapi/cli -- upload ${{ inputs.attestationVariant }} attestation-report "${path}"
-        done
+        report=$(bazel run //internal/api/attestationconfigapi/cli -- compare ${{ inputs.attestationVariant }} ${report})
+
+        path=$(realpath "${report}")
+        cat "${path}"
+
+        bazel run //internal/api/attestationconfigapi/cli -- upload ${{ inputs.attestationVariant }} attestation-report "${path}"
--- a/.github/actions/e2e_verify/action.yml
+++ b/.github/actions/e2e_verify/action.yml
@@ -96,7 +96,7 @@ runs:
       run: |
         reports=attestation-report-*.json
 
-        report=$(bazel run //internal/api/attestationconfigapi/cli -- compare ${{ inputs.attestationVariant }} ${report})
+        report=$(bazel run //internal/api/attestationconfigapi/cli -- compare ${{ inputs.attestationVariant }} ${reports})
 
         path=$(realpath "${report}")
         cat "${path}"
--- a/internal/api/attestationconfigapi/cli/compare.go
+++ b/internal/api/attestationconfigapi/cli/compare.go
@@ -21,9 +21,9 @@ import (
 
 func newCompareCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:     "compare {aws-sev-snp|azure-sev-snp|azure-tdx|gcp-sev-snp} [FILE] [FILE] ...",
-		Short:   "Compare two or more attestation reports and return the lowest version",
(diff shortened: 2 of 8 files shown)


.github/actions/e2e_verify/action.yml
```diff
@@ -94,13 +94,11 @@ runs:
...
-            exit 1
-        fi
+        reports=attestation-report-*.json
 
-        for file in "${reports[@]}"; do
-          path=$(realpath "${file}")
-          cat "${path}"
-          bazel run //internal/api/attestationconfigapi/cli -- upload ${{ inputs.attestationVariant }} attestation-report "${path}"
-        done
+        report=$(bazel run //internal/api/attestationconfigapi/cli -- compare ${{ inputs.attestationVariant }} ${report})
```
derpsteb, line 99:
```suggestion
        report=$(bazel run //internal/api/attestationconfigapi/cli -- compare ${{ inputs.attestationVariant }} ${reports})
```

is this a typo or intended? if it's a typo then the `()` in L97 need to stay, otherwise the glob is not extended.

derpsteb:
Changes make sense to me. I'll let burgerdev approve as he has a closer eye on the e2e tests.

.github/actions/e2e_verify/action.yml
daniel-weisse: Why would the glob not be expanded?
Locally, this works fine

internal/api/attestationconfigapi/cli/compare.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/spf13/cobra"
+)
+
+func newCompareCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "compare {aws-sev-snp|azure-sev-snp|azure-tdx|gcp-sev-snp} [FILE] [FILE] ...",
+		Short:   "Compare two or more attestation reports and return the lowest version",
+		Long:    "Compare two or more attestation reports and return the lowest version.",
+		Example: "cli compare azure-sev-snp report1.json report2.json",
+		Args:    cobra.MatchAll(cobra.MinimumNArgs(3), arg0isAttestationVariant()),
```
burgerdev, line 28: This could be generalized to `compare VARIANT FILE...` ("returns the minimum version of all given reports").

internal/api/attestationconfigapi/cli/compare.go
```diff
@@ -0,0 +1,101 @@
...
+	for _, file := range files[1:] {
+		report, err := readReport(file, fs)
+		if err != nil {
+			return "", fmt.Errorf("reading tdx report: %w", err)
+		}
+
+		if client.IsInputNewerThanOtherVersion(attestationVariant, lowestReport, report) {
+			lowestVersion = file
+			lowestReport = report
+		}
```
burgerdev, line 81: Not sure if this is relevant in practice, but the ordering relation on reports is unsound: if a report has one value older and one newer, what do we do with it?

daniel-weisse: This command is supposed to perform some pre-ordering before we actually upload anything.
While I think this case is unlikely, I think its still fine to upload such a version as it will have to occur 6 times in a row to be considered as a new stable version

internal/api/attestationconfigapi/cli/compare.go
burgerdev: Sorry, what I meant to say is that `compare` should accept `1..n` files, not `2..n`.

internal/api/attestationconfigapi/cli/compare.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/google/go-tdx-guest/proto/tdx"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+)
+
+func newCompareCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "compare VARIANT FILE [FILE...]",
+		Short:   "Returns the minimum version of all given attestation reports.",
+		Long:    "Compare two or more attestation reports and return the report with the minimum version of all given attestation reports.",
```
burgerdev, line 26:
```suggestion
		Long:    "Compare a list of attestation reports and return the report with the minimum version.",
```

burgerdev (approved):

PR #3239, opened 2024-07-09
dev-docs: set an MTU on the VPN route

burgerdev:
### Context

In a test setup with Constellation on GCP and a VM on Azure, I observed failures when trying to connect via TLS from the VM to the API server. The issue turned out to be packets dropped in XFRM, because they were too large for the 1500 MTU of the tunnel connection. 

### Proposed change(s)

Enabling path MTU over the GCP load balancer is likely impossible, so I'm adding a config knob to set the MTU manually. The conservative default setting should be enough to work with MTU 1500.

### Additional info

- [AB#4384](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/4384)

### Checklist

  - [x] tested manually on standard 1.29 cluster.
- [x] Link to Milestone


3u13r (approved):
LGTM

PR #3253, opened 2024-07-12
renovate: exclude stackit from Terraform group

burgerdev:
### Context

The stackit provider introduced a bug in `v0.24.0` and we'd like to avoid renovate bundling this version with the other Terraform providers. Excluding it from the group should result in a dedicated upgrade PR, which we can then ignore.

### Checklist
- [x] Link to Milestone

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -71,6 +71,7 @@
         "pinDigest",
         "rollback",
       ],
+      // stackit is excluded from the group so that we can ignore v0.24.1: https://github.com/stackitcloud/terraform-provider-stackit/issues/464.
       "excludeDepNames": ["stackit"],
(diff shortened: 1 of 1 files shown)


renovate.json5
```diff
@@ -71,6 +71,7 @@
         "pinDigest",
         "rollback",
       ],
+      "excludeDepNames": ["stackit"],
```
katexochen, line 74: Maybe we can add a comment why we include it, or even a todo to revert this in the future?

katexochen (approved):

renovate.json5
burgerdev: Thanks, done.

PR #3257, opened 2024-07-16
Allow upgrades on azure without Terraform changes on LBs created from within Kubernetes

3u13r:
### Proposed change(s)

- Remove worker nodes from any existing load balancer backend, since any VM NIC can only be attached to one backend group. Also the worker backends were unused anyway.
- Let the cloud-controller-manager create a separate "kubernetes-lb" to create LBs for Kubernetes services.
- Don't inline (most) of the network security rules
  - Sadly, we cannot fix the upgrade process in one go, as we need 2 state changes on the same resource. So the upgrade to this release still removes the network security rules added by the cloud controller manager, but which should heal itself in the upgrade process one it's restarted. After we execute the instructions mentioned in the TODO, there will finally be no terraform changes after "constellation create" -> "create LB svc" -> "terraform plan". 

Azure, snp, 1:2, lb: https://github.com/edgelesssys/constellation/actions/runs/11220341955
Azure, snp, 1:2, upgrade: https://github.com/edgelesssys/constellation/actions/runs/11220359763

### Checklist

- [x] Link to Milestone

### DIFF
--- a/docs/docs/reference/migration.md
+++ b/docs/docs/reference/migration.md
@@ -9,9 +9,9 @@ Use [`constellation config migrate`](./cli.md#constellation-config-migrate) to a
 ### Azure
 
 * To allow seamless upgrades on Azure when Kubernetes services of type `LoadBalancer` are deployed, the target  
- load balancer in which the `cloud-controller-maanger` created the load balancing rules was changed. Instead of using the load balancer, 
- created and maintained by the CLI's Terraform code, the `cloud-controller-mananger` now creates its own load balancer in Azure.  
- If inside your Constellation there are services of type `LoadBalancer`, please remove them before the upgrade and re-apply them  
+ load balancer in which the `cloud-controller-manager` creates load balancing rules was changed. Instead of using the load balancer
+ created and maintained by the CLI's Terraform code, the `cloud-controller-manager` now creates its own load balancer in Azure.
+ If your Constellation has services of type `LoadBalancer`, please remove them before the upgrade and re-apply them
  afterward. 
 
 
--- a/docs/docs/reference/migration.md
+++ b/docs/docs/reference/migration.md
@@ -3,7 +3,19 @@
 This document describes breaking changes and migrations between Constellation releases.
 Use [`constellation config migrate`](./cli.md#constellation-config-migrate) to automatically update an old config file to a new format.
 
-## Migrating from Azure's service principal authentication to managed identity authentication
+
+## Migrations to v2.19.0
+
+### Azure
+
+* To allow seamless upgrades on Azure when Kubernetes services of type `LoadBalancer` are deployed, we changed the target
+ load balancer in which the `cloud-controller-maanger` create the load balancing rules. Instead of using the load balancer we
+ create and maintain in the CLI's Terraform code, the `cloud-controller-mananger` now creates its own load balancer in Azure.
+ If inside your Constellation there are services of type `LoadBalancer`, please remove them before the upgrade and re-apply them
+ afterwards. 
+
+
+## Migrating from Azure's service principal authentication to managed identity authentication (during the upgrade to Constellation v2.8.0)
 
 - The `provider.azure.appClientID` and `provider.azure.appClientSecret` fields are no longer supported and should be removed.
 - To keep using an existing UAMI, add the `Owner` permission with the scope of your `resourceGroup`.
--- a/docs/docs/reference/migration.md
+++ b/docs/docs/reference/migration.md
@@ -8,11 +8,11 @@ Use [`constellation config migrate`](./cli.md#constellation-config-migrate) to a
 
 ### Azure
 
-* To allow seamless upgrades on Azure when Kubernetes services of type `LoadBalancer` are deployed, we changed the target
- load balancer in which the `cloud-controller-maanger` create the load balancing rules. Instead of using the load balancer we
- create and maintain in the CLI's Terraform code, the `cloud-controller-mananger` now creates its own load balancer in Azure.
- If inside your Constellation there are services of type `LoadBalancer`, please remove them before the upgrade and re-apply them
- afterwards. 
+* To allow seamless upgrades on Azure when Kubernetes services of type `LoadBalancer` are deployed, the target  
+ load balancer in which the `cloud-controller-maanger` created the load balancing rules was changed. Instead of using the load balancer, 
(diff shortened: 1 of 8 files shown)


terraform/infrastructure/azure/main.tf
```diff
@@ -232,6 +241,28 @@ resource "azurerm_network_security_group" "security_group" {
   }
 }
 
+resource "azurerm_network_security_rule" "nsg_rule" {
+  for_each = {
+    for o in concat(
+      local.ports,
+      [{ name = "nodeports", port = local.ports_node_range, priority = 200 }]
```
burgerdev, line 248: Why do we need to expose node ports via LB? Do we need to expose the `ports_node_range` for UDP, too?

terraform/infrastructure/azure/main.tf
```diff
@@ -214,10 +220,13 @@ resource "azurerm_network_security_group" "security_group" {
...
-      local.ports,
-      [{ name = "nodeports", port = local.ports_node_range, priority = 200 }]
-    )
+    # we keep this rule for one last release since the azurerm provider does not 
+    # support moving security rules that are inlined (like this) to the external resource one.
+    # Even worse, just defining the azurerm_network_security_group without the 
+    # "security_rule" block will NOT remove all the rules but do nothing.
+    # TODO(@3u13r): remove the "security_rule" block in the next release after this code has landed.
+    # So either after 2.18 or after 2.17.X if cerry-picked release.
+    for_each = [{ name = "konnectivity", priority = 1000, port = 8132 }]
```
burgerdev, line 229: Do I understand correctly that we don't need this rule anymore, but we need to migrate it before we can remove it?

3u13r: yes, we need to keep the block to signal to Azure that we want to move the security rules out of the dynamic block. When we remove this block, existing Constellations will keep the konnectivity security rule as long as they don't delete it manually. But since the block doesn't exist anymore, new Constellation will only have the rules we want.

terraform/infrastructure/azure/main.tf
3u13r: I don't know if I fully get your question. The intent of the code is to allow LBs to route traffic to node on the nodeports. The default case when using an LB is that the node opens a nodeport and the LB routes to the nodeip:nodeport. Good catch with the UDP though.

burgerdev: I read this block as adding a firewall rule for the entire NodePort range by default, but I expected individual exceptions for load balancers that are actually deployed.

3u13r: Good point. Azure has those LB to VM allow rule that exists by default and the CCM creates the needed "port to LB public IP" rule. Removed the nodeports from here.

docs/docs/reference/migration.md
```diff
@@ -3,7 +3,19 @@
...
+
+## Migrations to v2.19.0
+
+### Azure
+
+* To allow seamless upgrades on Azure when Kubernetes services of type `LoadBalancer` are deployed, we changed the target
+ load balancer in which the `cloud-controller-maanger` create the load balancing rules. Instead of using the load balancer we
+ create and maintain in the CLI's Terraform code, the `cloud-controller-mananger` now creates its own load balancer in Azure.
+ If inside your Constellation there are services of type `LoadBalancer`, please remove them before the upgrade and re-apply them
+ afterwards.
```
daniel-weisse, line 15:
```suggestion
* To allow seamless upgrades on Azure when Kubernetes services of type `LoadBalancer` are deployed, the target
 load balancer in which the `cloud-controller-maanger` created the load balancing rules was changed. Instead of using the load balancer, created and maintained by the CLI's Terraform code, the `cloud-controller-mananger` now creates its own load balancer in Azure.
 If inside your Constellation there are services of type `LoadBalancer`, please remove them before the upgrade and re-apply them
 afterward. 
```

docs/docs/reference/migration.md
```diff
@@ -3,7 +3,19 @@
...
 Use [`constellation config migrate`](./cli.md#constellation-config-migrate) to automatically update an old config file to a new format.
 
-## Migrating from Azure's service principal authentication to managed identity authentication
+
+## Migrations to v2.19.0
+
+### Azure
+
+* To allow seamless upgrades on Azure when Kubernetes services of type `LoadBalancer` are deployed, the target  
+ load balancer in which the `cloud-controller-maanger` created the load balancing rules was changed. Instead of using the load balancer,
```
burgerdev, line 12:
```suggestion
 load balancer in which the `cloud-controller-maanger` creates load balancing rules was changed. Instead of using the load balancer
```

docs/docs/reference/migration.md
```diff
@@ -3,7 +3,19 @@
...
-## Migrating from Azure's service principal authentication to managed identity authentication
+
+## Migrations to v2.19.0
+
+### Azure
+
+* To allow seamless upgrades on Azure when Kubernetes services of type `LoadBalancer` are deployed, the target  
+ load balancer in which the `cloud-controller-maanger` created the load balancing rules was changed. Instead of using the load balancer, 
+ created and maintained by the CLI's Terraform code, the `cloud-controller-mananger` now creates its own load balancer in Azure.  
+ If inside your Constellation there are services of type `LoadBalancer`, please remove them before the upgrade and re-apply them
```
burgerdev, line 14:
```suggestion
 If your Constellation has services of type `LoadBalancer`, please remove them before the upgrade and re-apply them  
```

burgerdev (approved):

docs/docs/reference/migration.md
```diff
@@ -3,7 +3,19 @@
...
 
-## Migrating from Azure's service principal authentication to managed identity authentication
+
+## Migrations to v2.19.0
+
+### Azure
+
+* To allow seamless upgrades on Azure when Kubernetes services of type `LoadBalancer` are deployed, the target  
+ load balancer in which the `cloud-controller-maanger` created the load balancing rules was changed. Instead of using the load balancer, 
+ created and maintained by the CLI's Terraform code, the `cloud-controller-mananger` now creates its own load balancer in Azure.
```
thomasten, line 13:
```suggestion
 created and maintained by the CLI's Terraform code, the `cloud-controller-manager` now creates its own load balancer in Azure.  
```

thomasten (approved):

PR #3308, opened 2024-08-09
mkosi: add su dependency

msanft:
### Context

Our OS images were not building successfully on off-the-shelf Ubuntu installations due to missing `newuidmap` and `newgidmap` packages. This image didn't occur in CI since the runners come with the required packages preinstalled.

### Proposed change(s)

- Add the `su` dependency to our Nix overlay for mkosi, sourcing the relevant newuidmap, newgidmap binaries.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -53,6 +53,9 @@
           zstd # zstd
           xz # xz
 
+          # newuidmap, newgidmap
+          su
+
           # utils
           gnused # sed
           gnugrep # grep


msanft:
After some consideration with @katexochen, it became clear that we actually need to source this binary from the host, as we cannot provide suid binaries through Nix directly on non-NixOS systems.

PR #3311, opened 2024-08-13
docs: remove date from "feature status of clouds" page

thomasten:
We often forget to update it and it's not that valuable.


3u13r (approved):
LGTM

PR #3315, opened 2024-08-19
docs: sort clouds table alphabetical

m1ghtym0:
### Proposed change(s)

- Sort clouds in comparison table alphabetical
- Keep OpenStack as "generic open-source" to the right


thomasten (approved):
Thanks.
Should we sort the sections below the table the same way?
Please backport to 2.17 before merge.

PR #3330, opened 2024-08-29
terraform-provider: document MAA patching more prominently

msanft:
### Context

The [Terraform provider documentation](https://registry.terraform.io/providers/edgelesssys/constellation) did not mention that the MAA needs to be manually patched previously - only the [Constellation usage documentation](https://docs.edgeless.systems/constellation/workflows/terraform-provider#quick-setup) on the Terraform provider did. However, customers will most likely only look at the former.

### Proposed change(s)

- Link the Constellation documentation on how to patch the MAA in the Terraform provider documentation.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/terraform-provider-constellation/docs/data-sources/attestation.md
+++ b/terraform-provider-constellation/docs/data-sources/attestation.md
@@ -42,7 +42,8 @@ See the [full list of CSPs](https://docs.edgeless.systems/constellation/overview
 ### Optional
 
 - `insecure` (Boolean) DON'T USE IN PRODUCTION Skip the signature verification when fetching measurements for the image.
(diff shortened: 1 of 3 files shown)


daniel-weisse (approved):

thomasten:
Is this placed prominently enough? Will users look at the maa_url property?
"Needs to be patched manually" sounds like the URL needs to be patched, but it's the policy that needs to be patched, right?

msanft:
> Is this placed prominently enough? Will users look at the maa_url property? "Needs to be patched manually" sounds like the URL needs to be patched, but it's the policy that needs to be patched, right?

Added an additional warning if an MAA URL is supplied, so there's no chance to miss it now. Also refined the wording.

thomasten (approved):

PR #3337, opened 2024-09-05
api: make attestationconfigapi public

elchead:
### Context

The continuum project relies also relies on the attestationconfigapi. So far, the code was duplicated. As agreed upon in chat, the package should be made public to be reused in other projects.

### Proposed change(s)

- move the attestationconfigapi outside of internal
- make the attestationconfigapi cli subpackage internal

### Checklist

- [ ] Link to Milestone


elchead:
Testing e2e verify:
https://github.com/edgelesssys/constellation/actions/runs/10788448192

api/attestationconfigapi/fetcher.go
```diff
@@ -18,9 +18,23 @@ import (
...
+
+// Variant is a cloud provider specific attestation variant.
+type Variant interface {
+	String() string
+}
+
 // Fetcher fetches config API resources without authentication.
 type Fetcher interface {
-	FetchLatestVersion(ctx context.Context, attestation variant.Variant) (Entry, error)
+	FetchLatestVersion(ctx context.Context, attestation Variant) (Entry, error)
```
elchead, line 37: continuum for example implements its own variant. I wanted to avoid also exposing this pkg.

api/attestationconfigapi/fetcher.go
```diff
@@ -18,9 +18,23 @@ import (
...
 const cosignPublicKey = constants.CosignPublicKeyReleases
 
+var (
+	// AzureSEVSNP is the Azure SEV-SNP variant.
+	AzureSEVSNP Variant = variant.AzureSEVSNP{}
+	// AWSSEVSNP is the AWS SEV-SNP variant.
+	AWSSEVSNP Variant = variant.AWSSEVSNP{}
+	// GCPSEVSNP is the GCP SEV-SNP variant.
+	GCPSEVSNP Variant = variant.GCPSEVSNP{}
+)
```
daniel-weisse, line 28: Why is this needed?

api/attestationconfigapi/fetcher.go
```diff
@@ -60,7 +74,7 @@ func newFetcherWithClientAndVerifier(client apifetcher.HTTPClient, cosignVerifie
 }
 
 // FetchLatestVersion returns the latest versions of the given type.
-func (f *fetcher) FetchLatestVersion(ctx context.Context, variant variant.Variant) (Entry, error) {
+func (f *fetcher) FetchLatestVersion(ctx context.Context, variant Variant) (Entry, error) {
```
daniel-weisse, line 77:
```suggestion
func (f *fetcher) FetchLatestVersion(ctx context.Context, variant string) (Entry, error) {
```
This might require some more changes to existing code, but why not expect `string` as input?

api/attestationconfigapi/fetcher.go
```diff
@@ -71,31 +85,39 @@ func (f *fetcher) FetchLatestVersion(ctx context.Context, variant variant.Varian
...
 
 // fetchVersionList fetches the version list information from the config API.
-func (f *fetcher) fetchVersionList(ctx context.Context, variant variant.Variant) (List, error) {
-	fetchedList, err := apifetcher.FetchAndVerify(ctx, f.HTTPClient, f.cdnURL, List{Variant: variant}, f.verifier)
+func (f *fetcher) fetchVersionList(ctx context.Context, attestationVariant Variant) (List, error) {
+	parsedVariant, err := variant.FromString(attestationVariant.String())
+	if err != nil {
+		return List{}, fmt.Errorf("parsing variant: %w", err)
+	}
+	fetchedList, err := apifetcher.FetchAndVerify(ctx, f.HTTPClient, f.cdnURL, List{Variant: parsedVariant}, f.verifier)
```
daniel-weisse, line 93:
```suggestion
func (f *fetcher) fetchVersionList(ctx context.Context, variant variant.Variant) (List, error) {
	fetchedList, err := apifetcher.FetchAndVerify(ctx, f.HTTPClient, f.cdnURL, List{Variant: variant}, f.verifier)
```
nit: functions that aren't exported can use our internal type as input. This should also save on some code duplication.

Also applies to `fetchVersion`

daniel-weisse (commented):
Can we rename `attestationconfigapi` to just `attestationconfig`?
The package is already in a directory called `api` and the name is needlessly long.

api/attestationconfigapi/fetcher.go
```diff
@@ -18,9 +18,23 @@ import (
...
+	// AWSSEVSNP is the AWS SEV-SNP variant.
+	AWSSEVSNP Variant = variant.AWSSEVSNP{}
+	// GCPSEVSNP is the GCP SEV-SNP variant.
+	GCPSEVSNP Variant = variant.GCPSEVSNP{}
+)
+
+// Variant is a cloud provider specific attestation variant.
+type Variant interface {
+	String() string
+}
```
msanft, line 33: We should drop this and make it an [`fmt.Stringer`](https://pkg.go.dev/fmt#Stringer)

api/attestationconfigapi/fetcher.go
```diff
@@ -71,31 +85,39 @@ func (f *fetcher) FetchLatestVersion(ctx context.Context, variant variant.Varian
 }
 
 // fetchVersionList fetches the version list information from the config API.
-func (f *fetcher) fetchVersionList(ctx context.Context, variant variant.Variant) (List, error) {
-	fetchedList, err := apifetcher.FetchAndVerify(ctx, f.HTTPClient, f.cdnURL, List{Variant: variant}, f.verifier)
+func (f *fetcher) fetchVersionList(ctx context.Context, attestationVariant Variant) (List, error) {
+	parsedVariant, err := variant.FromString(attestationVariant.String())
+	if err != nil {
+		return List{}, fmt.Errorf("parsing variant: %w", err)
+	}
```
msanft, line 92: So when Continuum has its own variants (and they don't exactly match what we have here in Constellation), this will break. I think this package that can only work with a certain set of variants *should* export the package specifying its variants.

msanft (changes requested):

api/attestationconfigapi/fetcher.go
daniel-weisse: Took a look at the code
From what I can tell `variant` is only used to parse strings in this context.
We should just refactor this to work on strings instead of the custom variant type.
We will still receive appropriate error messages if objects do not exist for those strings.

We can provide constants for Constellation's supported variants, e.g. `azure-sev-snp` and `aws-sev-snp`.
I think that was intended by providing [these variables](https://github.com/edgelesssys/constellation/blob/c28e8fbd43c121017f89e28f2d4bf7b8ba796ee9/api/attestationconfigapi/fetcher.go#L21-L28)

api/attestationconfigapi/fetcher.go
elchead: it is not needed, but since it's public a third party should be provided with valid defaults since `attestation/variant` is internal?

api/attestationconfigapi/fetcher.go
elchead: I did this in an earlier commit, but it didn't make the the intention so clear. It doesn't suffice to just pass any stringer object.

api/attestationconfigapi/fetcher.go
elchead: because it requires changing a lot of code

api/attestationconfigapi/fetcher.go
elchead: correct. I only kept this interface to not have to change a lot of places where this is used.

daniel-weisse: I would rather have changes in a lot of places in this PR, than expose an API that doesn't make sense.
To me, it looks like using the variant interface/type in this package is not needed, so we should remove it.

api/attestationconfigapi/fetcher.go
msanft: Either we make this a string and accept that it will fail late when running into a 404 on the API call (but have a simple interface without many exports), or we export the proper types that can be used here.

msanft:
I'm kind of thinking that this should actually be in its very own repo. If we keep "lending" packages around our projects, it will get messy (circular) at some point.

api/attestationconfigapi/fetcher.go
elchead: I think that the variant interface provides a safer design with the exposed variants, but I can update it to a string.

elchead:
> I'm kind of thinking that this should actually be in its very own repo. If we keep "lending" packages around our projects, it will get messy (circular) at some point.

In the long term yes, but I'm not sure if we should start this for one tiny library now.

api/attestationconfigapi/fetcher.go
daniel-weisse: My issue with the variant interface (which later gets converted to a `variant.Variant` type is, that this only _really_ makes sense in the context of Constellation.
If we expose this package as "public API", then it doesn't make sense to keep this restriction, but hide it behind some named interfaces.
This is especially confusing when other projects also provide their own `Variant` types, that might (or might not) fit into here.

If we just use a string, one can provide whatever they want. If the object does not exist, an error message is returned.
As already mentioned, we can provide the supported values as constants.

katexochen:
What are the stability guarantees the new public API will provide?

elchead:
> What are the stability guarantees the new public API will provide?

The API is backed by an S3 bucket that is versioned (v1). Older SDK versions should always remain compatible with the v1 S3 bucket. Any breaking changes, for newer SDK versions, would be stored in a v2 subpath of the bucket.
wdyt?

elchead:
We decided to not do this and instead copy the code if needed.

PR #3342, opened 2024-09-09
image: update mkosi to 24.3

msanft:
### Context

This updates mkosi to a next-version of v24.3, which is now available in nixpkgs. This removes the non-hermetic `uidmap` dependency, which is a great advantage. It will also be less of an effort to upgrade to v25 going forward.

### Proposed change(s)

- Update mkosi to v24.3, making changes to its configuration files as necessary.
- Changes required are keeping `/var/cache` around (which is reproducible for our images, so no problem), as mkosi needs files from it in the build process. mkosi now additionally requires an explicit option to fetch the signing keys for the package repositories from the internet. A hack was required to satisfy the Bazel package, which should probably be solved properly at some point.

### Additional info

- [AB#4553](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/4553)

### Checklist

  - [Image Build](https://github.com/edgelesssys/constellation/actions/runs/10559044434)
  - [E2E Test with the image](https://github.com/edgelesssys/constellation/actions/runs/10616352636/job/29426610705)  - see boot logs, failure is unrelated.
- [x] Link to Milestone

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -36,11 +36,11 @@
     },
     "nixpkgsUnstable": {
       "locked": {
-        "lastModified": 1725194671,
-        "narHash": "sha256-tLGCFEFTB5TaOKkpfw3iYT9dnk4awTP/q4w+ROpMfuw=",
+        "lastModified": 1725816686,
+        "narHash": "sha256-0Kq2MkQ/sQX1rhWJ/ySBBQlBJBUK8mPMDcuDhhdBkSU=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "b833ff01a0d694b910daca6e2ff4a3f26dee478c",
+        "rev": "add0443ee587a0c44f22793b8c8649a0dbc3bb00",
         "type": "github"
       },
       "original": {
@@ -83,11 +83,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1718003967,
-        "narHash": "sha256-PJbw9L8NgRKYkKqc1Pywv1glSyT+iE1D79GuF7ec4tc=",
+        "lastModified": 1725522308,
+        "narHash": "sha256-AtsFZ92WkVkH8fd0Xa0D6/PR84/dtOH6gpM4mtn32Hk=",
         "owner": "edgelesssys",
         "repo": "uplosi",
-        "rev": "b7edeef9b9eb1226998c86ffc8fbf9eacce94549",
+        "rev": "c8a482de299b5442f558d7381f0cbf898e0de0e1",
         "type": "github"
       },
       "original": {
--- a/flake.lock
+++ b/flake.lock
@@ -36,11 +36,11 @@
     },
     "nixpkgsUnstable": {
       "locked": {
-        "lastModified": 1717399147,
-        "narHash": "sha256-eCWaE/q1VItpFAxxLVt171MdtDcjEnwi6QB/yuF73JU=",
+        "lastModified": 1725194671,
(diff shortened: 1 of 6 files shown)


image/initrd/mkosi.conf
```diff
@@ -32,7 +33,6 @@ Packages=udev
          cryptsetup
 
 RemoveFiles=/var/log
-RemoveFiles=/var/cache
```
burgerdev, line 35: > Changes required are keeping `/var/cache` around (which is reproducible for our images, so no problem), as mkosi needs files from it in the build process. 

What exactly is `/var/cache` being used for, and is it also needed for the initrd?

I'm also slightly worried about `/var/cache` becoming irreproducible in the future.

image/base/mkosi.postinst
```diff
@@ -15,3 +15,19 @@ cp "${BUILDROOT}/usr/share/constellation/packagemanifest" "${OUTPUTDIR}/"
 
 # copy rpmdb to outputs
 cp "${BUILDROOT}"/var/lib/rpm/{rpmdb.sqlite-wal,rpmdb.sqlite-shm,rpmdb.sqlite,.rpm.lock} "${OUTPUTDIR}/"
+
+# FIXME(msanft):
+# Hack to satisfy Bazel's [output expectations](./BUILD.bazel).
```
burgerdev, line 20: Did this work before the update?

flake.lock
```diff
@@ -83,11 +83,11 @@
...
       },
       "locked": {
-        "lastModified": 1718003967,
-        "narHash": "sha256-PJbw9L8NgRKYkKqc1Pywv1glSyT+iE1D79GuF7ec4tc=",
+        "lastModified": 1725522308,
+        "narHash": "sha256-AtsFZ92WkVkH8fd0Xa0D6/PR84/dtOH6gpM4mtn32Hk=",
         "owner": "edgelesssys",
         "repo": "uplosi",
-        "rev": "b7edeef9b9eb1226998c86ffc8fbf9eacce94549",
+        "rev": "c8a482de299b5442f558d7381f0cbf898e0de0e1",
```
burgerdev, line 90: The commit description says we're upgrading mkosi, but the diff looks like we're upgrading only uplosi. Is the upgrade transitive, or am I missing something?

msanft: Yeah - This is a little confusing. The bump to `nixpkgsUnstable` above actuall brings in the [commit that updated mkosi](https://github.com/NixOS/nixpkgs/pull/338962) in nixpkgs.

image/base/mkosi.postinst
msanft: Yes. I assume it's due to a change in how `mkosi` calls `dnf` now, somehow not passing the variant to it. I spend quite some time looking at the [relevant code](https://github.com/systemd/mkosi/blob/v22/mkosi/installer/dnf.py) and couldn't pin down the exact reason though.

image/initrd/mkosi.conf
msanft: Good point. During testing of the build, I've run into errors which seemed to be related to this, but verifying it now, it seems that removing them is not a problem. Updated the commit to leave the `RemoveFiles=` instructions as they were.

.bazelversion
```diff
@@ -1 +1 @@
-7.2.1
+7.1.2
```
burgerdev, line 1: This is accidental, right?

burgerdev (approved):

PR #3355, opened 2024-09-13
image: re-enable autologin for debug and console images

3u13r:
In mkosi v24 --autologin no longer works for ttyS consoles. Since the CSPs use those exclusively for their serial consoles, we need to replace this with another solution. Since it seems the simplest, I just created the systemd unit myself and made it conditional on the kernel cmdline.

Console:
~https://github.com/edgelesssys/constellation/actions/runs/10856407864~
~https://github.com/edgelesssys/constellation/actions/runs/10874367218~
~https://github.com/edgelesssys/constellation/actions/runs/10874697856~
https://github.com/edgelesssys/constellation/actions/runs/10892773346

Debug:
~https://github.com/edgelesssys/constellation/actions/runs/10856416476~
~https://github.com/edgelesssys/constellation/actions/runs/10874954823~
https://github.com/edgelesssys/constellation/actions/runs/10892782667

Sanity Check / nightly:
~https://github.com/edgelesssys/constellation/actions/runs/10856431020~
~https://github.com/edgelesssys/constellation/actions/runs/10874956962~
https://github.com/edgelesssys/constellation/actions/runs/10892766154

~I have not yet tested the images, but if the console still isn't there I might have missed `enable serial-getty@tty0.service` in `20-constellation-base.preset`~ Tested all 3 images now. Feel free to re-test.

Just to be extra sure, I think we should add a manual serial console enabled check for the next release in the release steps. What do you think?

### Checklist

- [x] Link to Milestone

### DIFF
--- a/image/sysroot-tree/usr/lib/systemd/system/serial-getty@ttyS0.service.d/autologin.conf
+++ b/image/sysroot-tree/usr/lib/systemd/system/serial-getty@ttyS0.service.d/autologin.conf
@@ -1,0 +1,12 @@
+[Unit]
+Description=autologin
+ConditionPathExists=/proc/cmdline
+ConditionKernelCommandLine=|constellation.console
+ConditionKernelCommandLine=|constellation.debug
+
+[Service]
+ExecStart=
+ExecStart=-/sbin/agetty -o '-p -f -- \\u' --keep-baud --autologin root 115200,57600,38400,9600 - $TERM
+
+[Install]
+WantedBy=multi-user.target
--- a/image/system/variants.bzl
+++ b/image/system/variants.bzl
@@ -86,7 +86,6 @@ csp_settings = {
         },
     },
     "qemu": {
-        "autologin": True,
         "kernel_command_line_dict": {
             "console": "ttyS0",
             "constel.csp": "qemu",
@@ -136,10 +135,8 @@ attestation_variant_settings = {
 
 stream_settings = {
     "console": {
-        "autologin": True,
     },
     "debug": {
-        "autologin": True,
         "kernel_command_line": "constellation.debug",
     },
     "nightly": {},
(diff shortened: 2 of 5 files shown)


msanft:
FYI: corrected the link to the nightly build in the PR description, it pointed to the debug one. Testing the images now. Change LGTM I think

msanft:
Unfortunately, it doesn't seem to work yet. Instead of being shown a login prompt, I just get no prompt at all, but also cannot give any input to the VM.

image/system/variants.bzl
```diff
@@ -86,7 +86,6 @@ csp_settings = {
         },
     },
     "qemu": {
-        "autologin": True,
```
burgerdev, line 89: This doesn't need a `constellation.console` Kernel command line arg?

image/sysroot-tree/usr/lib/systemd/system/serial-getty@ttyS0.service.d/autologin.conf
```diff
@@ -0,0 +1,12 @@
...
+ConditionPathExists=/proc/cmdline
+ConditionKernelCommandLine=|constellation.console
+ConditionKernelCommandLine=|constellation.debug
+
+[Service]
+ExecStart=
+ExecStart=-/sbin/agetty -o '-p -f -- \\u' --keep-baud --autologin root 115200,57600,38400,9600 - $TERM
+
+[Install]
+WantedBy=multi-user.target
```
burgerdev, line 12: Whitespace issue.

burgerdev (approved):
Thanks for fixing this, I'm going to try out the image now.

image/system/variants.bzl
msanft: QEMU should provide `hvc0`, which mkosi enables by default now if I'm not mistaken

burgerdev: But that alone does not grant autologin, does it?

msanft: It should, but we'd need to keep the `--autologin` option for that. Or what's your concern?

burgerdev: Yes, my concern is that we can't log into QEMU images anymore if we take away the mkosi autologin and add our own based on the cmdline.

3u13r: I think I broke the fact that you could log into any qemu image even stable and nightly images. I will (try) to make this PR have the same behavior as before (i.e. enabling console for all images), though I'm not sure what the original reason was except for we can because it doesn't hurt security on qemu, but now that we have https://github.com/edgelesssys/constellation/tree/main/dev-docs/howto/bare-metal there might be a use-case to also have qemu images where one cannot log in.

3u13r:
Re-verified:
Console is accessible on miniconstellation with nightly and console image. Note that on console images `constellation.console` exists twice in the cmdline. 
Console is not accessible on GCP with nightly image.

burgerdev (approved):

PR #3357, opened 2024-09-16
terraform: allow GCP console access without opening debugd port

msanft:
### Context

On GCP, access to the serial console needs to be enabled via a metadata attribute. Previously, this has always been set to the `debug` variable, that also opens up the debugd port, which is unnecessary and potentially security-relevant for non-debug images, where arbitrary other services could bind to the debugd port by accident.
 
### Proposed change(s)

- Add a separate flag to enable serial console access without opening the debugd port.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/terraform-provider-constellation/examples/full/gcp/main.tf
+++ b/terraform-provider-constellation/examples/full/gcp/main.tf
@@ -76,6 +76,7 @@ module "gcp_infrastructure" {
   }
   image_id               = data.constellation_image.bar.image.reference
   debug                  = false
+  console_access         = false
   zone                   = local.zone
   region                 = local.region
   project                = local.project_id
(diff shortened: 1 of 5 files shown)


burgerdev (approved):
The change LGTM, but I don't quite understand the argument in the description: are you saying that we've been exposing the debug port when using a non-debug console image before? I don't see how, and this PR does not seem to address that.

msanft:
> The change LGTM, but I don't quite understand the argument in the description: are you saying that we've been exposing the debug port when using a non-debug console image before? I don't see how, and this PR does not seem to address that.

If someone were to try and get console access to a GCP console image before, he would have to enable the `debug` variable, which also exposed the port. This PR adds a knob so that you can get console access without opening the port.

elchead (approved):

daniel-weisse:
Why is this necessary?
The serial console is only usable for debug (and console) images anyways, where the security risk of opening the debug port is acceptable, since the images themselves are not secure.

3u13r (changes requested):
As discussed via message, I'd like to discuss this in-person first.

msanft:
> Why is this necessary? The serial console is only usable for debug (and console) images anyways, where the security risk of opening the debug port is acceptable, since the images themselves are not secure.

IMO, having an exposed port for an undefined service could grant attackers access to our infrastructure and is not desirable

PR #3361, opened 2024-09-17
operator: use GCP REST API for instance templates

msanft:
### Context

GCP upgrades were broken previously as new nodes didn't use the correct instance types. This was caused by the Protobuf-based GCP Go SDK [isn't aware of the `confidential_instance_type` field](https://github.com/googleapis/google-cloud-go/issues/10873?notification_referrer_id=NT_kwDOA3axdbQxMjQyMTc4MTc5ODo1ODExMDMyNQ#issuecomment-2353205902), which thus wasn't copied when creating a new instance template during an upgrade. This can be circumvented by using the [REST API implementation](https://github.com/googleapis/google-api-go-client), which is aware of the field.

### Proposed change(s)

- Use the GCP REST API (the corresponding Go SDK, really) for managing instance templates during an upgrade.

### Additional info

- [AB#4660](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/4660)
- [Instance template created before this change](https://console.cloud.google.com/compute/instanceTemplates/details/constell-e130e0b4-control-plane-c2aca755-1?project=constellation-331613) is missing `confidential_instance_type`
- [Instance template created after this change](https://console.cloud.google.com/compute/instanceTemplates/details/constell-911e01f7-worker-c26e910e-1?project=constellation-331613) has it set.

### Checklist

  - [GCP (SEV-SNP) upgrade test](https://github.com/edgelesssys/constellation/actions/runs/10904937959) 
- [x] Link to Milestone


msanft:
From my manual testing, it seems that this patch is not sufficient to make SEV-SNP upgrades on GCP work again. Anyhow, I verified that this fix alone is correct by checking that instance templates now use the correct confidential instance type, so I think this fix should be merged in either case.

operators/constellation-node-operator/internal/cloud/gcp/client/scalinggroup.go
```diff
@@ -49,29 +50,22 @@ func (c *Client) SetScalingGroupImage(ctx context.Context, scalingGroupID, image
...
-		InstanceTemplateResource: instanceTemplate,
-	})
-	if err != nil {
+	instanceTemplate.Name = newTemplateName
+	if _, err := c.instanceTemplateAPI.Insert(project, instanceTemplate); err != nil {
 		return fmt.Errorf("cloning instance template: %w", err)
 	}
-	if err := op.Wait(ctx); err != nil {
-		return fmt.Errorf("waiting for cloned instance template: %w", err)
-	}
```
burgerdev, line 70: The REST API also returns an operation - don't we need to do something with it?

burgerdev (approved):

operators/constellation-node-operator/internal/cloud/gcp/client/scalinggroup.go
burgerdev: Ah, there's no `Wait` on the rest op.

daniel-weisse (approved):

PR #3353, opened 2024-09-12
helm: fix kubeadm bugs caused by CoreDNS installation

burgerdev:
### Context

Putting the CoreDNS deployment under Helm management (#3236) introduced two bugs:

1. The `kubeadm upgrade` preflight checks started to fail because the version can't be parsed from the pinned image (see kubernetes/kubernetes#122967).
2. `kubeadm upgrade` overwrites the Helm chart resources because there's no way to skip the addon phase (see kubernetes/kubeadm#1318).

Both of these issues arise because Helm charts are installed before the Kubernetes upgrade, and both issues have not been caught because the upgrade process ignores errors from kubeadm (see edgelesssys/issues#134).

### Proposed change(s)

- Ignore CoreDNS-related preflight errors.
- Rename the CoreDNS ConfigMap during upgrades. This is working around the missing kubeadm phase feature: if there's no `kube-system/coredns` ConfigMap, kubeadm does not try to manage CoreDNS. This is more temporary code that will be deleted as soon as all eligible upgrade origin versions have Helm-managed CoreDNS (>2.19). 

### Additional info

- [AB#4654](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/4654)

### Checklist

- [x] Run the E2E tests that are relevant to this PR's changes
  - [x] Azure e2e
  - [x] Azure upgrade
- [x] Add labels (e.g., for changelog category)
- [x] Is PR title adequate for changelog?
- [x] Link to Milestone

PR #3366, opened 2024-09-18
helm: revert parts of CoreDNS Helm chart packaging

burgerdev:
### Context

We are trying to move CoreDNS from a cluster addon managed by `kubeadm` to a Helm chart managed alongside the Constellation microservices.

After #3236,  upgrading a cluster reverted CoreDNS to the version installed by `kubeadm`. This required #3353 as a workaround - the cluster pretends it does not have CoreDNS to fool kubeadm, details in that PR's description.

This workaround had an unexpected consequence, though: kubeadm has a preflight check for CoreDNS, which now starts failing. Since this check is done in the upgrade-agent, which is part of the image, it does not benefit from the `--ignore`s added in #3353 and fails. Now the `kubeadm upgrade` step is never executed, which makes us miss important migrations in the worst case.

### Proposed change(s)

Extend the migration to two upgrades.

1. upgrade-agent ignores CoreDNS problems in 2.18+
2. CoreDNS is installed as Helm chart in 2.19+

### Additional info

- [AB#4627](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/4627)

### Checklist

  - ~[ ] [GCP upgrade](https://github.com/edgelesssys/constellation/actions/runs/10924437871)~
  - [x] [GCP upgrade](https://github.com/edgelesssys/constellation/actions/runs/10935780631)
  - [x] [GCP e2e](https://github.com/edgelesssys/constellation/actions/runs/10924451622)
- [x] Link to Milestone


elchead (approved):
we should probably add a TODO(v2.19) to the helm chart installation /  add a ticket. but I think you have this under control @burgerdev

burgerdev:
Note to reviewers: 

There are three commits, 

* revert of #3236 
* revert of #3353 
* cherry-pick the upgrade-agent parts of #3353

Limiting the diff to the last one should be sufficient.

> we should probably add a TODO(v2.19) to the helm chart installation / add a ticket. but I think you have this under control @burgerdev

I'll add an AB ticket, thanks for the reminder.

3u13r (approved):
While the upgrade test is still failing, the code itself LGTM

burgerdev:
Interesting - failing upgrades were not even a symptom of the issue addressed here, and this branch is rebased on #3361. The failure is also during initial cluster setup at v2.17.0, which should be sufficiently tested by now. Let's see what the retry does.

msanft (approved):
Thanks for explaining this so thoroughly!

PR #3373, opened 2024-09-24
ci: update incorrect variable naming in mini e2e test

daniel-weisse:
### Context
The `rgname` variable set by `e2e/miniconstellation/main.sh.in` is not parsed back to the GitHub workflow if `terraform apply` failed

### Proposed change(s)

- Define resource group name outside of terraform to ensure there is always a valid value, even if `terraform apply` fails
- Only run tests in West Europe to get around potential quota issues

### Additional info
- Resource group is now [correctly set](https://github.com/edgelesssys/constellation/actions/runs/11008951631/job/30567585420#step:5:2918) on delete
- [e2e test](https://github.com/edgelesssys/constellation/actions/runs/11010915855/job/30573868498)
- Fixes https://github.com/edgelesssys/issues/issues/888

### DIFF
--- a/.github/actions/e2e_mini/action.yml
+++ b/.github/actions/e2e_mini/action.yml
@@ -61,6 +61,5 @@ runs:
       # clean up if e2e test failed or if the run was cancelled
       if: (failure() && steps.e2e-test.conclusion == 'failure') || cancelled()
       run: |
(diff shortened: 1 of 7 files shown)


burgerdev (approved):
Thanks!

PR #3375, opened 2024-09-24
openstack: Fix a crash in the bootstrapper during node join discovery

3u13r:
### Context
The bootstrapper still likes to query any nodes in addition to the LB endpoint when searching for join endpoints.

Openstack has a catalog discovery feature that gets URL endpoints during authentication (see: https://github.com/jtopjian/gophercloud/blob/master/openstack/endpoint_location.go#L63). Since we did not set the region we received more endpoints than expected and some even invalid ones.

### Proposed change(s)

- Passthrough openstack region name via IMDS to instance to use it in the cloud API client.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/internal/cloud/openstack/imds.go
+++ b/internal/cloud/openstack/imds.go
@@ -172,6 +172,20 @@ func (c *imdsClient) userDomainName(ctx context.Context) (string, error) {
 	return c.userDataCache.UserDomainName, nil
 }
 
+func (c *imdsClient) regionName(ctx context.Context) (string, error) {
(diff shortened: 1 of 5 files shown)


burgerdev (commented):
I don't see where the actual crash is being fixed or handled - the bootstrapper should be able to join the cluster even if the metadata API does not return or returns wrong addresses, shouldn't it?

3u13r:
I also not remember the actual crash location. I'll spin up a cluster to document this more thoroughly.

burgerdev (approved):
I tested the branch and the change lgtm. Let's follow up on the resilience topic.

msanft (approved):

PR #3376, opened 2024-09-24
Init

flxflx:
This creates a document that describes the security protocols of Constellation. This should eventually move to the docs.


dev-docs/security-overview.md
```diff
@@ -0,0 +1,74 @@
...
+1. The new node sends a certificate signing request (CSR) to the JoinService.
+1. The JoinService issues a corresponding certificate and sends it to the new node. The JoinService uses the signing key of the Kubernetes cluster for this, which is [generated by kubeadm](https://kubernetes.io/docs/setup/best-practices/certificates/).
+1. The JoinService sends a [kubeadm token](https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm-token/) to the new node.
+1. The JoinService sends the hashes of the to-be-installed Kubernetes binaries to the new node.
+1. The JoinService sends encryption key for the new node's local storage. This key is generated and managed by the cluster's KeyService.
+1. The JoinService sends the certificate of the cluster's control plane to the new node.
+
+After this, the aTLS connection is closed and the node is marked as "initialized". 
+
+The Bootstrapper on the new node uses the kubeadm token to download the configuration of the cluster from the Kubernetes API server. Subsequently, it downloads, verifies, and runs the given Kubernetes binaries accordingly. The kubeadm token is never used after this.
```
3u13r, line 64: > download the configuration of the cluster from the Kubernetes API server. Subsequently, it downloads, verifies, and runs the given Kubernetes binaries accordingly

That's not correct. The components are part of the `IssueJoinTicketResponse`: https://github.com/edgelesssys/constellation/blob/fix/openstack/set-region-in-cloud-client/joinservice/joinproto/join.proto#L47

dev-docs/security-overview.md
3u13r, line 1: Please break up the sentences in the paragraphs into different lines for easier review/github diff generation.

dev-docs/security-overview.md
```diff
@@ -0,0 +1,74 @@
...
+
+After this, the aTLS connection is closed and the node is marked as "initialized". 
+
+The Bootstrapper on the new node uses the kubeadm token to download the configuration of the cluster from the Kubernetes API server. Subsequently, it downloads, verifies, and runs the given Kubernetes binaries accordingly. The kubeadm token is never used after this. 
+
+The kubelet on the new node uses its own certificate and the certificate of the cluster's control plane (which the new node both received from the JoinService) to establish an mTLS connection with the cluster's Kubernetes API server. Once connected, the new node registers itself as control-plane node or worker node of the cluster. This process uses the standard Kubernetes mechanisms for adding nodes. 
+
+In Constellation, a virtual private network (VPN) exists between all nodes of a cluster. To join this VPN, the new node generates WireGuard credentials for itself and writes the public part to etcd via the mTLS connection with the Kubernetes API server. Whenever nodes within a Constellation cluster are talking to each other for the first time, they check etcd for the other node's public WireGuard credentials.
```
3u13r, line 68: > Whenever nodes within a Constellation cluster are talking to each other for the first time, they check etcd for the other node's public WireGuard credentials.

There is no synchronous mechanism to look up the public key before a packet is send out. The Cilium agent on the node watches for changes and eventually sets the correct public key and IPs on the WireGuard interface of the node. 
What we additionally do is that we forbid that unencrypted traffic leaves the node when it's source and destination reside in the same node subnet.

dev-docs/security-overview.md
```diff
@@ -0,0 +1,74 @@
...
+
+After this, the aTLS connection is closed and the node is marked as "initialized". This marker is irrevocably reflected in the node's remote attestation. This mechanism prevents a node from joining different clusters. 
+
+## Kubernetes bootstrapping on the first node
+
+The Bootstrapper on the first node downloads and verifies the Kubernetes binaries, using the hashes it received from the CLI. These binaries include for example [kubelet](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/), [kube-apiserver](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/), and [etcd](https://kubernetes.io/docs/tasks/administer-cluster/configure-upgrade-etcd/). With these, the Bootstrapper creates a single-node Kubernetes cluster. Etcd is a distributed key-value store that Kubernetes uses to store configuration data for services. The etcd agent runs on each control-plane node of a cluster. The agents use mTLS for communication between them. Etcd uses the Raft protocol (over mTLS) to distribute state between nodes.
```
3u13r, line 42: Neither the kube-api server nor the etcd is something that we regard as a Kubernetes component: https://github.com/edgelesssys/constellation/blob/main/internal/versions/versions.go#L269

dev-docs/chain-of-trust.jpg
3u13r, line 1: I think we should link to https://docs.edgeless.systems/constellation/architecture/attestation#chain-of-trust instead of embedding the jpg of the diagram here.

3u13r (changes requested):

dev-docs/security-overview.md
flxflx: These are now also part of the pinned components:

- https://github.com/edgelesssys/constellation/pull/2719
- https://github.com/edgelesssys/constellation/pull/2764

3u13r: Ok sure. I would've not considered them "components that the bootstrapper downloads" since there's an indirection of who downloads the kube-api server, but sure fine with me.

dev-docs/chain-of-trust.jpg
flxflx: Let's keep it for now and remove once we integrate with the docs.

flxflx:
@3u13r I addressed your comments and filled-in a range of blanks.

PR #3413, opened 2024-10-14
ci: upgrade to ubuntu-24.04 fixups

daniel-weisse:
### Context
renovate did not upgrade all of our workflows to ubuntu-24.04, since in some workflows, we configure the runner images through user input.
The upgrade also broke our e2e weekly cleanup workflow, since the new image no longer comes with terraform pre-installed

### Proposed change(s)

- Upgrade the runner image for e2e tests to ubuntu-24.04
  - [workflow run](https://github.com/edgelesssys/constellation/actions/runs/11322563015)
- Explicitly install Terraform in the e2e weekly cleanup workflow
  - [workflow run](https://github.com/edgelesssys/constellation/actions/runs/11322472066)


burgerdev (approved):
Thanks!

PR #3449, opened 2024-10-22
docs: clarify for Azure TDX with Terraform provider

elchead:
### Context

The current provider docs don't distinguish Azure SEV-SNP and TDX, even though the steps are not the same.
Also, I noticed that the TF code doesn't validate the instance type for the given attestation variant. Thus, a user might forget to change this field when using TDX. The validation cannot be easily included because infra and constellation resources are strictly separated in TF, but a comment seems like a good enough hint.

### Proposed change(s)

- add a comment to adjust the instance type for the TDX variant
- separate the instructions for Azure SEV-SNP and TDX

### Checklist

- [ ] Link to Milestone

### DIFF
--- a/docs/docs/workflows/terraform-provider.md
+++ b/docs/docs/workflows/terraform-provider.md
@@ -37,7 +37,7 @@ This example shows how to set up a Constellation cluster with the reference IAM 
 
   <Tabs groupId="variant">
   <TabItem value="sevsnp" label="SEV-SNP">
-  When creating a cluster on Azure SEV-SNP, you need to manually patch the policy of the MAA provider before creating the Constellation cluster, as this feature isn't available in Azure's Terraform provider yet. The Constellation CLI provides a utility for patching, but you
+  When creating a cluster with SEV-SNP, you need to manually patch the policy of the MAA provider before creating the Constellation cluster, as this feature isn't available in Azure's Terraform provider yet. The Constellation CLI provides a utility for patching, but you
   can also do it manually.
 
   ```bash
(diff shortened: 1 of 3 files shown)


docs/docs/workflows/terraform-provider.md
```diff
@@ -34,7 +34,10 @@ This example shows how to set up a Constellation cluster with the reference IAM
   Optionally, you can prefix the `terraform apply` command with `TF_LOG=INFO` to collect [Terraform logs](https://developer.hashicorp.com/terraform/internals/debugging) while applying the configuration. This may provide helpful output in debugging scenarios.
   </TabItem>
   <TabItem value="azure" label="Azure">
-  When creating a cluster on Azure, you need to manually patch the policy of the MAA provider before creating the Constellation cluster, as this feature isn't available in Azure's Terraform provider yet. The Constellation CLI provides a utility for patching, but you
+
```
msanft, line 37: Instead of the triple-nested tabs here, I'd prefer to have an info block indicating that the patching is necessary for SEV-SNP only.

Then, we could simply share the text for applying for both (since it's the same for SNP and TDX).

terraform-provider-constellation/examples/full/azure/main.tf
```diff
@@ -21,7 +21,7 @@ locals {
   location             = "northeurope"
   control_plane_count  = 3
   worker_count         = 2
-  instance_type        = "Standard_DC4as_v5"
+  instance_type        = "Standard_DC4as_v5" // needs to be changed for azure-tdx
```
msanft, line 24:
```suggestion
  instance_type        = "Standard_DC4as_v5" // Adjust if using TDX
```

msanft: `needs to be changed` reads to me as if this could have also done by us.

msanft (changes requested):

msanft (approved):

thomasten (approved):
LGTM. Please backport to the 2.19 docs before merge.

PR #3454, opened 2024-10-22
terraform: fix security rule reconciliation on Azure

elchead:
### Context

In https://github.com/edgelesssys/constellation/pull/3257, the migration to defining the network security rules as separate resources was started.
Unfortunately, the concurrent usage of inline and security rule resources causes problems (see https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/network_security_group). Terraform cannot resolve which declarative state of the rules should be applied if both inline and resources exist, thus making it a concurrency race which resource overwrites the others.

This bug was first observed as part of the weekly Azure TDX Terraform provider test, but it did indeed also fail during the release test with the CLI (https://github.com/edgelesssys/constellation/actions/runs/11402745527/job/31778136903).
The observed behavior was that the Kube API server was not reachable anymore, because the network security rule was missing.
This bug applies also to Azure SEV-SNP, but since the bug is a concurrency race, its manifestation is not deterministic. In fact, the correct behavior happened to occur during the release test on SEV-SNP (https://github.com/edgelesssys/constellation/actions/runs/11402745527/job/31778136329#step:19:874).

The problem can occur on any cluster on Azure and every `terraform apply` could cause this bug.

### Proposed change(s)
By leaving the inline security rules block of the network security group empty, this resource does not reconcile the rules.
Thus, the solution is to fully migrate to separate resources for the security rules.

I propose to do a patch release where users from 2.18 are required to upgrade to 2.19.1 directly.

- remove the inline network security rules on Azure

### Additional info

I tested setting up a new cluster with this change and subsequent applies didn’t result in any terraform diff as expected.
The cluster remained reachable.

Upgrade test from v2.18.0 -> v2.19.1
✅  [Azure TDX](https://github.com/edgelesssys/constellation/actions/runs/11513095092/job/32049960071)
✅ [Azure SEV-SNP](https://github.com/edgelesssys/constellation/actions/runs/11497662606)

For upgrades from v2.18.0 on Azure, the old security rules are orphaned and should be manually cleaned up by the user:

```
#!/bin/bash
name="<insert>"  # the name provided in the config
id="<insert>"    # the cluster id

rules=(
  "kubernetes"
  "bootstrapper"
  "verify"
  "recovery"
  "join"
  "debugd"
  "konnectivity"
)

for rule in "${rules[@]}"; do
  echo "Deleting rule: $rule"
  az network nsg rule delete \
    --resource-group "$name-rg" \
    --nsg-name "$name-$id" \
    --name "$rule"
done

echo "All specified rules have been deleted."
```

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/e2e-upgrade.yml
+++ b/.github/workflows/e2e-upgrade.yml
@@ -40,7 +40,7 @@ on:
         type: string
         required: false
       toMicroservices:
-        description: Microservice version to target for the upgrade, empty for upgrade target's default version. When simulating a patch upgrade, you must set this value because the default will not be the simulated version.
+        description: Microservice version to target for the upgrade, empty for upgrade target's default version.
         type: string
         required: false
       simulatedTargetVersion:
--- a/.github/workflows/e2e-upgrade.yml
+++ b/.github/workflows/e2e-upgrade.yml
@@ -132,57 +132,6 @@ jobs:
 
           echo "cloudProvider=${cloudProvider}" | tee -a "$GITHUB_OUTPUT"
 
-  build-target-cli:
-    name: Build upgrade target version CLI
-    runs-on: ubuntu-24.04
-    permissions:
-      id-token: write
-      checks: write
-      contents: read
-      packages: write
-    steps:
-      - name: Checkout
-        if: inputs.gitRef == 'head'
-        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
-        with:
-          fetch-depth: 0
-          ref: ${{ !github.event.pull_request.head.repo.fork && github.head_ref || '' }}
-
-      - name: Checkout ref
-        if: inputs.gitRef != 'head'
-        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
-        with:
-          fetch-depth: 0
-          ref: ${{ inputs.gitRef }}
-
-      - name: Setup Bazel & Nix
-        uses: ./.github/actions/setup_bazel_nix
-
-      - name: Log in to the Container registry
-        uses: ./.github/actions/container_registry_login
-        with:
-          registry: ghcr.io
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Simulate patch upgrade
-        if: inputs.simulatedTargetVersion != ''
-        run: |
-          echo ${{ inputs.simulatedTargetVersion }} > version.txt
-
-      - name: Build CLI
-        uses: ./.github/actions/build_cli
-        with:
-          enterpriseCLI: true
-          outputPath: "build/constellation"
-          push: true
-
-      - name: Upload CLI binary
-        uses: actions/upload-artifact@0b2256b8c012f0828dc542b3febcab082c67f72b # v4.3.4
-        with:
-          name: constellation-upgrade-${{ inputs.attestationVariant }}
-          path: build/constellation
-
   create-cluster:
     name: Create upgrade origin version cluster
     runs-on: ubuntu-24.04
@@ -279,7 +228,6 @@ jobs:
       packages: write
     needs:
       - generate-input-parameters
-      - build-target-cli
       - create-cluster
     steps:
       - name: Checkout
@@ -298,6 +246,32 @@ jobs:
 
       - name: Setup Bazel & Nix
         uses: ./.github/actions/setup_bazel_nix
+
+      - name: Log in to the Container registry
+        uses: ./.github/actions/container_registry_login
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # applying the version manipulation here so that the upgrade test tool is also on the simulated target version
+      - name: Simulate patch upgrade
+        if: inputs.simulatedTargetVersion != ''
+        run: |
+          echo ${{ inputs.simulatedTargetVersion }} > version.txt
+
+      - name: Build CLI
+        uses: ./.github/actions/build_cli
+        with:
+          enterpriseCLI: true
+          outputPath: "build/constellation"
+          push: true
+
+      - name: Upload CLI binary # is needed for the cleanup step
+        uses: actions/upload-artifact@0b2256b8c012f0828dc542b3febcab082c67f72b # v4.3.4
+        with:
+          name: constellation-upgrade-${{ inputs.attestationVariant }}
+          path: build/constellation
 
       - name: Login to AWS
         uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v4.0.2
@@ -335,11 +309,6 @@ jobs:
         with:
           azure_credentials: ${{ secrets.AZURE_E2E_IAM_CREDENTIALS }}
 
-      - name: Download CLI
-        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
-        with:
-          name: constellation-upgrade-${{ inputs.attestationVariant }}
-          path: build
 
       - name: Download Working Directory (Pre-test)
         uses: ./.github/actions/artifact_download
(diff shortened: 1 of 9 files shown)


elchead:
@3u13r I understand that your intention for the dangling konnectivity rule in #3257 was to enforce that the other rules are deleted, but this should not be necessary afais. Or do you see any issue?

3u13r:
> I understand that your intention for the dangling konnectivity rule in https://github.com/edgelesssys/constellation/pull/3257 was to enforce that the other rules are deleted, but this should not be necessary afais. Or do you see any issue?

No, if the posted CI runs succeeds this solution is cleaner in general and is likely to cause less problems in the future.

elchead:
I'm not sure where the best place for the migration script would be (see Additional Information in the description).

.github/workflows/e2e-upgrade.yml
```diff
@@ -404,15 +404,11 @@ jobs:
...
-          if [[ -n ${KUBERNETES} ]]; then
-            KUBERNETES_FLAG="--target-kubernetes=$KUBERNETES"
-          fi
-
           sudo sh -c 'echo "127.0.0.1 license.confidential.cloud" >> /etc/hosts'
-          bazel run --test_timeout=14400 //e2e/internal/upgrade:upgrade_test -- --want-worker "$WORKERNODES" --want-control "$CONTROLNODES" --target-image "$IMAGE" "$KUBERNETES_FLAG" "$MICROSERVICES_FLAG"
+          echo "CLI VERSION:"
+          echo $(./build/constellation version)
+          CLI=$(realpath ./build/constellation)
+          bazel run --test_timeout=14400 //e2e/internal/upgrade:upgrade_test -- --want-worker "$WORKERNODES" --want-control "$CONTROLNODES" --target-image "$IMAGE" --target-kubernetes "$KUBERNETES" --target-microservices "$MICROSERVICES" --cli "$CLI"
```
elchead, line 411: the old code was broken because an empty string "" as substitution for a flag breaks the flag parsing

e2e/internal/upgrade/upgrade.go
```diff
@@ -297,10 +297,10 @@ func getCLIPath(cliPathFlag string) (string, error) {
 	pathCLI := os.Getenv("PATH_CLI")
 	var relCLIPath string
 	switch {
-	case pathCLI != "":
-		relCLIPath = pathCLI
 	case cliPathFlag != "":
```
elchead, line 300: a flag should take precedence over ENV

3u13r (changes requested):
We should also document the not necessarily required but at least recommended migration steps in `docs/docs/reference/migration.md`

.github/workflows/e2e-upgrade.yml
elchead: I decided to keep the explicit cli passing, such that the identical CLI is used for "config migrate" and the rest of the upgrade workflow

elchead:
As discussed with @3u13r, we decided to build the target CLI inside the upgrade job itself, because separating this created several consistency issues for the simulated patch version scenario
The main reason being that the upgrade tool itself didn't have the patched version, but referred to its own embedded version to determine some target versions.

Moreover, the extra job didn't save any time because the CLI is built inside the bazel upgrade target too.

docs/docs/reference/migration.md
```diff
@@ -3,51 +3,80 @@
 This document describes breaking changes and migrations between Constellation releases.
 Use [`constellation config migrate`](./cli.md#constellation-config-migrate) to automatically update an old config file to a new format.
 
+## Migrations to v2.19.1
+
+### Azure
+
+* During the upgrade, security rules are migrated and the old ones are recommended to be cleaned up manually by the user. This step is recommended but not strictly required. The below script shows how to programatically delete the old rules through the Azure CLI:
```
elchead, line 10: I'd recommend to not make this optional, because we could not change the security rule name back and remove the priority offset in Terraform otherwise.

docs/docs/reference/migration.md
```diff
@@ -3,51 +3,80 @@
 This document describes breaking changes and migrations between Constellation releases.
 Use [`constellation config migrate`](./cli.md#constellation-config-migrate) to automatically update an old config file to a new format.
 
+## Migrations to v2.19.1
+
+### Azure
+
+* During the upgrade, security rules are migrated and the old ones need to be cleaned up manually by the user. The below script shows how to delete them through the Azure CLI:
```
elchead, line 10: Once approved, I will backport this to the v2.19 docs

docs/docs/reference/migration.md
```diff
@@ -3,51 +3,80 @@
...
+
+```bash
+#!/bin/bash
+name="<insert>"  # the name provided in the config
+id="<insert>"    # the cluster id
```
3u13r, line 15:
```suggestion
uid="<insert>"    # the cluster id can be retrieved via `yq '.infrastructure.uid' constellation-state.yaml`
```

docs/docs/reference/migration.md
```diff
@@ -3,51 +3,80 @@
...
+  "debugd"
+  "konnectivity"
+)
+
+for rule in "${rules[@]}"; do
+  echo "Deleting rule: $rule"
+  az network nsg rule delete \
+    --resource-group "$name-rg" \
+    --nsg-name "$name-$id" \
+    --name "$rule"
```
3u13r, line 32:
```suggestion
  echo "Deleting rule: ${rule}"
  az network nsg rule delete \
    --resource-group "${name-rg}" \
    --nsg-name "${name}-${uid}" \
    --name "${rule}"
```

docs/docs/reference/migration.md
```diff
@@ -3,51 +3,80 @@
...
+  "recovery"
+  "join"
+  "debugd"
+  "konnectivity"
+)
+
+for rule in "${rules[@]}"; do
+  echo "Deleting rule: $rule"
+  az network nsg rule delete \
+    --resource-group "$name-rg" \
```
3u13r, line 30: We need to also define the rg name as the other variables above since it can be specified in the constellation-conf.yaml und `.provider.azure.resourceGroup`

3u13r (changes requested):

docs/docs/reference/migration.md
```diff
@@ -3,51 +3,81 @@
...
+  "bootstrapper"
+  "verify"
+  "recovery"
+  "join"
+  "debugd"
+  "konnectivity"
+)
+
+  echo "Deleting rule: ${rule}"
+  az network nsg rule delete \
```
3u13r, line 29: I think the loop went missing in the last commit

3u13r (changes requested):

docs/docs/reference/migration.md
```diff
@@ -3,51 +3,81 @@
...
-
+ afterward.
 
 ## Migrating from Azure's service principal authentication to managed identity authentication (during the upgrade to Constellation v2.8.0)
 
-- The `provider.azure.appClientID` and `provider.azure.appClientSecret` fields are no longer supported and should be removed.
-- To keep using an existing UAMI, add the `Owner` permission with the scope of your `resourceGroup`.
-- Otherwise, simply [create new Constellation IAM credentials](../workflows/config.md#creating-an-iam-configuration) and use the created UAMI.
-- To migrate the authentication for an existing cluster on Azure to an UAMI with the necessary permissions:
+* The `provider.azure.appClientID` and `provider.azure.appClientSecret` fields are no longer supported and should be removed.
```
3u13r, line 51: I don't think we currently enforce any rules regarding `*` vs `-`. They seem to be interchangeable. Did your linter change this, or do you have a reason for this preference? I know that docs currently favor `*` (i.e. are used more). I think we can revert this for now to not have the last commit that changed the line updated and revisit this discussion later.

3u13r (approved):
LGTM

docs/docs/reference/migration.md
thomasten: We agreed on markdownlint, which (unfortunately) picks the character of the topmost list item on the page. So if you wanted to keep the diff small but still use the linter, you would replace the first item with `-` in this case.

thomasten (approved):

elchead:
Upgrade test passed again: https://github.com/edgelesssys/constellation/actions/runs/11615051562

PR #3456, opened 2024-10-23
ci: use GOPROXY fallbacks

burgerdev:
### Context

Golang builds use a module proxy to fetch dependencies. The default configuration is `GOPROXY=https://proxy.golang.org,direct`, which means "always use proxy.golang.org, and fall back to direct if the result is a 404". If the proxy fails differently (network disconnect, 502, ...), fetching fails.

### Proposed change(s)

- Use fallbacks on all proxy errors (this is done by using `|` instead of `,`).
- Add goproxy.io as additional fallback.

### Related issue
- Fixes edgelesssys/issues#1004

### Checklist
- [x] Link to Milestone

### DIFF
--- a/bazel/bazelrc/ci.bazelrc
+++ b/bazel/bazelrc/ci.bazelrc
@@ -62,6 +62,10 @@ build --remote_local_fallback
 # Docs: https://bazel.build/reference/command-line-reference#flag--grpc_keepalive_time
 build --grpc_keepalive_time=30s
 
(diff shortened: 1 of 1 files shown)


daniel-weisse (approved):

burgerdev:
[Unit test](https://github.com/edgelesssys/constellation/actions/runs/11482038364/job/31954110588) was successful.

Issue #3438, opened 2024-10-17
Customers should not be able to create clusters with wrong flavors (stackit)


h3adex:
### Use case

Is it possible to display a warning when a customer is trying to create a constellation cluster using a wrong flavor id? Currently the customer is able to apply any flavor. Because it is not a confidential one the setup process is stuck in connecting state. Prior creating the cluster we should warn the user if he is using any flavor which is not listed in the list of available stackit flavors.

Available confidential flavors in your codebase:
https://github.com/edgelesssys/constellation/blob/50df35438e1c28c05ad3eed6561a5ccbd92b55a5/internal/config/instancetypes/stackit.go#L10

Once stackit is adding new flavors, I'll create PR to add them. But currently this is the correct list. I can also review/test the PR for this feature request. I might also write some Go-Code to implement it, if you give me a hint where to start.

<img width="634" alt="Screenshot 2024-10-17 at 13 20 45" src="https://github.com/user-attachments/assets/b91e1624-ab11-49c3-a8dc-577306570b57">

### Would you be willing to implement this feature?

- [X] Yes, I could contribute this feature.

msanft:
Hey @h3adex,

That's certainly a good addition! If you want to contribute this, you could add a check to the existing [instance type validation code](https://github.com/edgelesssys/constellation/blob/b8c26f5989a21d7d87abbd7653e3992fbd097e8a/internal/config/validation.go#L523). Additionally, we could add a comment in our [Terraform provider example](https://github.com/edgelesssys/constellation/blob/b8c26f5989a21d7d87abbd7653e3992fbd097e8a/terraform-provider-constellation/examples/full/stackit/main.tf#L28), indicating that a confidential instance type needs to be selected.

Thanks a lot!

h3adex:
PR is merged. Closed this issue. TYVM!

PR #3463, opened 2024-10-28
config: only allow confidential instances on stackit

h3adex:
### Context
When creating a confidential Kubernetes cluster on Stackit, customers have occasionally confused confidential flavor IDs with non-confidential ones which resulted in a "stuck" creation period.

### Proposed Change(s)
I have implemented an instance type check for the Stackit cloud provider to prevent this confusion.

### Related Issue
[#3438](https://github.com/edgelesssys/constellation/issues/3438)

### Additional Info
I have rebuilt the Constellation CLI and confirmed that the change works correctly on a Linux x86 machine.
Using ```bazel build //cli:cli_oss```

### Checklist

### DIFF
--- a/internal/config/validation.go
+++ b/internal/config/validation.go
@@ -550,15 +550,16 @@ func validInstanceTypeForProvider(insType string, attestation variant.Variant, p
 		}
 	case variant.QEMUVTPM{}, variant.QEMUTDX{}:
 		// only allow confidential instances on stackit cloud using QEMU vTPM
-		if cloud := provider.OpenStack.Cloud; strings.ToLower(cloud) == "stackit" {
-			for _, instanceType := range instancetypes.STACKITInstanceTypes {
-				if insType == instanceType {
-					return true
+		if provider.OpenStack != nil {
+			if cloud := provider.OpenStack.Cloud; strings.ToLower(cloud) == "stackit" {
+				for _, instanceType := range instancetypes.STACKITInstanceTypes {
+					if insType == instanceType {
+						return true
+					}
 				}
+				return false
 			}
-			return false
 		}
-
 		return true
 	}
 	return false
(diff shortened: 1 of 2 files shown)


msanft (approved):
LGTM. Also thanks a lot for adding a test!

internal/config/validation.go
```diff
@@ -549,6 +549,16 @@ func validInstanceTypeForProvider(insType string, attestation variant.Variant) b
 			}
 		}
 	case variant.QEMUVTPM{}, variant.QEMUTDX{}:
+		// only allow confidential instances on stackit cloud using QEMU vTPM
+		if cloud := provider.OpenStack.Cloud; strings.ToLower(cloud) == "stackit" {
```
msanft, line 553: We need an `if provider.OpenStack != nil` here, as there's more scenarios where we'll have `case variant.QEMUVTPM{}, variant.QEMUTDX{}`, such as MiniConstellation, and else we'd run into a nil panic there here when dereferencing `provider.OpenStack`.

We might also want "success" test case for these variants without the OpenStack / STACKIT case here.

msanft (changes requested):

msanft:
The failing CI is somewhat expected on your branch, as the required tokens aren't available there. I'll verify that these succeed on my machine once we're ready to merge, so don't worry about fixing that for now. :)

internal/config/validation.go
h3adex: I've added a provider check and two test cases. Although I'm not entirely sure which provider Miniconstellation is using, I made an educated guess: `ProviderConfig{QEMU: &QEMUConfig{}}` with the variant `variant.QEMUTDX{}`.

daniel-weisse (approved):

msanft:
All tests succeed, merging.

PR #3482, opened 2024-11-18
ci: run `terraform apply` in provider example test with timeout

daniel-weisse:
### Context
GitHub actions are run with a 6 hour internal timeout, after which running steps are terminated.
The termination signal is not properly forwarded to the running Terraform process, leading to resources not being cleaned up and state information potentially being lost.
For us, this only happens when there is an internal error during the `constellation apply` stage of the terraform provider, which is ran without any timeouts.
Since `terraform apply` does not support any global timeout flags, we may hit the 6 hour mark during this step.

### Proposed change(s)

- Run `terraform apply` with `timeout` to stop execution of apply after 1 hour, or 4 hours for upgrades

### Related issue
- Should fix these clean up failures caused by hitting GitHub's timeout:
  - https://github.com/edgelesssys/constellation/actions/runs/11866687017/job/33073770054
  - https://github.com/edgelesssys/constellation/actions/runs/11866687017/job/33073769883
  - https://github.com/edgelesssys/constellation/actions/runs/11866687017/job/33073769769


.github/workflows/e2e-test-provider-example.yml
```diff
@@ -407,7 +407,7 @@ jobs:
         shell: bash
         run: |
           terraform init --upgrade
-          terraform apply -auto-approve
+          timeout 4h terraform apply -auto-approve
```
3u13r, line 410: Why is this timeout that long, we don't wait here until all nodes are upgraded.

3u13r (changes requested):
> this only happens when there is an internal error during the constellation apply stage of the terraform provider, which is ran without any timeouts.

This would also affect other people using our provider in the Terraform CI, right? Moreover, I think the proper fix is to have a default, user editable timeout in the `constellation apply` step.
Having said that, we can do this as a follow up, since we want to fix CI issues here.

PR #3483, opened 2024-11-18
ci: update workload identity provider url

daniel-weisse:
### Context
We moved the workload pool to a different project

### Proposed change(s)

- Update the GCP workload identity provider URL

### DIFF
--- a/.github/actions/login_gcp/action.yml
+++ b/.github/actions/login_gcp/action.yml
@@ -22,7 +22,7 @@ runs:
     - name: Authorize GCP access
       uses: google-github-actions/auth@71fee32a0bb7e97b4d33d548e7d957010649d8fa # v2.1.3
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-cleanup-weekly.yml
```diff
@@ -4,7 +4,6 @@ on:
   schedule:
     - cron: "0 0 * * 0" # At 00:00 every Sunday UTC
   workflow_dispatch:
-
```
3u13r, line 7: I think the double newline is fine. Also omitting this doesn't add a commit to the file's history. So I think we should omit this for now.

3u13r (changes requested):

3u13r (approved):
LGTM, see: https://github.com/edgelesssys/constellation/actions/runs/11893749723/job/33144539351

PR #3496, opened 2024-11-26
bootstrapper: only err if no control plane IPs available

msanft:
### Context

Previously we errored out of the entire join if retrieval of either LB IP or control plane public IP failed. This resulted in the entire "use either IP" logic not working as intended.

This also fixes the problem encountered by STACKIT folks, where public IP retrieval fails and thus no nodes can join the cluster.

### Proposed change(s)

- Log a warning only if the IP retrievals fail, and only errors out of the join if no IP can be found at all.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/bootstrapper/internal/joinclient/joinclient.go
+++ b/bootstrapper/internal/joinclient/joinclient.go
@@ -173,13 +173,13 @@ func (c *JoinClient) tryJoinWithAvailableServices() (ticket *joinproto.IssueJoin
 
 	endpoint, _, err := c.metadataAPI.GetLoadBalancerEndpoint(ctx)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to get load balancer endpoint: %w", err)
+		c.log.Warn("Failed to get load balancer endpoint", "err", err)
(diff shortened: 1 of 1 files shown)


bootstrapper/internal/joinclient/joinclient.go
```diff
@@ -173,13 +173,13 @@ func (c *JoinClient) tryJoinWithAvailableServices() (ticket *joinproto.IssueJoin
 
 	endpoint, _, err := c.metadataAPI.GetLoadBalancerEndpoint(ctx)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to get load balancer endpoint: %w", err)
+		c.log.Warn("Failed to get load balancer endpoint", "err", err)
```
burgerdev, line 176: As far as I understand we could leave this block unchanged, because joining via node IP does not work anyway due to node-to-node strict mode. Otoh, this way it's nicely symmetric.

burgerdev (approved):

daniel-weisse (approved):

PR #3514, opened 2024-12-02
ci: ensure autoscaling test actuallly waits for all Pods to be ready

daniel-weisse:
### Context
Our e2e test for autoscaling uses `kubectl wait--for condition=available` to check if all Pods were scheduled correctly before checking if all expected nodes have joined the cluster.
Since we don't use a specially prepared deployment, `kubectl wait` only requires 75% of the desired Pods to actually be ready, or scheduled at all, to report available for the deployment.

### Proposed change(s)

- Use a deployment with `maxUnavailable` Pods set to `0` to ensure `kubectl wait` actually waits for all Pods

### Related issue
- Fixes https://github.com/edgelesssys/issues/issues/592
- Fixes https://github.com/edgelesssys/issues/issues/864
- 

### Additional info

- [e2e test](https://github.com/edgelesssys/constellation/actions/runs/12115264859)


burgerdev (approved):
Nice, thanks!

PR #3522, opened 2024-12-03
constellation-node-operator: don't bail out on listing errors

msanft:
### Context

If the GCP project has scaling groups for which our checks can't be performed (which is the case for regional scaling groups, as they "don't exist" for the operator, if deployed in another region) . In that case, we should not bail out directly but go on with the next group. An error should only be thrown if there are no matching groups at all.

### Proposed change(s)

- Don't error directly, but only if no scaling groups are found.

### Checklist

  - [x] [E2E](https://github.com/edgelesssys/constellation/actions/runs/12135753706)
- [x] Link to Milestone

### DIFF
--- a/operators/constellation-node-operator/internal/cloud/gcp/client/scalinggroup.go
+++ b/operators/constellation-node-operator/internal/cloud/gcp/client/scalinggroup.go
@@ -105,6 +105,8 @@ func (c *Client) GetAutoscalingGroupName(scalingGroupID string) (string, error) 
 // ListScalingGroups retrieves a list of scaling groups for the cluster.
 func (c *Client) ListScalingGroups(ctx context.Context, uid string) ([]cspapi.ScalingGroup, error) {
 	results := []cspapi.ScalingGroup{}
+	var retErr error
+
 	iter := c.instanceGroupManagersAPI.AggregatedList(ctx, &computepb.AggregatedListInstanceGroupManagersRequest{
 		Project: c.projectID,
 	})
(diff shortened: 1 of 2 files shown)


operators/constellation-node-operator/internal/cloud/gcp/client/scalinggroup.go
daniel-weisse, line 1: Instead of simply dropping all errors, lets collect the errors and return them (if there are any) in case no scaling groups can be found

msanft:
[Another E2E](https://github.com/edgelesssys/constellation/actions/runs/12137211435)

operators/constellation-node-operator/internal/cloud/gcp/client/scalinggroup.go
```diff
@@ -103,8 +103,7 @@ func (c *Client) GetAutoscalingGroupName(scalingGroupID string) (string, error)
 }
 
 // ListScalingGroups retrieves a list of scaling groups for the cluster.
-func (c *Client) ListScalingGroups(ctx context.Context, uid string) ([]cspapi.ScalingGroup, error) {
-	results := []cspapi.ScalingGroup{}
+func (c *Client) ListScalingGroups(ctx context.Context, uid string) (results []cspapi.ScalingGroup, retErr error) {
```
3u13r, line 106: Since you explicitly return the error in all cases anyway and I think that named returns sometime are a footgun, I'd define the retErr in the function itself and explicitly return in all cases (as already done)

3u13r (changes requested):

operators/constellation-node-operator/internal/cloud/gcp/client/scalinggroup.go
```diff
@@ -129,7 +131,7 @@ func (c *Client) ListScalingGroups(ctx context.Context, uid string) ([]cspapi.Sc
 			}
 			template, err := c.instanceTemplateAPI.Get(c.projectID, templateURI[len(templateURI)-1])
 			if err != nil {
-				return nil, fmt.Errorf("getting instance template: %w", err)
+				retErr = errors.Join(retErr, fmt.Errorf("getting instance template %q: %w", templateURI[len(templateURI)-1], err))
```
daniel-weisse, line 134:
```suggestion
				retErr = errors.Join(retErr, fmt.Errorf("getting instance template %q: %w", templateURI[len(templateURI)-1], err))
				continue
```
Need to skip to the next loop iteration in case of errors

daniel-weisse (approved):

3u13r (approved):
LGTM

PR #3557, opened 2024-12-19
feat: implement RFC 16 to allow emergency node access

miampf:
### Context

This PR aims to implement [RFC 16: Node access](https://github.com/edgelesssys/constellation/blob/main/rfc/016-node-access.md).

### Proposed change(s)

This PR only implements part of the RFC. Currently, the following is implemented:
- The `openssh-server` package was added to the node image
- OpenSSH was configured to only allow public key authentication and use a CA public key as a [user certificate](https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Certificate-based_Authentication#SSH_User_Certificates)
  - The derivation of this certificate will be handled in another PR.
- A new terraform variable `emergency_ssh` was added to allow control over load balancing ports. Currently, this is implemented (and tested) for
    - [x] azure
    - [x] aws
    - [x] gcp
    - [ ] openstack
- The emergency ssh e2e tests on this PR don't work since they reference the latest debug image build from main, which does not include the SSH server. The "all at once" test under additional information uses the same workflow, but with an image built from this branch.

### Additional info

- working E2E tests:
  - [azure](https://github.com/edgelesssys/constellation/actions/runs/13438121227)
  - [aws](https://github.com/edgelesssys/constellation/actions/runs/13440141872/job/37552326083)
  - [gcp](https://github.com/edgelesssys/constellation/actions/runs/13697211528/job/38302373167)
  - [all at once (`e2e-ssh` workflow)](https://github.com/edgelesssys/constellation/actions/runs/13789602661)

### Checklist

- [x] Run the E2E tests that are relevant to this PR's changes
- [x] Update [docs](https://github.com/edgelesssys/constellation/tree/main/docs)
- [x] Add labels (e.g., for changelog category)
- [x] Is PR title adequate for changelog?

PR #3568, opened 2025-01-07
cli: add `ssh` command to securely connect with nodes over ssh

miampf:
### Context

RFC 16 requires key derivation for the client and all nodes. This PR aims to implement that. For the main PR of this RFC, see #3557 (WIP).

### Proposed change(s)

- Add grpc functionality to request a CA key from the join service
  - This CA key is derived in the key service
  - The join client in the bootstrapper than gets the derived key in the `IssueJoinTicketResponse`

### Additional info

- The RFC specifies that the CA key should be derived directly from the master secret. This is not the case here. Instead, the key service requests a key from a KMS which in turn is used to derive the CA key. If the cKMS is used, the key sent by the KMS is derived using the master secret.

### Checklist

  - Documentation updates regarding the use of this feature will be done in #3557 
- [x] Link to Milestone


miampf:
Used the `feature` label instead of `no changelog` since this PR adds the `constellation ssh` subcommand for key derivation.

cli/internal/cmd/ssh.go
```diff
@@ -0,0 +1,120 @@
...
+	fh := file.NewHandler(afero.NewOsFs())
+	debugLogger, err := newDebugFileLogger(cmd, fh)
+	if err != nil {
+		return err
+	}
+
+	// NOTE(miampf): Since other KMS aren't fully implemented yet, this commands assumes that the cKMS is used and derives the key accordingly.
+	var mastersecret secret
+	if err = fh.ReadJSON(fmt.Sprintf("%s.json", constants.ConstellationMasterSecretStoreName), &mastersecret); err != nil {
+		return err
```
burgerdev, line 63: This is the outermost error, so it should be somewhat digestible for users. Maybe print or wrap an error message along the lines of "mastersecret is required for SSH access"?

The same goes for the public key, in case it does not exist or does not parse as public key.

cli/internal/cmd/ssh.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/edgelesssys/constellation/v2/internal/file"
+	"github.com/edgelesssys/constellation/v2/internal/kms/setup"
+	"github.com/edgelesssys/constellation/v2/internal/kms/uri"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+
+	"golang.org/x/crypto/ssh"
+)
+
+type secret struct {
```
burgerdev, line 27: Don't we have this type somewhere already?

cli/internal/cmd/ssh.go
```diff
@@ -0,0 +1,120 @@
...
+	if err = fh.ReadJSON(fmt.Sprintf("%s.json", constants.ConstellationMasterSecretStoreName), &mastersecret); err != nil {
+		return err
+	}
+
+	mastersecret_uri := uri.MasterSecret{Key: mastersecret.Key, Salt: mastersecret.Salt}
+	kms, err := setup.KMS(cmd.Context(), uri.NoStoreURI, mastersecret_uri.EncodeToURI())
+	if err != nil {
+		return err
+	}
+	key, err := kms.GetDEK(cmd.Context(), crypto.DEKPrefix, 256)
```
burgerdev, line 71: Please use a suffix name specific to SSH CA (and make it a constant over at constants.go).

cli/internal/cmd/ssh.go
```diff
@@ -0,0 +1,120 @@
...
+	if err != nil {
+		return err
+	}
+
+	ca, err := ssh.NewSignerFromSigner(priv)
+	if err != nil {
+		return err
+	}
+
+	debugLogger.Debug("CA KEY generated", "key", string(ssh.MarshalAuthorizedKey(ca.PublicKey())))
```
burgerdev, line 86:
```suggestion
	debugLogger.Debug("SSH CA KEY generated", "public-key", string(ssh.MarshalAuthorizedKey(ca.PublicKey())))
```

cli/internal/cmd/ssh.go
```diff
@@ -0,0 +1,120 @@
...
+		ValidBefore:     uint64(time.Now().Add(24 * time.Hour).Unix()),
+		ValidPrincipals: []string{"root"},
+		Permissions:     permissions,
+	}
+	if err := certificate.SignCert(rand.Reader, ca); err != nil {
+		return err
+	}
+
+	debugLogger.Debug("Signed certificate", "certificate", string(ssh.MarshalAuthorizedKey(&certificate)))
+	fh.Write(fmt.Sprintf("%s/ca_cert.pub", constants.TerraformWorkingDir), ssh.MarshalAuthorizedKey(&certificate), file.OptOverwrite, file.OptMkdirAll)
```
burgerdev, line 116: I think this should not `MkdirAll` - either we have a terraform directory and it's not needed, or we have none, which probably means we're not in a Constellation workspace and should not create terraform dirs.

cli/internal/cmd/ssh.go
```diff
@@ -0,0 +1,120 @@
...
+	fmt.Printf("You can now connect to a node using 'ssh -F %s/ssh_config -i <your private key> <node ip>'.\nYou can obtain the private node IP via the web UI of your CSP.\n", constants.TerraformWorkingDir)
```
burgerdev, line 117: Node names should also work, right? Maybe that's easier to explain.

joinservice/joinproto/join.proto
```diff
@@ -9,14 +9,17 @@ option go_package = "github.com/edgelesssys/constellation/v2/joinservice/joinpro
 service API {
   // IssueJoinTicket issues a join ticket for a new node.
   rpc IssueJoinTicket(IssueJoinTicketRequest) returns (IssueJoinTicketResponse);
-  // IssueRejoinTicket issues a join ticket for a node that has previously joined the cluster.
-  rpc IssueRejoinTicket(IssueRejoinTicketRequest) returns (IssueRejoinTicketResponse);
+  // IssueRejoinTicket issues a join ticket for a node that has previously
+  // joined the cluster.
+  rpc IssueRejoinTicket(IssueRejoinTicketRequest)
+      returns (IssueRejoinTicketResponse);
```
burgerdev, line 15: Please don't reformat existing code if you're not actually changing it - that makes it harder to inspect the history.

keyservice/internal/server/server.go
```diff
@@ -82,3 +85,27 @@ func (s *Server) GetDataKey(ctx context.Context, in *keyserviceproto.GetDataKeyR
...
+func (s *Server) GetCAKey(ctx context.Context, in *keyserviceproto.GetCAKeyRequest) (*keyserviceproto.GetCAKeyResponse, error) {
+	log := s.log.With("peerAddress", grpclog.PeerAddrFromContext(ctx))
+
+	key, err := s.conKMS.GetDEK(ctx, crypto.DEKPrefix, 256)
+	if err != nil {
+		log.With(slog.Any("error", err)).Error("Failed to get key for derivation")
+		return nil, status.Errorf(codes.Internal, "%v", err)
+	}
+
+	_, priv, err := ed25519.GenerateKey(bytes.NewReader(key))
```
burgerdev, line 99: Does the keyservice really need to know about SSH? We could keep most interfaces as they are if we just used `GetDataKey` and derived the SSH public key in the joinclient.

cli/internal/cmd/ssh.go
miampf: Agreed, I'll also add a check if the directory exists

cli/internal/cmd/ssh.go
miampf: To what exactly should I apply this suffix? Do you mean the file name?

burgerdev: The key derivation uses an [`info`](https://pkg.go.dev/golang.org/x/crypto/hkdf#New) parameter to derive different keys for different use cases. You are passing the `crypto.DEKPrefix` as `info` here. Normal data keys use `crypto.DEKPrefix + "<custom context for this key>"`, and I'm suggesting to do the same.

cli/internal/cmd/ssh.go
miampf: Node names dont seem to work, would leave that at the current message

internal/crypto/crypto_test.go
```diff
@@ -121,6 +122,39 @@ func TestGenerateRandomBytes(t *testing.T) {
...
+		wantErr bool
+	}{
+		"invalid key": {
+			key:     make([]byte, 0),
+			wantErr: true,
+		},
+		"valid key": {
+			key: nullKey,
+		},
+	}
```
burgerdev, line 142: Negative test cases you could add:

* nil input
* longer input

cli/internal/cmd/ssh_test.go
```diff
@@ -0,0 +1,97 @@
...
+			fh:           newFsWithDirectory(),
+			masterSecret: someMasterSecret,
+			wantErr:      true,
+		},
+		"no master secret": {
+			fh:      newFsWithDirectory(),
+			pubKey:  someSSHPubKey,
+			wantErr: true,
+		},
+		"malformatted public key": {
```
burgerdev, line 57: Do you mean "malformed"?

cli/internal/cmd/ssh_test.go
```diff
@@ -0,0 +1,97 @@
...
+			if tc.masterSecret != "" {
+				require.NoError(tc.fh.Write(constants.MasterSecretFilename, []byte(tc.masterSecret)))
+			}
+
+			err := generateKey(context.Background(), someSSHPubKeyPath, tc.fh, logger.NewTest(t))
+			if tc.wantErr {
+				assert.Error(err)
+			} else {
+				assert.NoError(err)
+			}
```
burgerdev, line 94: Maybe test that a certificate is written and valid?

internal/crypto/crypto_test.go
miampf: Should longer input be treated as an error? If the input is too long, only the first `ed25519.SeedSize` bytes will be used currently and the rest will be discarded.

joinservice/joinproto/join.proto
```diff
@@ -45,6 +45,8 @@ message IssueJoinTicketResponse {
   string kubernetes_version = 9;
   // kubernetes_components is a list of components to install on the node.
   repeated components.Component kubernetes_components = 10;
+  // emergency_ca_key is an ssh ca key that can be used to connect to a node in case of an emergency.
+  bytes emergency_ca_key = 11;
```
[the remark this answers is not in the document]
daniel-weisse: This is now the authorized public key used to access the nodes via ssh, correct?
Can we adjust the name accordingly?

bootstrapper/internal/joinclient/joinclient.go
```diff
@@ -271,6 +271,10 @@ func (c *JoinClient) startNodeAndJoin(ticket *joinproto.IssueJoinTicketResponse,
 		return fmt.Errorf("writing kubelet key: %w", err)
 	}
 
+	if err := c.fileHandler.Write(constants.SSHCAKeyPath, ticket.EmergencyCaKey, file.OptMkdirAll); err != nil {
+		return fmt.Errorf("writing ca key: %w", err)
```
daniel-weisse, line 275:
```suggestion
		return fmt.Errorf("writing ssh ca key: %w", err)
```
nit

cli/internal/cmd/ssh.go
```diff
@@ -0,0 +1,125 @@
...
+
+	"golang.org/x/crypto/ssh"
+)
+
+var permissions = ssh.Permissions{
+	Extensions: map[string]string{
+		"permit-port-forwarding": "yes",
+		"permit-pty":             "yes",
+	},
+}
```
daniel-weisse, line 33: inline this since it's only use is in line 112

cli/internal/cmd/ssh.go
```diff
@@ -0,0 +1,125 @@
...
+// NewSSHCmd returns a new cobra.Command for the ssh command.
+func NewSSHCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "ssh",
+		Short: "Prepare your cluster for emergency ssh access",
+		Long:  "Prepare your cluster for emergency ssh access and sign a given key pair for authorization.",
+		Args:  cobra.ExactArgs(0),
+		RunE:  runSSH,
+	}
+	cmd.Flags().String("key", "", "The path to an existing ssh public key.")
```
daniel-weisse, line 44:
```suggestion
	cmd.Flags().String("key", "", "the path to an existing ssh public key")
```
Flags descriptions should start lowercase and not end in a punctuation mark if they are a single sentence

cli/internal/cmd/ssh.go
```diff
@@ -0,0 +1,125 @@
...
+	if os.IsNotExist(err) {
+		return fmt.Errorf("directory %q does not exist", constants.TerraformWorkingDir)
+	}
+	if err != nil {
+		return err
+	}
+
+	// NOTE(miampf): Since other KMS aren't fully implemented yet, this commands assumes that the cKMS is used and derives the key accordingly.
+	var mastersecret uri.MasterSecret
+	if err = fh.ReadJSON(fmt.Sprintf("%s.json", constants.ConstellationMasterSecretStoreName), &mastersecret); err != nil {
```
daniel-weisse, line 75:
```suggestion
	if err = fh.ReadJSON(constants.MasterSecretFilename, &mastersecret); err != nil {
```

internal/crypto/crypto_test.go
```diff
@@ -121,6 +122,47 @@ func TestGenerateRandomBytes(t *testing.T) {
...
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			assert := assert.New(t)
+
+			_, err := GenerateEmergencySSHCAKey(tc.key)
+			if tc.wantErr {
+				assert.NotNil(err)
+			} else {
+				assert.Nil(err)
+			}
```
daniel-weisse, line 161:
```suggestion
			if tc.wantErr {
				assert.Error(err)
			} else {
				assert.NoError(err)
			}
```

internal/crypto/crypto.go
```diff
@@ -62,6 +64,19 @@ func GenerateRandomBytes(length int) ([]byte, error) {
 	return nonce, nil
 }
 
+// GenerateEmergencySSHCAKey creates a CA that is used to sign keys for emergency ssh access.
+func GenerateEmergencySSHCAKey(key []byte) (ssh.Signer, error) {
+	_, priv, err := ed25519.GenerateKey(bytes.NewReader(key))
```
daniel-weisse, line 69:
```suggestion
func GenerateEmergencySSHCAKey(seed []byte) (ssh.Signer, error) {
	_, priv, err := ed25519.GenerateKey(bytes.NewReader(seed))
```
nit: just for clarity. The `ed25519` also refers to this input as seed

joinservice/internal/server/server.go
```diff
@@ -100,6 +102,18 @@ func (s *Server) IssueJoinTicket(ctx context.Context, req *joinproto.IssueJoinTi
...
+	ssheCADerivationKey, err := s.dataKeyGetter.GetDataKey(ctx, constants.SSHCAKeySuffix, ed25519.SeedSize)
+	if err != nil {
+		log.With(slog.Any("error", err)).Error("Failed to get emergency SSH CA derivation key")
+		return nil, status.Errorf(codes.Internal, "getting emergency SSH CA derivation key: %s", err)
+	}
+	ca, err := crypto.GenerateEmergencySSHCAKey(ssheCADerivationKey)
+	if err != nil {
+		log.With(slog.Any("error", err)).Error("Failed to derive ssh CA key from derivation key")
+		return nil, status.Errorf(codes.Internal, "generating ssh emergency CA key: %s", err)
+	}
```
daniel-weisse, line 115:
```suggestion
	sshCAKeySeed, err := s.dataKeyGetter.GetDataKey(ctx, constants.SSHCAKeySuffix, ed25519.SeedSize)
	if err != nil {
		log.With(slog.Any("error", err)).Error("Failed to get seed material to derive SSH CA key")
		return nil, status.Errorf(codes.Internal, "getting emergency SSH CA key seed material: %s", err)
	}
	ca, err := crypto.GenerateEmergencySSHCAKey(sshCAKeySeed)
	if err != nil {
		log.With(slog.Any("error", err)).Error("Failed to derive ssh CA key from seed material")
		return nil, status.Errorf(codes.Internal, "generating ssh emergency CA key: %s", err)
	}
```
similar to my comment for the `crypto` package

joinservice/joinproto/join.proto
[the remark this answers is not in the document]
miampf: This is the public key of the CA that is then used to verify the signature of a private key and certificate used to authenticate with the node, so it still is a CA key. I can make the public key part more clear, but I think the CA part should stay.

cli/internal/cmd/ssh.go
```diff
@@ -0,0 +1,125 @@
...
+	kms, err := setup.KMS(ctx, uri.NoStoreURI, mastersecretURI.EncodeToURI())
+	if err != nil {
+		return fmt.Errorf("setting up KMS: %s", err)
+	}
+	key, err := kms.GetDEK(ctx, crypto.DEKPrefix+constants.SSHCAKeySuffix, ed25519.SeedSize)
+	if err != nil {
+		return fmt.Errorf("retrieving key from KMS: %s", err)
+	}
+
+	ca, err := crypto.GenerateEmergencySSHCAKey(key)
```
daniel-weisse, line 89:
```suggestion
	sshCAKeySeed, err := kms.GetDEK(ctx, crypto.DEKPrefix+constants.SSHCAKeySuffix, ed25519.SeedSize)
	if err != nil {
		return fmt.Errorf("retrieving key from KMS: %s", err)
	}

	ca, err := crypto.GenerateEmergencySSHCAKey(sshCAKeySeed)
```
same as my previous two comments

e2e/miniconstellation/.terraform.lock.hcl
daniel-weisse, line 1: This change does not belong in this PR, I think

cli/internal/cmd/ssh.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	if err := certificate.SignCert(rand.Reader, ca); err != nil {
+		return fmt.Errorf("signing certificate: %s", err)
+	}
+
+	debugLogger.Debug("Signed certificate", "certificate", string(ssh.MarshalAuthorizedKey(&certificate)))
+	if err := fh.Write(fmt.Sprintf("%s/ca_cert.pub", constants.TerraformWorkingDir), ssh.MarshalAuthorizedKey(&certificate), file.OptOverwrite); err != nil {
+		return fmt.Errorf("writing certificate: %s", err)
+	}
+	fmt.Printf("You can now connect to a node using 'ssh -F %s/ssh_config -i <your private key> <node ip>'.\nYou can obtain the private node IP via the web UI of your CSP.\n", constants.TerraformWorkingDir)
```
daniel-weisse, line 122: Why is the ssh key written to the terraform directory?
That seems a bit out of place to me. Writing to the current working directory should be enough

daniel-weisse, line 122: nit: pass the `cobra.Command` and use `cmd.Printf` to write the message.
That would theoretically also allow testing against this output in case we ever need to do that.

daniel-weisse (changes requested):
Overall structure looks good to me.
Mostly notes on style and naming

cli/internal/cmd/ssh.go
miampf: It is written into the terraform directory, since the corresponding `ssh_config` for the client is generated by terraform and I felt like having certificate and corresponding ssh config in the same directory makes sense.

joinservice/joinproto/join.proto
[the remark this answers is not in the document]
daniel-weisse: Lets go with `authorized_ca_public_key`

cli/internal/cmd/ssh.go
```diff
@@ -0,0 +1,122 @@
...
+// NewSSHCmd returns a new cobra.Command for the ssh command.
+func NewSSHCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "ssh",
+		Short: "Prepare your cluster for emergency ssh access",
+		Long:  "Prepare your cluster for emergency ssh access and sign a given key pair for authorization.",
+		Args:  cobra.ExactArgs(0),
+		RunE:  runSSH,
+	}
+	cmd.Flags().String("key", "", "the path to an existing ssh public key.")
```
daniel-weisse, line 36:
```suggestion
	cmd.Flags().String("key", "", "the path to an existing ssh public key")
```

e2e/miniconstellation/.terraform.lock.hcl
daniel-weisse, line 1: The change is back again.
Checking the terraform code, there shuldn't be the need to import the random provider for the miniConstellation e2e test.
Do you perhaps have some file that is gitignored which requires the random provider?

miampf: I think this happened accidentally while switching from one branch to another. I'll remove the file from the git history and clean up my repo in the hopes that that fixes it.

miampf: Managed to remove the random import, however, some hashes still change on generate even tho I should not have any hidden files that influence the miniconstellation terraform. I am unsure if this is intended?

daniel-weisse: Possibly some configuration problems with your setup.
If the CI doesn't complain its fine.
In general, I would recommend just dropping the changes to that file and not committing them at all

miampf: Dropped all file changes by just copying the file from main

daniel-weisse (approved):
looks good from my side
Please let others review as well before merging

miampf:
Not a fan of the new PR title since this does not encapsulate all the changes + lets it seem like the feature is fully implemented, which it is not. Any chance that you'd be open to a different title @daniel-weisse ?

daniel-weisse:
I changed the PR title to be compliant with [our PR title conventions](https://github.com/edgelesssys/constellation/blob/main/dev-docs/workflows/pull-request.md#conventions).
Since the PR title is used for the user facing changelog, the important thing is that this PR implements the CLI command. Any internal changes are secondary and not that relevant for the user.

cli/internal/cmd/ssh.go
```diff
@@ -0,0 +1,122 @@
...
+
+	keyPath, err := cmd.Flags().GetString("key")
+	if err != nil {
+		return fmt.Errorf("retrieving path to public key from flags: %s", err)
+	}
+
+	return generateKey(cmd, keyPath, fh, debugLogger)
+}
+
+func generateKey(cmd *cobra.Command, keyPath string, fh file.Handler, debugLogger debugLog) error {
```
burgerdev, line 56: This function should rather be called `writeCertificateForKey`.

burgerdev (approved):
lgtm, thanks!

Thanks for the review, @daniel-weisse!

PR #3587, opened 2025-01-17
ci: allow unrestricted user namespaces in image builds

msanft:
### Context

Since Ubuntu 23.10, creating namespaces, e.g. via `unshare`, requires privileges. Since `unshare` is used by `mkosi` in our image build pipeline, we need to re-enable them explicitly here before building the image, aligning with the behavior before Ubuntu 23.10.

### Proposed change(s)

- Explicitly enable unprivileged user namespaces in the image build action.

### Additional info

- [Action before this update](https://github.com/edgelesssys/constellation/actions/runs/12818102529)
- [Action after this update](https://github.com/edgelesssys/constellation/actions/runs/12825033842)

### Checklist

- [x] Link to Milestone

### DIFF
--- a/.github/workflows/build-os-image.yml
+++ b/.github/workflows/build-os-image.yml
@@ -167,6 +167,12 @@ jobs:
         with:
           clouds_yaml: ${{ secrets.STACKIT_IMAGE_UPLOAD_CLOUDS_YAML }}
 
+      - name: Allow unrestricted user namespaces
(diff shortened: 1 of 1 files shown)


daniel-weisse (approved):

burgerdev:
For the record, this was tested: https://github.com/edgelesssys/constellation/actions/runs/12825033842.

PR #3656, opened 2025-02-20
gcp: support projects with no default permissions

3u13r:
### Context

When creating VMs we used to rely on the following points:
1. The default project service account is automatically attached to every VM in the project, which doesn't override it
2. The default service account used to have many permissions due having the Editor role attached by default (see: https://cloud.google.com/iam/docs/service-account-types#default)
3. Though, the scopes added to the VM filtered how this service account could be used

Since GCP now doesn't give the project service account Editor by default, we need to create our own service account with minimal permissions. Note that the filtering over scopes is deprecated, one should simply attach a service account with minimal permissions and maximum scope.

Not only the Bootstrapper on the VM but also the constellation-operator and join-service automatically used the project service account per default, since we didn't explicitly set the service account already create for (and used by) other cluster components.
Since we want the VMs permissions to be as narrow as possible, we need to pass the in-cluster service account to also the constellation-operator and join-service. 

### Proposed change(s)

- Create a second very narrow, read-only service account and attach it to VMs
- Don't filter via scope anymore
- Attach the in-cluster service account to the constellation-operator and join-service

### Checklist

   - [ ]  ~~e2e upgrade. gcp, snp: https://github.com/edgelesssys/constellation/actions/runs/13443379097~~
   - [ ]  ~~e2e upgrade. gcp, snp: https://github.com/edgelesssys/constellation/actions/runs/13755459817~~
   - [ ] ~~https://github.com/edgelesssys/constellation/actions/runs/13762021363~~
   - [x] https://github.com/edgelesssys/constellation/actions/runs/13763556980
- [x] Link to Milestone

How to test:
* Create/upgrade your gcp iam credentials
* Successfully start a cluster, neither the bootstrapper, operator or the joinservice should error


cli/internal/cloudcmd/iam.go
msanft, line 1: For the sake of not further complicating the iam create command, should we make it default to use the `serviceAccountID` for the `serviceAccountVMID` if the latter is not present? (i.e. the current behavior)

msanft: Just making it a prefix is a great idea. Would also resemble what we [currently have on AWS](https://docs.edgeless.systems/constellation/reference/cli#synopsis-20).

3u13r: I guess then I add a prefix flag to the command. While this is counter-intuitive behavior: should we make it have precedence over the `serviceAccountID` if both are set, so we can set both in the CI and deprecate the old on slowly?

msanft: Id not have both but remove serviceAccountID directly. (we can have breaking changes)

Having both would indeed be counter-intuitive. Or what was your point for that?

terraform-provider-constellation/examples/full/gcp/main.tf
```diff
@@ -45,11 +45,11 @@ resource "random_bytes" "measurement_salt" {
 
 module "gcp_iam" {
   // replace $VERSION with the Constellation version you want to use, e.g., v2.14.0
-  source             = "https://github.com/edgelesssys/constellation/releases/download/$VERSION/terraform-module.zip//terraform-module/iam/gcp"
-  project_id         = local.project_id
-  service_account_id = "${local.name}-sa"
-  zone               = local.zone
-  region             = local.region
+  source      = "https://github.com/edgelesssys/constellation/releases/download/$VERSION/terraform-module.zip//terraform-module/iam/gcp"
+  sproject_id = local.project_id
```
elchead, line 49:
```suggestion
  project_id = local.project_id
```

elchead (approved):
owned files lgtm

daniel-weisse (commented):
Getting an error when running e2e tests during `constellation iam create`:
```
Error: "account_id" ("e2e-13852850608-1-fe70a4d4-sa-vm") must be between 6 and 30 characters long
  with google_service_account.service_account_vm,
  on main.tf line 22, in resource "google_service_account" "service_account_vm":
  22:   account_id   = local.sa_vm_name
Attempting to roll back.
Rollback succeeded.
Error: terraform apply: exit status 1
Error: "account_id" ("e2e-13852850608-1-fe70a4d4-sa-vm") must be between 6 and 30 characters long
  with google_service_account.service_account_vm,
  on main.tf line 22, in resource "google_service_account" "service_account_vm":
  22:   account_id   = local.sa_vm_name
```

https://github.com/edgelesssys/constellation/actions/runs/13852850608/job/38763459800

cli/internal/cloudcmd/iamupgrade.go
```diff
@@ -22,6 +22,9 @@ import (
 // UpgradeRequiresIAMMigration returns true if the given cloud provider requires an IAM migration.
 func UpgradeRequiresIAMMigration(provider cloudprovider.Provider) bool {
```
msanft, line 23: We don't call this function anywhere atm, do we?

3u13r:
Thanks for the review! 
New e2e tests:
gcp, sev-snp, nop: https://github.com/edgelesssys/constellation/actions/runs/13858160864
gcp, sev-snp, upgrade: https://github.com/edgelesssys/constellation/actions/runs/13858179541

msanft (approved):
Ignored all the nits for now so we can get this PR through with.

My only observation is that we seem to (partly) attach the roles on K8s resource level now. Wouldn't it be enough to just attach the accounts to the VMs on Terraform level?
Just for my understanding.

Otherwise, good to merge for me.

PR #3657, opened 2025-02-20
cli: re-introduce iam upgrade check

3u13r:
### Context

We still maintain the function in which we manually add the CSP if we did changes that require an `constellation iam upgrade`, though the function wasn't used for some time. Either we drop this safety check completely (as it also only applies to the cli and not the terraform) and is one less manual step when changing code and releasing, or we use it. 

I'm currently of the opinion to drop this, but wanted to show what the full feature looked like.

### Proposed change(s)

- re-introduce iam upgrade check

### Checklist

upgrade to this branch: (gcp-snp, 1:1) https://github.com/edgelesssys/constellation/actions/runs/13679706679
nop e2e gcp-snp: https://github.com/edgelesssys/constellation/actions/runs/13679730897

### DIFF
--- a/cli/internal/cmd/apply.go
+++ b/cli/internal/cmd/apply.go
@@ -368,7 +368,7 @@ func (a *applyCmd) apply(
 	// Check current Terraform state, if it exists and infrastructure upgrades are not skipped,
 	// and apply migrations if necessary.
 	if !a.flags.skipPhases.contains(skipInfrastructurePhase) {
-		if err := a.runTerraformApply(cmd, conf, stateFile, upgradeDir); err != nil {
+		if err := a.runTerraformApply(cmd, conf, stateFile, upgradeDir, a.flags.yes); err != nil {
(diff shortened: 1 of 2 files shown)


elchead:
Good catch. What is your proposed way of telling users that they need to upgrade the IAM setup for this upgrade?
Is it the release notes? IMO we can drop this warning in the CLI then

daniel-weisse (approved):
IMO we should make this as obvious as possible for the user.
While documenting it in just one place is simpler for the devs, I see no harm in additionally keeping the check in the CLI.
Since these IAM upgrades are somewhat of a "breaking" change we should document they should ideally be mentioned in the release notes, have a small entry in the migration page, and have this additional check during upgrades with the CLI

We can't easily supply something similar for Terraform, but here people should pay a bit more attention to the release notes anyways.

elchead (approved):
I don't have a strong opinion either, but I agree that the dev effort for maintaining the warning is also limited. Let's merge then

PR #3663, opened 2025-02-24
constellation-node-operator: upgrade control plane nodes first

msanft:
### Context

Since Kubernetes v1.31, the requirement that Kubeletes of version `n` are incompatible with outdated (version `n-1`) Kube-API servers / control planes is actually enforced / errors if not fulfilled. Thus, we need to make sure during the upgrade that this case doesn't occur in a Constellation cluster.

### Proposed change(s)

- Sort the nodes in the upgrade process of the operator, such that control plane nodes get upgraded first.
- Also, workers now wait during the upgrade if there are still outdated control plane nodes.

### Additional info

- [AB#5311](https://dev.azure.com/Edgeless/ae37573d-ccde-4af2-ab1e-001e587197d1/_workitems/edit/5311)
- [GCP 2:8 Upgrade Test](https://github.com/edgelesssys/constellation/actions/runs/13496779987). The high worker count ensures that the probability for a worker node being upgraded before a control plane node is reasonably high.

### Checklist

- [x] Link to Milestone

### DIFF
--- a/operators/constellation-node-operator/controllers/nodeversion_controller.go
+++ b/operators/constellation-node-operator/controllers/nodeversion_controller.go
@@ -214,7 +214,7 @@ func (r *NodeVersionReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{Requeue: shouldRequeue}, nil
 	}
 
-	newNodeConfig := newNodeConfig{desiredNodeVersion, groups.Outdated, pendingNodeList.Items, scalingGroupByID, newNodesBudget}
+	newNodeConfig := newNodeConfig{desiredNodeVersion, groups.Outdated, groups.Donors, pendingNodeList.Items, scalingGroupByID, newNodesBudget}
 	if err := r.createNewNodes(ctx, newNodeConfig); err != nil {
 		logr.Error(err, "Creating new nodes")
 		return ctrl.Result{Requeue: shouldRequeue}, nil
@@ -614,6 +614,15 @@ func (r *NodeVersionReconciler) createNewNodes(ctx context.Context, config newNo
 	if config.newNodesBudget < 1 || len(config.outdatedNodes) == 0 {
 		return nil
 	}
+	// We need to look at both the outdated nodes *and* the nodes that have already
+	// been moved to the donors here because even if a CP node has already been moved to
+	// the donors, we still want to defer worker upgrades until the new CP node is actually joined.
+	hasOutdatedControlPlanes := false
+	for _, entry := range append(config.outdatedNodes, config.donors...) {
+		if nodeutil.IsControlPlaneNode(&entry) {
+			hasOutdatedControlPlanes = true
+		}
+	}
 	outdatedNodesPerScalingGroup := make(map[string]int)
 	for _, node := range config.outdatedNodes {
 		// skip outdated nodes that got assigned an heir in this Reconcile call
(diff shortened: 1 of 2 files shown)


operators/constellation-node-operator/controllers/nodeversion_controller.go
```diff
@@ -614,6 +614,17 @@ func (r *NodeVersionReconciler) createNewNodes(ctx context.Context, config newNo
...
+		// We need to look at both the outdated nodes *and* the nodes that have already
+		// been moved to the donors here because even if a CP node has already been moved to
+		// the donors, we still want to defer worker upgrades until the new CP node is actually joined.
+		for _, entry := range append(config.outdatedNodes, config.donors...) {
+			if nodeutil.IsControlPlaneNode(&entry) {
+				return true
+			}
+		}
+		return false
+	}()
```
daniel-weisse, line 627:
```suggestion
	// We need to look at both the outdated nodes *and* the nodes that have already
	// been moved to the donors here because even if a CP node has already been moved to
	// the donors, we still want to defer worker upgrades until the new CP node is actually joined.
	hasOutdatedControlPlanes := false
	for _, entry := range append(config.outdatedNodes, config.donors...) {
		if nodeutil.IsControlPlaneNode(&entry) {
			hasOutdatedControlPlanes = true
		}
	}
```
Why inline this function if its not reused?

daniel-weisse (commented):
looks good to me if tests pass

internal/versions/versions.go
```diff
@@ -377,7 +377,7 @@ var VersionConfigs = map[ValidK8sVersion]KubernetesVersion{
 				InstallPath: patchFilePath("kube-scheduler"),
 			},
 			{
-				Url:         "data:application/json;base64,W3sib3AiOiJyZXBsYWNlIiwicGF0aCI6Ii9zcGVjL2NvbnRhaW5lcnMvMC9pbWFnZSIsInZhbHVlIjoicmVnaXN0cnkuazhzLmlvL2V0Y2Q6My41LjE1LTBAc2hhMjU2OmE2ZGM2M2U2ZThjZmEwMzA3ZDc4NTE3NjJmYTZiNjI5YWZiMThmMjhkOGFhM2ZhYjVhNmU5MWI0YWY2MDAyNmEifV0=",
+				Url:         "data:application/json;base64,W3sib3AiOiJyZXBsYWNlIiwicGF0aCI6Ii9zcGVjL2NvbnRhaW5lcnMvMC9pbWFnZSIsInZhbHVlIjoicmVnaXN0cnkuazhzLmlvL2V0Y2Q6My41LjE2LTBAc2hhMjU2OmM2YTlkMTFjYzVjMDRiMTE0Y2NkZWYzOWE5MjY1ZWVlZjgxOGUzZDAyZjUzNTliZTAzNWFlNzg0MDk3ZmRlYzUifV0=",
```
3u13r, line 380: I think this should not be part of this PR. Is the main not generated?

3u13r (changes requested):

internal/versions/versions.go
msanft: [burgerdev/k8s-1.31](https://github.com/edgelesssys/constellation/tree/burgerdev/k8s-1.31) is not. I think it doesn't matter whether this is changed here or in that PR.

msanft:
Manual 1:4 test on GCP succeeded too

daniel-weisse (approved):

internal/versions/versions.go
burgerdev: My bad, mistaken rebase conflict resolution. Should be fixed now.

operators/constellation-node-operator/controllers/nodeversion_controller.go
```diff
@@ -637,7 +646,7 @@ func (r *NodeVersionReconciler) createNewNodes(ctx context.Context, config newNo
 			requiredNodesPerScalingGroup[scalingGroupID] = outdatedNodesPerScalingGroup[scalingGroupID] - pendingJoiningNodesPerScalingGroup[scalingGroupID]
 		}
 	}
-	for scalingGroupID := range requiredNodesPerScalingGroup {
+	for _, scalingGroupID := range config.scalingGroupByID.getScalingGroupIDsSorted() {
```
burgerdev, line 649: I don't quite understand why we need to sort the groups - is this an optimization? The `continue` block added below looks like it would already do the trick.

burgerdev (approved):

operators/constellation-node-operator/controllers/nodeversion_controller.go
msanft: Yeah, you're probably right. I took this from Leo's PR initially without questioning it, but thinking about it - you seem right. Will remove this.

msanft:
[New E2E Test](https://github.com/edgelesssys/constellation/actions/runs/13541199418)

PR #3772, opened 2025-04-16
cli: support http proxy

thomasten:
### Context

This adds support for using the CLI behind an http proxy.
This is "best effort" for now and won't be documented.

### Proposed change(s)

- use default grpc dialer

### Additional info

- Tested manually. Won't add an e2e test for now.


msanft (approved):

elchead (approved):
owned files lgtm

PR #3773, opened 2025-04-16
docs: change exemplary GCP zones to ones that provide SNP machines

thomasten:
europe-west2 and us-west1 only have SEV-ES


3u13r (approved):
I personally wouldn't touch more than the latest released docs version (and "next"). Especially since there are versions where we don't support SNP yet and the docs are perfectly fine. But since this is more of a nit, approving anyway.

PR #3786, opened 2025-04-28
feat: use SSH host certificates

miampf:
### Context

Currently, if you use the emergency SSH access, you have to manually validate each nodes host key (TOFU). This PR fixes that by introducing [host certificates](https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Certificate-based_Authentication#SSH_Host_Certificates).

### Proposed change(s)

- Each node gets a host certificate that is used to verify it's authenticity

### Additional info

- For some reason, wildcards as a principal do not work which means that if you try to connect using the load balancers host name, verification would fail. However, connecting with the LBs public IP is not a problem so this is recommended in the documentation.

### Checklist

- [x] Link to Milestone


cli/internal/cmd/ssh.go
```diff
@@ -74,7 +74,12 @@ func writeCertificateForKey(cmd *cobra.Command, keyPath string, fh file.Handler,
 		return fmt.Errorf("generating SSH emergency CA key: %s", err)
 	}
 
-	debugLogger.Debug("SSH CA KEY generated", "public-key", string(ssh.MarshalAuthorizedKey(ca.PublicKey())))
+	marshalledKey := string(ssh.MarshalAuthorizedKey(ca.PublicKey()))
+	debugLogger.Debug("SSH CA KEY generated", "public-key", marshalledKey)
+	knownHostsContent := fmt.Sprintf("@cert-authority * %s", marshalledKey)
+	if err := fh.Write("./known_hosts", []byte(knownHostsContent), file.OptMkdirAll); err != nil {
+		return fmt.Errorf("reading known hosts file (does \"./known_hosts\" exist?): %w", err)
```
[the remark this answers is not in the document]
miampf: The file is read by the SSH CLI when connecting

joinservice/internal/server/server.go
```diff
@@ -114,6 +120,21 @@ func (s *Server) IssueJoinTicket(ctx context.Context, req *joinproto.IssueJoinTi
 		return nil, status.Errorf(codes.Internal, "generating ssh emergency CA key: %s", err)
 	}
 
+	// FIXME: the file exists on the node but - logically - not on the container
```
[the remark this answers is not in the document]
miampf: Sorry, leftover comment, will remove this

bootstrapper/internal/interfaces/interfaces.go
```diff
@@ -0,0 +1,40 @@
...
+package interfaces
+
+import (
+	"net"
+)
+
+// GetNetworkInterfaces retrieves all network interface addresses.
+func GetNetworkInterfaces() ([]string, error) {
+	var principalList []string
+	ifaces, err := net.Interfaces()
```
burgerdev, line 16: If you pass `[]net.Interface` as input to the function, you should be able to add some unit tests.

bootstrapper/internal/interfaces/interfaces.go
```diff
@@ -0,0 +1,40 @@
...
+
+package interfaces
+
+import (
+	"net"
+)
+
+// GetNetworkInterfaces retrieves all network interface addresses.
+func GetNetworkInterfaces() ([]string, error) {
+	var principalList []string
```
burgerdev, line 15: Refering to `principals` does not make much sense in this package.

internal/crypto/crypto.go
```diff
@@ -77,6 +78,41 @@ func GenerateEmergencySSHCAKey(seed []byte) (ssh.Signer, error) {
 	return ca, nil
 }
 
+// GenerateSignedSSHHostKey creates a host key that is signed by the given CA.
+func GenerateSignedSSHHostKey(principals []string, ca ssh.Signer) (*pem.Block, *ssh.Certificate, error) {
+	hostKeyPub, hostKey, err := ed25519.GenerateKey(nil)
```
burgerdev, line 83: Why do we generate the key here, instead of receiving a public key and signing that?

internal/crypto/crypto.go
```diff
@@ -77,6 +78,41 @@ func GenerateEmergencySSHCAKey(seed []byte) (ssh.Signer, error) {
...
+	}
+
+	certificate := ssh.Certificate{
+		CertType:        ssh.HostCert,
+		ValidPrincipals: principals,
+		ValidAfter:      uint64(time.Now().Unix()),
+		ValidBefore:     ssh.CertTimeInfinity,
+		Reserved:        []byte{},
+		Key:             hostKeySSHPub,
+		KeyId:           "host_key",
```
burgerdev, line 103: Maybe the first principal would make more sense here.

joinservice/internal/server/server.go
```diff
@@ -114,6 +120,20 @@ func (s *Server) IssueJoinTicket(ctx context.Context, req *joinproto.IssueJoinTi
 		return nil, status.Errorf(codes.Internal, "generating ssh emergency CA key: %s", err)
 	}
 
+	principalList := req.HostCertificatePrincipals
+	additionalPrincipals, err := s.fileHandler.Read(constants.SSHAdditionalPrincipalsPath)
```
burgerdev, line 124: This file is only written by the initserver right now. If we want to use it on all control plane nodes, we need to write it in the joinclient, too.

daniel-weisse:
Do these certificates work across node reboots?

bootstrapper/internal/initserver/initserver.go
```diff
@@ -247,6 +249,77 @@ func (s *Server) Init(req *initproto.InitRequest, stream initproto.API_InitServe
...
+	if err != nil {
+		if e := s.sendLogsWithMessage(stream, status.Errorf(codes.Internal, "getting network interfaces: %s", err)); e != nil {
+			err = errors.Join(err, e)
+		}
+		return err
+	}
+	// Needed since go doesn't implicitly convert slices of structs to slices of interfaces
+	interfacesForFunc := make([]addresses.NetInterface, len(interfaces))
+	for i := range interfaces {
+		interfacesForFunc[i] = &interfaces[i]
```
daniel-weisse, line 262: Same question here

bootstrapper/internal/initserver/initserver.go
daniel-weisse, line 1: I think the whole error return logic in this file can be simplified a lot using the following logic:

```Go
if err != nil {
    return errors.Join(err, s.sendLogsWithMessage(stream, status.Errorf(codes.<error_code>, fmt.Errorf("<error message>: %w", err)))
}
```

daniel-weisse (commented):
looks good to me if tests pass

daniel-weisse (approved):

msanft (approved):

bootstrapper/internal/joinclient/joinclient.go
```diff
@@ -209,6 +214,68 @@ func (c *JoinClient) requestJoinTicket(serviceEndpoint string) (ticket *joinprot
...
+			c.log.With(slog.Any("error", err)).Error("Failed to read SSH host key file")
+			return nil, nil, err
+		}
+
+		hostKey, err := ssh.ParsePrivateKey(hostKeyData)
+		if err != nil {
+			c.log.With(slog.Any("error", err)).Error("Failed to parse SSH host key file")
+			return nil, nil, err
+		}
+		hostKeyPubSSH = hostKey.PublicKey()
```
burgerdev, line 276: It looks like this branch assumes an existing `constants.SSHHostKeyPath` file, but 

* `err` might have been something non-nil, other than `os.ErrNotExist`
* the path may be a directory (in principle, obviously unlikely)

bootstrapper/internal/joinclient/joinclient.go
```diff
@@ -209,6 +214,68 @@ func (c *JoinClient) requestJoinTicket(serviceEndpoint string) (ticket *joinprot
...
+		if err != nil {
+			c.log.With(slog.Any("error", err)).Error("Failed to format SSH host key")
+			return nil, nil, err
+		}
+		if err := c.fileHandler.Write(constants.SSHHostKeyPath, pem.EncodeToMemory(pemHostKey), file.OptMkdirAll); err != nil {
+			c.log.With(slog.Any("error", err)).Error("Failed to write SSH host key")
+			return nil, nil, err
+		}
+	} else {
+		hostKeyData, err := c.fileHandler.Read(constants.SSHHostKeyPath)
```
burgerdev, line 265: In which situation do we have an SSH key on the filesystem while running the join client? Don't we need to join before we can even access the state?

miampf: IIRC, in testing I experienced that the join client of the worker (or a control plane? I don't quite remember) crashed and that the error was an existing key, which is why I added this. But I have also kinda lost track a bit since this PR underwent some major restructuring multiple times so maybe this was resolved in the process. I will test this again next week and tell you the results :)

burgerdev (approved):
Thanks!

miampf:
[emergency ssh e2e test](https://github.com/edgelesssys/constellation/actions/runs/15876065372/job/44763942524)

PR #3819, opened 2025-05-08
ci: run e2e cleanup daily

daniel-weisse:
### Context
No reason to only run this once a week. Resources should be cleaned up as quick as possible

### Proposed change(s)

- Run clean up once a day
- Don't abort a clean up run if a step fails, instead continue to the end
- Clean up resources from release runs

### DIFF
--- a/.github/actions/e2e_cleanup_timeframe/e2e-cleanup.sh
+++ b/.github/actions/e2e_cleanup_timeframe/e2e-cleanup.sh
@@ -2,7 +2,7 @@
 
 # get_e2e_test_ids_on_date gets all workflow IDs of workflows that contain "e2e" on a specific date.
 function get_e2e_test_ids_on_date {
-  ids="$(gh run list --created "$1" --status failure --json createdAt,workflowName,databaseId --jq '.[] | select(.workflowName | contains("e2e") and (contains("MiniConstellation") | not)) | .databaseId' -L1000 -R edgelesssys/constellation || exit 1)"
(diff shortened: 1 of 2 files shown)


.github/actions/e2e_cleanup_timeframe/e2e-cleanup.sh
```diff
@@ -14,20 +14,20 @@ function download_tfstate_artifact {
...
 function delete_resources {
   if [[ -d "$1/constellation-terraform" ]]; then
-    cd "$1/constellation-terraform" || exit 1
-    terraform init > /dev/null || exit 1 # first, install plugins
-    terraform destroy -auto-approve || exit 1
-    cd ../../ || exit 1
+    cd "$1/constellation-terraform" || return 1
+    terraform init > /dev/null || return 1 # first, install plugins
+    terraform destroy -auto-approve || return 1
+    cd ../../ || return 1
```
thomasten, line 20: we need to make sure that we leave the directory if one of the terraform commands fail

thomasten (approved):
Thanks for fixing this!

daniel-weisse:
I reduced the cleanup timeframe down to 4 days.
There is a quirk that can cause errors when trying to remove GCP resources after the IAM resources were already successfully removed.
This could lead to many cleanup runs with errors in that timeframe.
Since the job should now be a bit more reliable, I'll make sure to check the failure if anything does go wrong

